PR #18, opened 2023-05-18
event flows

cjcobb23:
A bunch of different event flows. Leave a comment if there is a flow not here that you would like to see.


cjcobb23:
superseded by https://github.com/axelarnetwork/axelar-amplifier/pull/40

PR #35, opened 2023-07-20
feat(CommandBatcher): add ABI encoding logic

eguajardo:
https://axelarnetwork.atlassian.net/browse/AXE-1294

depends on this one first: https://github.com/axelarnetwork/poc-axelar-service-registry/pull/16


contracts/command-batcher/Cargo.toml
```diff
@@ -40,6 +40,12 @@ thiserror = { version = "1.0.31" }
 serde_json = "1.0.89"
 either = "1.8.1"
 hex = { version = "0.4.3", default-features = false, features = [] }
+connection-router = { workspace = true, features = ["library"] }
+ethabi = { version = "18.0.0", default-features = false, features = [] }
+sha3 = { version = "0.10", default-features = false, features = [] }
+multisig = { workspace = true, features = ["library"] }
+axelar-wasm-std = { workspace = true }
+k256 = { version = "0.11.1", features = ["ecdsa"] }
```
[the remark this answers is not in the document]
eguajardo: this is what we discussed in a standup some time ago, first I tried this crate in the multisig contract: https://crates.io/crates/libsecp256k1 which is the one being used by the NEAR gateway but the build size when using it is much greater than the limit allowed. I ended up using `k256` because that's the one being used by cosmwasm itself: https://github.com/CosmWasm/cosmwasm/blob/0dae968db91734045239f45e982cddb1e5f11115/packages/crypto/Cargo.toml#L22.
There is another crate that seems promising but I haven't tried it yet: https://crates.io/crates/secp256k1 which is a wrapper of the C implementation used by bitcoin, not sure if there is an easy way to verify the implementation hasn't been modified in any way before being wrapped
I think we should keep it this way for the devnet deployment and circle back in another iteration. Thoughts?

contracts/command-batcher/src/encoding/evm.rs
```diff
@@ -0,0 +1,554 @@
...
+            threshold,
+            Token::Array(signatures),
+        ])
+        .into()
+    }
+}
+
+fn evm_address(pub_key: &[u8]) -> HexBinary {
+    let pub_key = PublicKey::from_sec1_bytes(pub_key)
+        .expect("violated invariant: pub_key is not a valid secp256k1 public key");
```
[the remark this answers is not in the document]
eguajardo: I wasn't sure if this should be an error or a panic, my reasoning was that if it gets to the point of executing this code (keygen, key rotation and signing already happened) and this fails, then it's most likely a bug in code, because the actual validation should have happened first time the public key was introduced (I need to add this validation on the next PR where it includes keygen for the batcher contract)

contracts/command-batcher/src/encoding/evm.rs
```diff
@@ -0,0 +1,554 @@
...
+
+    Keccak256::digest(message_id.as_bytes())
+        .as_slice()
+        .try_into()
+        .expect("violated invariant: Keccak256 length is not 32 bytes")
+}
+
+// TODO: This will make it incompatible with current version of destination chain gateways,
+// they rely on tx hash and event index as well, but just to emit events
+// https://github.com/axelarnetwork/axelar-cgp-solidity/blob/main/contracts/AxelarGateway.sol#L466
```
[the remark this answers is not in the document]
cjcobb23: Can we just add dummy fields to be safe? Set each of them to zero or something.

contracts/command-batcher/src/encoding/evm.rs
```diff
@@ -0,0 +1,557 @@
...
+    ethabi::encode(&[
+        Token::String(source_chain),
+        Token::String(source_address),
+        Token::Address(destination_address),
+        Token::FixedBytes(payload_hash.into()),
+    ])
+    .into()
+}
+
+fn batch_id(block_height: u64, data: &HexBinary) -> HexBinary {
```
cjcobb23, line 251: What do we need the block height?

contracts/command-batcher/src/encoding/evm.rs
```diff
@@ -0,0 +1,557 @@
...
+    id_hasher.update(data.as_slice());
+
+    id_hasher.finalize().as_slice().into()
+}
+
+fn msg_to_sign(data: &HexBinary) -> HexBinary {
+    let msg = Keccak256::digest(data.as_slice());
+
+    let unsigned = [
+        "\x19Ethereum Signed Message:\n32".as_bytes(), // Keccek256 hash length = 32
```
cjcobb23, line 264: Why do we add this prefix? Probably worth adding a comment.

contracts/command-batcher/src/encoding/evm.rs
```diff
@@ -0,0 +1,557 @@
...
+use super::traits;
+
+#[derive(Debug)]
+pub struct Message {
+    pub id: String,
+    pub source_address: String,
+    pub source_chain: String,
+    pub destination_address: ethereum_types::Address,
+    pub payload_hash: [u8; 32],
+}
```
cjcobb23, line 25: What do we need this type for?

contracts/command-batcher/src/encoding/traits.rs
```diff
@@ -0,0 +1,21 @@
...
+
+use axelar_wasm_std::Snapshot;
+use cosmwasm_std::{HexBinary, Uint256};
+use multisig::types::{PublicKey, Signature};
+
+use super::{evm::Message, Data};
+
+pub trait CommandBatch {
+    fn new(block_height: u64, messages: Vec<Message>, destination_chain_id: Uint256) -> Self;
+}
```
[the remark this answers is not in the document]
eguajardo: I changed how traits are defined and moved some structs out of the evm encoding, I think that should solve it

contracts/command-batcher/src/encoding/evm.rs
eguajardo: honestly I don't remember, but this is how core is creating the ID right now. I tried to make the output as similar to core as possible. Maybe @sammy1991106 can enlighten us?

contracts/command-batcher/src/encoding/evm.rs
eguajardo: its the standard for EVM signed data: https://eips.ethereum.org/EIPS/eip-191. I just added the comment

contracts/command-batcher/src/encoding/evm.rs
eguajardo: you are right it's not needed, it was used before as intermediary struct for creating validated input, but then I added the type `Command` and it's easier to just go directly from router `Message` to `Command` since it's the only place where is needed. Removed

contracts/command-batcher/src/encoding/evm.rs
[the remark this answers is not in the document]
eguajardo: hm... if the end goal is to adapt the destination gateway, I would prefer to just create a dummy gateway where it doesn't use those fields instead for the e2e test. But if we have to deconstruct the message ID into tx hash and event index for the voting verifier anyways, then I think we should probably just use those fields here as well and make it compliant with the current Gateway version. Thoughts?

contracts/command-batcher/src/encoding/evm.rs
cjcobb23: Unless there's a good reason, I'd rather remove it because I feel like it makes things confusing.

contracts/command-batcher/Cargo.toml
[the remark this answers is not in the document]
eguajardo: task to keep track of this: https://axelarnetwork.atlassian.net/browse/AXE-1366

contracts/command-batcher/src/encoding/evm.rs
```diff
@@ -0,0 +1,483 @@
...
+    }
+}
+
+pub struct Builder;
+impl traits::Builder for Builder {
+    fn build_batch(
+        messages: Vec<Message>,
+        destination_chain_id: Uint256,
+    ) -> Result<CommandBatch, ContractError> {
+        let message_ids: Vec<String> = messages.iter().map(|msg| msg.id.clone()).collect();
```
cjcobb23, line 49: We should probably sort these message ids, so that way the batch id is not affected by the order of message ids

contracts/command-batcher/src/encoding/evm.rs
```diff
@@ -0,0 +1,483 @@
...
+                })
+            })
+            .collect::<Result<Vec<Operator>, ContractError>>()?;
+        operators.sort_by(|a, b| a.address.cmp(&b.address));
+
+        Ok(Proof {
+            operators,
+            threshold: snapshot.quorum.into(),
+        })
+    }
```
cjcobb23, line 100: Can we somehow ensure that every `Proof` always has the operators and signatures sorted like this? Maybe make a constructor?

contracts/command-batcher/src/encoding/evm.rs
eguajardo: I did the sorting inside the batch_id function itself, I didn't want to do it from here, because it would also change the order of other stuff that also depends on message_ids, like the command vector, but I would assume that it should respect the order in which the messages were relayed

contracts/command-batcher/src/encoding/evm.rs
eguajardo: removed traits completely and added `new` constructors for types

contracts/command-batcher/src/encoding/evm.rs
```diff
@@ -0,0 +1,477 @@
...
+    error::ContractError,
+    types::{Command, CommandBatch, CommandType, Data, Operator, Proof},
+};
+
+impl TryFrom<Message> for Command {
+    type Error = ContractError;
+
+    fn try_from(msg: Message) -> Result<Self, Self::Error> {
+        Ok(Command {
+            command_type: CommandType::ApproveContractCall, // TODO: this would change when other command types are supported
```
sammy1991106, line 21:
```suggestion
            type: CommandType::ApproveContractCall, // TODO: this would change when other command types are supported
```

contracts/command-batcher/src/encoding/evm.rs
```diff
@@ -0,0 +1,477 @@
...
+            command_params: command_params(
```
sammy1991106, line 22:
```suggestion
            params: command_params(
```

contracts/command-batcher/src/encoding/evm.rs
```diff
@@ -0,0 +1,477 @@
...
+}
+
+fn batch_id(message_ids: &[String]) -> HexBinary {
+    let mut message_ids = message_ids
+        .iter()
+        .map(|id| id.as_bytes())
+        .collect::<Vec<&[u8]>>();
+    message_ids.sort();
+
+    Keccak256::digest(message_ids.concat()).as_slice().into()
```
sammy1991106, line 206: Use `.join` with a separator instead.

contracts/command-batcher/src/encoding/evm.rs
```diff
@@ -0,0 +1,477 @@
...
+        let encoded_data = data.encode();
+
+        let id = batch_id(&message_ids);
+        let msg_to_sign = msg_to_sign(&encoded_data);
+
+        Ok(Self {
+            id,
+            message_ids,
+            data,
+            msg_to_sign,
```
sammy1991106, line 64: Do we really need to store this as part of the `CommandBatch`? Can't we just add another method to it `signing_payload(&self)` to calculate it whenever needed?

contracts/command-batcher/src/encoding/evm.rs
```diff
@@ -0,0 +1,477 @@
...
+        })
+    }
+}
+
+impl Proof {
```
sammy1991106, line 69: Again, move this to where `Proof` is defined.

contracts/command-batcher/src/types.rs
```diff
@@ -1,19 +1,48 @@
...
+    pub command_params: HexBinary,
+}
+
+impl Display for CommandType {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            CommandType::ApproveContractCall => write!(f, "approveContractCall"),
+        }
+    }
+}
```
sammy1991106, line 25:
```suggestion
#[cw_serde]
pub enum CommandType {
    ApproveContractCall,
}

impl Display for CommandType {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        match self {
            CommandType::ApproveContractCall => write!(f, "approveContractCall"),
        }
    }
}

#[cw_serde]
pub struct Command {
    pub id: HexBinary,
    pub command_type: CommandType,
    pub command_params: HexBinary,
}
```

contracts/command-batcher/src/encoding/evm.rs
```diff
@@ -0,0 +1,477 @@
...
+                operator.address.as_slice(),
+            )));
+            weights.push(Token::Uint(ethereum_types::U256::from_big_endian(
+                &operator.weight.to_be_bytes(),
+            )));
+
+            if let Some(signature) = &operator.signature {
+                signatures.push(Token::Bytes(signature.into()));
+            }
+        });
```
sammy1991106, line 117: Use `.fold` instead.

contracts/command-batcher/src/encoding/evm.rs
```diff
@@ -0,0 +1,477 @@
...
+            msg_to_sign,
+        })
+    }
+}
+
+impl Proof {
+    pub fn new(
+        snapshot: Snapshot,
+        signers: HashMap<String, Signature>,
+        pub_keys: HashMap<String, multisig::types::PublicKey>,
```
sammy1991106, line 73: Add a method to `CommandBatch` to take these as parameters instead.

contracts/command-batcher/src/encoding/evm.rs
```diff
@@ -0,0 +1,477 @@
...
+    pub fn encode_execute_data(&self, data: &Data) -> HexBinary {
+        ethabi::encode(&[
+            Token::Bytes(data.encode().into()),
+            Token::Bytes(self.encode().into()),
+        ])
+        .into()
+    }
+}
+
+impl Data {
```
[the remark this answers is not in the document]
eguajardo: I added this here since here is where all logic related to evm encoding is. I ended up moving `Data` here instead since you also pointed out that destination_chain_id type may change depending of implementation. Thoughts?

contracts/command-batcher/src/encoding/evm.rs
eguajardo: this is here because I wanted to decouple whatever is specific to evm encoding from the rest of the code

contracts/command-batcher/src/encoding/evm.rs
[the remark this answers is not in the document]
sammy1991106: It's just awkward to have to navigate back-and-forth between files to see the logic of methods and the definition of the structure. I also don't think this `Data` structure and even the `CommandBatch` will be re-used by any other non-evm chain. Most if not all types defined in `types.rs` are evm-specific.

contracts/command-batcher/src/encoding/evm.rs
sammy1991106: `ty`

contracts/command-batcher/src/encoding/evm.rs
sammy1991106: As discussed yesterday, this structure should be deleted.

contracts/command-batcher/src/encoding/evm.rs
eguajardo: done

contracts/command-batcher/src/encoding/evm.rs
```diff
@@ -0,0 +1,467 @@
...
+        })?;
+    let pub_key = pub_key.to_encoded_point(false);
+
+    Ok(Keccak256::digest(&pub_key.as_bytes()[1..]).as_slice()[12..].into())
+}
+
+fn sorted_operators(signers: Vec<Signer>) -> Result<Vec<Operator>, ContractError> {
+    let mut operators = signers
+        .into_iter()
+        .map(|signer| {
```
sammy1991106, line 169: Implement `TryInto<Signer> for Operator`

contracts/command-batcher/src/encoding/evm.rs
```diff
@@ -0,0 +1,467 @@
...
+            Ok(Operator {
+                address: evm_address((&signer.pub_key).into())?,
+                weight: signer.weight,
+                signature: signer.signature,
+            })
+        })
+        .collect::<Result<Vec<Operator>, ContractError>>()?;
+    operators.sort_by(|a, b| a.address.cmp(&b.address));
```
sammy1991106, line 177: I'd suggest implementing `std::cmp::Ord` for `Operator` and then just use `.sort`.

contracts/command-batcher/src/encoding/evm.rs
```diff
@@ -0,0 +1,467 @@
...
+                )));
+                weights.push(Token::Uint(ethereum_types::U256::from_big_endian(
+                    &operator.weight.to_be_bytes(),
+                )));
+
+                if let Some(signature) = &operator.signature {
+                    signatures.push(Token::Bytes(signature.into()));
+                }
+
+                (addresses, weights, signatures)
```
sammy1991106, line 106: I don't think this is right based on the code at https://github.com/axelarnetwork/axelar-cgp-solidity/blob/9c7a260c848011f27d6e7ecb1cba88de79206ccc/contracts/auth/AxelarAuthWeighted.sol#L90. Should only add address and weight to their corresponding Vec if signature is present.

contracts/command-batcher/src/encoding/evm.rs
```diff
@@ -0,0 +1,467 @@
...
+        let unsigned = [
+            "\x19Ethereum Signed Message:\n32".as_bytes(), // Keccek256 hash length = 32
+            msg.as_slice(),
+        ]
+        .concat();
+
+        Keccak256::digest(unsigned).as_slice().into()
+    }
+
+    pub fn encode_execute_data(&self, multisig: Multisig) -> Result<HexBinary, ContractError> {
```
sammy1991106, line 81:
```suggestion
    pub fn encode_execute_data(&self, quorum: Uint256, signers: Vec<Signer>) -> Result<HexBinary, ContractError> {
```
cuz the multisig state is another thing that's awkward and irrelavant to this method.

sammy1991106 (changes requested):

contracts/command-batcher/src/encoding/evm.rs
cjcobb23: actually we need all of the addresses, but only a subset of signatures

contracts/command-batcher/src/encoding/evm.rs
```diff
@@ -0,0 +1,467 @@
...
+        Ok(ethabi::encode(&[
+            Token::Bytes(self.data.encode().into()),
+            Token::Bytes(self.encode_proof(multisig)?.into()),
+        ])
+        .into())
```
cjcobb23, line 86: I think we need the function selector here. We need the abi format for the gateway too.

cjcobb23 (changes requested):

contracts/command-batcher/src/encoding/evm.rs
sammy1991106: you are right

milapsheth:
Add unit tests to verify the data is encoded correctly. See the go code for reference
https://github.com/axelarnetwork/axelar-core/blob/main/x/evm/types/types_test.go#L95

Similarly, check if the commands get encoded correctly into `data`.

contracts/command-batcher/src/encoding/evm.rs
```diff
@@ -0,0 +1,588 @@
...
+        messages: Vec<Message>,
+        destination_chain_id: Uint256,
+    ) -> Result<Self, ContractError> {
+        let message_ids: Vec<String> = messages.iter().map(|msg| msg.id.clone()).collect();
+
+        let data = Data {
+            destination_chain_id,
+            commands: messages
+                .into_iter()
+                .map(|msg| msg.try_into())
```
sammy1991106, line 69:
```suggestion
                .map(TryInto::try_into)
```

sammy1991106 (approved):

cjcobb23 (approved):

PR #36, opened 2023-07-21
feat: refactor service registry

cjcobb23:
* add declare chain support message
* service registration and worker authorization require governance vote
* add bond message

AXE-1285


contracts/service-registry/src/contract.rs
```diff
@@ -100,60 +136,125 @@ pub mod execute {
...
+                            BondingState::Bonded { amount }
+                            | BondingState::RequestedUnbonding { amount }
+                            | BondingState::Unbonding {
+                                amount,
+                                unbonded_at: _,
+                            } => BondingState::Bonded {
+                                amount: amount + bond,
+                            },
+                            BondingState::Unbonded {} => BondingState::Bonded { amount: bond },
+                        };
```
sammy1991106, line 248: extract into a method of `BondingState`? `fn add_bond(self) -> Self` It can be easily unit-tested as well that way.

contracts/service-registry/src/contract.rs
```diff
@@ -163,126 +264,152 @@ pub mod execute {
...
+                Ok(BondingState::Unbonding {
+                    unbonded_at: env.block.time,
+                    amount,
+                })
+            }
+            BondingState::Bonded { amount } if !can_unbond => {
+                Ok(BondingState::RequestedUnbonding { amount })
+            }
+            _ => Err(ContractError::InvalidBondingState(worker.bonding_state)),
+        }?;
```
sammy1991106, line 317: Again, this should be extracted into a method of `BondingState` `fn unbond(self) -> Self` and be unit tested.

contracts/service-registry/src/contract.rs
```diff
@@ -163,126 +264,152 @@ pub mod execute {
...
+            BondingState::Bonded { amount } | BondingState::RequestedUnbonding { amount }
+                if can_unbond =>
+            {
+                Ok(BondingState::Unbonding {
+                    unbonded_at: env.block.time,
+                    amount,
+                })
+            }
+            BondingState::Bonded { amount } if !can_unbond => {
+                Ok(BondingState::RequestedUnbonding { amount })
```
sammy1991106, line 314: I don't understand why is there such a state `RequestedUnbonding`.

sammy1991106 (changes requested):

contracts/service-registry/src/contract.rs
cjcobb23: This is for when a worker wants to unbond, but they are still part of the snapshot being used to sign command batches. We can't let the unbonding period start until they are no longer in use, but we need some way of not including them in future snapshots. So a worker that moves to this state won't be included in future snapshots (returned via GetWorkers)

contracts/service-registry/src/state.rs
```diff
@@ -20,50 +30,234 @@ pub struct Service {
...
+        unbonded_at: Timestamp,
+    },
+    Unbonded,
 }
 
-pub struct WorkerIndexes<'a> {
-    pub service_name: MultiIndex<'a, String, Worker, &'a Addr>,
-}
+impl BondingState {
+    pub fn add_bond(self, bond: Uint128) -> Self {
```
sammy1991106, line 68: 👍 so much better and easier to be tested

sammy1991106 (approved):

contracts/service-registry/src/msg.rs
```diff
@@ -4,44 +4,58 @@ use cosmwasm_std::{Addr, Uint128};
...
 pub enum QueryMsg {
-    #[returns(ActiveWorkers)]
-    GetActiveWorkers { service_name: String },
-}
-
-#[cw_serde]
-pub struct ActiveWorkers {
-    pub workers: Vec<Worker>,
+    #[returns(Vec<Worker>)]
+    GetBondedWorkers {
```
haiyizxx, line 57: the new name is a bit confusing, the query returns bonded + authorized workers,

haiyizxx (approved):

PR #37, opened 2023-07-23
feat(voting-verifier): implement vote and tally function

haiyizxx:


packages/axelar-wasm-std/src/voting.rs
```diff
@@ -76,16 +160,201 @@ impl WeightedPoll {
...
+
+        if self.status != PollStatus::InProgress {
+            return Err(Error::PollNotInProgress {});
+        }
+
+        self.voted.insert(sender.clone());
+
+        for (tally, vote) in self.votes.iter_mut().zip(votes.iter()) {
+            if *vote {
+                *tally += self.snapshot.get_participant_weight(sender).unwrap();
```
cjcobb23, line 220: Just a thought, since we are looping through the tally for all of the messages here, we can calculate whether each message has achieved quorum and determine whether the poll is finished. Would probably be useful to do that and at least emit such in an event.

contracts/voting-verifier/src/execute.rs
```diff
@@ -88,16 +94,66 @@ pub fn verify_messages(
...
+        .zip(poll_result.results.iter())
+        .filter_map(|(message, verified)| match *verified {
+            true => Some(message),
+            false => None,
+        })
+        .collect::<Vec<&Message>>();
+
+    for message in messages {
+        VERIFIED_MESSAGES.save(deps.storage, &message.id, message)?;
+    }
```
cjcobb23, line 146: Do we need to check for message mismatch here? What if I create two polls at the same time, with same message ids but different messages?

packages/axelar-wasm-std/src/voting.rs
```diff
@@ -34,12 +64,68 @@ impl From<u64> for PollID {
     }
 }
 
+impl Mul for PollID {
+    type Output = Self;
+
+    fn mul(self, rhs: Self) -> Self::Output {
+        Self(self.0 * rhs.0)
+    }
+}
```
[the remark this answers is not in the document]
haiyizxx: required by `One` trait

contracts/voting-verifier/src/events.rs
```diff
@@ -119,7 +119,33 @@ impl fmt::Display for EvmMessage {
...
+    fn from(other: Voted) -> Self {
+        Event::new("voted")
+            .add_attribute("poll_id", other.poll_id)
+            .add_attribute("voter", other.voter)
+    }
+}
+
+pub struct PollEnded<'a> {
+    pub poll_id: PollID,
+    pub results: &'a Vec<bool>,
```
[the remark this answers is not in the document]
haiyizxx: the function returns results

contracts/voting-verifier/src/events.rs
```diff
@@ -31,12 +31,38 @@ impl From<PollStarted> for Event {
             .add_attribute("poll_id", other.poll_id)
             .add_attribute("source_gateway_address", other.source_gateway_address)
             .add_attribute("confirmation_height", other.confirmation_height.to_string())
```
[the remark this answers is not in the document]
haiyizxx: not universal atm, can make it universal by
a. maintain a chain -> param map
b. have chain params store in some other contract

contracts/voting-verifier/src/execute.rs
```diff
@@ -142,7 +141,9 @@ pub fn end_poll(deps: DepsMut, env: Env, poll_id: PollID) -> Result<Response, Co
         .collect::<Vec<&Message>>();
 
     for message in messages {
-        VERIFIED_MESSAGES.save(deps.storage, &message.id, message)?;
+        if !is_message_verified(deps.as_ref(), message)? {
+            VERIFIED_MESSAGES.save(deps.storage, &message.id, message)?;
+        }
```
cjcobb23, line 146: One message mismatch will abort the whole batch. I think it's ok though, if this actually happens we will need to intervene and migrate things.

packages/axelar-wasm-std/src/voting.rs
```diff
@@ -217,7 +217,10 @@ impl Poll for WeightedPoll {
 
         for (tally, vote) in self.votes.iter_mut().zip(votes.iter()) {
             if *vote {
-                *tally += self.snapshot.get_participant_weight(sender).unwrap();
+                *tally += self
```
cjcobb23, line 229:
```
        for (tally, vote) in self.votes.iter_mut().zip(votes.iter()) {
            if *vote {
                *tally += Uint256::from(participant.weight);
            }
        }
```

packages/axelar-wasm-std/src/voting.rs
```diff
@@ -191,7 +191,7 @@ impl Poll for WeightedPoll {
         &mut self,
         block_height: u64,
         sender: &Addr,
-        votes: &[bool],
+        votes: Vec<bool>,
     ) -> Result<PollStatus, Error> {
         if !self.snapshot.is_participant(sender) {
```
cjcobb23, line 205:
```
        let participant = self
            .snapshot
            .participants
            .get(&sender.to_string())
            .ok_or(Error::NotParticipant {})?;
```

contracts/voting-verifier/src/msg.rs
```diff
@@ -19,10 +22,10 @@ pub struct InstantiateMsg {
 pub enum ExecuteMsg {
     // Computes the results of a poll
     // For all verified messages, calls MessagesVerified on the verifier
-    EndPoll { poll_id: String },
+    EndPoll { poll_id: PollID },
```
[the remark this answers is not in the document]
haiyizxx: it's an integer

contracts/voting-verifier/src/execute.rs
```diff
@@ -88,16 +94,67 @@ pub fn verify_messages(
...
+
+    Ok(Response::new()
+        .add_event(
+            PollEnded {
+                poll_id: poll_result.poll_id,
+                results: &poll_result.results,
+            }
+            .into(),
+        )
+        .set_data(to_binary(&EndPollResponse { poll_result })?))
```
sammy1991106, line 157: I agree with Chris on that it's weird for the `PollEnded` event to take `results` in as a reference.

contracts/voting-verifier/src/execute.rs
```diff
@@ -144,11 +201,24 @@ fn create_poll(
...
     let id = POLL_ID.incr(store)?;
 
-    let poll = voting::WeightedPoll::new(id.into(), snapshot, block_height + expiry, poll_size);
+    let poll = WeightedPoll::new(id, snapshot, block_height + expiry, poll_size);
     POLLS.save(store, id, &poll)?;
 
     Ok(id)
 }
+
+fn clear_pending_message(
```
sammy1991106, line 213:
```suggestion
fn remove_pending_message(
```
or delete

contracts/voting-verifier/src/execute.rs
```diff
@@ -88,16 +94,67 @@ pub fn verify_messages(
...
+
+    let poll_result = poll.tally(env.block.height)?;
+    POLLS.save(deps.storage, poll_id, &poll)?;
+    let messages = clear_pending_message(deps.storage, poll_id)?;
+    if messages.len() != poll_result.results.len() {
+        panic!(
+            "poll {} results and pending messages have different length",
+            poll_id
+        )
+    }
```
sammy1991106, line 132:
```suggestion

    assert!(
        messages.len() == poll_result.results.len(),
        "poll {} results and pending messages have different length",
        poll_id
    );
```

packages/axelar-wasm-std/src/voting.rs
```diff
@@ -86,16 +169,208 @@ impl WeightedPoll {
...
+        &mut self,
+        block_height: u64,
+        sender: &Addr,
+        votes: Vec<bool>,
+    ) -> Result<PollStatus, Error> {
+        if !self.snapshot.is_participant(sender) {
+            return Err(Error::NotParticipant {});
+        }
+
+        if block_height > self.expires_at {
```
sammy1991106, line 209:
```suggestion
        if block_height >= self.expires_at {
```

packages/axelar-wasm-std/src/voting.rs
```diff
@@ -76,16 +153,201 @@ impl WeightedPoll {
...
+
+    fn cast_vote(
+        &mut self,
+        block_height: u64,
+        sender: &Addr,
+        votes: &[bool],
+    ) -> Result<PollStatus, Error> {
+        if !self.snapshot.is_participant(sender) {
+            return Err(Error::NotParticipant {});
+        }
```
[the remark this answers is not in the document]
sammy1991106: Agree with CJ, but I'd extract it into a method of `Snapshot` `.get_participant(&self)`.

packages/axelar-wasm-std/src/voting.rs
sammy1991106: Agree with CJ, and I'd also do something like 
```
        self.votes
            .iter_mut()
            .zip(votes.iter())
            .filter(|(_, vote)| **vote)
            .for_each(|(tally, _)| {
                *tally += self.snapshot.get_participant_weight(sender).expect("msg")
            });
```

packages/axelar-wasm-std/src/voting.rs
haiyizxx: I generally avoid access fields directly. participants can be private and encapsulated. only make fields public for necessary public interface. 

But we can agree on our standard practices.

packages/axelar-wasm-std/src/voting.rs
```diff
@@ -86,16 +169,208 @@ impl WeightedPoll {
...
+            voted: HashSet::new(),
         }
     }
 }
 
 impl Poll for WeightedPoll {
-    fn tally() -> Result<PollResult, Error> {
-        todo!()
+    fn tally(&mut self, block_height: u64) -> Result<PollResult, Error> {
+        if block_height < self.expires_at && self.voted.len() != self.snapshot.participants.len() {
```
sammy1991106, line 179: Why is there that second check?

sammy1991106 (changes requested):

contracts/voting-verifier/src/execute.rs
haiyizxx: sure, can clone

packages/axelar-wasm-std/src/voting.rs
haiyizxx: can tally early if all participants voted

packages/axelar-wasm-std/src/voting.rs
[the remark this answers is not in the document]
haiyizxx: makes sense

packages/axelar-wasm-std/src/voting.rs
sammy1991106: ah gotcha.. Can you change the `!=` to `<` instead then? I know it's the same, but confusing.

packages/axelar-wasm-std/src/voting.rs
haiyizxx: combined 💡  from two of you

cjcobb23 (approved):

packages/axelar-wasm-std/src/voting.rs
```diff
@@ -83,19 +164,216 @@ impl WeightedPoll {
...
+        if self.status != PollStatus::InProgress {
+            return Err(Error::PollNotInProgress {});
+        }
+
+        self.voted.insert(sender.clone());
+
+        self.votes
+            .iter_mut()
+            .zip(votes.iter())
+            .filter(|(_, vote)| **vote)
```
sammy1991106, line 232:
```suggestion
            .zip(votes.into_iter())
            .filter(|(_, vote)| *vote)
```

sammy1991106 (approved):

packages/axelar-wasm-std/src/voting.rs
```diff
@@ -76,16 +160,201 @@ impl WeightedPoll {
...
     }
 }
 
 impl Poll for WeightedPoll {
-    fn tally() -> Result<PollResult, Error> {
-        todo!()
+    fn tally(&mut self, block_height: u64) -> Result<PollResult, Error> {
+        if block_height < self.expires_at && self.voted.len() != self.snapshot.participants.len() {
+            return Err(Error::PollNotEnded {});
+        }
```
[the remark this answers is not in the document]
haiyizxx: we now have threshold for each message, so cannot just check the threshold for a poll.

PR #50, opened 2023-07-31
feat(multisig-prover): add contract entry points

eguajardo:
https://axelarnetwork.atlassian.net/browse/AXE-1295


contracts/command-batcher/src/msg.rs
```diff
@@ -1,13 +1,30 @@
...
 pub enum ExecuteMsg {
     // Start building a proof that includes specified messages
     // Queries the gateway for actual message contents
-    ConstructProof { message_ids: Vec<String> },
+    ConstructProof {
+        message_ids: Vec<String>,
+    },
+    KeyGen {
+        pub_keys: HashMap<String, HexBinary>,
+    },
```
cjcobb23, line 27: Can we name this `RotateSnapshot`, to be more compatible with the changes we have planned?

contracts/command-batcher/src/contract.rs
```diff
@@ -1,44 +1,343 @@
...
+    }
+
+    fn snapshot(
+        querier: QuerierWrapper,
+        block: BlockInfo,
+        config: &Config,
+    ) -> Result<Snapshot, ContractError> {
+        let query_msg = service_registry::msg::QueryMsg::GetActiveWorkers {
+            service_name: config.service_name.clone(),
+            chain_name: config.chain_name.to_string(),
```
cjcobb23, line 166: We need a way to distinguish the workers who are signing from the workers who are voting. Right now, they are the same, since both components use the same service name and chain name. I think they should keep using the same service, so that way their stake can be reused. But the chain name has to be different in some way. I sort of think we should append "-signing" to the chain name here (and maybe do the same for voting), so workers can signal whether they support signing and/or voting.

contracts/command-batcher/src/contract.rs
```diff
@@ -1,44 +1,343 @@
...
+
+        let snapshot = snapshot(deps.querier, env.block, &config)?;
+
+        for participant in snapshot.participants.keys() {
+            if !pub_keys.contains_key(participant) {
+                return Err(ContractError::PublicKeyNotFound {
+                    participant: participant.to_owned(),
+                });
+            }
+        }
```
cjcobb23, line 144: This checks that every snapshot participant has a matching public key. Should we also check that every public key has a matching participant in the snapshot?

contracts/command-batcher/src/contract.rs
```diff
@@ -1,44 +1,343 @@
...
+
+        let keygen_msg = multisig::msg::ExecuteMsg::KeyGen {
+            key_id: config.service_name,
```
cjcobb23, line 147: Can we just pass in the key_id as a string?

cjcobb23: And then store it with the snapshot?

contracts/command-batcher/src/contract.rs
```diff
@@ -1,44 +1,343 @@
...
+            snapshot,
+            pub_keys,
+        };
+
+        Ok(Response::new().add_message(WasmMsg::Execute {
+            contract_addr: config.multisig.into(),
+            msg: to_binary(&keygen_msg)?,
+            funds: vec![],
+        }))
```
cjcobb23, line 156: Need an event here

contracts/command-batcher/src/contract.rs
```diff
@@ -1,44 +1,343 @@
...
+    match parse_reply_execute_data(reply) {
+        Ok(MsgExecuteContractResponse { data: Some(data) }) => {
+            let command_batch_id = REPLY_ID_TO_BATCH.load(deps.storage, reply_id)?;
+
+            let session_id =
+                from_binary(&data).map_err(|_| ContractError::InvalidContractReply {
+                    reason: "invalid multisig session ID".to_string(),
+                })?;
+
+            let proof_id = ProofID::new(&command_batch_id, &session_id);
```
cjcobb23, line 218: Can we make the proof_id deterministic, so we don't have to process the reply? It could be a pair of `(key_id, msg_to_sign)`

cjcobb23: This would require changes to the multisig contract, since session_id is just a counter, so I think out of scope for this PR. But it would be nice for the future if the multisig contract derived session_id from the input (msg to sign, and key_id). Then we don't need to actually process the reply from the multisig contract here, and avoid the complexity associated with that.

In general, I'd like to see this whole batch creation and signing flow become idempotent.

contracts/command-batcher/src/state.rs
```diff
@@ -1 +1,27 @@
...
+    pub chain_name: ChainName,
+}
+
+pub const CONFIG: Item<Config> = Item::new("config");
+pub const COMMANDS_BATCH: Map<&BatchID, CommandBatch> = Map::new("command_batch");
+pub const PROOF_BATCH_MULTISIG: Map<&ProofID, (BatchID, Uint64)> =
+    Map::new("batch_multisig_session");
+
+pub const REPLY_ID_COUNTER: Item<u64> = Item::new("reply_id_counter");
+pub const REPLY_ID_TO_BATCH: Map<u64, BatchID> = Map::new("reply_id_to_batch");
```
cjcobb23, line 27: Can this just be an `Item`? We really only ever need to store one of these at a time. Once the reply is processed, we can reuse the Item again for the next call. We don't need to store an entry in this map for every submessage reply that has ever occurred. It's a transient thing, and can be deleted after we process the reply.

contracts/command-batcher/src/state.rs
```diff
@@ -1 +1,27 @@
...
+#[cw_serde]
+pub struct Config {
+    pub admin: Addr,
+    pub gateway: Addr,
+    pub multisig: Addr,
+    pub service_registry: Addr,
+    pub destination_chain_id: Uint256,
+    pub signing_threshold: Threshold,
+    pub service_name: String,
+    pub chain_name: ChainName,
```
cjcobb23, line 18: I don't think we can use `ChainName` here, since that's a type that's meant to be validated when converting from string.

contracts/command-batcher/src/msg.rs
```diff
@@ -1,13 +1,32 @@
...
+use crate::{encoding::Data, types::ProofID};
 
-use crate::types::{Data, Proof};
+#[cw_serde]
+pub struct InstantiateMsg {
+    pub gateway_address: String,
+    pub multisig_address: String,
+    pub service_registry_address: String,
+    pub destination_chain_id: Uint256,
+    pub signing_threshold: (Uint64, Uint64),
```
cjcobb23, line 15: There is a `Threshold` type in axelar-wasm-std, should use that

contracts/command-batcher/src/contract.rs
cjcobb23: Actually, I think it's fine to just have the same validator set for voting and signing for a particular chain. Makes the security easier to reason about, since there's one set of workers securing the connection, regardless of which way messages are flowing.

contracts/command-batcher/src/contract.rs
```diff
@@ -1,44 +1,350 @@
...
-pub fn execute(
-    _deps: DepsMut,
+pub fn instantiate(
+    deps: DepsMut,
     _env: Env,
-    _info: MessageInfo,
+    info: MessageInfo,
+    msg: InstantiateMsg,
+) -> Result<Response, ContractError> {
+    let admin = deps.api.addr_validate(info.sender.as_str())?;
```
cjcobb23, line 35: admin should be a parameter in the message, so the sender can specify a different admin than themselves

contracts/command-batcher/src/contract.rs
eguajardo: Agree this should be out of scope for now. Before we implement deterministic id for the multisig we should implement deletion of failed sessions (completed too?), otherwise I think we won't be able to retry multisig for same batch

contracts/command-batcher/src/state.rs
eguajardo: good point, fixed it

contracts/command-batcher/src/state.rs
eguajardo: in that case I think it's ok `ChainName` here since this is the state struct, instead I changed it to String in the Instance msg and validated during instantiation before storing it in the config

contracts/command-batcher/src/state.rs
```diff
@@ -1 +1,27 @@
...
+use cw_storage_plus::{Item, Map};
 
+use crate::types::{BatchID, CommandBatch, ProofID};
+
+#[cw_serde]
+pub struct Config {
+    pub admin: Addr,
+    pub gateway: Addr,
+    pub multisig: Addr,
+    pub service_registry: Addr,
```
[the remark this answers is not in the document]
eguajardo: these are all strings in the instantiation msg, they are validated before being stored in the config

contracts/command-batcher/src/msg.rs
eguajardo: this is the unvalidated message, the Threshold type is used after validation to store it in the config

contracts/command-batcher/src/contract.rs
eguajardo: added the key_id param, but why do we need this in the snapshot itself? snapshot in the current state is also used for voting which doesn't need keys

contracts/command-batcher/src/contract.rs
cjcobb23: I think we should add still add some event here, especially because these contracts are all changing quite a bit. I don't think we need the owner, but the other attributes are good.

contracts/command-batcher/src/contract.rs
cjcobb23: I think it's a good sanity check. I think the caller would want to know if one of the public keys doesn't match a participant in the snapshot, which would cause the public key to not be used.

contracts/command-batcher/src/msg.rs
cjcobb23: ah I see. Can you try just adding this to the threshold type:
```
#[serde(try_from="(Uint64, Uint64)")]
pub struct Threshold {
```
https://serde.rs/container-attrs.html#try_from

contracts/command-batcher/src/contract.rs
cjcobb23: I don't think we need to delete completed sessions, but yes we should have some way of restarting a failed session

contracts/command-batcher/src/msg.rs
eguajardo: this worked, here is the commit: https://github.com/axelarnetwork/axelar-amplifier/pull/50/commits/1c7491909b20f13e1a08c5651bc9130fa3f03f62. I had to add `into` as well to match both serialization and deserialization otherwise it fails. With this change it will mean that whenever we serialize, the JSON output will be a tuple instead of the Threshold struct, not sure if that's fine for whoever will read the serialized JSON. Thoughts?

contracts/command-batcher/src/contract.rs
```diff
@@ -1,44 +1,481 @@
...
+        QueryMsg::GetProof { proof_id } => to_binary(&query::get_proof(deps, proof_id)?),
     }
 }
 
 pub mod query {
     use super::*;
 
-    pub fn get_proof(_proof_id: String) -> StdResult<GetProofResponse> {
-        todo!()
+    pub fn get_proof(deps: Deps, proof_id: String) -> StdResult<GetProofResponse> {
```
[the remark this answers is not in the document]
eguajardo: I think we first need to implement deterministic session_id, no? if we link batch->session_id there could be multiple entries of the same batch pointing to different session_ids, how would the contract know which one to get from the query? I don't think we can assume it's the last one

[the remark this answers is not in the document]
cjcobb23: Hmm I see, we could include an optional session_id value in the argument, and if it's not filled, just return a list of all matching proofs. I think it's unlikely that there will be two different sessions for the same batch, and it would make it more convenient if relayers don't have to keep track of proof id.

contracts/command-batcher/src/contract.rs
eguajardo: task to remove proof_id: https://axelarnetwork.atlassian.net/browse/AXE-1514

contracts/command-batcher/src/contract.rs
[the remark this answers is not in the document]
eguajardo: task to remove proof_id: https://axelarnetwork.atlassian.net/browse/AXE-1514
task to make session_id deterministic: https://axelarnetwork.atlassian.net/browse/AXE-1513

contracts/multisig-prover/src/contract.rs
```diff
@@ -1,44 +1,479 @@
...
+        let participants = active_workers
+            .into_iter()
+            .map(Worker::try_into)
+            .collect::<Result<Vec<Participant>, service_registry::ContractError>>()
+            .map_err(
+                |err: service_registry::ContractError| ContractError::InvalidParticipants {
+                    reason: err.to_string(),
+                },
+            )?
+            .try_into()
```
cgorenflo, line 192: what does this convert into?

contracts/multisig-prover/src/contract.rs
```diff
@@ -1,44 +1,479 @@
...
+            .map_err(|err: axelar_wasm_std::nonempty::Error| {
+                ContractError::InvalidParticipants {
+                    reason: err.to_string(),
+                }
+            })?;
```
cgorenflo, line 197: `change_context`

contracts/multisig-prover/src/contract.rs
```diff
@@ -1,44 +1,479 @@
...
+
+        Ok(Snapshot::new(
+            block
+                .time
+                .try_into()
+                .expect("violated invariant: block time cannot be invalid"),
+            block
+                .height
+                .try_into()
+                .expect("violated invariant: block height cannot be invalid"),
```
cgorenflo, line 207:
```suggestion
                .expect("violated invariant: block height cannot be zero"),
```

contracts/multisig-prover/src/contract.rs
```diff
@@ -1,44 +1,479 @@
...
+                ContractError::InvalidParticipants {
+                    reason: err.to_string(),
+                }
+            })?;
+
+        Ok(Snapshot::new(
+            block
+                .time
+                .try_into()
+                .expect("violated invariant: block time cannot be invalid"),
```
cgorenflo, line 203:
```suggestion
                .expect("violated invariant: block time cannot be zero"),
```

contracts/multisig-prover/src/contract.rs
```diff
@@ -1,44 +1,479 @@
...
+        }))?;
+
+        let status = match multisig.state {
+            MultisigState::Pending => ProofStatus::Pending,
+            MultisigState::Completed => {
+                let execute_data = batch
+                    .encode_execute_data(multisig.quorum, multisig.signers)
+                    .map_err(|e| {
+                        StdError::generic_err(format!("failed to encode execute data: {}", e))
+                    })?;
```
cgorenflo, line 272: in general, don't create hierarchical errors manually. That's exactly what error_stack is trying to solve. Create an error for the specific inner reason, then change the context to the more general error. Also, try to use generic errors only to convert external errors and then go with typed errors from there

contracts/multisig-prover/src/encoding/evm.rs
```diff
@@ -55,6 +55,18 @@ impl TryFrom<Signer> for Operator {
...
+impl BatchID {
+    pub fn new(message_ids: &[String]) -> BatchID {
+        let mut message_ids = message_ids
+            .iter()
+            .map(|id| id.as_bytes())
+            .collect::<Vec<&[u8]>>();
+        message_ids.sort();
+
+        Keccak256::digest(message_ids.concat()).as_slice().into()
+    }
```
cgorenflo, line 67: point of discussion: message ids are in a sense user input. Is it possible to create two batches with the same id but different messages?

contracts/multisig-prover/src/state.rs
```diff
@@ -1 +1,27 @@
...
+    pub chain_name: ChainName,
+}
+
+pub const CONFIG: Item<Config> = Item::new("config");
+pub const KEY_ID: Item<String> = Item::new("key_id");
+pub const COMMANDS_BATCH: Map<&BatchID, CommandBatch> = Map::new("command_batch");
+pub const PROOF_BATCH_MULTISIG: Map<&ProofID, (BatchID, Uint64)> =
+    Map::new("batch_multisig_session");
+
+pub const REPLY_BATCH: Item<BatchID> = Item::new("reply_tracker");
```
cgorenflo, line 27: clever!

contracts/multisig-prover/src/encoding/evm.rs
```diff
@@ -55,6 +55,18 @@ impl TryFrom<Signer> for Operator {
     }
 }
 
+impl BatchID {
```
cgorenflo, line 58: why is this defined here, when the rest of BatchID is somewhere completely different

cgorenflo (changes requested):
please cut your PRs into smaller pieces in the future. This is way too big and complex to review it with confidence

contracts/multisig-prover/src/encoding/evm.rs
```diff
@@ -55,6 +55,18 @@ impl TryFrom<Signer> for Operator {
...
+    pub fn new(message_ids: &[String]) -> BatchID {
+        let mut message_ids = message_ids
+            .iter()
+            .map(|id| id.as_bytes())
+            .collect::<Vec<&[u8]>>();
+        message_ids.sort();
+
+        Keccak256::digest(message_ids.concat()).as_slice().into()
```
sammy1991106, line 66: Should join with a delimiter insatead of simply concat together.

contracts/multisig-prover/src/contract.rs
```diff
@@ -1,44 +1,479 @@
...
+    state::{Config, COMMANDS_BATCH, CONFIG, KEY_ID, PROOF_BATCH_MULTISIG, REPLY_BATCH},
+    types::{BatchID, CommandBatch, ProofID},
 };
 
+const START_MULTISIG_REPLY_ID: u64 = 1;
+
 #[cfg_attr(not(feature = "library"), entry_point)]
-pub fn execute(
-    _deps: DepsMut,
+pub fn instantiate(
```
sammy1991106, line 29: It's too hard to review it when `instantiate` and all execute methods are done in one single PR. Please split in the future. One method at a time.

contracts/multisig-prover/src/contract.rs
```diff
@@ -1,44 +1,479 @@
...
+        multisig,
+        service_registry,
+        destination_chain_id: msg.destination_chain_id,
+        signing_threshold: msg.signing_threshold,
+        service_name: msg.service_name,
+        chain_name: ChainName::from_str(&msg.chain_name).map_err(
+            |e: connection_router::ContractError| ContractError::InvalidInput {
+                reason: e.to_string(),
+            },
+        )?,
```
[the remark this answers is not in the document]
sammy1991106: Agree with @cgorenflo. Also, I am not a fan of errors containing a reason string. It's too generic.

contracts/multisig-prover/src/contract.rs
```diff
@@ -1,44 +1,479 @@
...
+        message_ids: Vec<String>,
+    ) -> Result<Response, ContractError> {
+        let key_id = KEY_ID.load(deps.storage)?;
+
+        let config = CONFIG.load(deps.storage)?;
+
+        let batch_id = BatchID::new(&message_ids);
+
+        let query = gateway::msg::QueryMsg::GetMessages { message_ids };
+        let messages: Vec<Message> = deps.querier.query(&QueryRequest::Wasm(WasmQuery::Smart {
```
sammy1991106, line 89: What does the gateway do in the query if it cannot find some of the given message ids?

contracts/multisig-prover/src/contract.rs
```diff
@@ -1,44 +1,479 @@
...
+            config.signing_threshold,
+            participants,
+        ))
     }
 }
 
 #[cfg_attr(not(feature = "library"), entry_point)]
-pub fn query(_deps: Deps, _env: Env, msg: QueryMsg) -> StdResult<Binary> {
+pub fn reply(deps: DepsMut, _env: Env, reply: Reply) -> Result<Response, ContractError> {
+    match parse_reply_execute_data(reply) {
```
sammy1991106, line 216: Should match the `reply.id` instead to see it it equals `START_MULTISIG_REPLY_ID`

contracts/multisig-prover/src/contract.rs
```diff
@@ -1,44 +1,479 @@
...
+            let session_id =
+                from_binary(&data).map_err(|_| ContractError::InvalidContractReply {
+                    reason: "invalid multisig session ID".to_string(),
+                })?;
+
+            let proof_id = ProofID::new(&command_batch_id, &session_id);
+
+            PROOF_BATCH_MULTISIG.save(deps.storage, &proof_id, &(command_batch_id, session_id))?;
+
+            Ok(Response::new().add_event(Event::ProofUnderConstruction { proof_id }.into()))
```
sammy1991106, line 229: extract into a method or mod like `execute` and `query`.

contracts/multisig-prover/src/contract.rs
sammy1991106, line 1: The entire file has almost 500 lines. We should consider extracting the `execute` and `query` mods into their own files. @cgorenflo @cjcobb23 thought?

contracts/multisig-prover/src/contract.rs
sammy1991106: same question

contracts/multisig-prover/src/contract.rs
```diff
@@ -1,44 +1,479 @@
...
+            )?
+            .try_into()
+            .map_err(|err: axelar_wasm_std::nonempty::Error| {
+                ContractError::InvalidParticipants {
+                    reason: err.to_string(),
+                }
+            })?;
+
+        Ok(Snapshot::new(
+            block
```
sammy1991106, line 200: The `timestamp` and `height` in core represent when the snapshot is created in core. Looking at this usage here, I am quite confused for why we need them anymore.

contracts/multisig-prover/src/error.rs
```diff
@@ -6,9 +6,27 @@ pub enum ContractError {
     #[error("{0}")]
     Std(#[from] StdError),
 
+    #[error("caller is not authorized")]
+    Unauthorized {},
```
sammy1991106, line 10:
```suggestion
    Unauthorized,
```

contracts/multisig-prover/src/error.rs
```diff
@@ -6,9 +6,27 @@ pub enum ContractError {
...
+    InvalidPublicKey { reason: String },
+
+    #[error("invalid input: {reason}")]
+    InvalidInput { reason: String },
+
+    #[error("invalid participants: {reason}")]
+    InvalidParticipants { reason: String },
+
+    #[error("no messages found")]
+    NoMessagesFound {},
```
sammy1991106, line 25:
```suggestion
    NoMessagesFound,
```

contracts/multisig-prover/src/error.rs
```diff
@@ -6,9 +6,27 @@ pub enum ContractError {
...
+
+    #[error("invalid contract reply: {reason}")]
+    InvalidContractReply { reason: String },
+
+    #[error("wrong chain")]
+    WrongChain {},
```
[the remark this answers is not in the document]
sammy1991106:
```suggestion
    #[error("caller is not authorized")]
    Unauthorized,

    #[error("message is invalid: {reason}")]
    InvalidMessage { reason: String },

    #[error("public key is invalid: {reason}")]
    InvalidPublicKey { reason: String },

    #[error("invalid input: {reason}")]
    InvalidInput { reason: String },

    #[error("invalid participants: {reason}")]
    InvalidParticipants { reason: String },

    #[error("no messages found")]
    NoMessagesFound {},

    #[error("invalid contract reply: {reason}")]
    InvalidContractReply { reason: String },

    #[error("wrong chain")]
    WrongChain {},
```

sammy1991106, line 31:
```suggestion
    WrongChain,
```

contracts/multisig-prover/src/events.rs
```diff
@@ -1,16 +1,44 @@
...
+    },
 }
 
 impl From<Event> for cosmwasm_std::Event {
     fn from(other: Event) -> Self {
         match other {
             Event::ProofUnderConstruction { proof_id } => {
                 cosmwasm_std::Event::new("proof_under_construction")
-                    .add_attribute("proof_id", proof_id.to_hex())
+                    .add_attribute("proof_id", proof_id.to_string())
```
sammy1991106, line 25:
```suggestion
                    .add_attribute("proof_id", to_string(proof_id).expect("..."))
```

contracts/multisig-prover/src/types.rs
```diff
@@ -40,3 +82,54 @@ pub struct Operator {
...
+    fn from(id: &[u8]) -> Self {
+        Self(id.into())
+    }
+}
+
+impl ProofID {
+    pub fn new(batch_id: &BatchID, session_id: &Uint64) -> Self {
+        let mut hasher = Keccak256::new();
+
+        hasher.update(Into::<&[u8]>::into(batch_id));
```
sammy1991106, line 130: Do we need to separator between batch_id and session_id? Asking cuz I'm not sure what exactly does `hasher.update` do.

contracts/multisig/src/error.rs
```diff
@@ -39,6 +39,9 @@ pub enum ContractError {
     #[error("Key ID {key_id:?} already exists")]
     DuplicateKeyID { key_id: String },
 
+    #[error("number of participants does not match number of public keys")]
+    PublicKeysMismatchParticipants {},
```
sammy1991106, line 43:
```suggestion
    PublicKeysMismatchParticipants,
```

sammy1991106 (changes requested):

contracts/multisig-prover/src/contract.rs
```diff
@@ -1,44 +1,479 @@
...
+                pub_keys: pub_keys.clone(),
+            })?,
+            funds: vec![],
+        };
+
+        let event = Event::SnapshotRotated {
+            key_id,
+            snapshot,
+            pub_keys,
+        };
```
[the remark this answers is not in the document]
eguajardo: in any case, the message sent to multisig here is a normal message not a submessage, so the whole thing reverts if any fails

contracts/multisig-prover/src/contract.rs
eguajardo: from `Vec` to `axelar_wasm_std::nonempty::Vec`

contracts/multisig-prover/src/contract.rs
eguajardo: it used to be `zero` before, but I changed it because from the contract perspective `axelar_wasm_std` is an external library and it shouldn't assume that the error resulting from `try_into` will always be because it's zero

contracts/multisig-prover/src/encoding/evm.rs
eguajardo: if the content of the batches is different, the workers would have rejected one set of messages or the other during voting

contracts/multisig-prover/src/encoding/evm.rs
eguajardo: somehow I messed this up when I merged conflicts from previous PR and kept the old way of doing it

contracts/multisig-prover/src/contract.rs
cgorenflo: what else could it be?

contracts/multisig-prover/src/contract.rs
eguajardo: as it is right now, it will fail

contracts/multisig-prover/src/contract.rs
```diff
@@ -1,44 +1,479 @@
...
+            message_ids: batch.message_ids,
+            data: batch.data,
+            status,
+        })
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use anyhow::Error;
```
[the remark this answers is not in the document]
eguajardo: moved to https://axelarnetwork.atlassian.net/browse/AXE-1613

contracts/multisig-prover/src/contract.rs
eguajardo: moved to https://axelarnetwork.atlassian.net/browse/AXE-1613

contracts/multisig-prover/src/contract.rs
eguajardo: moved to https://axelarnetwork.atlassian.net/browse/AXE-1613

contracts/multisig-prover/src/error.rs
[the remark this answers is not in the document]
eguajardo: moved to https://axelarnetwork.atlassian.net/browse/AXE-1613

contracts/multisig-prover/src/contract.rs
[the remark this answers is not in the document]
eguajardo: moved to https://axelarnetwork.atlassian.net/browse/AXE-1613

contracts/multisig-prover/src/events.rs
```diff
@@ -1,16 +1,46 @@
...
-    ProofUnderConstruction { proof_id: HexBinary },
+    ProofUnderConstruction {
+        proof_id: ProofID,
+    },
+    SnapshotRotated {
+        key_id: String,
+        snapshot: Snapshot,
+        pub_keys: HashMap<String, HexBinary>,
+    },
 }
```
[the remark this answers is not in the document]
cjcobb23: We should address this in a later PR though.

contracts/multisig-prover/src/events.rs
```diff
@@ -20,12 +18,12 @@ pub enum Event {
...
-            Event::ProofUnderConstruction { proof_id } => {
-                cosmwasm_std::Event::new("proof_under_construction").add_attribute(
-                    "proof_id",
-                    to_string(&proof_id).expect("violated invariant: proof_id is not serializable"),
-                )
-            }
+            Event::ProofUnderConstruction {
+                multisig_session_id: proof_id,
+            } => cosmwasm_std::Event::new("proof_under_construction").add_attribute(
+                "proof_id",
```
cjcobb23, line 24: we should name the attribute "multisig_session_id" right?

eguajardo: lol yes

contracts/multisig-prover/src/events.rs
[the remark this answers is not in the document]
eguajardo: https://axelarnetwork.atlassian.net/browse/AXE-1624

contracts/multisig-prover/src/types.rs
```diff
@@ -26,9 +28,58 @@ pub struct Command {
...
+    type Output = BatchID;
+
+    fn from_vec(value: Vec<u8>) -> StdResult<Self::Output> {
+        Ok(from_binary(&value.into()).expect("violated invariant: BatchID is not deserializable"))
+    }
+}
+
+impl BatchID {
+    pub fn new(message_ids: &[String]) -> BatchID {
+        let mut message_ids = message_ids.to_vec();
```
sammy1991106, line 73:
```suggestion
    pub fn new(message_ids: impl AsRef<[String]>) -> BatchID {
    let mut message_ids = message_ids.as_ref().to_vec();
```

contracts/multisig/src/error.rs
cgorenflo: not resolved

cgorenflo (approved):

contracts/multisig-prover/src/execute.rs
```diff
@@ -0,0 +1,161 @@
...
+    }))?;
+
+    if messages.len() != length {
+        return Err(ContractError::MessagesCountMismatch);
+    }
+
+    let chain_name = Some(chain_name);
+    if messages
+        .iter()
+        .any(|msg| ChainName::from_str(&msg.destination_chain).ok() != chain_name)
```
sammy1991106, line 78:
```suggestion
    if messages
        .iter()
        .any(|msg| ChainName::from_str(&msg.destination_chain).expect("violated invariant: message with invalid chain found") != chain_name)
```

contracts/multisig-prover/src/execute.rs
```diff
@@ -0,0 +1,161 @@
...
+
+    let wasm_msg = wasm_execute(config.multisig, &start_sig_msg, vec![])?;
+
+    Ok(Response::new().add_submessage(SubMsg::reply_on_success(wasm_msg, START_MULTISIG_REPLY_ID)))
+}
+
+fn get_messages(
+    querier: QuerierWrapper,
+    message_ids: Vec<String>,
+    gateway: String,
```
sammy1991106, line 60:
```suggestion
    gateway: Addr,
```

contracts/multisig-prover/src/reply.rs
```diff
@@ -0,0 +1,40 @@
...
+                })?;
+
+            MULTISIG_SESSION_BATCH.save(
+                deps.storage,
+                multisig_session_id.u64(),
+                &command_batch_id,
+            )?;
+
+            Ok(Response::new().add_event(
+                Event::ProofUnderConstruction {
```
sammy1991106, line 27: Add `command_batch_id` to the event `ProofUnderConstruction` as well.

contracts/multisig-prover/src/execute.rs
```diff
@@ -0,0 +1,161 @@
...
+
+    let active_workers: Vec<Worker> = querier.query(&QueryRequest::Wasm(WasmQuery::Smart {
+        contract_addr: config.service_registry.to_string(),
+        msg: to_binary(&query_msg)?,
+    }))?;
+
+    let participants = active_workers
+        .into_iter()
+        .map(Worker::try_into)
+        .collect::<Result<Vec<Participant>, service_registry::ContractError>>()
```
sammy1991106, line 136:
```suggestion
        .collect::<Result<Vec<Participant>, _>>()
```

contracts/multisig-prover/src/lib.rs
```diff
@@ -2,7 +2,10 @@ pub mod contract;
 pub mod encoding;
 pub mod error;
 pub mod events;
+pub mod execute;
 pub mod msg;
+pub mod query;
+pub mod reply;
```
sammy1991106, line 8: Why do these mods need to be public?

contracts/multisig-prover/src/types.rs
```diff
@@ -26,9 +28,58 @@ pub struct Command {
...
+    fn from(id: &[u8]) -> Self {
+        Self(id.into())
+    }
+}
+
+impl<'a> From<&'a BatchID> for &'a [u8] {
+    fn from(id: &'a BatchID) -> Self {
+        id.0.as_slice()
+    }
+}
```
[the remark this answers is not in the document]
sammy1991106: Actually, why is needed to get a reference of `[u8]` from the `BatchID`?

contracts/multisig-prover/src/execute.rs
```diff
@@ -0,0 +1,161 @@
...
+) -> Result<Vec<Message>, ContractError> {
+    let length = message_ids.len();
+
+    let query = gateway::msg::QueryMsg::GetMessages { message_ids };
+    let messages: Vec<Message> = querier.query(&QueryRequest::Wasm(WasmQuery::Smart {
+        contract_addr: gateway,
+        msg: to_binary(&query)?,
+    }))?;
+
+    if messages.len() != length {
```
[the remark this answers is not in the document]
eguajardo: see @cjcobb23 comment here: https://github.com/axelarnetwork/axelar-amplifier/pull/50#discussion_r1303070842. Maybe we should change it to panic, if some time in the future we change the gateway query behavior to something else, we should probably review the prover as well to decide if something needs to be changed as well. A panic will ensure we pay attention to this part.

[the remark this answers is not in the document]
sammy1991106: Yes, then it should definitely panic IMO. Also, if you don't want to rely on any semantics of the gateway, then the code here should also check if `messages` are all distinct and contain all the message IDs given. @cjcobb23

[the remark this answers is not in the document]
cjcobb23: Eh, I think we can assume the gateway is not going to return duplicates or random messages that you didn't ask for. I think those would be legitimate bugs in the gateway. It doesn't really hurt to add checks for that, but I don't feel its totally necessary.
 
We definitely shouldn't assume order is the same, but I'm not sure if the code currently relies on the order actually.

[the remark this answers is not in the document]
eguajardo: from my point of view, the order is only relevant for deriving the batch ID, but the prover doesn't rely on the order of messages received, it sorts them itself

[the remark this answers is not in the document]
sammy1991106: I'm approving the PR, but I don't like us making assumptions about something and being extra cautious about others. This to me, is always a failure of the API design. Either it should just be get_message so that we no longer worry about it all together, or get_messages should return Vec<Option<Message>>.

sammy1991106 (approved):

PR #51, opened 2023-08-02
feat(voting-verifier): is_verified query

haiyizxx:


contracts/voting-verifier/src/msg.rs
```diff
@@ -43,6 +43,9 @@ pub struct Poll {
 pub enum QueryMsg {
     #[returns(Poll)]
     GetPoll { poll_id: PollID },
+
+    #[returns(Vec<(String, bool)>)]
```
cjcobb23, line 47: We should probably define this type somewhere since it's used in so many contracts, but not for this PR

contracts/voting-verifier/src/contract.rs
```diff
@@ -64,6 +66,20 @@ pub fn execute(
...
-pub fn query(_deps: Deps, _env: Env, _msg: QueryMsg) -> StdResult<Binary> {
-    todo!()
+pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> StdResult<Binary> {
+    match msg {
+        QueryMsg::IsVerified { messages } => {
+            let messages = messages
+                .into_iter()
+                .map(state::Message::try_from)
+                .collect::<Result<Vec<state::Message>, connection_router::ContractError>>()
+                .map_err(|e| StdError::generic_err(e.to_string()))?;
```
cjcobb23, line 76: If we add:
```
impl From<ContractError> for StdError {
    fn from(value: ContractError) -> Self {
        Self::generic_err(value.to_string())
    }
}
```
to connection_router/src/error.rs, we can get rid of this `map_err`

contracts/voting-verifier/src/error.rs
```diff
@@ -38,3 +38,9 @@ pub enum ContractError {
     #[error("{0}")]
     VoteError(#[from] voting::Error),
 }
+
+impl From<ContractError> for StdError {
+    fn from(value: ContractError) -> Self {
+        Self::generic_err(value.to_string())
+    }
+}
```
cjcobb23, line 46: Can we add this to connection_router errors too?

cjcobb23 (approved):

contracts/voting-verifier/src/msg.rs
haiyizxx: yes, thinking the same

contracts/voting-verifier/src/execute.rs
```diff
@@ -191,7 +191,7 @@ fn take_snapshot(
     ))
 }
 
-fn is_message_verified(deps: Deps, message: &Message) -> Result<bool, ContractError> {
+pub fn is_message_verified(deps: Deps, message: &Message) -> Result<bool, ContractError> {
```
sammy1991106, line 194: I'd suggest moving this method to the query mod.

contracts/voting-verifier/src/query.rs
```diff
@@ -0,0 +1,16 @@
...
+use crate::error::ContractError;
+use crate::execute::is_message_verified;
+
+pub fn verification_statuses(deps: Deps, messages: Vec<Message>) -> StdResult<Vec<(String, bool)>> {
+    messages
+        .into_iter()
+        .map(|message| {
+            is_message_verified(deps, &message).map(|verified| (message.id.to_string(), verified))
+        })
+        .collect::<Result<Vec<(String, bool)>, ContractError>>()
```
sammy1991106, line 14:
```suggestion
        .collect::<Result<Vec<_>, _>>()
```

contracts/voting-verifier/src/contract.rs
```diff
@@ -64,6 +64,19 @@ pub fn execute(
...
 #[cfg_attr(not(feature = "library"), entry_point)]
-pub fn query(_deps: Deps, _env: Env, _msg: QueryMsg) -> StdResult<Binary> {
-    todo!()
+pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> StdResult<Binary> {
+    match msg {
+        QueryMsg::IsVerified { messages } => {
+            let messages = messages
+                .into_iter()
+                .map(state::Message::try_from)
+                .collect::<Result<Vec<state::Message>, _>>()?;
```
sammy1991106, line 73:
```suggestion
                .collect::<Result<Vec<_>, _>>()?;
```

contracts/voting-verifier/tests/tests.rs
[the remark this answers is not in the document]
haiyizxx: ya, workflow updated

contracts/voting-verifier/src/query.rs
```diff
@@ -0,0 +1,26 @@
...
+
+pub fn is_message_verified(deps: Deps, message: &Message) -> Result<bool, ContractError> {
+    match VERIFIED_MESSAGES.may_load(deps.storage, &message.id)? {
+        Some(stored) if stored != *message => {
+            Err(ContractError::MessageMismatch(message.id.to_string()))
+        }
+        Some(_) => Ok(true),
+        None => Ok(false),
+    }
+}
```
cgorenflo, line 26: If I understand correctly how this will work, the whole call will fail if a single msg ID is false. I'd rather expect an unknown msg ID to give me `false`. To that end, I don't think either of these functions should have `Result` as return value

cgorenflo (changes requested):

contracts/voting-verifier/src/query.rs
haiyizxx: The whole call will fail if the caller provides a different message with an already verified id, means the message batch is invalid.

An unknown msg id will return false.

cgorenflo (approved):

PR #55, opened 2023-08-11
feat(voting-verifier): remove end poll

cjcobb23:
* determine poll results dynamically
* emit event when message is verified

See https://github.com/axelarnetwork/axelar-amplifier/pull/51 as well for how messages will be queried

AXE-1481


cgorenflo (changes requested):
While I like to remove end poll, this solution gets us into the same trouble we used to have in the vote module: The vote to reach quorum costs a lot more gas, and due to the time delay between simulation and execution is most likely going to fail due to lacking fees. We can't just tag the poll cleanup onto voting

cjcobb23:
closing, we can leave end poll for now and look into calling it from the end blocker

PR #58, opened 2023-08-18
feat(voting-verifier): confirm worker set update

cjcobb23:
* Add ConfirmWorkerSet execute msg
* Add IsWorkerSetConfirmed query

AXE-1536

### DIFF
--- a/contracts/connection-router/src/types.rs
+++ b/contracts/connection-router/src/types.rs
@@ -15,10 +15,7 @@ pub const ID_SEPARATOR: char = ':';
 
 #[cw_serde]
 #[serde(try_from = "String")]
-#[serde(into = "String")]
-pub struct MessageID {
-    value: String,
-}
+pub struct MessageID(String);
 
 impl FromStr for MessageID {
     type Err = ContractError;
@@ -27,34 +24,32 @@ impl FromStr for MessageID {
         if !s.contains(ID_SEPARATOR) || s.is_empty() {
             return Err(ContractError::InvalidMessageID {});
         }
-        Ok(MessageID {
-            value: s.to_lowercase(),
-        })
+        Ok(MessageID(s.to_lowercase()))
     }
 }
 
 impl TryFrom<String> for MessageID {
     type Error = ContractError;
     fn try_from(value: String) -> Result<Self, Self::Error> {
-        MessageID::from_str(value.as_str())
+        value.as_str().parse()
     }
 }
 
 impl From<MessageID> for String {
     fn from(d: MessageID) -> Self {
-        d.value
+        d.0
     }
 }
 
 impl<'a> MessageID {
     pub fn as_str(&'a self) -> &'a str {
-        self.value.as_str()
+        self.0.as_str()
     }
 }
 
 impl fmt::Display for MessageID {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "{}", self.value)
+        write!(f, "{}", self.0)
     }
 }
 
@@ -65,7 +60,7 @@ impl<'a> PrimaryKey<'a> for MessageID {
     type SuperSuffix = Self;
 
     fn key(&self) -> Vec<Key> {
-        vec![Key::Ref(self.value.as_bytes())]
+        vec![Key::Ref(self.0.as_bytes())]
     }
 }
 
@@ -74,13 +69,12 @@ impl KeyDeserialize for MessageID {
 
     fn from_vec(value: Vec<u8>) -> StdResult<Self> {
         let value = String::from_utf8(value).map_err(StdError::invalid_utf8)?;
-        Ok(Self { value })
+        Ok(Self(value))
     }
 }
 
 #[cw_serde]
 #[serde(try_from = "String")]
-#[serde(into = "String")]
 pub struct ChainName(String);
 
 impl FromStr for ChainName {
--- a/contracts/connection-router/src/types.rs
+++ b/contracts/connection-router/src/types.rs
@@ -14,6 +14,8 @@ use crate::ContractError;
 pub const ID_SEPARATOR: char = ':';
 
 #[cw_serde]
+#[serde(try_from = "String")]
+#[serde(into = "String")]
 pub struct MessageID {
     value: String,
 }
@@ -28,6 +30,13 @@ impl FromStr for MessageID {
         Ok(MessageID {
             value: s.to_lowercase(),
         })
+    }
+}
+
+impl TryFrom<String> for MessageID {
+    type Error = ContractError;
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        MessageID::from_str(value.as_str())
     }
 }
 
@@ -71,6 +80,7 @@ impl KeyDeserialize for MessageID {
 
 #[cw_serde]
 #[serde(try_from = "String")]
+#[serde(into = "String")]
 pub struct ChainName(String);
 
 impl FromStr for ChainName {
--- a/contracts/voting-verifier/src/contract.rs
+++ b/contracts/voting-verifier/src/contract.rs
@@ -6,7 +6,7 @@ use connection_router::state;
 
 use crate::error::ContractError;
 use crate::msg::{ExecuteMsg, InstantiateMsg, QueryMsg};
-use crate::state::{Config, CONFIG};
+use crate::state::{Config, CONFIG, CONFIRMED_WORKER_SETS};
 use crate::{execute, query};
 
 #[cfg_attr(not(feature = "library"), entry_point)]
@@ -23,6 +23,7 @@ pub fn instantiate(
         voting_threshold: msg.voting_threshold,
         block_expiry: msg.block_expiry,
         confirmation_height: msg.confirmation_height,
+        source_chain: msg.source_chain,
     };
     CONFIG.save(deps.storage, &config)?;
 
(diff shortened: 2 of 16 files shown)


contracts/voting-verifier/src/error.rs
```diff
@@ -37,6 +37,9 @@ pub enum ContractError {
 
     #[error("{0}")]
     VoteError(#[from] voting::Error),
+
+    #[error("worker set already confirmed")]
+    WorkerSetAlreadyConfirmed {},
```
cgorenflo, line 42:
```suggestion
    WorkerSetAlreadyConfirmed,
```

contracts/voting-verifier/src/events.rs
```diff
@@ -29,12 +30,32 @@ pub struct PollStarted {
...
+            serde_json::to_string(&msgs).expect("failed to serialize messages"),
+        ),
+        PollData::WorkerSet(worker_set_confirmation) => (
+            "worker_set_confirmation".into(),
+            serde_json::to_string(&worker_set_confirmation)
+                .expect("failed to serialize worker set confirmation"),
+        ),
+    }
+}
+
```
cgorenflo, line 55: suggestion: make this a `From<PollData> for Attribute` implementation and then decompose the attribute when you build the event like so:
```
let Attribute{ key, value } = other.data.into();
```

contracts/voting-verifier/src/execute.rs
```diff
@@ -13,17 +14,62 @@ use service_registry::msg::QueryMsg;
...
+    let config = CONFIG.load(deps.storage)?;
+    let snapshot = take_snapshot(deps.as_ref(), &env, &config.source_chain)?;
+    let participants = snapshot.get_participants();
+
+    if CONFIRMED_WORKER_SETS
+        .may_load(deps.storage, new_operators.hash())?
+        .is_some()
+    {
+        return Err(ContractError::WorkerSetAlreadyConfirmed {});
+    }
```
cgorenflo, line 46: do this first?

contracts/voting-verifier/src/execute.rs
```diff
@@ -13,17 +14,62 @@ use service_registry::msg::QueryMsg;
...
+
+    let id = create_poll(
+        deps.storage,
+        env.block.height,
+        config.block_expiry,
+        snapshot,
+        1,
+        PollType::ConfirmWorkerSet,
+    )?;
```
cgorenflo, line 55: have multiple `create_X_poll` instead? Then we can absorb this weird `1`, since the worker set poll always has a single element

contracts/voting-verifier/src/state.rs
```diff
@@ -18,14 +19,26 @@ pub struct Config {
     pub voting_threshold: Threshold,
     pub block_expiry: u64, // number of blocks after which a poll expires
     pub confirmation_height: u64,
+    pub source_chain: ChainName,
+}
+
+#[cw_serde]
+pub enum PollType {
+    Messages,
+    ConfirmWorkerSet,
```
cgorenflo, line 28: make use of rust's power and combine poll and type:
```
pub enum Poll {
   Messages(WeightedPoll),
   ConfirmWorkerSet(WeightedPoll),
}
```

contracts/voting-verifier/src/state.rs
```diff
@@ -18,14 +19,26 @@ pub struct Config {
...
+pub const POLLS: Map<PollID, (WeightedPoll, PollType)> = Map::new("polls");
 
 pub const PENDING_MESSAGES: Map<PollID, Vec<Message>> = Map::new("pending_messages");
 
 pub const VERIFIED_MESSAGES: Map<&MessageID, Message> = Map::new("verified_messages");
 
 pub const CONFIG: Item<Config> = Item::new("config");
+
+// primary key is hash of Operators struct
+pub const CONFIRMED_WORKER_SETS: Map<Vec<u8>, ()> = Map::new("confirmed_worker_sets");
```
sammy1991106, line 42: Do we not have some better type for hash than `Vec<u8>`?

packages/axelar-wasm-std/src/operators.rs
```diff
@@ -0,0 +1,18 @@
+use cosmwasm_schema::cw_serde;
+use cosmwasm_std::{to_vec, HexBinary, Uint256};
+
+use sha3::{Digest, Keccak256};
+#[cw_serde]
+pub struct Operators {
+    pub addresses: Vec<HexBinary>, // TODO, switch to evm address
+    pub weights: Vec<Uint256>,
```
sammy1991106, line 8:
```suggestion
    pub weights: Vec<(HexBinary, Uint256)>, // TODO, switch to evm address
```

packages/axelar-wasm-std/src/operators.rs
```diff
@@ -0,0 +1,18 @@
+use cosmwasm_schema::cw_serde;
+use cosmwasm_std::{to_vec, HexBinary, Uint256};
+
+use sha3::{Digest, Keccak256};
+#[cw_serde]
```
sammy1991106, line 5:
```suggestion

#[cw_serde]
```

contracts/voting-verifier/src/events.rs
```diff
@@ -29,12 +30,32 @@ pub struct PollStarted {
...
 
+pub enum PollData {
+    Messages(Vec<EvmMessage>),
+    WorkerSet(WorkerSetConfirmation),
+}
+
+fn get_data_attribute(data: PollData) -> (String, String) {
+    match data {
+        PollData::Messages(msgs) => (
+            "messages".into(),
```
sammy1991106, line 45: I don't think using the attribute name to differentiate is the right way to go. I'd much prefer either 1) the poll type has its own attribute or 2) using a different event type. For the actual data, `data` is good.

contracts/voting-verifier/src/events.rs
```diff
@@ -77,6 +95,24 @@ impl TryFrom<Vec<Message>> for EvmMessages {
...
+#[cw_serde]
+pub struct WorkerSetConfirmation {
+    pub tx_id: String,
+    pub log_index: u64,
+    pub operators: Operators,
+}
+
+impl WorkerSetConfirmation {
+    pub fn new(message_id: MessageID, operators: Operators) -> Result<Self, ContractError> {
+        let (tx_id, log_index) = parse_message_id(message_id.to_string())?;
```
sammy1991106, line 107: Why does `parse_message_id` take a `String` instead of `&MessageID` or even `AsRef<MessageID>`?

contracts/voting-verifier/src/events.rs
sammy1991106: Out of these 2, I prefer 2)

sammy1991106 (changes requested):

contracts/voting-verifier/src/contract.rs
```diff
@@ -23,6 +23,7 @@ pub fn instantiate(
         voting_threshold: msg.voting_threshold,
         block_expiry: msg.block_expiry,
         confirmation_height: msg.confirmation_height,
+        source_chain: msg.source_chain,
```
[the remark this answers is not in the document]
cjcobb23: It's already linked to a source chain. There is a source_gateway_address field already, and the contract is meant to work for a single source chain.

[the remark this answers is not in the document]
sammy1991106: That's not necessarily true cuz we can use CREATE2 to deterministically deploy gateways onto EVM chains so that they all have the exact same address. However, if we decide to link this contract to a specific chain, then the source chain check at https://github.com/axelarnetwork/axelar-amplifier/blob/b5b50fb142795700288a7c7df9f19aae02d8276c/contracts/voting-verifier/src/contract.rs#L56 absolutely needs to change. At https://github.com/axelarnetwork/axelar-amplifier/blob/b5b50fb142795700288a7c7df9f19aae02d8276c/contracts/voting-verifier/src/execute.rs#L128, it also has to change so that source chain isn't derived from messages.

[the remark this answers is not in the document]
cjcobb23: The gateways today though don't all have the same address. We can deploy with the same address, but I wouldn't want to add that as a requirement at the protocol layer. 

We also have just generally been planning to have one voting verifier per chain, for now at least. We can revisit that in the future, because there are benefits to having just one voting verifier. But for now, adding the source chain here doesn't break anything about the model we've been working off of.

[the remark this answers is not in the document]
sammy1991106: @cjcobb23 I am OK with either way, but I need to see consistency within the code of this particular contract. Right now, the messages verification part isn't consistent with the fact that this contract is linked to one single source chain.

contracts/connection-router/src/types.rs
```diff
@@ -14,6 +14,8 @@ use crate::ContractError;
 pub const ID_SEPARATOR: char = ':';
 
 #[cw_serde]
+#[serde(try_from = "String")]
+#[serde(into = "String")]
```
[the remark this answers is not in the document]
cjcobb23: Switched it to `MessageID(String)` and `into` isn't needed anymore

contracts/voting-verifier/src/events.rs
cjcobb23: I went with 2, lmk what you think.

contracts/voting-verifier/src/events.rs
```diff
@@ -81,7 +81,7 @@ impl From<PollStarted> for Event {
                     "messages",
                     to_string(&data).expect("failed to serialize messages"),
                 )
-                .add_attributes(<PollMetadata as Into<Vec<Attribute>>>::into(metadata)),
+                .add_attributes(Vec::<_>::from(metadata)),
```
sammy1991106, line 84: even better! 👍

sammy1991106 (approved):

cgorenflo (approved):

PR #64, opened 2023-08-22
feat(multisig): public key registration

cjcobb23:
## Description
* Add a method to register public keys
* Add a query to fetch public keys
* Distinguish keys and signatures by type

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Other Notes
AXE-1533


contracts/multisig/src/msg.rs
```diff
@@ -22,7 +22,11 @@ pub enum ExecuteMsg {
     KeyGen {
         key_id: String,
         snapshot: Snapshot,
-        pub_keys: HashMap<String, HexBinary>,
+        pub_keys: HashMap<String, (KeyType, HexBinary)>,
+    },
+    RegisterPublicKey {
+        public_key: HexBinary,
+        key_type: KeyType,
```
cgorenflo, line 29: why not use the `PublicKey` enum instead of the type + hex tuple?

contracts/multisig/src/state.rs
```diff
@@ -19,3 +20,5 @@ pub fn get_key(store: &dyn Storage, key_id: &KeyID) -> Result<Key, ContractError
             key_id: key_id.to_string(),
         })
 }
+
+pub const PUB_KEYS: Map<(Addr, KeyType), PublicKey> = Map::new("registered_pub_keys");
```
cgorenflo, line 24: in this case I would use the raw binary data for the public key, to prevent errors where KeyType and PublicKey don't match up. Then, we can define a conversion from ((addr, keytype), key) to (addr, typed_key)

cgorenflo, line 24: add a comment that the key type is part of the key so that signers can register multiple keys with different types

cgorenflo (changes requested):

contracts/multisig/src/types.rs
```diff
@@ -2,31 +2,85 @@ use std::{collections::HashMap, fmt};
...
-use crate::ContractError;
+use crate::{secp256k1::ecdsa_verify, ContractError};
 
 pub trait VerifiableSignature {
     fn verify(&self, msg: &MsgToSign, pub_key: &PublicKey) -> Result<bool, ContractError>;
 }
 
 #[cw_serde]
-pub struct PublicKey(HexBinary);
+pub enum KeyType {
```
sammy1991106, line 15: Discussed with Chris as well, here's what we came up with to make these nicer. Warning: pseudocode.
```
enum KeyType {
  ECDSA,
  ED25519,
}

enum Signature {
  ECDSA(HexBinary)
  ED25519(HexBinary)
}

enum PublicKey {
    ECDSA(HexBinary)
    ED25519(HexBinary)
}

trait KeyTyped {
  fn matches<T>(&self, other: T) -> bool where T: KeyTyped {
    self.key_type() == other.key_type()
  };

  fn key_type(&self) -> KeyType;
}

impl KeyTyped for PublicKey {
  fn key_type(&self) -> KeyType {
    match self {
      ECDSA(_) => KeyType::ECDSA,
      ED25519(_) => KeyType::ED25519,
    }
  }
}

impl KeyTyped for Signature {
  fn key_type(&self) -> KeyType {
    match self {
      ECDSA(_) => KeyType::ECDSA,
      ED25519(_) => KeyType::ED25519,
    }
  }
}

impl Signature {
  pub fn verify(&self, pub_key: PublicKey) -> Result<bool, Error> {
    if !self.matches(pub_key) {
      return Err(Error::new("Key type mismatch"));
    }

    match self {
      ECDSA(sig) => {
        todo!()
      },
      ED25519(sig) => {
        todo!()
      },
    }
  }
}
```

contracts/multisig/src/types.rs
```diff
@@ -55,32 +109,64 @@ impl MsgToSign {
...
-    fn from(original: &Signature) -> Self {
+impl From<&ECDSASignature> for Vec<u8> {
+    fn from(original: &ECDSASignature) -> Self {
         original.0.to_vec()
     }
 }
 
-impl<'a> From<&'a Signature> for &'a [u8] {
-    fn from(original: &'a Signature) -> Self {
+impl<'a> From<&'a ECDSASignature> for &'a [u8] {
```
sammy1991106, line 150: implement AsRef instead.

contracts/multisig/src/msg.rs
sammy1991106: Agree with Chris

contracts/multisig/src/events.rs
```diff
@@ -23,6 +23,11 @@ pub enum Event {
     SigningCompleted {
         session_id: Uint64,
     },
+    PublicKeyRegistered {
+        worker: Addr,
+        public_key: PublicKey,
+        key_type: KeyType,
```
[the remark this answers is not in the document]
sammy1991106: Don't need key_type here.

sammy1991106 (changes requested):

ampd/tests/report.rs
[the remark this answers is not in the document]
cjcobb23: I was wrong, it was supposed to be 2.

contracts/multisig-prover/src/encoding/evm.rs
```diff
@@ -51,7 +51,7 @@ impl TryFrom<Signer> for Operator {
         Ok(Self {
             address: evm_address((&signer.pub_key).into())?,
             weight: signer.weight,
-            signature: signer.signature,
+            signature: signer.signature.map(|Signature::ECDSA(sig)| sig),
```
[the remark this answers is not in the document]
cjcobb23: this was from before, but it's due to the way Signer is used. It's returned by the `GetMultisig` query, which queries the status of a specific signing session. So it returns the signer, and the signature (if present).

I actually refactored this in a later PR to move the signature out of the Signer struct.

contracts/multisig/src/types.rs
```diff
@@ -2,31 +2,85 @@ use std::{collections::HashMap, fmt};
...
 
 pub trait VerifiableSignature {
     fn verify(&self, msg: &MsgToSign, pub_key: &PublicKey) -> Result<bool, ContractError>;
 }
 
 #[cw_serde]
-pub struct PublicKey(HexBinary);
+pub enum KeyType {
+    ECDSA,
+}
```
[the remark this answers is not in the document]
cjcobb23: We need `KeyType` on it's own at least for the query, and possibly for the execute msg

contracts/multisig-prover/src/execute.rs
```diff
@@ -95,7 +96,11 @@ pub fn rotate_snapshot(
         msg: to_binary(&multisig::msg::ExecuteMsg::KeyGen {
             key_id: key_id.clone(),
             snapshot: snapshot.clone(),
-            pub_keys: pub_keys.clone(),
+            pub_keys: pub_keys
+                .clone()
+                .into_iter()
+                .map(|(k, v)| (k, (KeyType::ECDSA, v)))
```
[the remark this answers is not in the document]
cjcobb23: rotate snapshot is actually deleted in a later PR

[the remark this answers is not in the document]
cjcobb23: We won't need `rotate_snapshot` later on, it's actually deleted in a later PR. We can just pass `PublicKey` in KeyGen execute message, if we want to keep `KeyGen`

contracts/multisig/src/types.rs
```diff
@@ -2,31 +2,85 @@ use std::{collections::HashMap, fmt};
...
+use crate::{secp256k1::ecdsa_verify, ContractError};
 
 pub trait VerifiableSignature {
     fn verify(&self, msg: &MsgToSign, pub_key: &PublicKey) -> Result<bool, ContractError>;
 }
 
 #[cw_serde]
-pub struct PublicKey(HexBinary);
+pub enum KeyType {
+    ECDSA,
```
eguajardo, line 16: I noticed that the JSON representation of this looks like `"key_type":"e_c_d_s_a"`. I suggest we use serde rename macro (https://serde.rs/variant-attrs.html#rename) or rename_all so it doesn't have underscores

sammy1991106: Good catch @eguajardo. First, it should be `Ecdsa` instead of `ECDSA`. Second, should add `#[serde(tag = "type")]` to the enum.

contracts/multisig/src/types.rs
sammy1991106: @cjcobb23 Also, you need to check out this https://serde.rs/enum-representations.html. I personally prefer it being internally tagged, i.e. `#[serde(tag = "type")]`.

cjcobb23: I implemented this, take a look

contracts/multisig/src/secp256k1.rs
```diff
@@ -23,7 +23,7 @@ impl TryFrom<HexBinary> for PublicKey {
             });
         }
 
-        Ok(PublicKey::unchecked(other))
+        Ok(ECDSAPublicKey::unchecked(other))
```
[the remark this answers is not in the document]
cjcobb23: These were here prior to the PR, but I can see the usefulness of having to call `unchecked` versus just wrapping. Feel like it makes it more obvious that it should be checked prior to calling this.

contracts/multisig/src/contract.rs
```diff
@@ -95,13 +104,26 @@ pub mod execute {
...
     ) -> Result<Response, ContractError> {
         let mut session = SIGNING_SESSIONS
             .load(deps.storage, session_id.into())
             .map_err(|_| ContractError::SigningSessionNotFound { session_id })?;
 
         let key = KEYS.load(deps.storage, &session.key_id)?;
+        let signature: Signature = match key
+            .pub_keys
+            .iter()
+            .find(|pk| pk.0 == &info.sender.to_string())
```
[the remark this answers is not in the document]
eguajardo: was changed by CJ to `.find(|&(addr, _)| addr == &info.sender.to_string())`. Looks good to me

cgorenflo (approved):

PR #67, opened 2023-08-25
feat(multisig-prover): start update worker set

cjcobb23:
## Description
detect if worker set needs to be updated
abi encode new worker set and include in batch
depends on https://github.com/axelarnetwork/axelar-amplifier/pull/64

## Todos
update the worker set automatically as part of batch creation

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Other Notes
AXE-1534


contracts/multisig-prover/src/state.rs
```diff
@@ -24,3 +28,42 @@ pub const COMMANDS_BATCH: Map<&BatchID, CommandBatch> = Map::new("command_batch"
 pub const MULTISIG_SESSION_BATCH: Map<u64, BatchID> = Map::new("multisig_session_batch");
 
 pub const REPLY_BATCH: Item<BatchID> = Item::new("reply_tracker");
+
+#[cw_serde]
+pub struct WorkerSet {
+    pub signers: Vec<Signer>,
+    pub threshold: Uint256,
```
cgorenflo, line 35: which threshold is this?

contracts/multisig-prover/src/state.rs
```diff
@@ -24,3 +28,42 @@ pub const COMMANDS_BATCH: Map<&BatchID, CommandBatch> = Map::new("command_batch"
...
+    pub id: u64, // for randomness
+}
+
+impl WorkerSet {
+    pub fn new(
+        snapshot: Snapshot,
+        pub_keys: Vec<multisig::types::PublicKey>,
+        env: Env,
```
cgorenflo, line 43: pass in the block height instead

contracts/multisig/src/contract.rs
```diff
@@ -251,14 +256,16 @@ pub mod query {
...
+                    Signer {
+                        address: participant.address,
+                        weight: participant.weight.into(),
+                        pub_key,
+                    },
+                    session.signatures.get(&address).cloned(),
+                )
             })
-            .collect::<Vec<Signer>>();
+            .collect::<Vec<(Signer, Option<Signature>)>>();
```
cgorenflo, line 268: `let signers_with_sigs = ...`

contracts/multisig/src/msg.rs
```diff
@@ -51,12 +51,11 @@ pub struct Signer {
...
     pub pub_key: PublicKey,
-    pub signature: Option<Signature>,
 }
 
 #[cw_serde]
 pub struct Multisig {
     pub state: MultisigState,
     pub quorum: Uint256,
-    pub signers: Vec<Signer>,
+    pub signers: Vec<(Signer, Option<Signature>)>,
```
cgorenflo, line 60:
```suggestion
    pub signers_with_sigs: Vec<(Signer, Option<Signature>)>,
```

cgorenflo: why are signatures optional?

cgorenflo (changes requested):

contracts/multisig-prover/src/state.rs
cjcobb23: signing threshold. Defines how many signatures are needed before relaying to the external gateway

contracts/multisig/src/msg.rs
cjcobb23: This is the response for the `GetMultisig` query, and the behavior is to return all of the signers, even if they haven't signed a particular message yet. So the signature is present iff they signed.

contracts/multisig-prover/src/encoding/evm.rs
```diff
@@ -44,34 +46,69 @@ impl TryFrom<Message> for Command {
...
+        let params = transfer_operatorship_params(&worker_set)?;
+        Ok(Command {
+            ty: CommandType::TransferOperatorship,
+            params,
+            id: worker_set.hash(),
+        })
+    }
+}
+
+impl TryFrom<(Signer, Option<Signature>)> for Operator {
```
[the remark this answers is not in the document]
cjcobb23: The signature field was removed from `Signer` in this PR.

I agree there are a lot of structs that are meant to represent pretty much the same thing (a worker). Most of them (except WorkerSet) were there prior to this PR. I'd like to refactor them, but I'd rather separate that refactoring into a separate task.

contracts/multisig-prover/src/encoding/evm.rs
```diff
@@ -44,34 +46,69 @@ impl TryFrom<Message> for Command {
...
+        weights: zipped,
+        threshold: snapshot.quorum.into(),
+    })
+}
+
 impl CommandBatch {
     pub fn new(
         messages: Vec<Message>,
         destination_chain_id: Uint256,
+        new_worker_set: Option<WorkerSet>,
```
sammy1991106, line 95: I'd prefer something like following instead of having everything possible in the `.new` method.
```
let command_batch_builder = CommandBatchBuilder::default();

command_batch_builder.add_message( ... );
command_batch_builder.add_new_worker_set( ... );

let command_batch = command_batch_builder.build();
```

sammy1991106: We are likely going to have more types of commands, and keep adding things to `new` is confusing to me.

contracts/multisig-prover/src/encoding/evm.rs
```diff
@@ -44,34 +46,69 @@ impl TryFrom<Message> for Command {
...
     ) -> Result<Self, ContractError> {
         let message_ids: Vec<String> = messages.iter().map(|msg| msg.id.clone()).collect();
+        let mut commands = messages
+            .into_iter()
+            .map(TryInto::try_into)
+            .collect::<Result<Vec<Command>, ContractError>>()?;
+        if new_worker_set.is_some() {
+            commands.push(new_worker_set.clone().unwrap().try_into()?);
```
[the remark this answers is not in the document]
sammy1991106: Pushing it into `.commands` without touching the `.message_ids` is weird to me.

contracts/multisig-prover/src/types.rs
```diff
@@ -63,10 +65,13 @@ impl KeyDeserialize for BatchID {
 }
 
 impl BatchID {
-    pub fn new(message_ids: &[String]) -> BatchID {
+    pub fn new(message_ids: &[String], new_worker_set: Option<WorkerSet>) -> BatchID {
```
sammy1991106, line 68: I don't like seeing `new_worker_set` here. I prefer treating it as a message initiated from Axelar.

contracts/multisig-prover/src/encoding/evm.rs
```diff
@@ -150,6 +187,40 @@ impl CommandBatch {
     }
 }
 
+fn transfer_operatorship_params(worker_set: &WorkerSet) -> Result<HexBinary, ContractError> {
+    let mut operators: Vec<(HexBinary, Uint256)> = worker_set
+        .signers
+        .iter()
+        .map(|s| {
+            (
+                evm_address((&s.pub_key).into()).expect("couldn't convert pubkey to evm address"),
```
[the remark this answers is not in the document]
sammy1991106: I partially agree with @cgorenflo. Any Ecdsa public key can be converted into a valid Ethereum address so it only makes sense for the evm_address method to return an error when the key type isn't right.

contracts/multisig-prover/src/encoding/evm.rs
```diff
@@ -192,7 +263,9 @@ fn evm_address(pub_key: &[u8]) -> Result<HexBinary, ContractError> {
     Ok(Keccak256::digest(&pub_key.as_bytes()[1..]).as_slice()[12..].into())
 }
 
-fn sorted_operators(signers: Vec<Signer>) -> Result<Vec<Operator>, ContractError> {
+fn sorted_operators(
+    signers: Vec<(Signer, Option<Signature>)>,
+) -> Result<Vec<Operator>, ContractError> {
```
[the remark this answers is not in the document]
sammy1991106: Agree with @cgorenflo.

contracts/multisig-prover/src/contract.rs
```diff
@@ -41,6 +41,7 @@ pub fn instantiate(
         service_name: msg.service_name,
         chain_name: ChainName::from_str(&msg.chain_name)
             .map_err(|_| ContractError::InvalidChainName)?,
+        worker_set_diff_threshold: msg.worker_set_diff_threshold,
```
sammy1991106, line 44: What is this?

contracts/multisig-prover/src/execute.rs
```diff
@@ -80,6 +83,87 @@ fn get_messages(
...
+        let pub_key_query = multisig::msg::QueryMsg::GetPublicKey {
+            worker_address: worker.address.to_string(),
+            key_type: multisig::types::KeyType::ECDSA,
+        };
+        let pub_key: PublicKey = deps.querier.query(&QueryRequest::Wasm(WasmQuery::Smart {
+            contract_addr: config.multisig.to_string(),
+            msg: to_binary(&pub_key_query)?,
+        }))?;
+        pub_keys.push(pub_key);
+    }
```
sammy1991106, line 122: Do this with `.iter().map`?

contracts/multisig-prover/src/execute.rs
```diff
@@ -80,6 +83,87 @@ fn get_messages(
...
+            .iter()
+            .filter(|a_signer| {
+                !b.signers.iter().any(|b_signer| {
+                    a_signer.address == b_signer.address && a_signer.pub_key == b_signer.pub_key
+                })
+            })
+            .count()
+    };
+    count_new(new_workers, cur_workers) + count_new(cur_workers, new_workers) > max_diff
+}
```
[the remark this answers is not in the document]
sammy1991106: https://doc.rust-lang.org/std/collections/struct.BTreeSet.html#method.difference

contracts/multisig-prover/src/state.rs
sammy1991106: Agree with @cgorenflo. This goes again with my comment saying that I don't like passing exessive information to any function.

contracts/multisig-prover/src/state.rs
```diff
@@ -24,3 +28,42 @@ pub const COMMANDS_BATCH: Map<&BatchID, CommandBatch> = Map::new("command_batch"
...
+
+        Ok(WorkerSet {
+            signers,
+            threshold: snapshot.quorum.into(),
+            id: env.block.height,
+        })
+    }
+
+    pub fn hash(&self) -> HexBinary {
+        Keccak256::digest(serde_json::to_vec(&self).expect("couldn't serialize worker set"))
```
sammy1991106, line 63: Here it makes it cleaer that a BTreeSet makes way more sense than a Vec cuz order matter when it comes to the hash.

sammy1991106 (changes requested):

contracts/multisig-prover/src/types.rs
cjcobb23: I could just move `message_ids.push(new_worker_set.hash().to_string())` to the call site. A little weird though, because it's not actually a message id, and there's not really a way to generate a proper message id from a worker set. I'm ambivalent.

contracts/multisig-prover/src/encoding/evm.rs
```diff
@@ -44,34 +46,69 @@ impl TryFrom<Message> for Command {
...
         Ok(Self {
             address: evm_address((&signer.pub_key).into())?,
             weight: signer.weight,
-            signature: signer.signature.map(|Signature::ECDSA(sig)| sig),
+            signature: sig.map(|Signature::ECDSA(sig)| sig),
         })
     }
 }
 
+pub fn make_worker_set(
```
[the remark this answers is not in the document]
cjcobb23: It's not being used, deleted it. There is a similar method on `WorkerSet` actually

contracts/multisig-prover/src/contract.rs
cjcobb23: It's the max number of workers that can change without triggering a worker set update.

cgorenflo (approved):

PR #69, opened 2023-08-25
refactor: use try_from macro in ampd

cgorenflo:


packages/events-derive/src/lib.rs
```diff
@@ -41,32 +40,34 @@ pub fn try_from(arg: TokenStream, input: TokenStream) -> TokenStream {
         #input
 
         use error_stack::{IntoReport as _, ResultExt as _};
+        use events as _internal_events;
+        use core::convert::TryFrom as _internal_TryFrom;
```
sammy1991106, line 44: Why `_internal_TryFrom`?

ampd/src/handlers/evm_verify_msg.rs
```diff
@@ -38,7 +38,8 @@ pub struct Message {
 }
 
 #[derive(Deserialize, Debug)]
-struct Event {
+#[try_from("wasm-messages_poll_started")]
```
sammy1991106, line 41: so good!

ampd/src/handlers/evm_verify_msg.rs
```diff
@@ -160,18 +145,20 @@ where
...
             confirmation_height,
             participants,
-        } = match event.try_into().into_report()? {
-            Some(event) => event,
-            None => return Ok(()),
+        } = match event.try_into() as error_stack::Result<PollStartedEvent, events::Error> {
+            Err(report) if matches!(report.current_context(), EventTypeMismatch(_)) => {
+                return Ok(())
+            }
+            event => event.change_context(DeserializeEvent)?,
```
sammy1991106, line 161: This beomes a bit too much though.

ampd/src/handlers/evm_verify_msg.rs
```diff
@@ -160,18 +145,20 @@ where
...
             poll_id,
             source_chain,
             source_gateway_address,
             messages,
             confirmation_height,
             participants,
-        } = match event.try_into().into_report()? {
-            Some(event) => event,
-            None => return Ok(()),
+        } = match event.try_into() as error_stack::Result<PollStartedEvent, events::Error> {
```
sammy1991106, line 157:
```suggestion
        } = match event.try_into() as error_stack::Result<_, _> {
```

packages/events/src/event.rs
```diff
@@ -19,3 +27,50 @@ impl Event {
...
+    ))
+}
+
+fn base64_to_utf8(base64_str: &str) -> Result<String, DecodingError> {
+    STANDARD
+        .decode(base64_str)
+        .map_err(DecodingError::from)
+        .and_then(|bytes| String::from_utf8(bytes).map_err(DecodingError::from))
+        .into_report()
+}
```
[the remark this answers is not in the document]
cgorenflo: I don't think this gets around the `map_err` issue because the `?` would return incompatible error types

packages/events-derive/src/lib.rs
cgorenflo: to get around unexpected behaviour. For example, if the file you use that macro in already imports `TryFrom`, suddenly it wouldn't compile because of a duplicate import. Chances of a name collision with `_internal_TryFrom` are next to zero, so that should avoid this

ampd/src/handlers/evm_verify_msg.rs
cgorenflo: 🙂

ampd/src/handlers/evm_verify_msg.rs
cgorenflo: that's a side-effect of <error-stack::Report> unfortunately. Haven't found a better option so far

haiyizxx (approved):
looks good in general, I will give my approval as I will be away tmr. don't want block this one after conversations resolved

sammy1991106 (approved):

PR #70, opened 2023-08-26
feat(prover): confirm worker set

cjcobb23:
## Description
Confirm a worker set update via voting verifier, and then start using it.
Depends on https://github.com/axelarnetwork/axelar-amplifier/pull/67, which depends on https://github.com/axelarnetwork/axelar-amplifier/pull/64

## Todos
Refactor the code to remove any notion of key ID and key.

## Other Notes
AXE-1535


contracts/multisig-prover/src/state.rs
```diff
@@ -33,19 +35,20 @@ pub const REPLY_BATCH: Item<BatchID> = Item::new("reply_tracker");
 pub struct WorkerSet {
     pub signers: Vec<Signer>,
     pub threshold: Uint256,
-    pub id: u64, // for randomness
+    // for hash uniqueness. The same exact worker set could be in use at two different times,
+    // and we need to be able to distinguish between the two
+    pub nonce: u64,
```
cgorenflo, line 40: nice!

contracts/multisig-prover/src/execute.rs
```diff
@@ -121,29 +123,108 @@ pub fn update_worker_set(deps: DepsMut, env: Env) -> Result<Response, ContractEr
...
+                })
+                .collect(),
+        };
+
+        return Ok(Response::new().add_message(wasm_execute(
+            config.multisig,
+            &key_gen_msg,
+            vec![],
+        )?));
+    }
```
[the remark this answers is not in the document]
cjcobb23: We don't really even need the keygen execute msg at all anymore. All keygen is actually doing is storing the workerset in the multisig contract, so later on when we trigger signing sessions, we can just pass some id for the workerset, instead of the whole workerset. I would like to just remove keygen and store the workerset in the prover only, but I think that should be left to a later PR.

contracts/multisig-prover/src/execute.rs
```diff
@@ -121,29 +123,108 @@ pub fn update_worker_set(deps: DepsMut, env: Env) -> Result<Response, ContractEr
...
+    if !is_confirmed {
+        return Err(ContractError::WorkerSetNotConfirmed);
+    }
+
+    CURRENT_WORKER_SET.save(deps.storage, &worker_set)?;
+    NEXT_WORKER_SET.remove(deps.storage);
+    KEY_ID.save(deps.storage, &worker_set.hash().to_hex())?;
+
+    let key_gen_msg = multisig::msg::ExecuteMsg::KeyGen {
+        key_id: worker_set.id(), // TODO: replace key id with worker set id
```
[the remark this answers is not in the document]
cjcobb23: old comment, will remove

cgorenflo (approved):

contracts/multisig-prover/src/encoding/evm.rs
```diff
@@ -54,6 +55,29 @@ fn make_transfer_operatorship(worker_set: WorkerSet) -> Result<Command, Contract
     })
 }
 
+impl TryFrom<WorkerSet> for Operators {
```
sammy1991106, line 58: Implement the From trait instead? Doesn't seem that any error can occur.

contracts/multisig-prover/src/msg.rs
```diff
@@ -23,21 +23,19 @@ pub struct InstantiateMsg {
...
-    ConstructProof {
-        message_ids: Vec<String>,
-    },
-    RotateSnapshot {
-        pub_keys: HashMap<String, HexBinary>,
-        key_id: String,
-    },
+    ConstructProof { message_ids: Vec<String> },
     UpdateWorkerSet {},
+    ConfirmWorkerSet {},
```
sammy1991106, line 28:
```suggestion
    UpdateWorkerSet,
    ConfirmWorkerSet,
```

contracts/multisig-prover/src/msg.rs
```diff
@@ -23,21 +23,19 @@ pub struct InstantiateMsg {
...
 }
 
 #[cw_serde]
 #[derive(QueryResponses)]
 pub enum QueryMsg {
     #[returns(GetProofResponse)]
     GetProof { multisig_session_id: Uint64 },
+
+    #[returns(WorkerSet)]
+    GetWorkerSet {},
```
sammy1991106, line 38:
```suggestion
    GetWorkerSet,
```

contracts/multisig-prover/src/execute.rs
```diff
@@ -126,31 +128,116 @@ pub fn update_worker_set(deps: DepsMut, env: Env) -> Result<Response, ContractEr
...
+    let new_worker_set = WorkerSet::new(
+        participants.clone(),
+        snapshot.quorum.into(),
+        env.block.height,
+    )?;
+
+    let cur_worker_set = CURRENT_WORKER_SET.may_load(deps.storage)?;
 
-    let cur_worker_set = CURRENT_WORKER_SET.load(deps.storage)?;
+    let key_id = new_worker_set.id();
```
sammy1991106, line 147: key_id = new_worker_set.id()? This reads so weird that we really need to clear our terminologies. I'd rather seeing the `.hash` method on WorkerSet being used here than trying to understand that this isn't a bug.

contracts/multisig-prover/src/execute.rs
```diff
@@ -126,31 +128,116 @@ pub fn update_worker_set(deps: DepsMut, env: Env) -> Result<Response, ContractEr
...
+    let config = CONFIG.load(deps.storage)?;
 
-    NEXT_WORKER_SET.save(deps.storage, &new_worker_set)?;
-    let mut builder = CommandBatchBuilder::new(config.destination_chain_id);
-    builder.add_new_worker_set(new_worker_set)?;
+    let (worker_set, snapshot) = NEXT_WORKER_SET.load(deps.storage)?;
 
-    let batch = builder.build()?;
+    let query = voting_verifier::msg::QueryMsg::IsWorkerSetConfirmed {
+        new_operators: worker_set.clone().try_into()?,
```
sammy1991106, line 209: Here the operators are not nonced so a new worker set with the exact same operators as any previous one is immediately confirmed?

contracts/multisig-prover/src/execute.rs
```diff
@@ -126,31 +128,116 @@ pub fn update_worker_set(deps: DepsMut, env: Env) -> Result<Response, ContractEr
...
+        msg: to_binary(&query)?,
+    }))?;
+
+    if !is_confirmed {
+        return Err(ContractError::WorkerSetNotConfirmed);
+    }
+
+    CURRENT_WORKER_SET.save(deps.storage, &worker_set)?;
+    NEXT_WORKER_SET.remove(deps.storage);
+    KEY_ID.save(deps.storage, &worker_set.hash().to_hex())?;
```
sammy1991106, line 223: ok this is easier to comprehend, but why using that `.id()` method above then?

contracts/multisig-prover/src/state.rs
```diff
@@ -67,6 +69,11 @@ impl WorkerSet {
             .as_slice()
             .into()
     }
+
+    pub fn id(&self) -> String {
+        self.hash().to_hex()
+    }
```
sammy1991106, line 75:
```suggestion
```

contracts/multisig/src/contract.rs
```diff
@@ -46,7 +46,7 @@ pub fn execute(
         ExecuteMsg::KeyGen {
             key_id,
             snapshot,
-            pub_keys,
+            pub_keys_by_address: pub_keys,
```
sammy1991106, line 49: Why this renaming? If the renaming is legit, it should remain this way entering the execute mod.

sammy1991106 (changes requested):

contracts/multisig-prover/src/execute.rs
eguajardo: I think the id is used here because CJ was planning to remove the key_id everywhere in the future and use the worker set id instead when needed. Is that so @cjcobb23 ?

cjcobb23: yes we can remove the key id entirely, and just use a worker set id

contracts/multisig-prover/src/execute.rs
cjcobb23: Good catch. This is actually pretty tricky to deal with, because the nonce is not passed to the evm gateway.

There are a lot of ways to deal with this:

1. Make the voting verifier actually call `ConfirmWorkerSet` on the prover, after the worker set is confirmed. As far as I can tell, this is the closest to how we handle this in core. The main challenge here is this creates a circular dependency at instantiation time. The voting verifier would need to know the address of the prover, but the prover needs to know the address of the gateway, which needs to know the address of the voting verifier. If there was some sort of constant address deployer for cosmwasm contracts, this would be easier.
2. Prune old operator confirmations from the voting verifier. We really only need the most recent. The EVM gateway stores the last 16, so that would work as well. We could also still store the old ones, but only query the most recent (or last 16).

Either way, I do think we should tackle this in a separate PR.

cjcobb23: Well actually apparently the EVM gateway doesn't allow duplicate operators lol: https://github.com/axelarnetwork/axelar-cgp-solidity/blob/134d264144b5adc32c42b1389618e86c91fbdc75/contracts/auth/AxelarAuthWeighted.sol#L80

I think this doesn't really make sense, but given this is the case, we should just enforce the same thing in the prover.

eguajardo: Yeah I saw that line as well and thought it was weird, but Milap mentioned it's because in consensus the public keys are not reused during key rotation: https://axelarworkspace.slack.com/archives/C02T1BNNS9Y/p1693326543812459. So that check makes sense for consensus, but not anymore for amplifier. So I don't think we should enforce the same thing.

cjcobb23: I think until we change the evm gateway, we actually have to check this in amplifier. Otherwise amplifier could start a rotation that will never be confirmed by the evm gateway, which will cause problems.

contracts/multisig-prover/src/encoding/evm.rs
```diff
@@ -58,6 +58,28 @@ impl TryFrom<WorkerSet> for Command {
...
+            .iter()
+            .map(|s| {
+                (
+                    evm_address((&s.pub_key).into())
+                        .expect("couldn't convert pubkey to evm address"),
+                    s.weight,
+                )
+            })
+            .collect();
+        operators.sort_by_key(|op| op.0.clone());
```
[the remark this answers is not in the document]
eguajardo: Do you mean the `Operators` instead of the `WorkerSet`? the sort should be based on the evm address which is not part of the worker set. Should we have a separate refactor task for this one?

contracts/multisig-prover/src/execute.rs
```diff
@@ -126,31 +128,123 @@ pub fn update_worker_set(deps: DepsMut, env: Env) -> Result<Response, ContractEr
...
+
+            COMMANDS_BATCH.save(deps.storage, &batch.id, &batch)?;
+            REPLY_BATCH.save(deps.storage, &batch.id)?;
+
+            let start_sig_msg = multisig::msg::ExecuteMsg::StartSigningSession {
+                key_id: cur_worker_set.id(), // TODO remove the key_id
+                msg: batch.msg_to_sign(),
+            };
+
+            Ok(Response::new().add_submessage(SubMsg::reply_on_success(
```
haiyizxx, line 202: what about failure case?

contracts/multisig-prover/src/execute.rs
haiyizxx: same question

contracts/multisig-prover/src/execute.rs
```diff
@@ -126,31 +128,123 @@ pub fn update_worker_set(deps: DepsMut, env: Env) -> Result<Response, ContractEr
...
+                    s.address.to_string(),
+                    (KeyType::Ecdsa, s.pub_key.as_ref().into()),
+                )
+            })
+            .collect(),
     };
 
-    // TODO handle the reply
-    Ok(Response::new().add_message(wasm_execute(config.multisig, &start_sig_msg, vec![])?))
+    Ok(Response::new().add_message(wasm_execute(config.multisig, &key_gen_msg, vec![])?))
```
haiyizxx, line 247: why remove the todo?

contracts/multisig-prover/src/state.rs
```diff
@@ -36,19 +37,20 @@ pub const REPLY_BATCH: Item<BatchID> = Item::new("reply_tracker");
 pub struct WorkerSet {
     pub signers: BTreeSet<Signer>,
     pub threshold: Uint256,
-    pub nonce: u64, // for randomness
+    // for hash uniqueness. The same exact worker set could be in use at two different times,
+    // and we need to be able to distinguish between the two
+    pub nonce: u64,
```
haiyizxx, line 42: I prefer to change to `created_at` if we use current block height in `new`, or have a counter to track the nonce.

contracts/multisig-prover/src/execute.rs
eguajardo: do we want to handle the failure? I think it makes more sense if it reverts which is the case for failed submessages set to reply on success

haiyizxx: good to know 👍

contracts/multisig-prover/src/execute.rs
eguajardo: this is not a submessage so it doesn't get a reply

contracts/multisig-prover/src/execute.rs
```diff
@@ -126,31 +128,123 @@ pub fn update_worker_set(deps: DepsMut, env: Env) -> Result<Response, ContractEr
...
+        Some(cur_worker_set) => {
+            if !should_update_worker_set(
+                &new_worker_set,
+                &cur_worker_set,
+                config.worker_set_diff_threshold as usize,
+            ) {
+                return Err(ContractError::WorkerSetUnchanged);
+            }
+
+            NEXT_WORKER_SET.save(
```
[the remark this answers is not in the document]
eguajardo: good one, I believe we need to revert because if I'm not mistaken the external gateway only allows transfer operatorship to continue when the proof was signed by the current set: https://github.com/axelarnetwork/axelar-cgp-solidity/blob/main/contracts/AxelarGateway.sol#L502. We need to confirm the pending one before trying to confirm a new one or the new one will never be accepted by the external gateway. Is that right @cjcobb23 ?

contracts/multisig-prover/src/execute.rs
eguajardo: changed to `.id()` since key_id will be removed in another PR to use worker set id instead. Then `.id()` will make more sense

contracts/multisig-prover/src/execute.rs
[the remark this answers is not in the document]
eguajardo: I'm just wondering if we change to revert, would there be a scenario where we could get the worker set rotation stuck if one set couldn't be confirmed or something

contracts/multisig-prover/src/execute.rs
eguajardo: Task to remove key_id: https://axelarnetwork.atlassian.net/browse/AXE-1852

contracts/multisig-prover/src/execute.rs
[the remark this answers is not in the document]
cjcobb23: I think we should revert, and add a TODO item to come up with a way to abort a worker set update if it gets stuck. I can't think of a reason it would get stuck, and we can always migrate the contract if there is some reason that it does. But maybe there is a cleaner way. Either way, I think for this PR, revert is fine, and circle back later.

We should allow resigning the same worker set update though. I could see the proof id getting lost, or the signing event being missed for whatever reason, and it would be useful to be able to retrigger the event. So this would mean only revert if `NEXT_WORKER_SET` is not empty, and the stored value is different than the value we are trying to store. If this ends up being complicated, I'm ok leaving this for another PR too.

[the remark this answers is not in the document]
eguajardo: changed to revert

contracts/multisig-prover/src/execute.rs
eguajardo: Task to keep track of this: https://axelarnetwork.atlassian.net/browse/AXE-1869

sammy1991106: @cjcobb23 @eguajardo A lot of the solidity contracts were implemented with assumptions from the axelar-core's consensus code, which no longer holds true for the new cosmwasm implementation.

sammy1991106 (approved):

PR #72, opened 2023-08-29
feat(ampd): multisig contract signing started handler

haiyizxx:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


ampd/src/handlers/multisig.rs
```diff
@@ -72,21 +79,126 @@ impl TryFrom<&event_sub::Event> for Option<SigningStartedEvent> {
...
+{
+    type Err = Error;
+
+    async fn handle(&self, event: &event_sub::Event) -> error_stack::Result<(), Error> {
+        let SigningStartedEvent {
+            contract_address,
+            session_id,
+            pub_keys,
+            msg,
+        } = match event.try_into().into_report()? {
```
sammy1991106, line 148: use the macro that Chris created.

ampd/src/handlers/multisig.rs
```diff
@@ -72,21 +79,126 @@ impl TryFrom<&event_sub::Event> for Option<SigningStartedEvent> {
...
+                msg,
+                &pub_keys[&self.worker],
+            )
+            .await
+            .change_context(Error::Sign)?;
+
+        self.broadcast_signature(session_id.into(), signature.into())
+            .await?;
+
+        return Ok(());
```
sammy1991106, line 174:
```suggestion
        match pub_keys.get(&self.worker) {
            Some(pub_key) => {
                let signature = self
                    .signer
                    .sign(self.multisig.to_string().as_str(), msg, &pub_key)
                    .await
                    .change_context(Error::Sign)?;

                self.broadcast_signature(session_id.into(), signature.into())
                    .await?;

                Ok(())
            }
            None => Ok(()),
        }
```

ampd/src/handlers/multisig.rs
```diff
@@ -72,21 +79,126 @@ impl TryFrom<&event_sub::Event> for Option<SigningStartedEvent> {
...
+
+    async fn broadcast_signature(
+        &self,
+        session_id: Uint64,
+        signature: HexBinary,
+    ) -> error_stack::Result<(), Error> {
+        let msg = serde_json::to_vec(&ExecuteMsg::SubmitSignature {
+            session_id,
+            signature,
+        })
```
sammy1991106, line 118:
```suggestion
    async fn broadcast_signature(
        &self,
        session_id: impl Into<Uint64>,
        signature: impl Into<HexBinary>,
    ) -> error_stack::Result<(), Error> {
        let msg = serde_json::to_vec(&ExecuteMsg::SubmitSignature {
            session_id: session_id.into(),
            signature: signature.into(),
        })
```

sammy1991106 (changes requested):

ampd/src/handlers/multisig.rs
haiyizxx: was not merged before, will update

ampd/src/config.rs
```diff
@@ -18,6 +19,7 @@ pub struct Config {
     #[serde(with = "hex")]
     pub private_key: ECDSASigningKey,
     pub event_buffer_cap: usize,
+    pub multisig_contract: Option<TMAddress>,
```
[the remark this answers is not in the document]
haiyizxx: We impl Default for config, there is no default value for a contract address. 
It also allows someone run apmd if he wants to use some other proofing mechanism.

ampd/src/handlers/multisig.rs
```diff
@@ -52,23 +60,126 @@ where
...
+        } = match event.try_into() as error_stack::Result<_, _> {
+            Err(report) if matches!(report.current_context(), EventTypeMismatch(_)) => {
+                return Ok(());
+            }
+            result => result.change_context(DeserializeEvent)?,
+        };
+
+        if self.multisig != contract_address {
+            return Ok(());
+        }
```
cjcobb23, line 138: I think we need to check the prover address that triggered this signing somehow. I don't think we should sign any data submitted to the multisig contract without knowing where that data came from and what it is.

cjcobb23 (changes requested):

ampd/src/handlers/multisig.rs
haiyizxx: if we don't want permissionless signing, we should restrict on the smart contract level, not in ampd

cjcobb23: as in restrict which contracts can trigger signing sessions? We could require contracts to be approved via governance vote before they can trigger signing sessions. We'd need to add a new execute msg to the multisig contract to authorize contracts.

ampd/src/handlers/multisig.rs
```diff
@@ -52,23 +60,126 @@ where
...
+
```
cjcobb23, line 139: Can we add a logline here? Just something indicating a signing started event was received. Ampd was quite silent last time I ran it, and it was pretty hard to figure out if it was working or not

ampd/src/handlers/multisig.rs
```diff
@@ -52,23 +60,126 @@ where
...
+                    .sign(self.multisig.to_string().as_str(), msg, pub_key)
+                    .await
+                    .change_context(Error::Sign)?;
+
+                self.broadcast_signature(session_id.into(), signature.into())
+                    .await?;
+
+                Ok(())
+            }
+            None => Ok(()),
```
cjcobb23: might be useful to add some logging here too, possibly in both match cases

sammy1991106 (approved):

cgorenflo (approved):

cjcobb23 (approved):

PR #73, opened 2023-08-29
chore: upgrade error-stack version and remove deprecated dependencies

cgorenflo:


ampd/src/broadcaster/accounts.rs
```diff
@@ -30,13 +30,14 @@ where
...
         .account
-        .ok_or_else(|| Error::AccountNotFound {
-            address: address.clone(),
+        .ok_or_else(|| {
+            Error::AccountNotFound {
+                address: address.clone(),
+            }
+            .into()
         })
-        .into_report()
```
sammy1991106, line 36: One of the other reasons I didn't upgrade was also that I found `.into_report` quite nice haha. Anyway.

packages/report/src/result_ext.rs
```diff
@@ -0,0 +1,22 @@
+use error_stack::{Context, IntoReportCompat, Report};
+use eyre::Report as EyreReport;
+
+pub trait ResultCompatExt {
```
sammy1991106, line 4: nice!

packages/axelar-wasm-std/src/lib.rs
```diff
@@ -3,6 +3,7 @@ pub mod flagset;
 mod fn_ext;
 pub mod nonempty;
 pub mod operators;
+
```
sammy1991106, line 6: Any reason for adding this empty line? Just curious

sammy1991106 (approved):

packages/axelar-wasm-std/src/lib.rs
cgorenflo: might be that the error-stack import was on that line and I only deleted the text, not the whole line

haiyizxx (approved):

PR #76, opened 2023-08-31
feat: update snapshots to set constant weight to participants

eguajardo:
## Description
https://axelarnetwork.atlassian.net/browse/AXE-1538

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


packages/axelar-wasm-std/src/snapshot.rs
```diff
@@ -11,6 +11,22 @@ pub struct Participant {
     pub weight: nonempty::Uint256,
 }
 
+#[cw_serde]
+pub struct NonWeightedParticipant {
```
sammy1991106, line 15: Do we see a future where weighted and non-weighted participants both exist in the system? If no, it makes no sense to me to create this struct.

contracts/service-registry/src/state.rs
```diff
@@ -36,16 +36,24 @@ pub struct Worker {
...
                 weight: amount.try_into()?,
             }),
-            _ => Err(ContractError::InvalidBondingState(self.bonding_state)),
+            _ => Err(ContractError::InvalidBondingState(worker.bonding_state)),
+        }
+    }
+}
+
+impl From<Worker> for NonWeightedParticipant {
+    fn from(worker: Worker) -> Self {
```
sammy1991106, line 54: How can it be the case that only bonded workers can become a weighted participant but non-bonded workers can become a non-weighted participant?

sammy1991106 (changes requested):

sammy1991106:
The size of this PR is perfect!

contracts/service-registry/src/state.rs
```diff
@@ -36,16 +36,24 @@ pub struct Worker {
...
 
-    fn try_into(self) -> Result<Participant, ContractError> {
-        match self.bonding_state {
-            BondingState::Bonded { amount } => Ok(Participant {
-                address: self.address,
+    fn try_from(worker: Worker) -> Result<Participant, ContractError> {
+        match worker.bonding_state {
+            BondingState::Bonded { amount } => Ok(Self {
+                address: worker.address,
                 weight: amount.try_into()?,
```
sammy1991106, line 46: To me, this PR should have been as simple as here making it weight to be one, with a comment explaining the reason.

contracts/service-registry/src/state.rs
```diff
@@ -36,16 +36,24 @@ pub struct Worker {
     pub service_name: String,
 }
 
-impl TryInto<Participant> for Worker {
```
sammy1991106, line 39: Does clippy not catch this kind of stuff? `TryInto` should never have been implemented but `TryFrom`.

contracts/service-registry/src/state.rs
eguajardo: this was my original idea, but CJ wanted to have a separate struct for this. I guess it could be useful for some other devs that are developing their own provers and wanna start with constant weights first as well. I don't have a strong preference one way or the other

contracts/service-registry/src/state.rs
eguajardo: Found this: https://rust-lang.github.io/rust-clippy/master/index.html#/from_over_into, but I did some tests and apparently clippy will catch `Into` but not `TryInto` lol

contracts/service-registry/src/state.rs
sammy1991106: Maybe, but I don't like having useless but potentially-useful code around. It's confusing. People would question why it's there every time they see it.

contracts/service-registry/src/state.rs
```diff
@@ -36,16 +36,18 @@ pub struct Worker {
...
-    fn try_into(self) -> Result<Participant, ContractError> {
-        match self.bonding_state {
-            BondingState::Bonded { amount } => Ok(Participant {
-                address: self.address,
-                weight: amount.try_into()?,
+    fn try_from(worker: Worker) -> Result<Participant, ContractError> {
+        match worker.bonding_state {
+            BondingState::Bonded { amount: _ } => Ok(Self {
+                address: worker.address,
+                weight: Uint256::one()
```
sammy1991106, line 46: Can you add a comment here explaining the `one` weight?

sammy1991106 (approved):

contracts/service-registry/src/state.rs
```diff
@@ -36,16 +36,18 @@ pub struct Worker {
...
-            BondingState::Bonded { amount } => Ok(Participant {
-                address: self.address,
-                weight: amount.try_into()?,
+    fn try_from(worker: Worker) -> Result<Participant, ContractError> {
+        match worker.bonding_state {
+            BondingState::Bonded { amount: _ } => Ok(Self {
+                address: worker.address,
+                weight: Uint256::one() // Weight is set to one to ensure all workers have same weight. In future it should be derived from amount bonded
+                    .try_into()
+                    .expect("violated invariant: weight is zero"),
```
cgorenflo, line 48:
```suggestion
                    .expect("violated invariant: weight must not be zero"),
```
Otherwise it's not clear if the value is currently zero or if the invariant is that is should always be zero

cgorenflo (approved):

PR #80, opened 2023-09-05
feat: event decoding for evm verify worker set handler

sammy1991106:
## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

### DIFF
--- a/ampd/src/handlers/evm_verify_worker_set.rs
+++ b/ampd/src/handlers/evm_verify_worker_set.rs
@@ -13,7 +13,7 @@ pub struct Operators {
 }
 
 #[derive(Deserialize, Debug)]
-pub struct WorkerSet {
+pub struct WorkerSetConfirmation {
     pub tx_id: Hash,
     pub log_index: u64,
     pub operators: Operators,
@@ -25,7 +25,7 @@ pub struct WorkerSet {
 struct PollStartedEvent {
     #[serde(rename = "_contract_address")]
     contract_address: TMAddress,
-    worker_set: WorkerSet,
+    worker_set: WorkerSetConfirmation,
     poll_id: PollID,
     source_chain: connection_router::types::ChainName,
     source_gateway_address: EVMAddress,
--- a/ampd/src/handlers/evm_verify_worker_set.rs
+++ b/ampd/src/handlers/evm_verify_worker_set.rs
@@ -1,0 +1,124 @@
+use error_stack::ResultExt;
+use serde::Deserialize;
+
+use axelar_wasm_std::voting::PollID;
+use events_derive::try_from;
+
+use crate::types::{EVMAddress, Hash, TMAddress, U256};
+
+#[derive(Deserialize, Debug)]
+pub struct Operators {
+    pub weights: Vec<(EVMAddress, U256)>,
+    pub threshold: U256,
+}
+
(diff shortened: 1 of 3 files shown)


ampd/src/handlers/evm_verify_worker_set.rs
```diff
@@ -0,0 +1,124 @@
+use error_stack::ResultExt;
+use serde::Deserialize;
+
+use axelar_wasm_std::voting::PollID;
+use events_derive::try_from;
+
+use crate::types::{EVMAddress, Hash, TMAddress, U256};
+
+#[derive(Deserialize, Debug)]
+pub struct Operators {
```
cgorenflo, line 10: what's the difference between a worker and operator?

ampd/src/handlers/evm_verify_worker_set.rs
```diff
@@ -0,0 +1,124 @@
...
+    pub weights: Vec<(EVMAddress, U256)>,
+    pub threshold: U256,
+}
+
+#[derive(Deserialize, Debug)]
+pub struct WorkerSet {
+    pub tx_id: Hash,
+    pub log_index: u64,
```
cgorenflo, line 18: what is this information here? is that for verifying that a worker set has been updated?

ampd/src/handlers/evm_verify_worker_set.rs
sammy1991106: My understanding is that workers are anyone running external services (voting, signing, maybe others as well) and operators are those signer addresses set in the EVM gateway contract. I don't like these names either, but I'm following the cosmowasm contracts' events for now.

ampd/src/handlers/evm_verify_worker_set.rs
```diff
@@ -0,0 +1,124 @@
...
+    #[serde(rename = "_contract_address")]
+    contract_address: TMAddress,
+    worker_set: WorkerSet,
+    poll_id: PollID,
+    source_chain: connection_router::types::ChainName,
+    source_gateway_address: EVMAddress,
+    confirmation_height: u64,
+    expires_at: u64,
+    participants: Vec<TMAddress>,
+}
```
cgorenflo, line 35: do we need a completely separate event?

sammy1991106: Maybe we can unify with an enum later, but I'd prefer going with everything separate and then come back to refactor if possible.

ampd/src/handlers/evm_verify_worker_set.rs
sammy1991106: This is the information for us to locate the EVM event and do comparison for voting.

cgorenflo: Maybe call it `WorkerSetConfirmation` or something like that? It's weird to call it just `WorkerSet` with that information

ampd/src/handlers/evm_verify_worker_set.rs
cgorenflo: ok

haiyizxx (approved):

cgorenflo (approved):

PR #86, opened 2023-09-06
feat: add second subcommand to ampd

cgorenflo:


ampd/src/main.rs
```diff
@@ -59,6 +77,12 @@ async fn main() -> ExitCode {
     code
 }
 
+fn register_worker(args: Args) -> ExitCode {
+    println!("registering worker");
+    println!("args: {:?}", args);
+    ExitCode::SUCCESS
```
cjcobb23, line 83: Should there be a `TODO` here?

cgorenflo:
we'll revisit this in a less ad hoc manner

PR #93, opened 2023-09-12
chore: fail tests when there's warnings

fahimahmedx:
## Description
[Details](https://axelarnetwork.atlassian.net/jira/software/c/projects/AXE/boards/18?modal=detail&selectedIssue=AXE-1833)

Fail tests if there is a warning

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

### DIFF
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -8,7 +8,7 @@ on:
       - 'releases/**'
 
 env: 
-  RUSTFLAGS: "-D warnings"
+  RUSTFLAGS: "-D warnings --cfg tracing_unstable"
 
 name: Basic
 
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -53,7 +53,7 @@ jobs:
           command: wasm
           args: --locked --workspace --exclude ampd
         env:
-          RUSTFLAGS: "$-D warnings --cfg tracing_unstable -C link-arg=-s"
+          RUSTFLAGS: "-D warnings --cfg tracing_unstable -C link-arg=-s"
 
   lints:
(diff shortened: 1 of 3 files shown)


cgorenflo (approved):

cgorenflo (commented):
Why do you define RUSTFLAGS globally and inside the test job?

.github/workflows/basic.yml
```diff
@@ -45,7 +48,7 @@ jobs:
           command: wasm
           args: --locked --workspace --exclude ampd
```
milapsheth, line 49: ampd is being excluded here. should that be checked separately?

fahimahmedx: CJ mentions it's because the command is for compiling wasm contracts, and ampd doesn't have any wasm contracts.

contracts/multisig/src/msg.rs
```diff
@@ -38,7 +38,7 @@ pub enum QueryMsg {
     #[returns(Multisig)]
     GetMultisig { session_id: Uint64 },
 
-    #[returns(Key)]
```
sammy1991106, line 41: Do you know why Github keeps displaying these warnings, including the one below?

fahimahmedx: I tried researching about this when I was solving it but got no lead. My assumption is that because the module is only being used as a datatype (and not actually being initialized or used as a variable for example), then it doesn't actually use anything from the library itself.

cgorenflo (approved):

milapsheth (approved):

PR #100, opened 2023-09-14
feat(multisig-prover)!: add support for different encoding languages

cjcobb23:
Allows specification of encoding language when instantiating the multisig-prover. Moves all evm/abi specific encoding into it's own module. 

Originally, I thought of doing this with feature flags, but there are a number of drawbacks to using feature flags.
1. No support for mutually exclusive feature flags.
2. Complicates the build process. We would need to build multiple copies of the same contract, and it's unclear if that is even possible with the rust optimizer for cosmwasm.
3. Complicates tests and tools like Clippy. All of these things just assume the default feature set.

https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features
https://doc.rust-lang.org/cargo/reference/features.html#feature-combinations

So here we just pass an enum at instantiation to specify the encoding language, which is stored in the config object.

https://axelarnetwork.atlassian.net/browse/AXE-1862


contracts/multisig-prover/src/encoding/abi.rs
```diff
@@ -1,189 +1,138 @@
...
+
+            if let Some(signature) = &operator.signature {
+                signatures.push(Token::Bytes(<Vec<u8>>::from(signature.clone())));
+            }
 
-        Ok(calldata.into())
-    }
+            (addresses, weights, signatures)
+        },
+    );
```
cgorenflo, line 99:
```
        let (addresses, weights, signatures): (Vec<_>, Vec<_>, Vec<_>) = operators
            .into_iter()
            .map(|op| {
                (
                    Token::Address(ethereum_types::Address::from_slice(op.address.as_slice())),
                    Token::Uint(ethereum_types::U256::from_big_endian(
                        &op.weight.to_be_bytes(),
                    )),
                    op.signature.map(|sig| Token::Bytes(<Vec<u8>>::from(sig))),
                )
            })
            .multiunzip();
            
        let signatures = signatures.into_iter().filter_map(identity).collect();
```

contracts/multisig-prover/src/encoding/abi.rs
```diff
@@ -1,189 +1,138 @@
...
-}
+    let (commands_ids, commands_types, commands_params) = data.commands.iter().fold(
+        (vec![], vec![], vec![]),
+        |(mut commands_ids, mut commands_types, mut commands_params), command| {
+            commands_ids.push(Token::FixedBytes(command.id.to_vec()));
+            commands_types.push(Token::String(command.ty.to_string()));
+            commands_params.push(Token::Bytes(command.params.to_vec()));
+            (commands_ids, commands_types, commands_params)
+        },
+    );
```
cgorenflo, line 33:
```suggestion
let (commands_ids, commands_types, commands_params) = data
            .commands
            .iter()
            .map(|command| {
                (
                    Token::FixedBytes(command.id.to_vec()),
                    Token::String(command.ty.to_string()),
                    Token::Bytes(command.params.to_vec()),
                )
            })
            .multiunzip();
```

contracts/multisig-prover/src/encoding/abi.rs
```diff
@@ -1,189 +1,138 @@
...
-        ty: CommandType::TransferOperatorship,
-        params,
-        id: worker_set.hash(),
-    })
-}
+pub const GATEWAY_EXECUTE_FUNCTION_NAME: &str = "execute";
 
-impl TryFrom<Signer> for Operator {
-    type Error = ContractError;
+pub(super) fn encode(data: &Data) -> HexBinary {
```
sammy1991106, line 20: What's the thought here for making this function visible in the parent mod?

contracts/multisig-prover/src/encoding/mod.rs
```diff
@@ -1,4 +1,192 @@
...
+
+use crate::{
+    error::ContractError,
+    state::WorkerSet,
+    types::{BatchID, Command, CommandBatch, CommandType},
+};
+
+#[cw_serde]
+#[derive(Copy)]
+pub enum EncodingLanguage {
```
sammy1991106, line 18:
```suggestion
pub enum Encoder {
```

sammy1991106: These aren't really languages.

contracts/multisig-prover/src/encoding/mod.rs
```diff
@@ -1,4 +1,192 @@
...
+    }
+
+    pub fn add_message(&mut self, msg: Message) -> Result<(), ContractError> {
+        self.message_ids.push(msg.id.clone());
+        self.commands.push(make_command(msg, self.encoding)?);
+        Ok(())
+    }
+
+    pub fn add_new_worker_set(&mut self, worker_set: WorkerSet) -> Result<(), ContractError> {
+        self.new_worker_set = Some(worker_set.clone());
```
sammy1991106, line 82: I still don't like it that worker set update is treated differently than messages, but I understand it's not what this PR is concerned about.

contracts/multisig-prover/src/types.rs
```diff
@@ -81,6 +84,7 @@ pub struct CommandBatch {
     pub id: BatchID,
     pub message_ids: Vec<String>,
     pub data: Data,
+    pub encoding: EncodingLanguage,
```
sammy1991106, line 87:
```suggestion
    pub encoder: Encoder,
```

contracts/multisig-prover/src/encoding/abi.rs
cjcobb23: the parent mod has `pub fn encode(&self, encoding: EncodingLanguage)` function that calls this function. Other mods should just call the parent mod function.

sammy1991106: But do you need to make this `super` here to do that? The parent only always import the child, no?

cgorenflo: if he wouldn't the parent couldn't call `ecode`. Children can call private functions of their parents, not the other way around

contracts/multisig-prover/src/encoding/abi.rs
```diff
@@ -1,189 +1,136 @@
...
+                    &op.weight.to_be_bytes(),
+                )),
+                op.signature.map(|sig| Token::Bytes(<Vec<u8>>::from(sig))),
+            )
+        })
+        .multiunzip();
 
-        Ok(calldata.into())
-    }
+    let signatures = signatures.into_iter().flatten().collect();
```
cgorenflo, line 99: nice! Didn't know flatten would do that for options 👍

cgorenflo (approved):

contracts/multisig-prover/src/encoding/abi.rs
sammy1991106: @cgorenflo What I was trying to say is that the child doesn't seem to have to be exported as a pub mod in the parent, and then these functions here can just be `pub` instead of `pub(super)`. It looks good now though.

sammy1991106 (approved):

contracts/multisig-prover/src/encoding/abi.rs
cjcobb23: clippy actually caught this

PR #101, opened 2023-09-15
fix: correctly read and persist state to disk

cgorenflo:
https://axelarnetwork.atlassian.net/browse/AXE-1871


cjcobb23:
Something is wrong with the commit history here. The commits are the same as on main, but the hashes are different, so the PR is showing them as new commits. I think you need to rebase onto main, and remove those duplicate commits.

ampd/src/state.rs
```diff
@@ -78,104 +68,137 @@ impl StateUpdater {
...
-        self.flush()
+        self.state
     }
 
-    fn flush(self) -> Result<(), Error> {
-        info!("persisting state to disk");
-
-        let state =
-            serde_json::to_string(&self.state).change_context(Error::SerializationFailure)?;
+    pub fn set_public_key(&mut self, pub_key: PublicKey) {
```
sammy1991106, line 80: Lol I quite liked `AsMut` but it's ok to do it this way as well.

ampd/src/main.rs
```diff
@@ -119,16 +119,19 @@ fn parse_config(
...
 
     dirs::home_dir().map_or(path.to_path_buf(), |home| home.join(home_subfolder))
 }
+
+#[derive(Error, Debug)]
+enum Error {
+    #[error("failed to load config")]
+    LoadConfig,
+    #[error("catastrophic failure")]
+    Ampd,
```
sammy1991106, line 136:
```suggestion
    Panic,
```
or 
```suggestion
    Catastrophic,
```

ampd/src/lib.rs
```diff
@@ -20,7 +21,7 @@ use tofnd::grpc::{MultisigClient, SharableEcdsaClient};
 use types::TMAddress;
 
 use crate::config::Config;
-use crate::error::Error;
```
sammy1991106, line 23: But this mod is still there?

ampd/src/lib.rs
```diff
@@ -275,7 +281,24 @@ where
         };
 
         while (set.join_next().await).is_some() {}
-
-        res
+        state_rx.try_recv().change_context(Error::ReturnState)
```
sammy1991106, line 284: Now this run function wouldn't return back the error from the first terminated task, if any. I don't think this is right.

sammy1991106 (changes requested):

ampd/src/main.rs
cjcobb23, line 136: or
```suggestion
    Fatal
```

sammy1991106: Ah I like that! @cjcobb23

ampd/src/lib.rs
cgorenflo: what are you asking? The error cases have been merged into the lib mod (because they are the errors of this mod) and `LoggableError` and reporting extensions have been moved into the `packages` folder for reuse in contracts in a different PR

cgorenflo:
> Something is wrong with the commit history here. The commits are the same as on main, but the hashes are different, so the PR is showing them as new commits. I think you need to rebase onto main, and remove those duplicate commits.

Don't worry about that. This branch was originally built on top of another and those commits have been merged into main. Because of squashing git won't find the exact commits, but the diffs are still correct. Squashing this PR during merge will sort it out again

ampd/src/lib.rs
```diff
@@ -259,13 +260,18 @@ where
...
+        let (state_tx, mut state_rx) = oneshot::channel::<State>();
         let mut set = JoinSet::new();
         set.spawn(event_sub.run().change_context(Error::EventSub));
         set.spawn(event_processor.run().change_context(Error::EventProcessor));
         set.spawn(broadcaster.run().change_context(Error::Broadcaster));
-        set.spawn(state_updater.run().change_context(Error::StateUpdater));
+        set.spawn(async move {
+            state_tx
+                .send(state_updater.run().await)
+                .map_err(|_| report!(Error::ReturnState))
```
sammy1991106, line 271: Why not panic here?

ampd/src/lib.rs
```diff
@@ -275,7 +281,34 @@ where
         };
 
         while (set.join_next().await).is_some() {}
+        let state_update_result = state_rx.try_recv().change_context(Error::ReturnState);
```
sammy1991106, line 285: `try_recv` may return `Err(Empty)` if no pending value exists in the channel at the moment. How can we be sure that the state_updaer task has exited at this moment? It may be still waiting for the handlers to terminate first.

ampd/src/lib.rs
```diff
@@ -275,7 +281,34 @@ where
...
 
-        res
+        match (execution_result, state_update_result) {
+            (Err(mut err1), Err(err2)) => {
+                err1.extend_one(err2);
+                Err(err1)
+            }
+            (Err(err1), Ok(_)) => Err(err1),
+            (Ok(_), Err(err2)) => Err(err2),
```
sammy1991106, line 292: We still need to return the latest state to persist though even in the case of `execution_result` being an error.

ampd/src/main.rs
```diff
@@ -119,16 +136,19 @@ fn parse_config(
...
     };
 
     dirs::home_dir().map_or(path.to_path_buf(), |home| home.join(home_subfolder))
 }
+
+#[derive(Error, Debug)]
+enum Error {
+    #[error("failed to load config")]
+    LoadConfig,
+    #[error("catastrophic failure")]
```
cjcobb23, line 152:
```
#[error("fatal failure")]
```
catastrophic failure sounds a little extreme. Would make me worry that the cpu melted down or something like that.

ampd/src/state.rs
```diff
@@ -78,104 +72,137 @@ impl StateUpdater {
...
 
-        fs::write(self.state_path, state).change_context(Error::WriteFailure)?;
+pub fn load(path: impl AsRef<Path>) -> Result<State, Error> {
+    info!("loading state from disk");
 
-        Ok(())
+    match fs::read_to_string(path) {
+        Ok(state) => serde_json::from_str(&state).change_context(Error::InvalidState),
+        Err(_) => {
+            info!("state file does not exist, starting to track current state");
```
cjcobb23, line 97:
```
info!("state file does not exist, starting from current blockchain state");
```
I feel like the current wording is a bit confusing

cjcobb23 (approved):
Minor comments, but otherwise looks good

ampd/src/lib.rs
```diff
@@ -259,23 +271,55 @@ where
...
 
-        let res = match (set.join_next().await, token.is_cancelled()) {
+        let execution_result = match (set.join_next().await, token.is_cancelled()) {
             (Some(result), false) => {
                 token.cancel();
-                result.change_context(Error::Task)?
+                match result {
+                    Ok(result) => result,
+                    Err(err) => Err(err).change_context(Error::Task),
+                }
```
sammy1991106, line 293:
```suggestion
                result.unwrap_or_else(|err| Err(err).change_context(Error::Task))
```

sammy1991106 (approved):

sammy1991106 (approved):

ampd/src/lib.rs
```diff
@@ -259,23 +271,55 @@ where
...
+            Ok(())
+        });
 
-        let res = match (set.join_next().await, token.is_cancelled()) {
+        let execution_result = match (set.join_next().await, token.is_cancelled()) {
             (Some(result), false) => {
                 token.cancel();
-                result.change_context(Error::Task)?
+                match result {
+                    Ok(result) => result,
```
sammy1991106, line 291: I know this PR didn't change this part, but now I'm thinking about how could any of the tasks exit with OK if the token isn't cancelled.

ampd/src/lib.rs
```diff
@@ -40,7 +42,19 @@ const PREFIX: &str = "axelar";
...
+    let app = prepare_app(cfg, state).await;
+
+    match app {
+        Ok(app) => app
+            .run()
+            .await
+            .then(|(state, report)| (Some(state), report)),
+        Err(err) => (None, Err(err)),
+    }
+}
```
sammy1991106, line 55:
```suggestion
pub async fn run(cfg: Config, state: State) -> (State, Result<(), Error>) {
    let app = prepare_app(cfg, state.clone()).await;

    match app {
        Ok(app) => app.run().await.then(|(state, report)| (state, report)),
        Err(err) => (state, Err(err)),
    }
}
```
don't have a very strong opinion but this may also help simpliy code in main.rs. I'll leave it up to you.

ampd/src/lib.rs
cgorenflo: It's the correct behaviour. It just happens that currently we don't have any tasks that could shut down ampd, so we won't end up in this branch. But I'd rather keep it then break the app in the future because this match is not doing what it's supposed to

PR #105, opened 2023-09-16
refactor: use axelar-wasm-std ContractError

cgorenflo:
All contracts now return axelar-wasm-std ContractError at the highest level. Going forward, this allows us to stepwise integrate error-stack into each contract in an isolated manner

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -42,6 +42,8 @@ dependencies = [
 name = "aggregate-verifier"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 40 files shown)


cjcobb23 (approved):
LGTM. ~~I am wondering if we should just start returning `axelar_wasm_std::ContractError` everywhere. A lot of times a really low level function can still return an error defined in a different contract.~~ nevermind, error-stack handles this

haiyizxx (approved):

PR #106, opened 2023-09-19
refactor: introduce new Message type that can be deserialized safely

cgorenflo:
AXE-1885-Create-new-message-type

### DIFF
--- a/.github/workflows/basic.yaml
+++ b/.github/workflows/basic.yaml
@@ -8,7 +8,7 @@ on:
       - releases/**
 
 env:
-  RUSTFLAGS: -D warnings -W deprecated --cfg tracing_unstable
+  RUSTFLAGS: -D warnings -A deprecated --cfg tracing_unstable
 
 name: Basic
 
--- a/.github/workflows/basic.yaml
+++ b/.github/workflows/basic.yaml
@@ -67,7 +67,7 @@ jobs:
           command: wasm
           args: --locked --workspace --exclude ampd
         env:
-          RUSTFLAGS: -D warnings --cfg tracing_unstable -C link-arg=-s
+          RUSTFLAGS: -D warnings -A deprecated --cfg tracing_unstable -C link-arg=-s
 
   lints:
     name: Lints
--- a/.github/workflows/basic.yaml
+++ b/.github/workflows/basic.yaml
@@ -8,7 +8,7 @@ on:
       - releases/**
 
 env:
-  RUSTFLAGS: -D warnings --cfg tracing_unstable
+  RUSTFLAGS: -D warnings -W deprecated --cfg tracing_unstable
 
 name: Basic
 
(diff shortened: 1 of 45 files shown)


.github/workflows/basic.yaml
```diff
@@ -8,7 +8,7 @@ on:
       - releases/**
 
 env:
-  RUSTFLAGS: -D warnings --cfg tracing_unstable
+  RUSTFLAGS: -D warnings -W deprecated --cfg tracing_unstable
```
cgorenflo, line 11:
```suggestion
  RUSTFLAGS: -D warnings -A deprecated --cfg tracing_unstable
```

contracts/connection-router/src/error.rs
```diff
@@ -33,8 +33,8 @@ pub enum ContractError {
     #[error("chain is frozen")]
     ChainFrozen { chain: ChainName },
 
-    #[error("address of {0} is invalid")]
-    InvalidAddress(String),
+    #[error("address of is invalid")]
```
cgorenflo, line 36:
```suggestion
    #[error("address is invalid")]
```

contracts/connection-router/src/state.rs
```diff
@@ -77,6 +84,30 @@ impl<'a> IndexList<ChainEndpoint> for ChainEndpointIndexes<'a> {
 // Message represents a message for which the fields have been successfully validated.
 // This should never be supplied by the user.
 #[cw_serde]
+pub struct NewMessage {
+    pub id_on_chain: MessageID, // globally unique
```
sammy1991106, line 88: Consider https://github.com/axelarnetwork/axelar-amplifier/pull/107#discussion_r1330294293

contracts/connection-router/src/types.rs
```diff
@@ -41,9 +44,11 @@ impl From<MessageID> for String {
     }
 }
 
-impl<'a> MessageID {
-    pub fn as_str(&'a self) -> &'a str {
-        self.0.as_str()
+impl Deref for MessageID {
```
sammy1991106, line 47: Great!

contracts/connection-router/src/msg.rs
```diff
@@ -6,6 +6,7 @@ use crate::types::GatewayDirection;
 // Message is a type meant to be used in interfaces where the data can be provided by the user.
 // The fields have not necessarily been validated, and should be checked prior to further processing.
 #[cw_serde]
+#[deprecated(note = "use NewMessage instead")]
```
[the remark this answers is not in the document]
cgorenflo: yes. Of course, once the old one is deleted we can just call it `Message`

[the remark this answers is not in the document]
cjcobb23: I am a bit concerned with the ease of integration, since the types in this message are going to be used extensively by relayers. We would need to make sure to always serialize the types in a way that other languages can easily understand. If they always boil down to just a string or a byte array, that's fine. But we need to make sure that's the case, and document it clearly, etc.

cjcobb23 (approved):

sammy1991106 (approved):

PR #107, opened 2023-09-19
refactor: use new message struct in connection router

cgorenflo:
migrate the connection router to use the new message struct. Needed to introduce some auxiliary conversion functions until the old message structs can be completely eliminated.

### DIFF
--- a/contracts/connection-router/src/contract.rs
+++ b/contracts/connection-router/src/contract.rs
@@ -13,7 +13,7 @@ pub fn instantiate(
     _env: Env,
     _info: MessageInfo,
     msg: InstantiateMsg,
-) -> Result<Response, ContractError> {
+) -> Result<Response, axelar_wasm_std::ContractError> {
     let admin = deps.api.addr_validate(&msg.admin_address)?;
     let governance = deps.api.addr_validate(&msg.governance_address)?;
     CONFIG.save(
@@ -32,7 +32,7 @@ pub fn execute(
     _env: Env,
     info: MessageInfo,
     msg: ExecuteMsg,
-) -> Result<Response, ContractError> {
+) -> Result<Response, axelar_wasm_std::ContractError> {
     match msg {
         ExecuteMsg::RegisterChain {
             chain,
@@ -63,9 +63,11 @@ pub fn execute(
             info,
             msgs.into_iter()
                 .map(Message::try_from)
-                .collect::<Result<Vec<Message>, _>>()?,
+                .collect::<Result<Vec<Message>, _>>()
+                .map_err(axelar_wasm_std::ContractError::from)?,
         ),
     }
+    .map_err(axelar_wasm_std::ContractError::from)
 }
 
 pub mod execute {
--- a/contracts/connection-router/src/contract.rs
+++ b/contracts/connection-router/src/contract.rs
@@ -64,12 +64,12 @@ pub fn execute(
 }
 
 pub mod execute {
-
     use std::vec;
 
-    use axelar_wasm_std::flagset::FlagSet;
     use cosmwasm_std::{Addr, WasmMsg};
     use itertools::Itertools;
+
+    use axelar_wasm_std::flagset::FlagSet;
 
     use crate::state::NewMessage;
     use crate::{
@@ -193,8 +193,7 @@ pub mod execute {
 
         let mut wasm_msgs = vec![];
 
-        for (destination_chain, msgs_by_destination) in
-            &msgs.iter().group_by(|msg| msg.destination_chain.clone())
+        for (destination_chain, msgs) in &msgs.iter().group_by(|msg| msg.destination_chain.clone())
         {
             let destination_chain = chain_endpoints()
                 .may_load(deps.storage, destination_chain)?
@@ -208,9 +207,7 @@ pub mod execute {
 
             wasm_msgs.push(WasmMsg::Execute {
                 contract_addr: destination_chain.gateway.address.to_string(),
-                msg: to_binary(&ExecuteMsg::RouteMessages(
-                    msgs_by_destination.cloned().collect(),
-                ))?,
+                msg: to_binary(&ExecuteMsg::RouteMessages(msgs.cloned().collect()))?,
                 funds: vec![],
             });
         }
--- a/contracts/connection-router/src/contract.rs
+++ b/contracts/connection-router/src/contract.rs
@@ -5,7 +5,7 @@ use cosmwasm_std::{to_binary, Binary, Deps, DepsMut, Env, MessageInfo, Response,
 use crate::error::ContractError;
 use crate::events::{ChainRegistered, RouterInstantiated};
 use crate::msg::{ExecuteMsg, InstantiateMsg, QueryMsg};
-use crate::state::{chain_endpoints, Config, Message, CONFIG};
+use crate::state::{chain_endpoints, Config, CONFIG};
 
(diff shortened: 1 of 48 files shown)


contracts/connection-router/src/events.rs
```diff
@@ -109,46 +101,45 @@ impl From<ChainUnfrozen> for Event {
...
         ]
     }
 }
 
+pub fn make_message_event_new(event_name: &str, msg: NewMessage) -> Event {
+    let attrs: Vec<Attribute> = msg.into();
+    Event::new(event_name).add_attributes(attrs)
+}
+
+#[deprecated(note = "use make_message_event instead")]
```
cgorenflo, line 135:
```suggestion
#[deprecated(note = "use make_message_event_new instead")]
```

contracts/gateway/tests/mock.rs
```diff
@@ -109,7 +109,7 @@ pub fn mock_router_execute(
     match msg {
         connection_router::msg::ExecuteMsg::RouteMessages(msgs) => {
             for msg in msgs {
-                let msg = connection_router::state::Message::try_from(msg)?;
+                let msg: connection_router::state::Message = msg.try_into()?;
```
cgorenflo, line 112:
```suggestion
                let msg = connection_router::state::Message::try_from(msg)?;
```

contracts/connection-router/src/state.rs
```diff
@@ -81,26 +82,61 @@ impl<'a> IndexList<ChainEndpoint> for ChainEndpointIndexes<'a> {
...
 }
 
-// Message represents a message for which the fields have been successfully validated.
-// This should never be supplied by the user.
 #[cw_serde]
 pub struct NewMessage {
-    pub id_on_chain: MessageID, // globally unique
-    pub destination_address: Address,
+    pub id_on_chain: MessageID, // unique per chain
+    pub destination_address: ContractAddress,
```
sammy1991106, line 88: I almost wanna call this `chain_unique_id` or `cuid`, and the other one `cross_chain_unique_id` or `ccuid`.

contracts/connection-router/src/state.rs
```diff
@@ -180,29 +216,29 @@ impl From<Message> for msg::Message {
 }
 
 #[cw_serde]
-pub struct Address(nonempty::String);
+pub struct ContractAddress(nonempty::String);
```
sammy1991106, line 219: Not sure about this rename

contracts/connection-router/src/events.rs
```diff
@@ -109,46 +101,45 @@ impl From<ChainUnfrozen> for Event {
     }
 }
 
+impl From<NewMessage> for Vec<Attribute> {
+    fn from(other: NewMessage) -> Self {
+        vec![
+            ("id", other.id_on_chain).into(),
```
sammy1991106, line 107: Why `other.id_on_chain` here instead of `other.global_id()`?

contracts/connection-router/src/types.rs
```diff
@@ -78,6 +81,66 @@ impl KeyDeserialize for MessageID {
...
+/// cosmwasm cannot serialize tuples, so we need to convert [GlobalMessageId] into a struct
+#[cw_serde]
+struct GlobalMessageIdSerde {
+    chain_name: ChainName,
+    message_id: MessageID,
+}
+
+#[cw_serde]
+#[serde(from = "GlobalMessageIdSerde", into = "GlobalMessageIdSerde")]
+pub struct GlobalMessageId(ChainName, MessageID);
```
sammy1991106, line 93: Why not define these types in state.rs?

contracts/connection-router/src/state.rs
```diff
@@ -81,26 +82,61 @@ impl<'a> IndexList<ChainEndpoint> for ChainEndpointIndexes<'a> {
     }
 }
 
-// Message represents a message for which the fields have been successfully validated.
-// This should never be supplied by the user.
 #[cw_serde]
 pub struct NewMessage {
-    pub id_on_chain: MessageID, // globally unique
-    pub destination_address: Address,
+    pub id_on_chain: MessageID, // unique per chain
```
cjcobb23, line 87: Are we changing this to be only unique per chain now?

contracts/connection-router/src/state.rs
```diff
@@ -81,26 +82,61 @@ impl<'a> IndexList<ChainEndpoint> for ChainEndpointIndexes<'a> {
...
+    pub destination_address: ContractAddress,
     pub destination_chain: ChainName,
     pub source_chain: ChainName,
-    pub source_address: Address,
+    pub source_address: ContractAddress,
```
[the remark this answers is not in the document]
cgorenflo: I was going back and forth on this. It's just that cosmwasm already has an `Addr`, so I wanted to have a clearer distinction. But I'm fine with going back to `Address`

contracts/connection-router/src/state.rs
cgorenflo: I like `chain_unique_id`. I wouldn't go with acronyms when it's not clear what they stand for 😆

contracts/connection-router/src/events.rs
cgorenflo: As far as I can tell, the global ID is only important for us internally to guarantee uniqueness, it's not a property txs usually care about. Also, the chain data would be duplicated, that's more places data can get mangled. So I'd rather have this be the json representation of the message. In fact, it would be ideal if we could do something like `serde_json::to_string(message).to_attributes()`, but that would require parsing the string

contracts/connection-router/src/types.rs
cgorenflo: I think I defined the GlobalMessageId where the MessageID was defined, so that happened to be `types.rs`. I'd like to get rid of this catchall file entirely, but didn't want to put that in the same PR

contracts/connection-router/src/state.rs
cgorenflo: I strongly advocate for it for the following reasons:
- with the global id, the original producer is tied to the inner state checks of our contracts, because it needs to know to create this global id, instead of just putting the unique id it gets from the chain
- because source_chain and id can possibly clash, you can either guarantee safety of the type by making id private and only providing setters/getters (making it awkward to create) or making it easy to create messages but losing safety
- the global id is easily derived. As a bonus, the id is now responsible for it's own representation and acts as a composite key, so we can actually iterate over all messages of a particular chain if we need to

contracts/connection-router/src/contract.rs
```diff
@@ -194,22 +187,17 @@ pub mod execute {
...
-                .or_default()
-                .push(msg.clone().into());
+        if msgs.iter().any(|msg| msg.source_chain != source_chain.name) {
+            return Err(ContractError::WrongSourceChain);
         }
 
         let mut wasm_msgs = vec![];
-        for (destination_chain, msgs) in msgs_by_destination {
+
+        for (destination_chain, msgs_by_destination) in
```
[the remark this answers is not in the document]
sammy1991106: `msgs_of_destination` then

contracts/connection-router/src/state.rs
cjcobb23: I would actually call this `source_chain_id` or something with source_chain, to make it clear this is an id from the source chain

cjcobb23 (approved):

sammy1991106 (approved):

PR #109, opened 2023-09-19
refactor: make global id structured

cgorenflo:


contracts/connection-router/src/events.rs
```diff
@@ -104,8 +104,8 @@ impl From<ChainUnfrozen> for Event {
 impl From<NewMessage> for Vec<Attribute> {
     fn from(other: NewMessage) -> Self {
         vec![
-            ("id", other.id_on_chain).into(),
-            ("source_chain", other.source_chain).into(),
+            ("id", other.id.message_id).into(),
+            ("source_chain", other.id.chain).into(),
```
sammy1991106, line 108: I'd suggest merging these attributes into one as well. 
```suggestion
            ("id", other.id).into(),
```

contracts/connection-router/src/types.rs
```diff
@@ -83,39 +83,14 @@ impl KeyDeserialize for MessageID {
...
-    }
-}
-
-impl GlobalMessageId {
-    pub fn new(chain_name: ChainName, message_id: MessageID) -> Self {
-        Self(chain_name, message_id)
-    }
+pub struct GlobalMessageId {
+    pub chain: ChainName,
+    pub message_id: MessageID,
```
sammy1991106, line 88:
```suggestion
pub struct CrosschainUniqueId {
    pub chain: ChainName,
    pub message_id: ChainUniqueId,
```
I prefer calling things what they are instead of what they are for.

contracts/connection-router/src/types.rs
```diff
@@ -81,63 +81,38 @@ impl KeyDeserialize for MessageID {
...
-    }
-}
-
-impl GlobalMessageId {
-    pub fn new(chain_name: ChainName, message_id: MessageID) -> Self {
-        Self(chain_name, message_id)
-    }
+pub struct CrossChainUid {
+    pub chain: ChainName,
+    pub id: MessageID,
```
sammy1991106, line 88: I still find it weird that the `Message` type has a `uid` field, within which there's a field called `MessageID`.

sammy1991106 (approved):

PR #110, opened 2023-09-19
feat: multisig ed25519

eranrund:
## Description

This PR adds Ed25519 key support to the `multisig` contract.

## Todos

- [X] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Steps to Test

Run `cargo test` - happy to do any manual testing if you provide guidelines/instructions on how you approach that.

## Expected Behaviour

The `multisig` contract should now support Ed25519 keys.

## Other Notes

The `multisig-prover` contract is still hardcoded to Ecdsa keys. I am not sure if it is a priority to change that.

Linking to AXE-1900


contracts/multisig/src/key.rs
```diff
@@ -72,6 +88,10 @@ impl Signature {
             (Signature::Ecdsa(sig), PublicKey::Ecdsa(pub_key)) => {
                 ecdsa_verify(msg.into(), sig.as_ref(), pub_key.as_ref())
             }
+            (Signature::Ed25519(sig), PublicKey::Ed25519(pub_key)) => {
+                ed25519_verify(msg.into(), sig.as_ref(), pub_key.as_ref())
+            }
+            _ => Err(ContractError::KeyTypeMismatch),
```
cgorenflo, line 94: I don't like that this case now swallows any future additions without breaking (i.e. we might forget to extend it). Since it's ensured that the key type matches for both the key and the signature, the match should be modified to match on the key type, which eliminates any ambiguity and makes the discard case unnecessary.

contracts/multisig/src/key.rs
```diff
@@ -284,10 +333,122 @@ mod test {
...
+            Signature::try_from((KeyType::Ecdsa, ecdsa_test_data::signature())).unwrap();
+        let message = MsgToSign::try_from(ecdsa_test_data::message()).unwrap();
         let public_key = PublicKey::try_from((KeyType::Ecdsa, invalid_pub_key)).unwrap();
         let result = signature.verify(&message, &public_key).unwrap();
         assert_eq!(result, false);
     }
 }
+
+#[cfg(test)]
+mod ed25519_tests {
```
cgorenflo, line 346: I like the separation into multiple modules!

cgorenflo (changes requested):
I suggest some minor adjustments, otherwise looks good to me!

contracts/multisig/src/key.rs
eranrund: That makes sense, I'll update.

contracts/multisig/src/key.rs
eranrund: Thank you!

cgorenflo (approved):

contracts/multisig/src/ed25519.rs
```diff
@@ -0,0 +1,12 @@
+// TODO: Logic specific to secp256k1 will most likely be handled by core in the future.
```
cjcobb23, line 1: remove this comment

contracts/multisig/src/ed25519.rs
```diff
@@ -0,0 +1,12 @@
...
+use crate::ContractError;
+
+const ECDSA_SIGNATURE_LEN: usize = 64;
```
cjcobb23, line 4:
```suggestion
const ED25519_SIGNATURE_LEN: usize = 64;
```

cjcobb23 (approved):

PR #111, opened 2023-09-19
refactor: merge state.rs and types.rs

cgorenflo:


contracts/connection-router/src/error.rs
```diff
@@ -1,9 +1,8 @@
+use crate::state::ChainName;
 use axelar_wasm_std_derive::IntoContractError;
 use cosmwasm_std::StdError;
 use thiserror::Error;
```
haiyizxx, line 4:
```suggestion
use axelar_wasm_std_derive::IntoContractError;
use cosmwasm_std::StdError;
use thiserror::Error;

use crate::state::ChainName;
```

contracts/voting-verifier/src/events.rs
```diff
@@ -148,7 +147,7 @@ impl FromStr for MessageByTxEvent {
     }
 }
 
-fn parse_message_id(message_id: &MessageID) -> Result<(String, u64), ContractError> {
+fn parse_message_id(message_id: &MessageId) -> Result<(String, u64), ContractError> {
     // expected format: <source_chain>:<tx_id>:<index>
     let components = message_id.as_str().split(ID_SEPARATOR).collect::<Vec<_>>();
```
haiyizxx, line 152: not sure if it belongs to the scope of this pr. the function will break as message id does not contain <source_chain> anymore

haiyizxx: nvm, I found you updated in the following up pr

haiyizxx (approved):

contracts/connection-router/src/state.rs
```diff
@@ -84,31 +89,23 @@ impl<'a> IndexList<ChainEndpoint> for ChainEndpointIndexes<'a> {
 
 #[cw_serde]
 pub struct NewMessage {
-    pub id_on_chain: MessageID, // unique per chain
+    pub uid: CrossChainUid,
```
cjcobb23, line 92: This could probably just be called `id`. Reads a bit cleaner that way.

contracts/connection-router/src/state.rs
```diff
@@ -246,49 +245,319 @@ impl TryFrom<String> for Address {
...
+        let value = String::from_utf8(value).map_err(StdError::invalid_utf8)?;
+        Ok(Self(value))
+    }
+}
+
+#[cw_serde]
+pub struct CrossChainUid {
+    pub chain: ChainName,
+    pub id: MessageId,
+}
```
cjcobb23, line 316: I feel like `Message` needs to be somewhere in the name of this struct. `CrossChainMessageUid` or `CrossChainMesageId`. Otherwise it sounds like just a generic cross chain identifier, which it's not. It's for messages.

contracts/connection-router/src/state.rs
```diff
@@ -246,49 +245,319 @@ impl TryFrom<String> for Address {
...
+
+    fn from_vec(value: Vec<u8>) -> StdResult<Self> {
+        let value = String::from_utf8(value).map_err(StdError::invalid_utf8)?;
+        Ok(Self(value))
+    }
+}
+
+#[cw_serde]
+pub struct CrossChainUid {
+    pub chain: ChainName,
```
cjcobb23, line 314:
```suggestion
    pub src_chain: ChainName,
```

contracts/connection-router/src/state.rs
sammy1991106: Agree with @cjcobb23

contracts/connection-router/src/state.rs
sammy1991106: I don't think "src" needs to be there. The `CrossChainUid` does not have any notion of src or dest, but just about where the msg was initiated.

sammy1991106 (approved):

contracts/connection-router/src/state.rs
cjcobb23: It is the source chain though. When we need the source chain, we need to get it from here, since we removed `source_chain` from the encapsulating struct. Makes it clearer and easier to understand to just call it `source_chain` or `src_chain`.

contracts/connection-router/src/error.rs
cgorenflo: `cargo fmt` 😆

cgorenflo: no idea why that happened, fixing it

contracts/connection-router/src/state.rs
cgorenflo: yeah, there is going to be some renaming in the follow-ups

contracts/connection-router/src/state.rs
cgorenflo: Let's address the naming at the end of this string of refactorings 😆

contracts/connection-router/src/state.rs
cgorenflo: not sure where I land on this yet. It can be a bit confusing where to get the source chain from if it's not called that. On the other hand the id doesn't really have a notion of source, as Sammy pointed out. So it's kind of weird to to name it `source_chain`. I'd prefer to have another look at the naming once the code restructuring is completed

PR #112, opened 2023-09-20
refactor: migrate voting verifier to use NewMessage

cgorenflo:


contracts/connection-router/src/state.rs
```diff
@@ -89,7 +89,7 @@ impl<'a> IndexList<ChainEndpoint> for ChainEndpointIndexes<'a> {
 
 #[cw_serde]
 pub struct NewMessage {
-    pub uid: CrossChainUid,
+    pub cc_id: CrossChainId,
```
cjcobb23, line 92: this is a pretty cryptic name for this field, I think this needs a better name. just `id` is fine with me. `uid` is ok. `cc_id` is just too cryptic. This struct is public facing and throughout the entire codebase, so I think it's important to keep it clean.

cjcobb23 (changes requested):
I think the id field needs a better name, otherwise everything else looks good

contracts/connection-router/src/state.rs
haiyizxx, line 92: I thought you wanted to void acronym 😂

haiyizxx (approved):

cjcobb23 (approved):

PR #113, opened 2023-09-20
refactor: migrate aggregate verifier to use NewMessage

cgorenflo:


contracts/gateway/tests/tests.rs
cgorenflo, line 1: this one is a bit of a mess and will get cleaned up in the next PR. Unfortunately, the gateway tests make heavy use of the aggregate verifier interface, so a clean cut for the migration was hard to do

cjcobb23 (approved):

contracts/gateway/src/contract.rs
```diff
@@ -115,8 +119,9 @@ pub mod execute {
             msg: to_binary(&aggregate_verifier::msg::ExecuteMsg::VerifyMessages {
                 messages: unverified
                     .into_iter()
-                    .map(connection_router::msg::Message::from)
-                    .collect(),
+                    .map(|m| m.clone().try_into())
```
sammy1991106, line 122:
```suggestion
                    .map(|m| m.try_into())
```

cgorenflo: it's going to get deleted in the next PR anyway, so ignoring it here 😄

PR #115, opened 2023-09-20
feat(multisig-prover): encode commands via bcs

cjcobb23:
First of the sui encoding PRs. Encode the commands in bcs format.

### DIFF
--- a/contracts/multisig-prover/src/encoding/abi.rs
+++ b/contracts/multisig-prover/src/encoding/abi.rs
@@ -20,7 +20,7 @@ use ethabi::ethereum_types;
 
 pub const GATEWAY_EXECUTE_FUNCTION_NAME: &str = "execute";
 
-pub fn encode(data: &Data) -> HexBinary {
+pub fn encode(data: &Data) -> Result<HexBinary, ContractError> {
     let destination_chain_id = Token::Uint(ethabi::ethereum_types::U256::from_big_endian(
         &data.destination_chain_id.to_be_bytes(),
     ));
@@ -37,17 +37,17 @@ pub fn encode(data: &Data) -> HexBinary {
         })
         .multiunzip();
 
-    ethabi::encode(&[
+    Ok(ethabi::encode(&[
         destination_chain_id,
         Token::Array(commands_ids),
         Token::Array(commands_types),
         Token::Array(commands_params),
     ])
-    .into()
+    .into())
 }
 
-pub fn msg_to_sign(command_batch: &CommandBatch) -> HexBinary {
-    let msg = Keccak256::digest(encode(&command_batch.data).as_slice());
+pub fn msg_to_sign(command_batch: &CommandBatch) -> Result<HexBinary, ContractError> {
+    let msg = Keccak256::digest(encode(&command_batch.data)?.as_slice());
 
     // Prefix for standard EVM signed data https://eips.ethereum.org/EIPS/eip-191
     let unsigned = [
@@ -56,7 +56,7 @@ pub fn msg_to_sign(command_batch: &CommandBatch) -> HexBinary {
     ]
     .concat();
 
-    Keccak256::digest(unsigned).as_slice().into()
+    Ok(Keccak256::digest(unsigned).as_slice().into())
 }
 
 pub fn encode_execute_data(
@@ -65,7 +65,7 @@ pub fn encode_execute_data(
     signers: Vec<(Signer, Option<Signature>)>,
 ) -> Result<HexBinary, ContractError> {
     let param = ethabi::encode(&[
-        Token::Bytes(encode(&command_batch.data).into()),
+        Token::Bytes(encode(&command_batch.data)?.into()),
         Token::Bytes(encode_proof(quorum, signers)?.into()),
     ]);
 
@@ -589,7 +589,8 @@ mod test {
         let data = decode_data(&encoded_data);
         let res = data.encode(Encoder::Abi);
 
-        assert_eq!(res, encoded_data);
+        assert!(res.is_ok());
+        assert_eq!(res.unwrap(), encoded_data);
     }
 
     #[test]
@@ -615,7 +616,8 @@ mod test {
         let res = batch.msg_to_sign();
         let expected_msg = test_data::msg_to_sign();
 
-        assert_eq!(res, expected_msg);
+        assert!(res.is_ok());
+        assert_eq!(res.unwrap(), expected_msg);
     }
 
     #[test]
--- a/contracts/multisig-prover/src/encoding/bcs.rs
+++ b/contracts/multisig-prover/src/encoding/bcs.rs
@@ -1,0 +1,165 @@
+use bcs::to_bytes;
+use cosmwasm_std::HexBinary;
+use itertools::Itertools;
+
+use crate::error::ContractError;
+
+use super::Data;
+
+pub fn command_params(
+    source_chain: String,
+    source_address: String,
+    destination_address: String,
+    payload_hash: HexBinary,
+) -> Result<HexBinary, ContractError> {
+    let ret = to_bytes(&(
+        source_chain,
+        source_address,
+        <[u8; 30]>::try_from(&HexBinary::from_hex(&destination_address)?.to_vec()[..30])
+            .expect("couldn't convert destination_address to 30 byte array"), // TODO: is this right? Why are addresses 30 bytes?
+        payload_hash.to_vec(),
+    ))?;
+
+    Ok(ret.into())
+}
+
+pub fn encode(data: &Data) -> Result<HexBinary, ContractError> {
+    // destination chain id is u64
+    let destination_chain_id = &u64::from_le_bytes(
+        data.destination_chain_id.to_le_bytes()[..8]
+            .try_into()
+            .expect("Couldn't convert u256 to u64"),
+    );
+
+    let (commands_ids, command_types, command_params): (Vec<[u8; 32]>, Vec<String>, Vec<Vec<u8>>) =
+        data.commands
+            .iter()
+            .map(|command| {
+                (
+                    <[u8; 32]>::try_from(&command.id.to_vec()[..32])
+                        .expect("couldn't convert command id to 32 byte array"), // command-ids are fixed length sequences
+                    command.ty.to_string(),
+                    command.params.to_vec(),
+                )
+            })
+            .multiunzip();
+
+    Ok(to_bytes(&(
+        destination_chain_id,
+        commands_ids,
+        command_types,
+        command_params,
+    ))?
+    .into())
+}
+
+#[cfg(test)]
+mod test {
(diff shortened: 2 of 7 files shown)


contracts/multisig-prover/src/encoding/abi.rs
```diff
@@ -20,7 +20,7 @@ use ethabi::ethereum_types;
 
 pub const GATEWAY_EXECUTE_FUNCTION_NAME: &str = "execute";
 
-pub fn encode(data: &Data) -> HexBinary {
+pub fn encode(data: &Data) -> Result<HexBinary, ContractError> {
```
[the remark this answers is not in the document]
cjcobb23: yeah just wanted to match bcs. Probably in the future move all of these into a trait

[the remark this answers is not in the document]
haiyizxx: why not panic on bsc encoding, as we know the commands can be serialized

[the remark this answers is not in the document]
cjcobb23: well do you think we should panic if someone submits a gmp call with a destination address that is not a proper sui address?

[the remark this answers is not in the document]
cjcobb23: oh well for that I actually already panic lol. I actually don't have a strong preference, panic or return an error is fine with me.

contracts/multisig-prover/src/encoding/bcs.rs
```diff
@@ -0,0 +1,165 @@
...
+    source_address: String,
+    destination_address: String,
+    payload_hash: HexBinary,
+) -> Result<HexBinary, ContractError> {
+    let ret = to_bytes(&(
+        source_chain,
+        source_address,
+        <[u8; 32]>::try_from(&HexBinary::from_hex(&destination_address)?.to_vec()[..32])
+            .expect("couldn't convert destination_address to 32 byte array"), // TODO: is this right? Why are addresses 32 bytes?
+        payload_hash.to_vec(),
```
sammy1991106, line 20: I wanna ask the same question. Is this right? Why are addresses 30 bytes?

sammy1991106, line 20: Shouldn't it fail if `destination_address` has more than 30 bytes?

contracts/multisig-prover/src/encoding/bcs.rs
```diff
@@ -0,0 +1,165 @@
...
+    ))?;
+
+    Ok(ret.into())
+}
+
+pub fn encode(data: &Data) -> Result<HexBinary, ContractError> {
+    // destination chain id is u64
+    let destination_chain_id = &u64::from_le_bytes(
+        data.destination_chain_id.to_le_bytes()[..8]
```
sammy1991106, line 29: Again, should it fail if `destination_chain_id` has more than 8 bytes?

contracts/multisig-prover/src/encoding/bcs.rs
```diff
@@ -0,0 +1,165 @@
...
+            .try_into()
+            .expect("Couldn't convert u256 to u64"),
+    );
+
+    let (commands_ids, command_types, command_params): (Vec<[u8; 32]>, Vec<String>, Vec<Vec<u8>>) =
+        data.commands
+            .iter()
+            .map(|command| {
+                (
+                    <[u8; 32]>::try_from(&command.id.to_vec()[..32])
```
sammy1991106, line 39: same here.

contracts/multisig-prover/src/encoding/bcs.rs
cjcobb23: I fixed this but forgot to update the comment somehow. They're 32 bytes

contracts/multisig-prover/src/encoding/abi.rs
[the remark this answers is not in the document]
cjcobb23: I removed the result and just panic now

contracts/multisig-prover/src/encoding/bcs.rs
```diff
@@ -0,0 +1,206 @@
...
+}
+
+pub fn encode(data: &Data) -> HexBinary {
+    let destination_chain_id = chain_id_as_u64(data.destination_chain_id);
+
+    let (commands_ids, command_types, command_params): (Vec<[u8; 32]>, Vec<String>, Vec<Vec<u8>>) =
+        data.commands
+            .iter()
+            .map(|command| {
+                assert!(
```
cgorenflo, line 52: why is there an assert in production code?

contracts/multisig-prover/src/encoding/bcs.rs
sammy1991106: It's only OK to panic if some other stage of the message processing has already validated it. Otherwise, one can send any invalid-according-to-destination-chain message to us and make us panic.

contracts/multisig-prover/src/encoding/bcs.rs
```diff
@@ -0,0 +1,206 @@
...
+    );
+    let ret = to_bytes(&(
+        source_chain,
+        source_address,
+        <[u8; 32]>::try_from(&HexBinary::from_hex(&destination_address)?.to_vec()[..32])
+            .expect("couldn't convert destination_address to 32 byte array"),
+        payload_hash.to_vec(),
+    ))?;
+
+    Ok(ret.into())
```
sammy1991106, line 28:
```suggestion
    let ret = to_bytes(&(
        source_chain,
        source_address,
        <[u8; 32]>::try_from(HexBinary::from_hex(&destination_address)?.to_vec())?,
        payload_hash.to_vec(),
    ))?;

    Ok(ret.into())
```

contracts/multisig-prover/src/encoding/bcs.rs
sammy1991106: https://doc.rust-lang.org/std/primitive.array.html#impl-TryFrom%3CVec%3CT,+Global%3E%3E-for-Box%3C%5BT;+N%5D,+Global%3E @cjcobb23 you don't need to slice it yourself.

contracts/multisig-prover/src/encoding/bcs.rs
cgorenflo: ignore this, apparently that is somewhat common practice in rust

contracts/multisig-prover/src/encoding/bcs.rs
```diff
@@ -0,0 +1,165 @@
...
+    source_chain: String,
+    source_address: String,
+    destination_address: String,
+    payload_hash: HexBinary,
+) -> Result<HexBinary, ContractError> {
+    let ret = to_bytes(&(
+        source_chain,
+        source_address,
+        <[u8; 30]>::try_from(&HexBinary::from_hex(&destination_address)?.to_vec()[..30])
+            .expect("couldn't convert destination_address to 30 byte array"), // TODO: is this right? Why are addresses 30 bytes?
```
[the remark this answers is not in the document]
cjcobb23: lol I fixed this, they are 32 bytes

contracts/multisig-prover/src/encoding/bcs.rs
cjcobb23: So I thought about this more and I realized we shouldn't panic here. This is the only place we check the destination address. The abi module returns an error here if the destination address is not a valid evm address. So we should do the same in this module. There's an argument to be had to move the check to somewhere earlier in the pipeline. But for now, we should check here and return a result.

For command ids, we should panic, because command ids are not user input and should never be invalid.

contracts/multisig-prover/src/encoding/bcs.rs
```diff
@@ -0,0 +1,218 @@
...
+
+    let destination_address = <[u8; 32]>::try_from(
+        HexBinary::from_hex(&destination_address)?.to_vec(),
+    )
+    .map_err(|_| ContractError::InvalidMessage {
+        reason: format!(
+            "destination_address is not a valid Sui address: {}",
+            destination_address
+        ),
+    })?;
```
sammy1991106, line 32: I think it's me to blame, but what is `ret` lol?

contracts/multisig-prover/src/encoding/bcs.rs
```diff
@@ -0,0 +1,218 @@
...
+    assert!(
+        chain_id <= Uint256::from(u64::MAX),
+        "chain_id ({}) is greater than u64 max",
+        chain_id
+    );
+    u64::from_le_bytes(
+        chain_id.to_le_bytes()[..8]
+            .try_into()
+            .expect("Couldn't convert u256 to u64"),
+    )
```
sammy1991106, line 55:
```suggestion
    chain_id
        .to_string()
        .parse()
        .expect("chain_id is invalid")
```

contracts/multisig-prover/src/encoding/bcs.rs
sammy1991106:
```suggestion
command_id
        .to_vec()
        .try_into()
        .expect("couldn't convert command id to 32 byte array")
```

contracts/multisig-prover/src/encoding/bcs.rs
```diff
@@ -0,0 +1,218 @@
...
+        source_address,
+        destination_address,
+        payload_hash.to_vec(),
+    ))?;
+
+    Ok(ret.into())
+}
+
+// destination chain id must be u64 for sui
+fn chain_id_as_u64(chain_id: Uint256) -> u64 {
```
sammy1991106, line 45:
```suggestion
fn uint256_to_u64(chain_id: Uint256) -> u64 {
```

contracts/multisig-prover/src/encoding/bcs.rs
haiyizxx: ya, `command_params` can return error, the `encode` function should panic as we controls all inputs

contracts/multisig-prover/src/encoding/bcs.rs
cjcobb23: lol you hate my short variable names, it's supposed to stand for "return". I changed it though

contracts/multisig-prover/src/encoding/bcs.rs
```diff
@@ -0,0 +1,224 @@
...
+    ))
+    .expect("couldn't serialize command as bcs")
+    .into())
+}
+
+fn u256_to_u64(chain_id: Uint256) -> u64 {
+    chain_id.to_string().parse().expect("chain_id is invalid")
+}
+
+fn make_command_id(command_id: &HexBinary) -> [u8; 32] {
```
haiyizxx, line 48: I feel like

```
struct CommandId([u8;32]);
impl TryFrom<HexBinary> for CommandId {...}
```

sammy1991106 (approved):

PR #118, opened 2023-09-21
refactor: migrate gateway to use NewMessage

cgorenflo:


contracts/gateway/src/contract.rs
```diff
@@ -61,40 +50,37 @@ pub fn execute(
...
     use super::*;
 
-    fn contains_duplicates(msgs: &mut Vec<Message>) -> bool {
+    fn contains_duplicates(msgs: &mut Vec<NewMessage>) -> bool {
         let orig_len = msgs.len();
-        msgs.sort_unstable_by_key(|a| a.id.to_string());
-        msgs.dedup_by(|a, b| a.id == b.id);
+        msgs.sort_unstable_by_key(|a| a.cc_id.to_string());
+        msgs.dedup_by(|a, b| a.cc_id == b.cc_id);
         orig_len != msgs.len()
```
sammy1991106, line 64:
```suggestion
        msgs.iter()
            .map(|msg| msg.cc_id.clone())
            .collect::<HashSet<_>>()
            .len()
            != msgs.len()
```

sammy1991106: The sorting seems unnecessarily costly.

contracts/gateway/src/contract.rs
```diff
@@ -61,40 +50,37 @@ pub fn execute(
...
-        let query_response: Vec<(String, bool)> =
+        let query_response: Vec<(CrossChainId, bool)> =
             deps.querier.query(&QueryRequest::Wasm(WasmQuery::Smart {
                 contract_addr: verifier.to_string(),
                 msg: to_binary(&query_msg)?,
             }))?;
 
         Ok(msgs.into_iter().partition(|m| -> bool {
-            match query_response.iter().find(|r| m.id.to_string() == r.0) {
+            match query_response.iter().find(|r| m.cc_id == r.0) {
```
sammy1991106, line 83: I'm never a fan of these one-letter variable names.

contracts/gateway/src/events.rs
```diff
@@ -1,24 +1,26 @@
...
+    MessageRoutingFailed { msg: NewMessage },
 }
 
 impl From<GatewayEvent> for Event {
     fn from(other: GatewayEvent) -> Self {
         match other {
-            GatewayEvent::MessageVerified { msg } => make_message_event("message_verified", msg),
-            GatewayEvent::MessageRouted { msg } => make_message_event("message_routed", msg),
+            GatewayEvent::MessageVerified { msg } => {
+                make_message_event_new("message_verified", msg)
```
sammy1991106, line 16: this function name is so odd.

cjcobb23 (approved):

contracts/gateway/src/contract.rs
cjcobb23: I actually don't mind really short variable names in lambdas like this

contracts/gateway/src/contract.rs
cjcobb23: I find you're way cleaner @sammy1991106 , but sorting actually might be faster here, because it's probably a small vector, and the sorting is in place

cgorenflo: I left this as is on purpose, because dealing with duplicates is a separate issue on my board, and for that this whole message changes 😆

contracts/gateway/src/contract.rs
cgorenflo: I do prefer `msg` over `m`, i.e. if there is a good short abbreviation I'd rather use this than just a single letter, but it is acceptable to me as long as the expression is trivial

contracts/gateway/src/events.rs
cgorenflo: the `new` is tagged on to differentiate the existing function which I can't delete yet. Once the old message types are deleted, this will become `make_message_event`

contracts/gateway/src/contract.rs
sammy1991106: fair enough. I can't see how sorting is more efficient though. It's O(NlogN) v.s. O(N) to me.

contracts/gateway/src/contract.rs
sammy1991106: I'd much prefer `msg` over `m`, and `res` over `r`.

sammy1991106 (approved):

PR #119, opened 2023-09-21
refactor: move recoverable signature into multisig prover contract

cgorenflo:


contracts/multisig-prover/src/types.rs
```diff
@@ -90,15 +90,28 @@ pub struct CommandBatch {
 
 #[cw_serde]
 #[derive(Ord, PartialOrd, Eq)]
-pub struct Operator {
+pub struct Operator<T>
+where
+    T: Recoverability,
```
cjcobb23, line 95: Just want to confirm, we don't actually need this today, do we?

contracts/multisig/src/key.rs
```diff
@@ -13,8 +14,43 @@ pub enum KeyType {
...
+impl Recoverable {
+    const LEN: usize = 65;
+}
+
+#[cw_serde]
+#[derive(Ord, PartialOrd, Eq)]
+pub struct NonRecoverable;
+
+impl NonRecoverable {
+    const LEN: usize = 64;
```
cjcobb23, line 41: Are these sizes correct for all signature schemes, or just ecdsa?

contracts/multisig-prover/src/encoding/abi.rs
```diff
@@ -130,27 +147,37 @@ pub fn make_operators(worker_set: WorkerSet) -> Operators {
...
+    T: Recoverability,
+{
     Ok(Operator {
         address: evm_address(signer.pub_key.as_ref())?,
         weight: signer.weight,
         signature: None,
     })
 }
 
+fn shift27(recovery_byte: u8) -> u8 {
```
cjcobb23, line 177:
```suggestion
fn add27(recovery_byte: u8) -> u8 {
```

contracts/multisig-prover/src/encoding/abi.rs
```diff
@@ -130,27 +147,37 @@ pub fn make_operators(worker_set: WorkerSet) -> Operators {
...
 }
 
-fn make_evm_operators_with_sigs(
-    signers_with_sigs: Vec<(Signer, Option<Signature>)>,
-) -> Result<Vec<Operator>, ContractError> {
+fn make_evm_operators_with_sigs<T>(
+    signers_with_sigs: Vec<(Signer, Option<Signature<T>>)>,
+) -> Result<Vec<Operator<T>>, ContractError>
+where
+    T: Recoverability,
```
cjcobb23, line 154: Does this need to be generic here? I think we always want `Recoverable`?

contracts/multisig-prover/src/encoding/abi.rs
```diff
@@ -130,27 +147,37 @@ pub fn make_operators(worker_set: WorkerSet) -> Operators {
...
             }
             op
         })
     })
 }
 
-fn make_evm_operator(signer: Signer) -> Result<Operator, ContractError> {
+fn make_evm_operator<T>(signer: Signer) -> Result<Operator<T>, ContractError>
+where
+    T: Recoverability,
```
cjcobb23, line 168: Same question, shouldn't this always be `Recoverable`?

contracts/multisig/src/key.rs
cgorenflo: it might just be ecdsa. I probably have to move the generic element into the variant instead

contracts/multisig-prover/src/encoding/abi.rs
cgorenflo: some of the functions where used in both contexts. I can go through them again and check

contracts/multisig/src/key.rs
```diff
@@ -13,8 +14,43 @@ pub enum KeyType {
 
 #[cw_serde]
 #[derive(PartialOrd, Ord, Eq)]
-pub enum Signature {
-    Ecdsa(HexBinary),
+pub enum Signature<T>
+where
+    T: Recoverability,
```
milapsheth, line 19: This is adding more complexity to the Signature type. Recoverability is not core concept for a Signature scheme. It's very EVM specific. ed25519 doesn't have a recoverable variant for e.g.

Since recoverable signatures are only needed when encoding the proof, I'd suggest just using plain bytes for the type

cjcobb23 (approved):

milapsheth (approved):

PR #120, opened 2023-09-21
feat(multisig-prover): bcs transfer operatorship

cjcobb23:
https://axelarnetwork.atlassian.net/browse/AXE-1923


contracts/multisig-prover/src/encoding/bcs.rs
```diff
@@ -0,0 +1,77 @@
+use axelar_wasm_std::operators::Operators;
+use bcs::to_bytes;
+use cosmwasm_std::{HexBinary, Uint256};
+
+use crate::{error::ContractError, state::WorkerSet};
+
+pub fn make_operators(worker_set: WorkerSet) -> Operators {
```
sammy1991106, line 7: How is this different than the function in abi.rs? They look identical to me.

contracts/multisig-prover/src/encoding/bcs.rs
```diff
@@ -0,0 +1,77 @@
...
+    operators.sort_by_key(|op| op.0.clone());
+    Operators {
+        weights_by_addresses: operators,
+        threshold: worker_set.threshold,
+    }
+}
+
+fn u256_to_u128(val: Uint256) -> u128 {
+    u128::from_le_bytes(
+        val.to_le_bytes()[..16]
```
sammy1991106, line 22: Validate the length. Simply ignoring other bytes makes me unconfortable.

cgorenflo: didn't we have this in another PR (maybe it was different integer sizes). Seems like the asserts you've been using would be appropriate here as well

contracts/multisig-prover/src/encoding/bcs.rs
```diff
@@ -0,0 +1,98 @@
...
+            .try_into()
+            .expect("couldn't convert u256 to u128"),
+    )
+}
+
+pub fn transfer_operatorship_params(worker_set: &WorkerSet) -> Result<HexBinary, ContractError> {
+    let mut operators: Vec<(HexBinary, Uint256)> = worker_set
+        .signers
+        .iter()
+        .map(|s| (s.pub_key.clone().into(), s.weight))
```
cgorenflo, line 32:
```suggestion
        .map(|signer| (signer.pub_key.clone().into(), signer.weight))
```

cgorenflo (approved):

contracts/multisig-prover/src/encoding/bcs.rs
cjcobb23: We have this in another PR, and it does validate the length there. I can just change this when merging

sammy1991106 (approved):

PR #126, opened 2023-09-24
feat: append msg ids to error for duplicate messages

cgorenflo:


contracts/gateway/src/contract/execute.rs
```diff
@@ -102,3 +72,54 @@ pub fn route_outgoing_messages(
...
+    CONFIG
+        .load(deps.storage)
+        .change_context(ContractError::VerifierNotFound)
+        .map(|cfg| cfg.verifier)
+}
+
+fn load_router(deps: &DepsMut) -> error_stack::Result<Addr, ContractError> {
+    CONFIG
+        .load(deps.storage)
+        .change_context(ContractError::RouterNotFound)
```
sammy1991106, line 109: Smae here. It should be `ConfigNotFound` imo.

contracts/gateway/src/contract/execute.rs
```diff
@@ -102,3 +72,54 @@ pub fn route_outgoing_messages(
             .map(|msg| GatewayEvent::MessageRouted { msg }.into()),
     ))
 }
+
+fn partition_by_verified(
+    deps: DepsMut,
+    msgs: Vec<NewMessage>,
+) -> Result<(Vec<NewMessage>, Vec<NewMessage>), ContractError> {
+    let verifier = CONFIG.load(deps.storage)?.verifier;
```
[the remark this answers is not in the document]
sammy1991106: I'd still much prefer a pure execute mod without `deps`. To avoid repeated loading. In this case, `verifier` is already loaded in the function `verify_messages`.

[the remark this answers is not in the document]
cgorenflo: addressing this in a separate PR

contracts/gateway/src/error.rs
```diff
@@ -15,4 +15,7 @@ pub enum ContractError {
 
     #[error("batch contains duplicate message ids")]
     DuplicateMessageID,
+
+    #[error("could not contract config")]
```
cjcobb23, line 19:
```suggestion
    #[error("could not find contract config")]
```

contracts/gateway/src/error.rs
```diff
@@ -15,4 +15,7 @@ pub enum ContractError {
 
     #[error("batch contains duplicate message ids")]
     DuplicateMessageID,
```
cjcobb23, line 17:
```suggestion
    DuplicateMessageIds,
```

contracts/gateway/src/contract/execute.rs
```diff
@@ -102,3 +72,49 @@ pub fn route_outgoing_messages(
...
+        .change_context(ContractError::ConfigNotFound)?;
+    Ok(cfg)
+}
+
+fn ensure_unique_ids(msgs: &[NewMessage]) -> error_stack::Result<(), ContractError> {
+    let duplicates: Vec<_> = msgs
+        .iter()
+        .map(|m| &m.cc_id)
+        .duplicates()
+        .map(|cc_id| cc_id.to_string())
```
cjcobb23, line 113:
```suggestion
        .map(|m| m.cc_id.to_string())
        .duplicates()
```

contracts/gateway/src/contract/execute.rs
```diff
@@ -102,3 +72,49 @@ pub fn route_outgoing_messages(
...
+            Some((_, v)) => *v,
+            None => false,
+        }
+    }))
+}
+
+fn load_config(deps: &DepsMut) -> error_stack::Result<Config, ContractError> {
+    let cfg = CONFIG
+        .load(deps.storage)
+        .change_context(ContractError::ConfigNotFound)?;
```
cjcobb23, line 104: Shouldn't this be a panic? When would a config not exist?

contracts/gateway/src/error.rs
cgorenflo: thanks, missed that!

contracts/gateway/src/contract/execute.rs
cgorenflo: I forgot to add a comment here, there is a reason why I split the maps: I want the duplicates check to use the strongly typed ID instead of the string

contracts/gateway/src/contract/execute.rs
cgorenflo: I guess we could panic here. I don't feel strongly about it. It's not dependent on user input, but the error gives a nicer response if it ever should happen. I'll think about it some more

cgorenflo: yes you are right, this makes more sense as a panic. I'll change it in a follow up PR that touches this code anyway

sammy1991106 (approved):

PR #130, opened 2023-09-27
refactor: move Recoverable types down to Ecdsa instead of being directly on Signature

cgorenflo:


contracts/multisig/src/key.rs
```diff
@@ -1,55 +1,113 @@
...
+            });
+        }
+        Ok(NonRecoverable(value))
+    }
 }
 
 #[cw_serde]
 #[derive(Ord, PartialOrd, Eq)]
-pub struct Recoverable;
+pub struct Recoverable(HexBinary);
```
cjcobb23, line 79: I still think we don't actually need this type. I think we should just assume all of the signatures are non-recoverable, and just add the recovery bit in the multisig prover when encoding.

contracts/multisig/src/key.rs
```diff
@@ -1,55 +1,113 @@
...
 #[cw_serde]
 #[derive(PartialOrd, Ord, Eq)]
-pub enum Signature<T>
-where
-    T: Recoverability,
-{
-    Ecdsa(HexBinary, PhantomData<T>),
+pub enum Signature {
+    Ecdsa(NonRecoverable),
+    EcdsaRecoverable(Recoverable),
```
cjcobb23, line 19: I don't think we need this. I think this still creates confusion. I think we should just have Ecdsa, and the signatures are always non-recoverable.

contracts/multisig/src/key.rs
```diff
@@ -1,55 +1,113 @@
...
+
+impl NonRecoverable {
+    const LEN: usize = 64;
+
+    pub fn to_recoverable(
+        &self,
+        msg: &[u8],
+        pub_key: &PublicKey,
+        recovery_transform: impl FnOnce(u8) -> u8,
+    ) -> Result<Recoverable, ContractError> {
```
cjcobb23, line 34:
```suggestion
    pub fn to_recoverable(
        &self,
        msg: &[u8],
        pub_key: &PublicKey,
        recovery_transform: impl FnOnce(u8) -> u8,
    ) -> Result<HexBinary, ContractError> {
```

just return `HexBinary` and make it a free function, instead of a method

contracts/multisig/src/key.rs
milapsheth: yeah, recoverable is just an encoding format for ecdsa signatures, not a different type of signature scheme

cjcobb23 (approved):

PR #131, opened 2023-09-27
feat(multisig): constant gas usage for signature submission

eguajardo:
## Description
Signatures were previously part of the SigningSession struct, which caused that every time a new signature was submitted it would load all signatures, update and then store again all signatures with the new one. Now signatures have been moved to a separate Map to improve gas cost on subsequent signers.

For now it's not exactly constant gas usage for subsequent signers because we still need to load all signatures to calculate threshold and determine if the session should change state. This should be re-evaluated after doing performance tests.

https://axelarnetwork.atlassian.net/browse/AXE-1803

## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


eguajardo:
Did some tests, before this change each subsequent signature costed approx 6.5k extra gas, after the change it was approx 800, however there is a 10k increase when the signature that completes the session is submitted. I'll what can be done to improve this last signature

Before change:
| signature index            | gas cost | diff from previous |
|  :--------------------: | --------: | -----------------: |
| 1                                     | 156865   | 0
| 2                                    | 163377   | 6512
| 3                                    | 169902   | 6525
| 4 (threshold reached) | 176623   | 6721

After change:
| signature index            | gas cost | diff from previous |
|  :--------------------: | --------: | -----------------: |
| 1                                     | 153401   | 0
| 2                                    | 154195   | 794
| 3                                    | 155007   | 812
| 4 (threshold reached) | 165379   | 10372

cjcobb23:
> Did some tests, before this change each subsequent signature costed approx 6.5k extra gas, after the change it was approx 800, however there is a 10k increase when the signature that completes the session is submitted. I'll see what can be done to improve this last signature
> 
> Before change:
> 
> signature index	gas cost	diff from previous
> 1	156865	0
> 2	163377	6512
> 3	169902	6525
> 4 (threshold reached)	176623	6721
> After change:
> 
> signature index	gas cost	diff from previous
> 1	153401	0
> 2	154195	794
> 3	155007	812
> 4 (threshold reached)	165379	10372

hmm, does the 10k increase always occur when the threshold is reached? If the threshold is higher, does it increase then?

contracts/multisig/src/signing.rs
```diff
@@ -26,72 +26,102 @@ impl SigningSession {
...
-                    .participants
-                    .get(addr)
-                    .expect("violated invariant: signature submitted by non-participant")
-                    .weight
-                    .into()
-            })
-            .sum()
+    if signers_weight(signatures, snapshot)? >= snapshot.quorum.into() {
+        session.state = MultisigState::Completed;
+        SIGNING_SESSIONS.save(store, session.id.u64(), &session)?;
```
cjcobb23, line 121: Is this what's causing the 10k increase when the threshold is reached?

cjcobb23 (changes requested):
Requesting changes because I think we need to solve the 10k increase when the threshold is reached before merging

contracts/multisig/src/signing.rs
eguajardo: yes, I commented out the `SIGNING_SESSIONS.save(store, session.id.u64(), &session)?` and instead of a 10k increase its only like 200. Maybe we don't really need to store the state, we could re-calculate it on the fly whenever is needed?

cjcobb23: calculating on the fly sounds good to me

eguajardo: removed state from storage and instead recalculated everytime it is needed. Here is the new table of gas used:



signature index | gas cost | diff from previous
-- | -- | --
1 | 153127 | 0
2 | 155398 | 2271
3 | 156966 | 1568
4 (threshold reached) | 158666 | 1700

cgorenflo (approved):

PR #134, opened 2023-09-28
feat(rewards): add rewards contract interface

cjcobb23:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


contracts/rewards/src/msg.rs
cgorenflo, line 1: replace all the references to validators with "workers". We want to avoid confusion with consensus validators

contracts/rewards/src/msg.rs
```diff
@@ -0,0 +1,57 @@
...
+        event_id: String,
+        validator_address: String,
+    },
+    // Process rewards for the most recent epoch, if not yet processed, and send the required number of tokens to each validator
+    ProcessRewards {
+        // Address of contract for which to process rewards.
+        // For example, address of a voting verifier instance.
+        contract_address: String,
+    },
+    // Increase the number of AXL tokens
```
cgorenflo, line 44:
```suggestion
    /// Start a new reward pool for the given address if none exists, add tokens to an existing one otherwise.
```
I assume that we use cosmwasms baked in token transfer to attach funds?

contracts/rewards/src/msg.rs
```diff
@@ -0,0 +1,57 @@
...
+    StartValidatorEvent {
+        event_id: String,
+    },
+    // Log a specific validator as participating in a specific event
+    ValidatorParticipated {
+        event_id: String,
+        validator_address: String,
+    },
+    // Process rewards for the most recent epoch, if not yet processed, and send the required number of tokens to each validator
+    ProcessRewards {
```
cgorenflo, line 39: I think edge cases will be tricky to deal with, but let's experiment with it

contracts/rewards/src/msg.rs
cjcobb23: yes

contracts/rewards/src/msg.rs
```diff
@@ -0,0 +1,57 @@
...
+    epoch_duration: nonempty::Uint64,
+    // Total length of time over which rewards in a pool are distributed, specified in number of blocks. For example, if pool_duration is
+    // 1 million blocks, and epoch_duration is 500 blocks, every 500 blocks, 1 million / 500 ( = 5,000) tokens are distributed to participating
+    // validators. The tokens to be distributed are split equally amongst the participating validators
+    pool_duration: nonempty::Uint64,
+    // Participation threshold validators must meet to receive rewards in a given epoch, specified as a fraction between 0 and 1. Validators
+    // must participate in at least this fraction of all events in a given epoch to receive rewards. So, if participation_threshold is 9/10,
+    // and there are 100 events in a given epoch, validators must have participated in at least 90 events to receive rewards.
+    // Participation is reset at the beginning of each epoch, so participation in previous epochs does not affect rewards for future epochs.
+    participation_threshold: nonempty::Uint64,
```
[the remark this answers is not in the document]
cjcobb23: yeah it was supposed to be `Threshold`, I messed this up

contracts/rewards/src/msg.rs
```diff
@@ -0,0 +1,57 @@
...
+}
+
+#[cw_serde]
+pub struct RewardsParams {
+    // How often rewards are calculated, specified in number of blocks. Participation is calculated over this window. So if epoch is 500
+    // blocks, validators are rewarded for their participation within each 500 block window.
+    epoch_duration: nonempty::Uint64,
+    // Total length of time over which rewards in a pool are distributed, specified in number of blocks. For example, if pool_duration is
+    // 1 million blocks, and epoch_duration is 500 blocks, every 500 blocks, 1 million / 500 ( = 5,000) tokens are distributed to participating
+    // validators. The tokens to be distributed are split equally amongst the participating validators
```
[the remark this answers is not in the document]
cjcobb23: ah yeah, it's total amount * epoch_duration / pool_duration. will fix

contracts/rewards/src/msg.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    /// Start a new reward pool for the given address if none exists. Otherwise, add tokens to an existing reward pool.
+    /// Any attached funds a denom matching the rewards denom are added to the pool.
+    AddRewards {
+        /// Address of contract for which to reward participation. For example, address of a voting verifier instance.
+        contract_address: String,
+    },
+
+    /// Overwrites the currently stored params. Callable only by governance.
+    ModifyParams { params: RewardsParams },
```
fish-sammy, line 54:
```suggestion
    UpdateParams { params: RewardsParams },
```

contracts/rewards/src/msg.rs
```diff
@@ -0,0 +1,58 @@
...
+    participation_threshold: Threshold,
+}
+
+#[cw_serde]
+pub enum ExecuteMsg {
+    /// Called each time a new event is started, for which workers are rewarded for participating
+    StartRewardEvent { event_id: String },
+
+    /// Log a specific worker as participating in a specific event
+    WorkerParticipated {
```
fish-sammy, line 35: Need to rename this to a verb.
```suggestion
    RecordParticipation {
```

contracts/rewards/src/msg.rs
```diff
@@ -0,0 +1,58 @@
...
+}
+
+#[cw_serde]
+pub struct RewardsParams {
+    /// How often rewards are calculated, specified in number of blocks. Participation is calculated over this window. So if epoch is 500
+    /// blocks, workers are rewarded for their participation within each 500 block window.
+    epoch_duration: nonempty::Uint64,
+
+    /// Total length of time over which rewards in a pool are distributed, specified in number of blocks. For example, if pool_duration is
+    /// 1 million blocks, and epoch_duration is 500 blocks, and total size of the reward pool is 100k tokens, every 500 blocks,
```
fish-sammy, line 18: Does "total size of the reward pool" change every window as rewards are given out and are added? How does the calculation work when more rewards are added while the pool isn't empty.

contracts/rewards/src/msg.rs
fish-sammy:
```suggestion
    DistributeRewards {
```
The word process is vague.

fish-sammy: If event started at block 499, and participations come in at block 500+, what does it mean when it comes to the calculation the participation? No body participated for that event for that window? The rewards contract doesn't seem to know when do events end.

contracts/rewards/src/msg.rs
cjcobb23: Calculation would have to change as more rewards are added. Pretty sure the calculation then is:
```
total rewards remaining / epochs remaining / participating validators
```
where total rewards remaining is rewards not yet distributed and epochs remaining is number of epochs for which rewards have not yet been distributed. I can update the comment to this if you like, but I do feel like it's a little harder to wrap your head around this than the basic case example.

contracts/rewards/src/msg.rs
```diff
@@ -0,0 +1,58 @@
...
+    /// 100k * 500 / 1 million (= 50) tokens are distributed to participating workers. The tokens to be distributed are split equally amongst the participating workers
+    pool_duration: nonempty::Uint64,
+
+    /// Participation threshold workers must meet to receive rewards in a given epoch, specified as a fraction between 0 (exclusive) and 1 (exclusive). workers
```
cgorenflo, line 22:
```suggestion
    /// Participation threshold workers must meet to receive rewards in a given epoch, specified as a fraction between 0 (exclusive) and 1 (exclusive). Workers
```

cgorenflo (approved):

contracts/rewards/src/msg.rs
fish-sammy: But why even bother letting the rewards contract know when events begin? Why not just something like RecordParticipation { participants: Vec<...> } at the end of the event?

contracts/rewards/src/msg.rs
fish-sammy: What does epochs remaining mean? I don't see any param related to epochs when adding rewards.

cjcobb23: In the params, there is an `epoch_duration`

cjcobb23: The current epoch is just determined from the current block height, so we don't really need it for any of the messages.

PR #135, opened 2023-09-28
feat: let construct_proof update worker_set

fahimahmedx:
## Description
`construct_proof` updates the worker_set automatically if it needs to be updated. two test cases were added for checking if the worker set was updated, and for checking that it causes an error if there is no worker set initially.

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Steps to Test
`cargo test` test cases all pass

### DIFF
--- a/contracts/multisig-prover/src/error.rs
+++ b/contracts/multisig-prover/src/error.rs
@@ -50,6 +50,6 @@ pub enum ContractError {
     #[error("a worker set confirmation already in progress")]
     WorkerSetConfirmationInProgress,
 
-    #[error("there is no worker set")]
+    #[error("no worker set stored")]
     NoWorkerSet,
 }
--- a/contracts/multisig-prover/src/error.rs
+++ b/contracts/multisig-prover/src/error.rs
@@ -49,4 +49,7 @@ pub enum ContractError {
 
     #[error("a worker set confirmation already in progress")]
     WorkerSetConfirmationInProgress,
+
+    #[error("there is no worker set")]
+    NoWorkerSet,
 }
--- a/contracts/multisig-prover/src/execute.rs
+++ b/contracts/multisig-prover/src/execute.rs
@@ -28,17 +28,20 @@ pub fn construct_proof(
     message_ids: Vec<String>,
 ) -> Result<Response, ContractError> {
     let config = CONFIG.load(deps.storage)?;
-    let workers_info = get_workers_info(&deps, &env, &config)?;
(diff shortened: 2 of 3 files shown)


contracts/multisig-prover/src/error.rs
```diff
@@ -49,4 +49,7 @@ pub enum ContractError {
 
     #[error("a worker set confirmation already in progress")]
     WorkerSetConfirmationInProgress,
+
+    #[error("there is no worker set")]
```
cjcobb23, line 53:
```suggestion
    #[error("no worker set stored")]
```

contracts/multisig-prover/src/execute.rs
```diff
@@ -22,18 +22,30 @@ use crate::{
...
-pub fn construct_proof(deps: DepsMut, message_ids: Vec<String>) -> Result<Response, ContractError> {
-    let key_id = KEY_ID.load(deps.storage)?;
+pub fn construct_proof(
+    deps: DepsMut,
+    env: Env,
+    message_ids: Vec<String>,
+) -> Result<Response, ContractError> {
     let config = CONFIG.load(deps.storage)?;
+    let workers_info = get_workers_info(&deps, &env, &config)?;
+    let new_worker_set = get_next_worker_set(&deps, &env, &config, false)?;
```
cjcobb23, line 32: Can we move these closer to where they are used? Inside the match statement. Makes it easier to read

contracts/multisig-prover/src/execute.rs
```diff
@@ -146,6 +158,7 @@ fn get_next_worker_set(
     deps: &DepsMut,
     env: &Env,
     config: &Config,
+    called_by_update_worker_set: bool,
```
cjcobb23, line 161: I would avoid something like this, for two reasons: 1. at the callsite, it's not clear what a random boolean flag is for. And 2. It increases the complexity of the function, by adding more conditional branches.

What if we refactor this so this just returns an error if there is no current worker set? And then make a separate function, like `make_worker_set`, that is just:
```
    let workers_info = get_workers_info(deps, env, config)?;
    return WorkerSet::new(
            workers_info.pubkeys_by_participant,
            workers_info.snapshot.quorum.into(),
            env.block.height,
        )?;

```

cjcobb23 (changes requested):

cjcobb23 (approved):

PR #144, opened 2023-10-03
feat: sui verify message handler

haiyizxx:
## Todos

- [x] Unit tests
- [x] Manual tests
- [ ] Documentation
- [x] Connect epics/issues


ampd/src/queue/queued_broadcaster.rs
```diff
@@ -38,11 +39,12 @@ impl QueuedBroadcasterDriver {
     }
 }
 
+#[automock]
 #[async_trait]
 pub trait BroadcasterClient {
     async fn broadcast<T>(&self, tx: T) -> Result
     where
-        T: Msg + Send + Sync;
+        T: Msg + Send + Sync + 'static;
```
fish-sammy, line 47: Having `'static` is a requirment from automock I assume?

ampd/src/sui/json_rpc.rs
```diff
@@ -1,12 +1,14 @@
...
 use ethers::providers::{JsonRpcClient, ProviderError};
+use mockall::automock;
 use sui_json_rpc_types::{SuiTransactionBlockResponse, SuiTransactionBlockResponseOptions};
 use sui_types::digests::TransactionDigest;
 
 use crate::json_rpc::Client;
 
 type Result<T> = error_stack::Result<T, ProviderError>;
 
+#[automock]
```
fish-sammy, line 11: Get the method to return `Result<Vec<Option<SuiTransactionBlockResponse>>>` as discussed.

fish-sammy (changes requested):

ampd/src/queue/queued_broadcaster.rs
haiyizxx: yep, I think it's fine to add the restriction for async

ampd/src/handlers/voter.rs
```diff
@@ -0,0 +1,150 @@
...
+        assert!(!voter.is_voting_verifier(&TMAddress::random(PREFIX)));
+    }
+
+    #[async_test]
+    async fn should_submit_vote() {
+        let mut broadcaster = MockBroadcasterClient::new();
+
+        broadcaster
+            .expect_broadcast()
+            .once()
```
fish-sammy, line 139:
```suggestion
            .once().with(...)
```
to ensure it's voting for the correct thing.

ampd/src/sui/json_rpc.rs
```diff
@@ -22,18 +26,24 @@ where
...
+                    .map(TransactionDigest::base58_encode)
                     .collect::<Vec<_>>(),
                 SuiTransactionBlockResponseOptions::new().with_events(),
             ),
         )
         .await
+        .map(|vec: Vec<SuiTransactionBlockResponse>| {
+            vec.into_iter()
+                // None checkpoint means the transaction is not finalized
+                .map(|block| block.checkpoint.map(|_| block))
```
fish-sammy, line 45: Their docs mention `"This is only returned in the read api, not in the transaction execution api"`. What's the difference between the read API and the execution API? Also, it seems that this `checkpoint` is only Some(...) when the transaction is included as well as finalized. Can you rename the method `transaction_blocks` to `finalized_transaction_blocks` then?

ampd/src/types.rs
```diff
@@ -55,3 +55,21 @@ impl fmt::Display for TMAddress {
...
+    impl TMAddress {
+        pub fn random(prefix: &str) -> Self {
+            Self(
+                PublicKey::from(SigningKey::random(&mut OsRng).verifying_key())
+                    .account_id(prefix)
+                    .expect("failed to convert to account identifier"),
+            )
+        }
+    }
+}
```
fish-sammy, line 75: @cgorenflo weigh in your opinion here. The point was that I don't like including code in the production build for it that is only needed by testing. Not sure if this is the best way/location to do this though.

ampd/src/sui/json_rpc.rs
haiyizxx: sure good name. the execution api execute transaction to blockchain and wait for response

ampd/src/handlers/sui_verify_msg.rs
```diff
@@ -12,3 +29,322 @@ pub struct Message {
...
+        rpc_client: C,
+        broadcast_client: B,
+    ) -> Self {
+        Self {
+            worker,
+            voting_verifier,
+            rpc_client,
+            broadcast_client,
+        }
+    }
```
cgorenflo, line 73: not necessary, it doesn't do any translation or checks, so you might as well use the object initializer

ampd/src/handlers/sui_verify_msg.rs
```diff
@@ -12,3 +29,322 @@ pub struct Message {
...
+        }
+
+        let deduplicated_tx_ids: HashSet<_> = messages.iter().map(|msg| msg.tx_id).collect();
+        let transaction_blocks = self
+            .rpc_client
+            .finalized_transaction_blocks(deduplicated_tx_ids)
+            .await
+            .change_context(Error::TxReceipts)?;
+
+        let votes = messages
```
cgorenflo, line 130: add a comment that explains that the handler doesn't make any assumptions about id duplication in the event, but the rpc call fails(?) if there are duplicates, therefore you have to do this convolution of deduplication and mapping

cgorenflo (approved):

fish-sammy (approved):

PR #146, opened 2023-10-10
refactor: extract gateway and router apis into separate packages

cgorenflo:


packages/connection-router-api/src/msg.rs
cgorenflo, line 1: moved from router and replaced strings with `ChainName` and `Address`

packages/connection-router-api/src/primitives.rs
```diff
@@ -0,0 +1,182 @@
...
+
+// [cw_serde] has been expanded here because we need to implement PartialEq manually
+#[derive(
+    ::cosmwasm_schema::serde::Serialize,
+    ::cosmwasm_schema::serde::Deserialize,
+    Clone,
+    Debug,
+    ::cosmwasm_schema::schemars::JsonSchema,
+)]
+#[allow(clippy::derive_partial_eq_without_eq)]
```
haiyizxx, line 39: is it needed? it does not derive partial eq

haiyizxx (approved):

packages/connection-router-api/src/error.rs
```diff
@@ -0,0 +1,14 @@
+use thiserror::Error;
+
+/// A chain name must adhere to the following rules:
+/// 1. it can optionally start with an uppercase letter, followed by one or more lowercase letters
+/// 2. it can have an optional suffix of an optional dash and one or more digits ("1", "03", "-5" are all valid suffixes)
+pub const CHAIN_NAME_REGEX: &str = "^[A-Z]?[a-z]+(-?[0-9]+)?$";
```
fish-sammy, line 6: Have this unit tested, which also serves as examples for valid chain names.

packages/connection-router-api/src/primitives.rs
```diff
@@ -0,0 +1,187 @@
...
+    fn eq(&self, other: &Self) -> bool {
+        self.0.to_lowercase() == other.0.to_lowercase()
+    }
+}
+
+impl FromStr for ChainName {
+    type Err = Error;
+
+    fn from_str(chain_name: &str) -> Result<Self, Self::Err> {
+        fn matches_chain_name_pattern(s: &str) -> bool {
```
fish-sammy, line 62: It's a bit weird to define this function. It's not even used for more than once.

packages/connection-router-api/src/primitives.rs
```diff
@@ -0,0 +1,187 @@
...
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        value.parse()
+    }
+}
+
+impl Deref for Address {
+    type Target = str;
+
+    fn deref(&self) -> &Self::Target {
+        self.0.deref()
```
fish-sammy, line 113:
```suggestion
        self.0
```

fish-sammy (commented):
Do different apis deserve to be separate Rust projects? Why not just a api project and then these for different contracts all live in it?

cgorenflo:
> Do different apis deserve to be separate Rust projects? Why not just a api project and then these for different contracts all live in it?

Contracts can vary for different ecosystems, so I don't want to force other contracts to import apis they don't need. We can explore if we want to merge gateway and connection router, but even they are only loosely coupled by `RouteMessages`

packages/connection-router-api/src/primitives.rs
cgorenflo: it just seems natural to implement from_str if we also implement `TryFrom<String>` and it makes tests nicer to write

packages/connection-router-api/src/primitives.rs
cgorenflo: doesn't work, Addr is a nonempty::String, so I need to deref that

packages/connection-router-api/src/primitives.rs
fish-sammy: Hmm I meant this local function `matches_chain_name_pattern`.

packages/connection-router-api/src/error.rs
cgorenflo: what do you mean? there are  a bunch of tests for the chain names

fish-sammy (approved):

packages/connection-router-api/src/msg.rs
```diff
@@ -0,0 +1,47 @@
+use crate::primitives::*;
+use cosmwasm_schema::{cw_serde, QueryResponses};
+
+#[cw_serde]
+pub enum ExecuteMsg {
+    /*
+     * Governance Methods
+     * All of the below messages can only be called by governance
```
cjcobb23, line 8:
```suggestion
     * All of the below messages should only be called by governance
```
Since this is just an API, we can't enforce what the impl does

packages/connection-router-api/src/msg.rs
```diff
@@ -0,0 +1,47 @@
...
+    },
+    // Changes the gateway address associated with a particular chain
+    UpgradeGateway {
+        chain: ChainName,
+        contract_address: Address,
+    },
+
+    /*
+     * Router Admin Methods
+     * All of the below messages can only be called by the router admin
```
cjcobb23, line 23:
```suggestion
     * All of the below messages should only be called by the router admin
```

packages/gateway-api/src/msg.rs
```diff
@@ -0,0 +1,11 @@
+use connection_router_api::primitives::Message;
+use cosmwasm_schema::cw_serde;
+
+#[cw_serde]
+pub enum ExecuteMsg {
+    // Permissionless
+    VerifyMessages(Vec<Message>),
+
+    // Permissionless
+    RouteMessages(Vec<Message>),
```
cjcobb23, line 10: Maybe add some comments about what these are supposed to do

cjcobb23 (approved):

PR #147, opened 2023-10-11
feat(voting-verifier): constant gas usage when voting

eguajardo:
## Description
Poll struct in voting verifier previously had a set of those workers that submitted a vote, this caused all subsequent votes to cost a little more gas because it had to load all previous voters and store one more. The solution was to extract the set of workers that submitted a vote into it's own `Map` in storage

https://axelarnetwork.atlassian.net/browse/AXE-1999

## Todos

- [x] Unit tests
- [x] Manual tests
- [ ] Documentation
- [x] Connect epics/issues

### DIFF
--- a/contracts/voting-verifier/src/execute.rs
+++ b/contracts/voting-verifier/src/execute.rs
@@ -44,7 +44,7 @@ pub fn confirm_worker_set(
     }
 
     let config = CONFIG.load(deps.storage)?;
-    let snapshot = take_snapshot(deps.as_ref(), &env, &config.source_chain)?;
+    let snapshot = take_snapshot(deps.as_ref(), &config.source_chain)?;
     let participants = snapshot.get_participants();
 
     let poll_id = create_worker_set_poll(
@@ -127,7 +127,7 @@ pub fn verify_messages(
         return Ok(response);
     }
 
-    let snapshot = take_snapshot(deps.as_ref(), &env, &pending_messages[0].cc_id.chain)?;
+    let snapshot = take_snapshot(deps.as_ref(), &pending_messages[0].cc_id.chain)?;
     let participants = snapshot.get_participants();
     let id = create_messages_poll(
         deps.storage,
@@ -271,11 +271,7 @@ pub fn end_poll(deps: DepsMut, env: Env, poll_id: PollID) -> Result<Response, Co
         .set_data(to_binary(&EndPollResponse { poll_result })?))
 }
 
-fn take_snapshot(
-    deps: Deps,
-    env: &Env,
-    chain: &ChainName,
-) -> Result<snapshot::Snapshot, ContractError> {
+fn take_snapshot(deps: Deps, chain: &ChainName) -> Result<snapshot::Snapshot, ContractError> {
     let config = CONFIG.load(deps.storage)?;
 
     // todo: add chain param to query after service registry updated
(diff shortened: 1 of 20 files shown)


contracts/voting-verifier/src/execute.rs
```diff
@@ -344,3 +346,174 @@ fn remove_pending_message(
...
+        .count();
+
+    if block_height < poll.expires_at
+            // can tally early if all participants voted
+            && voters < poll.snapshot.get_participants().len()
+    {
+        return Err(ContractError::PollNotEnded);
+    }
+
+    Ok(poll.tally()?)
```
cgorenflo, line 389: if this construction is just about matching error types I'd prefer `poll.tally().into()`

packages/axelar-wasm-std/src/voting.rs
```diff
@@ -154,13 +153,12 @@ pub enum PollStatus {
 
 #[cw_serde]
 pub struct WeightedPoll {
-    poll_id: PollID,
-    snapshot: Snapshot,
-    expires_at: u64,
-    poll_size: u64,
-    votes: Vec<Uint256>, // running tally of weighted votes
-    status: PollStatus,
-    voted: HashSet<Addr>,
```
cjcobb23, line 163: I'm not sure how I feel about removing `voted`. Now some of the logic of voting (not adding votes from the same participant twice) is removed from this module and pushed to somewhere else. I'm not sure this is best from a code perspective. I would rather still keep track of who voted in this struct. Maybe we could do something like make this a `HashMap<Addr,bool>` and initialize all of the participants to `false`? We could use that same struct to track who the participants are as well, instead of using the `Snapshot`

cjcobb23 (changes requested):

packages/axelar-wasm-std/src/voting.rs
eguajardo: good suggestion, I'll try this out

eguajardo: changed to this:
```
#[cw_serde]
pub struct Participation {
    pub voted: bool,
}

#[cw_serde]
pub struct WeightedPoll {
    poll_id: PollID,
    quorum: nonempty::Uint256,
    expires_at: u64,
    poll_size: u64,
    votes: Vec<Uint256>, // running tally of weighted votes
    status: PollStatus,
    participation: HashMap<String, Participation>,
}
```

now everything remains in the voting.rs

contracts/voting-verifier/src/execute.rs
```diff
@@ -344,3 +346,174 @@ fn remove_pending_message(
...
+        .prefix(poll.poll_id)
+        .range(store, None, None, Order::Ascending)
+        .count();
+
+    if block_height < poll.expires_at
+            // can tally early if all participants voted
+            && voters < poll.snapshot.get_participants().len()
+    {
+        return Err(ContractError::PollNotEnded);
+    }
```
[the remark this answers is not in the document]
eguajardo: cc @haiyizxx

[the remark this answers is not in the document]
haiyizxx: The current feature is incomplete and requires further improvement. 
The poll itself does not have quorum, we'll need to wait until the poll concludes to tally each message to see if it reaches quorum.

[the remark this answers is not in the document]
eguajardo: discussed with Haiyi, we concluded we can allow the poll to end before expiration if a result has reached quorum. Here is a task to keep track of this: https://axelarnetwork.atlassian.net/browse/AXE-2201

PR #148, opened 2023-10-11
feat(rewards): state scaffolding

cjcobb23:
Just some basic functions needed for rewards. The logic will come in a later PR.

https://axelarnetwork.atlassian.net/browse/AXE-2159

### DIFF
--- a/contracts/rewards/src/error.rs
+++ b/contracts/rewards/src/error.rs
@@ -29,4 +29,7 @@ pub enum ContractError {
 
     #[error("error loading rewards watermark")]
     LoadRewardsWatermark,
+
+    #[error("invalid event id")]
+    InvalidEventId,
 }
--- a/contracts/rewards/src/error.rs
+++ b/contracts/rewards/src/error.rs
@@ -1,0 +1,9 @@
+use axelar_wasm_std_derive::IntoContractError;
+use cosmwasm_std::StdError;
+use thiserror::Error;
+
+#[derive(Error, Debug, PartialEq, IntoContractError)]
+pub enum ContractError {
+    #[error(transparent)]
+    Std(#[from] StdError),
+}
--- a/contracts/rewards/src/error.rs
+++ b/contracts/rewards/src/error.rs
@@ -1,9 +1,32 @@
 use axelar_wasm_std_derive::IntoContractError;
-use cosmwasm_std::StdError;
 use thiserror::Error;
 
 #[derive(Error, Debug, PartialEq, IntoContractError)]
 pub enum ContractError {
-    #[error(transparent)]
-    Std(#[from] StdError),
+    #[error("error saving params")]
+    SaveParams,
+
+    #[error("error saving epoch tally")]
+    SaveEpochTally,
+
+    #[error("error saving event")]
+    SaveEvent,
+
+    #[error("error saving rewards pool")]
+    SaveRewardsPool,
+
+    #[error("error saving rewards watermark")]
+    SaveRewardsWatermark,
+
+    #[error("error loading epoch tally")]
+    LoadEpochTally,
+
+    #[error("error loading event")]
+    LoadEvent,
+
+    #[error("error loading rewards pool")]
+    LoadRewardsPool,
+
+    #[error("error loading rewards watermark")]
+    LoadRewardsWatermark,
 }
--- a/contracts/rewards/src/msg.rs
+++ b/contracts/rewards/src/msg.rs
@@ -15,7 +15,7 @@ pub struct RewardsParams {
     pub epoch_duration: nonempty::Uint64,
 
     /// Total number of tokens distributed as rewards per epoch. Tokens are split equally amongst all participating workers for a given epoch
-    pub rewards_rate: nonempty::Uint256,
+    pub rewards_per_epoch: nonempty::Uint256,
 
     /// Participation threshold workers must meet to receive rewards in a given epoch, specified as a fraction between 0 (exclusive) and 1 (exclusive). Workers
     /// must participate in at least this fraction of all events in a given epoch to receive rewards. So, if participation_threshold is 9/10,
--- a/contracts/rewards/src/msg.rs
+++ b/contracts/rewards/src/msg.rs
@@ -12,16 +12,16 @@ pub struct InstantiateMsg {
 pub struct RewardsParams {
     /// How often rewards are calculated, specified in number of blocks. Participation is calculated over this window. So if epoch_duration is 500
     /// blocks, workers are rewarded for their participation within each 500 block window.
-    epoch_duration: nonempty::Uint64,
+    pub epoch_duration: nonempty::Uint64,
 
     /// Total number of tokens distributed as rewards per epoch. Tokens are split equally amongst all participating workers for a given epoch
-    rewards_rate: nonempty::Uint64,
+    pub rewards_rate: nonempty::Uint256,
 
     /// Participation threshold workers must meet to receive rewards in a given epoch, specified as a fraction between 0 (exclusive) and 1 (exclusive). Workers
     /// must participate in at least this fraction of all events in a given epoch to receive rewards. So, if participation_threshold is 9/10,
     /// and there are 100 events in a given epoch, workers must have participated in at least 90 events to receive rewards.
     /// Participation is reset at the beginning of each epoch, so participation in previous epochs does not affect rewards for future epochs.
-    participation_threshold: Threshold,
+    pub participation_threshold: Threshold,
 }
 
 #[cw_serde]
--- a/contracts/rewards/src/state.rs
+++ b/contracts/rewards/src/state.rs
@@ -13,22 +13,22 @@ pub struct Config {
 
 #[cw_serde]
 pub struct EpochTally {
-    epoch_num: u64,
-    contract: Addr,
-    total_events: u64,
-    participation: HashMap<Addr, u64>,
-    distributed_rewards: bool,
+    pub epoch_num: u64,
+    pub contract: Addr,
+    pub total_events: u64,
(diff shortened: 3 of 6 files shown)


contracts/rewards/src/error.rs
```diff
@@ -0,0 +1,9 @@
+use axelar_wasm_std_derive::IntoContractError;
+use cosmwasm_std::StdError;
+use thiserror::Error;
+
+#[derive(Error, Debug, PartialEq, IntoContractError)]
+pub enum ContractError {
+    #[error(transparent)]
+    Std(#[from] StdError),
```
cgorenflo, line 8: I'd prefer if you incorporated error-stack from the start, so we don't have to go back and redo the whole error reporting again. In that case this is not needed

contracts/rewards/src/state.rs
```diff
@@ -0,0 +1,383 @@
...
+use cosmwasm_schema::cw_serde;
+use cosmwasm_std::{Addr, Storage, Uint256};
+use cw_storage_plus::{Item, Map};
+
+use crate::{error::ContractError, msg::RewardsParams};
+
+#[cw_serde]
+pub struct Config {
+    pub params: RewardsParams,
+}
```
cgorenflo, line 12: why have two different structs that do the same thing? Can we either keep params or config but not both?

contracts/rewards/src/state.rs
```diff
@@ -0,0 +1,383 @@
...
+    fn save_last_checkpoint(&mut self, epoch: &Epoch) -> Result<(), ContractError>;
+
+    fn save_rewards_watermark(&mut self, epoch_num: u64) -> Result<(), ContractError>;
+
+    fn save_event(&mut self, event: &Event) -> Result<(), ContractError>;
+
+    fn save_epoch_tally(&mut self, tally: &EpochTally) -> Result<(), ContractError>;
+
+    fn save_rewards_pool(&mut self, pool: &RewardsPool) -> Result<(), ContractError>;
+}
```
cgorenflo, line 93: this makes a pretty big weakness of my store approach visible: it's a brittle trait that needs to be changed whenever another item needs to be stored and and can grow quite large and unwieldy. I think instead we should provide a test trait implementation (if `MemoryStorage` in cosmwasm-std doesn't suit our needs) of `Storage` in cosmwasm-std

cgorenflo (changes requested):

contracts/rewards/src/error.rs
fish-sammy: I thought this shouldn't be needed for the specific ContractError?

contracts/rewards/src/state.rs
fish-sammy: Have the same question.

contracts/rewards/src/state.rs
```diff
@@ -0,0 +1,383 @@
...
+#[cw_serde]
+pub struct Config {
+    pub params: RewardsParams,
+}
+
+#[cw_serde]
+pub struct EpochTally {
+    pub epoch_num: u64,
+    pub contract: Addr,
+    pub total_events: u64,
```
fish-sammy, line 18:
```suggestion
    pub event_count: u64,
```

contracts/rewards/src/state.rs
```diff
@@ -0,0 +1,383 @@
...
+    pub participation: HashMap<Addr, u64>,
+    pub rewards_rate: Uint256,
```
fish-sammy, line 20: This is the rewards that should be distributed for this entire epoch? In that case, I'll drop the word "rate" in there just to not confused with the per block reward rate thing.

contracts/rewards/src/state.rs
```diff
@@ -0,0 +1,383 @@
...
+pub struct Event {
+    event_id: String,
+    contract: Addr,
+    epoch_num: u64,
+}
+
+#[cw_serde]
+pub struct Epoch {
+    pub epoch_num: u64,
+    pub block_height_started: u64,
```
fish-sammy, line 46: What is this? The block where the epoch starts at?

contracts/rewards/src/state.rs
fish-sammy: I don't neceseccarily think that this is an issue. If it becomes one, we can simply divide this `Store` trait into multiple, each logically manages one thing. The point is that now we can mock data access in testing.

contracts/rewards/src/state.rs
```diff
@@ -0,0 +1,383 @@
...
+
+const LAST_CHECKPOINT: Item<Epoch> = Item::new("current_epoch");
+
+const TALLIES: Map<(Addr, u64), EpochTally> = Map::new("tallies");
+
+const EVENTS: Map<(String, Addr), Event> = Map::new("events");
+
+const POOLS: Map<Addr, RewardsPool> = Map::new("pools");
+
+const WATERMARK: Item<u64> = Item::new("rewards_watermark");
```
fish-sammy, line 105: Can you add comments explaining what these are and what are the keys? What is the watermark?

contracts/rewards/src/state.rs
cjcobb23: No this is the rewards rate that was in effect during that epoch. We need this in case the rewards rate changes for future epochs, but the rewards themselves haven't been distributed.

contracts/rewards/src/state.rs
cjcobb23: yes

contracts/rewards/src/state.rs
cjcobb23: yeah I think the params should just be stored by themselves as an `Item`. The config in our other contracts is a static set of values set at instantiation that never changes, but the params here are different, since they can be changed.

Also, I think we can combine the params and the last checkpoint into one `Item`

fish-sammy (approved):

contracts/rewards/src/error.rs
```diff
@@ -0,0 +1,32 @@
...
+
+    #[error("error loading event")]
+    LoadEvent,
+
+    #[error("error loading rewards pool")]
+    LoadRewardsPool,
+
+    #[error("error loading rewards watermark")]
+    LoadRewardsWatermark,
+}
```
cgorenflo, line 32: I don't really see these as contract level errors, but we don't really have a good example to point you at just yet. I'm ok with keeping it like that for now but we should revisit

contracts/rewards/src/msg.rs
```diff
@@ -12,16 +12,16 @@ pub struct InstantiateMsg {
 pub struct RewardsParams {
     /// How often rewards are calculated, specified in number of blocks. Participation is calculated over this window. So if epoch_duration is 500
     /// blocks, workers are rewarded for their participation within each 500 block window.
-    epoch_duration: nonempty::Uint64,
+    pub epoch_duration: nonempty::Uint64,
 
     /// Total number of tokens distributed as rewards per epoch. Tokens are split equally amongst all participating workers for a given epoch
-    rewards_rate: nonempty::Uint64,
+    pub rewards_rate: nonempty::Uint256,
```
cgorenflo, line 18: why not make it explicit?
```suggestion
    pub rewards_per_epoch: nonempty::Uint256,
```

contracts/rewards/src/state.rs
```diff
@@ -0,0 +1,389 @@
...
+use cw_storage_plus::{Item, Map};
+use error_stack::{Result, ResultExt};
+
+use crate::{error::ContractError, msg::RewardsParams};
+
+#[cw_serde]
+pub struct StoredParams {
+    pub params: RewardsParams,
+    /// epoch in which the params were updated
+    pub updated: Epoch,
```
cgorenflo, line 14:
```suggestion
    pub last_updated: Epoch,
```

contracts/rewards/src/state.rs
```diff
@@ -0,0 +1,389 @@
...
+}
+
+#[cw_serde]
+pub struct EpochTally {
+    pub epoch_num: u64,
+    pub contract: Addr,
+    pub event_count: u64,
+    pub participation: HashMap<Addr, u64>,
+    pub rewards_rate: Uint256,
```
cgorenflo, line 23:
```suggestion
    pub distributed_rewards: Uint256,
```
and move it to `Epoch`

contracts/rewards/src/state.rs
```diff
@@ -0,0 +1,389 @@
...
+            event_count: 0,
+            participation: HashMap::new(),
+            rewards_rate,
+        }
+    }
+}
+
+#[cw_serde]
+pub struct Event {
+    event_id: String,
```
cgorenflo, line 41:
```suggestion
    event_id: nonempty::String,
```

contracts/rewards/src/state.rs
```diff
@@ -0,0 +1,389 @@
...
+
+        assert!(loaded.is_ok());
+        assert!(loaded.as_ref().unwrap().is_some());
+        assert_eq!(loaded.unwrap().unwrap(), pool);
+
+        let loaded = store.load_rewards_pool(Addr::unchecked("a different contract"));
+        assert!(loaded.is_ok());
+        assert!(loaded.as_ref().unwrap().is_none());
+    }
+}
```
cgorenflo, line 389: I don't really see any benefit in these tests, they just tie down the implementation. I'd rather wait and test at the high level functions of the reward contract

contracts/rewards/src/error.rs
cjcobb23: yeah I wasn't really sure what was best here

contracts/rewards/src/state.rs
cjcobb23: yeah some of them are probably not necessary, but I think it's probably useful to test things like "same event id but different contract address loads a different event".

PR #149, opened 2023-10-12
refactor: change chain_name datatype from String to ChainName

fahimahmedx:
## Description
Updates instances of chain_name to use the ChainName datatype, and updated test cases.
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Steps to Test
run `cargo test`
## Expected Behaviour
all test cases pass

### DIFF
--- a/contracts/connection-router/src/msg.rs
+++ b/contracts/connection-router/src/msg.rs
@@ -1,4 +1,4 @@
-use crate::state::{GatewayDirection, Message};
+use crate::state::{ChainName, GatewayDirection, Message};
 use cosmwasm_schema::{cw_serde, QueryResponses};
 
 #[cw_serde]
@@ -17,12 +17,12 @@ pub enum ExecuteMsg {
      */
     // Registers a new chain with the router
     RegisterChain {
-        chain: String,
+        chain: ChainName,
         gateway_address: String,
     },
     // Changes the gateway address associated with a particular chain
     UpgradeGateway {
-        chain: String,
+        chain: ChainName,
         contract_address: String,
     },
 
@@ -32,12 +32,12 @@ pub enum ExecuteMsg {
      */
     // Freezes a chain, in the specified direction.
     FreezeChain {
-        chain: String,
+        chain: ChainName,
         direction: GatewayDirection,
     },
     // Unfreezes a chain, in the specified direction.
     UnfreezeChain {
-        chain: String,
+        chain: ChainName,
         direction: GatewayDirection,
     },
 
--- a/contracts/service-registry/src/contract.rs
+++ b/contracts/service-registry/src/contract.rs
@@ -252,11 +252,7 @@ pub mod execute {
             .ok_or(ContractError::WorkerNotFound)?;
 
         for chain in chains {
-            WORKERS_PER_CHAIN.save(
-                deps.storage,
-                (&service_name, &chain, &info.sender),
-                &(),
-            )?;
+            WORKERS_PER_CHAIN.save(deps.storage, (&service_name, &chain, &info.sender), &())?;
         }
 
         Ok(Response::new())
--- a/contracts/service-registry/src/contract.rs
+++ b/contracts/service-registry/src/contract.rs
@@ -110,6 +110,8 @@ pub fn execute(
 }
 
 pub mod execute {
+    use connection_router::state::ChainName;
+
     use crate::state::{AuthorizationState, WORKERS, WORKERS_PER_CHAIN};
 
     use super::*;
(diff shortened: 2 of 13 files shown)


contracts/service-registry/src/state.rs
```diff
@@ -126,7 +127,7 @@ pub enum AuthorizationState {
 // maps service_name -> Service
 pub const SERVICES: Map<&str, Service> = Map::new("services");
 // maps (service_name, chain_name, worker_address) -> ()
-pub const WORKERS_PER_CHAIN: Map<(&str, &str, &Addr), ()> = Map::new("workers_per_chain");
+pub const WORKERS_PER_CHAIN: Map<(&ChainName, &str, &Addr), ()> = Map::new("workers_per_chain");
```
cjcobb23, line 130: I'd be curious if this change actually increases gas cost for `DeclareChainSupport`, because it changes the structure of the map. Before, every insert in that method shared the same first element of the primary key (service name), but now the first element is different for every insert. So it will be worse cache performance.

I'd prefer to keep the order the same as before. If you have to define some additional things for the compiler to be happy, that's ok.

contracts/connection-router/src/msg.rs
```diff
@@ -17,12 +17,12 @@ pub enum ExecuteMsg {
      */
     // Registers a new chain with the router
     RegisterChain {
-        chain: String,
+        chain: ChainName,
```
cjcobb23, line 20: To keep this PR simple, let's just change the service registry (and anything else we need to change because we changed the service registry). Let's leave the connection-router alone.

contracts/service-registry/src/contract.rs
```diff
@@ -336,21 +338,23 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> Result<Binary, ContractErr
...
-        chain_name: String,
+        chain_name: ChainName,
     ) -> Result<Vec<Worker>, ContractError> {
         let service = SERVICES
             .may_load(deps.storage, &service_name)?
             .ok_or(ContractError::ServiceNotFound)?;
 
         let workers = WORKERS_PER_CHAIN
-            .prefix((&service_name, &chain_name))
+            .prefix((&chain_name, &service_name))
```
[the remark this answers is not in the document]
cjcobb23: Actually ignore this, I'd rather just switch the order back to how it was.

contracts/connection-router/src/msg.rs
fish-sammy: Why leaving the router alone makes this PR simpler?

contracts/service-registry/src/contract.rs
```diff
@@ -250,7 +252,7 @@ pub mod execute {
             .ok_or(ContractError::WorkerNotFound)?;
 
         for chain in chains {
-            WORKERS_PER_CHAIN.save(deps.storage, (&service_name, &chain, &info.sender), &())?;
+            WORKERS_PER_CHAIN.save(deps.storage, (&chain, &service_name, &info.sender), &())?;
```
fish-sammy, line 255: Why switch the order of chain and service_name?

contracts/service-registry/src/contract.rs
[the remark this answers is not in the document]
fish-sammy: Was there any reason for the switch?

contracts/service-registry/src/state.rs
```diff
@@ -126,7 +127,7 @@ pub enum AuthorizationState {
 // maps service_name -> Service
 pub const SERVICES: Map<&str, Service> = Map::new("services");
 // maps (service_name, chain_name, worker_address) -> ()
```
fish-sammy, line 129: Need to update the comment if the switch absolutely has to happen.

contracts/service-registry/src/contract.rs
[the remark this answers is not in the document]
fahimahmedx: This was switched since `WORKERS_PER_CHAIN` now has ChainName first instead of second.

I'm going to keep trying to figure out how to implement `KeyDeserialize` for the `&ChainName` so I can keep it as second, which would make the switch not necessary.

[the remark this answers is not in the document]
fish-sammy: I obviously understand that you had to switch cuz the underlying data object switched. I was asking for the reason for this switch from the beginning, and I guess you kinda answered it. Ya please try to avoid the switch.

cjcobb23 (approved):

PR #150, opened 2023-10-12
feat(rewards): add contract abstraction

cjcobb23:
* implement logic for updating epoch

https://axelarnetwork.atlassian.net/browse/AXE-2167

### DIFF
--- a/contracts/rewards/src/contract/execute.rs
+++ b/contracts/rewards/src/contract/execute.rs
@@ -91,35 +91,34 @@ mod test {
 
     use super::Contract;
 
-    /// Tests that the current epoch is computed correctly based on block height and epoch duration
+    /// Tests that the current epoch is correctly computed when a number of blocks less than the epoch duration passes.
+    /// The epoch number should not increase.
     #[test]
-    fn test_get_current_epoch() {
-        let epoch_duration = 100u64;
-        let rewards_per_epoch: nonempty::Uint256 = Uint256::from(100u128).try_into().unwrap();
-        let current_epoch = Epoch {
-            epoch_num: 1,
-            block_height_started: 250,
-            rewards: rewards_per_epoch.clone(),
-        };
-        let stored_params = StoredParams {
-            params: RewardsParams {
-                participation_threshold: (Uint64::new(1), Uint64::new(2)).try_into().unwrap(),
-                epoch_duration: epoch_duration.try_into().unwrap(),
-                rewards_per_epoch,
-            },
-            last_updated: current_epoch.clone(),
-        };
-        let mut contract = create_contract(stored_params.clone());
-
+    fn test_get_current_epoch_same_epoch() {
+        let (mut contract, current_epoch, stored_params) = setup();
+        let epoch_duration: u64 = stored_params.params.epoch_duration.into();
         // epoch shouldn't change
         let new_epoch = contract
             .get_current_epoch(current_epoch.block_height_started + 1)
             .unwrap();
         assert_eq!(new_epoch, current_epoch);
 
+        // epoch still shouldn't change
+        let new_epoch = contract
+            .get_current_epoch(current_epoch.block_height_started + epoch_duration - 1)
+            .unwrap();
+        assert_eq!(new_epoch, current_epoch);
+    }
+
+    /// Tests that the current epoch is computed correctly when a number of blocks greater than or equal to the
+    /// epoch duration, but less than 2 times the epoch duration, passes. The epoch number should increase by one.
+    #[test]
+    fn test_get_current_epoch_next_epoch() {
+        let (mut contract, current_epoch, stored_params) = setup();
+        let epoch_duration: u64 = stored_params.params.epoch_duration.into();
+
         // epoch should increase by one
-        let block =
-            current_epoch.block_height_started + u64::from(stored_params.params.epoch_duration);
+        let block = current_epoch.block_height_started + epoch_duration;
         let new_epoch = contract.get_current_epoch(block).unwrap();
         assert_ne!(new_epoch, current_epoch);
         assert_eq!(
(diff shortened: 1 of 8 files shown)


contracts/rewards/src/contract/execute.rs
```diff
@@ -0,0 +1,184 @@
...
+    fn get_current_epoch(&mut self, cur_block_height: u64) -> Result<Epoch, ContractError> {
+        let stored_params = self.store.load_params();
+        let epoch_duration: u64 = stored_params.params.epoch_duration.into();
+        let epoch = stored_params.last_updated;
+        if cur_block_height >= epoch.block_height_started + epoch_duration {
+            let new_epoch_num = ((cur_block_height - epoch.block_height_started) / epoch_duration)
+                + epoch.epoch_num;
+            let new_epoch = Epoch {
+                epoch_num: new_epoch_num,
+                block_height_started: cur_block_height - (cur_block_height % epoch_duration),
```
cgorenflo, line 36: this is not accurate as soon as the epoch_duration param changes

cgorenflo: you need to do `(cur_block_height - epoch.blockc_height_started) % epoch_duraction`, because we can only look ata time window when the duration has been fixed

contracts/rewards/src/contract/execute.rs
```diff
@@ -0,0 +1,184 @@
...
+        assert_ne!(new_epoch, current_epoch);
+        assert_eq!(
+            Epoch {
+                epoch_num: 4,
+                block_height_started: block - 10,
+                rewards: current_epoch.rewards.clone()
+            },
+            new_epoch
+        );
+    }
```
cgorenflo, line 175: make these different tests or find a way to parameterize them so it's easier to understand what's going on

contracts/rewards/src/contract/execute.rs
```diff
@@ -0,0 +1,211 @@
...
+        assert_ne!(new_epoch, current_epoch);
+        assert_eq!(
+            Epoch {
+                epoch_num: 2,
+                block_height_started: block - epoch_duration / 2,
+                rewards: current_epoch.rewards.clone()
+            },
+            new_epoch
+        );
+    }
```
cgorenflo, line 145: legibility of this test can be improved by parameterization, something like this:
```
let test_cases = vex![(height1, expected_num1, expected_start1), (height2, expected_num2, expected_start2)]
for case in test_cases.iter(){
   // assertions
}
```

contracts/rewards/src/contract/execute.rs
```diff
@@ -0,0 +1,211 @@
...
+        assert_eq!(
+            Epoch {
+                epoch_num: 5,
+                block_height_started: block,
+                rewards: current_epoch.rewards.clone()
+            },
+            new_epoch
+        );
+
+        // epoch should increase by more than one, but start of epoch is in the past
```
cgorenflo, line 167: what does "in the past" mean? Just "earlier than `block`"?

contracts/rewards/src/contract/execute.rs
cgorenflo, line 1: the tests already look a lot cleaner and I have a couple of suggestions how to make them even better. It looks to me like there are actually only two different tests:
1. the call is idempotent
2. based on the current epoch, duration parameter and block height you get the expected result. To that end I'd rename the test `test_get_current_epoch_same_epoch` to `get_current_epoch_same_epoch_is_idempotent` (and adjust the comments), and combine all the other tests into a single one and parameterize it along the way I sketched out in my other comment about `test_cases`

One more nit pick: Test names in rust don't usually start with `test_` since they already have a `#[test]` attribute and therefore are clearly denoted as tests

contracts/rewards/src/contract/execute.rs
```diff
@@ -0,0 +1,211 @@
...
+
+    fn create_contract(stored_params: StoredParams) -> Contract<state::MockStore> {
+        let mut store = state::MockStore::new();
+        store
+            .expect_load_params()
+            .returning(move || stored_params.clone());
+        Contract { store }
+    }
+
+    fn setup() -> (Contract<state::MockStore>, Epoch, StoredParams) {
```
cgorenflo, line 189: I'd prefer to have parameters for duration, epoch_num and block_height_started, so they are clearly visible in the tests

contracts/rewards/src/contract/execute.rs
cjcobb23: yes, the function is called at block x, but the epoch started at block y, and y < x

contracts/rewards/src/contract/execute.rs
```diff
@@ -0,0 +1,190 @@
...
+    fn get_current_epoch(&mut self, cur_block_height: u64) -> Result<Epoch, ContractError> {
+        let stored_params = self.store.load_params();
+        let epoch_duration: u64 = stored_params.params.epoch_duration.into();
+        let epoch = stored_params.last_updated;
+        if cur_block_height >= epoch.block_height_started + epoch_duration {
+            let elapsed = cur_block_height - epoch.block_height_started;
+            let epoch_num = ((elapsed) / epoch_duration) + epoch.epoch_num;
+            let block_height_started = cur_block_height - ((elapsed) % epoch_duration);
+
+            let new_epoch = Epoch {
```
fish-sammy, line 36: Are you not storing this?

contracts/rewards/src/contract/execute.rs
```diff
@@ -0,0 +1,190 @@
...
+                epoch_num,
+                block_height_started,
+                rewards: epoch.rewards,
```
fish-sammy, line 39: Wouldn't this simply be `epoch.block_height_started + epoch_count * epoch_duration` if you do `let epoch_count = ((cur_block_height - epoch.block_height_started) / epoch_duration);` at line 32?

fish-sammy, line 39: What is the `rewards` stored in the latest known epoch? Why is it carried to the new epoch?

fish-sammy (changes requested):

contracts/rewards/src/contract/execute.rs
cjcobb23: no we just compute dynamically, we only need to store the current epoch when the epoch duration changes

contracts/rewards/src/contract/execute.rs
cjcobb23: It's the rewards from the last time the parameters changed, so it's the most up to date. It's carried to each new epoch, because it hasn't changed. This is the rewards rate, i.e. 100AXL tokens each epoch.

fish-sammy (approved):

cgorenflo (approved):

PR #151, opened 2023-10-12
feat: implement the execute method route_to_nexus for the nexus gateway contract

fish-sammy:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

### DIFF
--- a/contracts/nexus-gateway/src/nexus.rs
+++ b/contracts/nexus-gateway/src/nexus.rs
@@ -1,0 +1,81 @@
+use axelar_wasm_std::nonempty;
+use connection_router::state::{
+    Address, ChainName, Message as RouterMessage, MessageId, ID_SEPARATOR,
+};
+use cosmwasm_std::{CosmosMsg, CustomMsg};
+use error_stack::{Report, Result, ResultExt};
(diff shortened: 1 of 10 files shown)


contracts/nexus-gateway/src/nexus.rs
```diff
@@ -0,0 +1,74 @@
...
+        .change_context_lazy(|| ContractError::InvalidMessageId(message_id.clone()))?;
+
+    Ok((tx_id, index))
+}
+
+impl TryFrom<RouterMessage> for Message {
+    type Error = Report<ContractError>;
+
+    fn try_from(msg: RouterMessage) -> Result<Self, ContractError> {
+        let (source_tx_id, source_tx_index) = parse_message_id(&msg.cc_id.id)?;
```
cjcobb23, line 54: Isn't there supposed to be some sort of fallback if the message id can't be parsed into this format? The source chain can use whatever message id format it wants, it might not fit into this format.

cjcobb23 (changes requested):

cjcobb23 (approved):

PR #156, opened 2023-10-16
feat(multisig): store block at which the session was completed

eguajardo:
## Description
This is a prerequisite to integrate signing with rewards. We need the block at which the session was completed in order to calculate the grace period for signing late.

https://axelarnetwork.atlassian.net/browse/AXE-2313

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


contracts/multisig/src/contract.rs
```diff
@@ -132,20 +133,32 @@ pub mod execute {
...
+            MultisigState::Completed { completed_at } => {
+                Ok(Response::new().add_event(event.into()).add_event(
+                    Event::SigningCompleted {
+                        session_id,
+                        completed_at,
+                    }
+                    .into(),
+                ))
+            }
         }
```
cgorenflo, line 162: this is going to fire with every late signature

contracts/multisig/src/signing.rs
```diff
@@ -25,19 +24,19 @@ impl SigningSession {
...
 }
 
 fn validate_session_signature(
     session: &SigningSession,
-    signatures: &HashMap<String, Signature>,
     key: &Key,
     signer: String,
     signature: &Signature,
 ) -> Result<(), ContractError> {
     // TODO: revisit again once expiration and/or rewards are introduced
```
cgorenflo, line 38: both are introduced now, should this comment change?

contracts/multisig/src/signing.rs
```diff
@@ -79,15 +77,32 @@ fn signers_weight(
...
+    // weight must be loaded outside `if`, otherwise it could cause gas cost inconsistency among signers
+    let weight = signers_weight(store, session.id.u64(), &key.snapshot)?;
+    if session.state == MultisigState::Pending && weight >= key.snapshot.quorum.into() {
+        session.state = MultisigState::Completed {
+            completed_at: block_height,
+        };
+    }
+
+    SIGNING_SESSIONS.save(store, session.id.u64(), session)
+}
```
cgorenflo, line 95: I'd like to keep storage management separate from decision logic. Consider loading and saving separately

contracts/multisig/src/signing.rs
```diff
@@ -79,15 +77,32 @@ fn signers_weight(
...
+
 pub fn sign(
```
cgorenflo, line 97: consider if it makes sense to follow this pattern (sketch)
```
let mut session = load_session(store)
session.sign(signature)
save_session(session)
```
or even better, keep the session immutable if that makes sense:
```
let session = load_session(store).sign(signature)
save_session(session)
```

contracts/multisig/src/contract.rs
eguajardo: I did this on purpose, I assumed either way it would be helpful to get back an event indicating the session was completed. I added the `completed_at` to the event so it can be determined if the completion was in the past. Thoughts?

contracts/multisig/src/signing.rs
eguajardo: I'm removing this on the next PR when I'm taking the grace period into account in this `if`

contracts/multisig/src/signing.rs
```diff
@@ -25,19 +24,19 @@ impl SigningSession {
...
 fn validate_session_signature(
     session: &SigningSession,
-    signatures: &HashMap<String, Signature>,
     key: &Key,
     signer: String,
     signature: &Signature,
 ) -> Result<(), ContractError> {
     // TODO: revisit again once expiration and/or rewards are introduced
-    if calculate_session_state(signatures, &key.snapshot)? == MultisigState::Completed {
+    if let MultisigState::Completed { .. } = session.state {
```
fish-sammy, line 39:
```suggestion
    if matches!(session.state, MultisigState::Completed { .. }) {
```

contracts/multisig/src/signing.rs
fish-sammy: Agree with Chris. The code atm is quite procedural.

contracts/multisig/src/signing.rs
```diff
@@ -25,19 +24,19 @@ impl SigningSession {
...
+            state: MultisigState::Pending,
         }
     }
 }
 
 fn validate_session_signature(
     session: &SigningSession,
-    signatures: &HashMap<String, Signature>,
     key: &Key,
     signer: String,
```
fish-sammy, line 35:
```suggestion
    signer: &str,
```
A validation method should not take away the ownership.

contracts/multisig/src/signing.rs
```diff
@@ -62,10 +61,9 @@ fn validate_session_signature(
     Ok(())
```
fish-sammy, line 61:
```suggestion
    match key.pub_keys.get(&signer) {
        Some(pub_key) if !signature.verify(&session.msg, pub_key)? => {
            Err(ContractError::InvalidSignature {
                session_id: session.id,
                signer,
            })
        }
        None => Err(ContractError::NotAParticipant {
            session_id: session.id,
            signer,
        }),
        _ => Ok(()),
    }
```
Replace the above match with.

contracts/multisig/src/signing.rs
fish-sammy: Agree with Chris

fish-sammy (changes requested):

contracts/multisig/src/contract.rs
cgorenflo: I think it's more confusing than helpful. Events don't get shown as a response when you call the contract from the cli for example, rather they get returned to specific subscription which could make the reasonable assumption that every signing can only completed once. If you want to return an event, it should be a different one, like `SignedAfterCompletion`

contracts/multisig/src/signing.rs
eguajardo: changed this to `&Addr`

contracts/multisig/src/signing.rs
eguajardo: did some restructuring, please take a look again

contracts/multisig/src/signing.rs
eguajardo: did some restructuring, please take a look again

contracts/multisig/src/signing.rs
```diff
@@ -183,111 +153,75 @@ mod tests {
...
-                SigningSession::new(Uint64::one(), config.key_id.clone(), config.message.clone());
-
-            // first run
-            let (_, result) = do_sign(&session, 0, &mut config).unwrap();
-            assert_eq!(result, MultisigState::Pending);
-
-            // second run
-            let (_, result) = do_sign(&session, 1, &mut config).unwrap();
-            assert_eq!(result, MultisigState::Completed);
+    fn test_success_validation() {
```
cgorenflo, line 202: no need to start test functions with `test_`, they are already clearly denoted by the attribute and general rust convention forgoes the prefix

cgorenflo (approved):

fish-sammy (approved):

PR #157, opened 2023-10-16
refactor: change payload hash to 32 byte array 

cgorenflo:


contracts/connection-router/src/state/payload_hex.rs
```diff
@@ -0,0 +1,19 @@
...
+}
+
+pub fn deserialize<'de, D>(deserializer: D) -> Result<[u8; 32], D::Error>
+where
+    D: Deserializer<'de>,
+{
+    HexBinary::deserialize(deserializer)?
+        .to_array::<32>()
+        .map_err(|err| Error::custom(err))
+}
```
fish-sammy, line 19: Why not?
```suggestion
pub fn serialize<S, const N: usize>(value: &[u8; N], serializer: S) -> Result<S::Ok, S::Error>
where
    S: Serializer,
{
    HexBinary::from(value).serialize(serializer)
}

pub fn deserialize<'de, D, const N: usize>(deserializer: D) -> Result<[u8; N], D::Error>
where
    D: Deserializer<'de>,
{
    HexBinary::deserialize(deserializer)?
        .to_array::<N>()
        .map_err(|err| Error::custom(err))
}
```

fish-sammy: Also, I am not a fan of this mod name. It's really just serializing `[u8: N]` as hex string. Shouldn't have to be tied with payload_hash.

contracts/voting-verifier/src/events.rs
```diff
@@ -145,7 +145,7 @@ impl TryFrom<Message> for TxEventConfirmation {
             destination_address: other.destination_address,
             destination_chain: other.destination_chain,
             source_address: other.source_address,
-            payload_hash: other.payload_hash,
+            payload_hash: HexBinary::from(other.payload_hash),
```
fish-sammy, line 148: Why is this event `TxEventConfirmation` not changed?

contracts/connection-router/src/state/payload_hex.rs
cgorenflo: nice! I wasn't aware you could do `const N`. The reason I went with `payload_hash` is that I had it as `hex` first, but that collides with the std libs hex crate and I couldn't figure out how to resolve that naming conflict. I'm open for suggestions

contracts/voting-verifier/src/events.rs
cgorenflo: Didn't want to broaden the scope of the PR and didn't check what exactly this is. Let me see if it makes sense to include it into this PR

contracts/connection-router/src/state.rs
```diff
@@ -98,7 +96,7 @@ pub struct Message {
     pub destination_address: Address,
     /// for better user experience, the payload hash gets encoded into hex at the edges (input/output),
     /// but internally, we treat it as raw bytes to enforce it's format.
-    #[serde(with = "payload_hex")]
+    #[serde(with = "axelar_wasm_std::hex")]
```
fish-sammy, line 99: 👍

fish-sammy (approved):

PR #158, opened 2023-10-16
feat: nexus gateway contract route messages to nexus only if they have not been routed yet

fish-sammy:
## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

### DIFF
--- a/contracts/nexus-gateway/src/contract/execute.rs
+++ b/contracts/nexus-gateway/src/contract/execute.rs
@@ -1,6 +1,6 @@
 use connection_router::state::Message;
 use cosmwasm_std::{Addr, Response};
-use error_stack::ResultExt;
+use error_stack::{report, ResultExt};
 
 use crate::error::ContractError;
 use crate::nexus;
@@ -24,7 +24,7 @@ where
             sender if sender == self.config.router => self
                 .route_to_nexus(msgs)
                 .change_context(ContractError::RouteToNexus),
-            _ => Err(ContractError::Unauthorized)?,
+            _ => Err(report!(ContractError::Unauthorized)),
         }
     }
 
@@ -38,9 +38,8 @@ where
             })
             .collect::<Result<Vec<_>>>()?;
 
-        for msg in &msgs {
-            self.store.set_message_routed(&msg.cc_id)?;
-        }
+        msgs.iter()
+            .try_for_each(|msg| self.store.set_message_routed(&msg.cc_id))?;
 
         let msgs = msgs
             .into_iter()
@@ -62,7 +61,7 @@ mod test {
     use super::*;
 
     #[test]
-    fn test_route_messages_unauthorized() {
+    fn route_messages_unauthorized() {
         let store = MockStore::new();
         let config = Config {
             nexus: Addr::unchecked("nexus"),
@@ -76,7 +75,7 @@ mod test {
     }
 
     #[test]
-    fn test_route_to_nexus_with_no_msg() {
+    fn route_to_nexus_with_no_msg() {
         let store = MockStore::new();
         let config = Config {
             nexus: Addr::unchecked("nexus"),
@@ -90,7 +89,7 @@ mod test {
     }
 
     #[test]
-    fn test_route_to_nexus_with_msgs_that_have_not_been_routed() {
+    fn route_to_nexus_with_msgs_that_have_not_been_routed() {
         let mut store = MockStore::new();
         let config = Config {
             nexus: Addr::unchecked("nexus"),
@@ -110,16 +109,10 @@ mod test {
             Message {
                 cc_id: CrossChainId {
                     chain: "sourceChain".parse().unwrap(),
-                    id: "0xab550540ea222040733516376bef3b8dd564c1cc35144e123db3dc04d3bd2fe4:0"
-                        .parse()
-                        .unwrap(),
+                    id: "0x2fe4:0".parse().unwrap(),
                 },
-                source_address: "0x05c0cada09A16b4e665894F688193050b67Cb860"
-                    .parse()
-                    .unwrap(),
-                destination_address: "0xae0Ee0A63A2cE6BaeEFFE56e7714FB4EFE48D419"
-                    .parse()
-                    .unwrap(),
+                source_address: "0xb860".parse().unwrap(),
+                destination_address: "0xD419".parse().unwrap(),
                 destination_chain: "destinationChain".parse().unwrap(),
                 payload_hash: decode(
                     "bb9b5566c2f4876863333e481f4698350154259ffe6226e283b16ce18a64bcf1",
@@ -130,16 +123,10 @@ mod test {
             Message {
                 cc_id: CrossChainId {
                     chain: "sourceChain".parse().unwrap(),
-                    id: "0x5c25214402887813674e5be1922a58f60ea14400380ec7840c8c4fa064826b33:10"
-                        .parse()
-                        .unwrap(),
+                    id: "0x6b33:10".parse().unwrap(),
                 },
-                source_address: "0x635E6496D51514f0b64265db24e1FC4AEff70725"
-                    .parse()
-                    .unwrap(),
-                destination_address: "0x3fC91A3afd70395Cd496C647d5a6CC9D4B2b7FAD"
-                    .parse()
-                    .unwrap(),
+                source_address: "0x0725".parse().unwrap(),
+                destination_address: "0x7FAD".parse().unwrap(),
                 destination_chain: "destinationChain".parse().unwrap(),
                 payload_hash: decode(
                     "bb9b5566c2f4876863333e481f4698350154259ffe6226e283b16ce18a64bcf1",
@@ -154,7 +141,7 @@ mod test {
     }
 
     #[test]
-    fn test_route_to_nexus_with_msgs_that_have_been_routed() {
+    fn route_to_nexus_with_msgs_that_have_been_routed() {
         let mut store = MockStore::new();
         let config = Config {
             nexus: Addr::unchecked("nexus"),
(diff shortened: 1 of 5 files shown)


contracts/nexus-gateway/src/state.rs
```diff
@@ -23,14 +31,29 @@ impl<'a> GatewayStore<'a> {
...
+        CONFIG
+            .save(self.storage, &config)
+            .change_context(ContractError::StoreFailure)
     }
 
-    fn load_config(&self) -> Config {
+    fn load_config(&self) -> Result<Config> {
         CONFIG
             .load(self.storage)
-            .expect("config must be set during contract instantiation")
```
cjcobb23, line 33: why this change? Feel like this should never happen.

contracts/nexus-gateway/src/contract/execute.rs
```diff
@@ -1,17 +1,54 @@
...
+        for msg in &msgs {
+            self.store.set_message_routed(&msg.cc_id)?;
+        }
+
+        let msgs = msgs
+            .into_iter()
+            .map(TryInto::try_into)
+            .collect::<Result<Vec<nexus::Message>>>()?;
+
+        Ok(Response::new().add_messages(msgs))
```
cjcobb23, line 50: This message will definitely error if the nexus fails to route the message right? Just want to make sure a message isn't set as routed here, but then not actually routed in the nexus.

fish-sammy: yes

contracts/nexus-gateway/src/state.rs
fish-sammy: This could happen if deserialization fails. For example, we have some custom deserializer implemented for the Config where we do some custom validation.

cjcobb23: yeah but shouldn't we panic in that situation?

contracts/nexus-gateway/src/contract/execute.rs
cjcobb23: This actually makes me wonder, why don't we track the messages in nexus, instead of here? Since nexus has the restriction of only once routing, why not just enforce that there?

contracts/nexus-gateway/src/contract/execute.rs
```diff
@@ -1,17 +1,54 @@
...
+        mut self,
+        sender: Addr,
+        msgs: Vec<Message>,
+    ) -> Result<Response<nexus::Message>> {
+        match sender {
+            sender if sender == self.config.nexus => todo!(),
+            sender if sender == self.config.router => self
+                .route_to_nexus(msgs)
+                .change_context(ContractError::RouteToNexus),
+            _ => Err(ContractError::Unauthorized)?,
```
cgorenflo, line 27: it's sad that this has to be a conditional match instead of a direct equality comparison, but I still prefer this match pattern over an `if.. else if` chain

cgorenflo, line 27: I think I prefer `Err(report!(ContractError::Unauthorized))` or `bail!(ContractError::Unauthorized)` over the use of `?` here

contracts/nexus-gateway/src/contract/execute.rs
```diff
@@ -1,17 +1,54 @@
...
+    pub fn route_messages(
+        mut self,
+        sender: Addr,
+        msgs: Vec<Message>,
+    ) -> Result<Response<nexus::Message>> {
+        match sender {
+            sender if sender == self.config.nexus => todo!(),
+            sender if sender == self.config.router => self
+                .route_to_nexus(msgs)
+                .change_context(ContractError::RouteToNexus),
```
cgorenflo, line 26: consider moving the context switch into `route_to_nexus`

contracts/nexus-gateway/src/contract/execute.rs
```diff
@@ -1,17 +1,54 @@
...
+                Ok(false) => Some(Ok(msg)),
+                Err(err) => Some(Err(err)),
+            })
+            .collect::<Result<Vec<_>>>()?;
+
+        for msg in &msgs {
+            self.store.set_message_routed(&msg.cc_id)?;
+        }
+
+        let msgs = msgs
```
cgorenflo, line 45: use `inspect` to make all of this a single expression

contracts/nexus-gateway/src/contract/execute.rs
```diff
@@ -20,18 +57,115 @@ mod test {
 
     use connection_router::state::{CrossChainId, Message};
 
+    use crate::state::{Config, MockStore};
+
     use super::*;
 
+    #[test]
+    fn test_route_messages_unauthorized() {
```
cgorenflo, line 65: we don't have to start test functions with `test_`. They already have the `#[test]` attribute and from what I've seen so far it seems to be convention to directly describe the test, e.g. `do_not_route_messages_from_unauthorized_sender`

contracts/nexus-gateway/src/contract/execute.rs
```diff
@@ -20,18 +57,115 @@ mod test {
...
+        let store = MockStore::new();
+        let config = Config {
+            nexus: Addr::unchecked("nexus"),
+            router: Addr::unchecked("router"),
+        };
+        let contract = Contract { store, config };
+
+        let res = contract.route_messages(Addr::unchecked("unauthorized"), vec![]);
+
+        assert!(res.is_err_and(|err| matches!(err.current_context(), ContractError::Unauthorized)));
```
cgorenflo, line 75: nice! I like the use of `is_err_and`

contracts/nexus-gateway/src/contract/execute.rs
```diff
@@ -20,18 +57,115 @@ mod test {
...
+                destination_address: "0x3fC91A3afd70395Cd496C647d5a6CC9D4B2b7FAD"
+                    .parse()
+                    .unwrap(),
+                destination_chain: "destinationChain".parse().unwrap(),
+                payload_hash: decode(
+                    "bb9b5566c2f4876863333e481f4698350154259ffe6226e283b16ce18a64bcf1",
+                )
+                .unwrap()
+                .into(),
+            },
```
cgorenflo, line 149: can you make this more compact? The exact content of messages doesn't really matter for these tests. (E.g. use shorter hexes so the parse().unwrap() doesn't wrap or extract the message creation into another function

contracts/nexus-gateway/src/state.rs
cgorenflo: panic if the error is due to a programming error, Err() if it is due to input data (or something circumstantial like unavailable services).
So if the some validation of the config fails that should be an Err

cjcobb23: But shouldn't we not save a config that can't be deserialized? Which would mean if we get here, then we saved a config that we shouldn't have saved, hence it would be a programming error.

My point is, I don't see how we can get to this state, where the config can't be loaded. It seems like a totally invalid contract state to me. I also don't know what can even be done on the contract if the config can't be loaded, the whole contract is basically dead at this point.

cgorenflo: at that point you are baking knowledge about which data came from the outside and which was created internally into the store. What if the config was never (de)serialized before? It's still very easy to convert this Err into a panic at a higher level were we have explicit knowledge about where the struct came from

cgorenflo: in conclusion: 
- should the contract panic if the config can't be loaded? Most likely, yes
- should the store know in which cases it's safe to panic vs returning an Err? I don't think so

contracts/nexus-gateway/src/contract/execute.rs
cgorenflo: good point!

fish-sammy: This contract is an extension of nexus as the name suggests. The code here is more upgradable than nexus.

contracts/nexus-gateway/src/contract/execute.rs
fish-sammy: sure

contracts/nexus-gateway/src/state.rs
fish-sammy: Same to me. It should be the contract instead of the store to determine when to panic.

contracts/nexus-gateway/src/contract/execute.rs
fish-sammy: I like it to be here cuz `route_to_nexus` reports the specifics about why `RouteToNexus` occurs.

contracts/nexus-gateway/src/contract/execute.rs
fish-sammy: Can't do `inspect` cuz `set_message_routed` returns a Result.

contracts/nexus-gateway/src/contract/execute.rs
fish-sammy: sure

contracts/nexus-gateway/src/contract/execute.rs
```diff
@@ -20,54 +56,135 @@ mod test {
...
+        ];
+        let res = contract.route_messages(Addr::unchecked("router"), msgs);
+
+        assert!(res.is_ok_and(|res| res.messages.len() == 2));
     }
 
     #[test]
-    fn test_route_to_nexus_with_msgs() {
+    fn route_to_nexus_with_msgs_that_have_been_routed() {
```
cjcobb23, line 144: might be good to test a mix of messages that have been routed and some that haven't

cjcobb23 (approved):

cgorenflo (approved):

PR #160, opened 2023-10-17
feat(multisig): require StartSigningSession to only be called by an authorized set of callers

fahimahmedx:
## Description
Requires an authorized caller to be able to start a signing session, as outlined in this [doc](https://www.notion.so/bright-ambert-2bd/Amplifier-Multisig-Access-Control-103a72112dc44c6185ca010be01f36c2):

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Steps to Test
`cargo test`

## Expected Behaviour
`cargo test` passes


contracts/multisig/src/contract.rs
```diff
@@ -218,6 +229,41 @@ pub mod execute {
             .into(),
         ))
     }
+
+    pub fn require_valid_caller(deps: &DepsMut, info: MessageInfo) -> Result<(), ContractError> {
```
cgorenflo, line 233: pass in the sender directly, `MessageInfo` isn't needed here

contracts/multisig/src/msg.rs
```diff
@@ -33,6 +33,12 @@ pub enum ExecuteMsg {
     RegisterPublicKey {
         public_key: PublicKey,
     },
+    AuthorizeCaller {
+        contract_address: Addr,
+    },
+    UnauthorizeCaller {
+        contract_address: Addr,
+    },
```
cjcobb23, line 41: Add some comments documenting these messages

contracts/multisig/src/state.rs
```diff
@@ -45,3 +45,5 @@ pub fn get_key(store: &dyn Storage, key_id: &KeyID) -> Result<Key, ContractError
 
 // key type is part of the key so signers can register multiple keys with different types
 pub const PUB_KEYS: Map<(Addr, KeyType), HexBinary> = Map::new("registered_pub_keys");
+
+pub const AUTHORIZED_CALLERS: Map<Addr, ()> = Map::new("authorized_callers");
```
cjcobb23, line 49: Add a comment describing what this is for

contracts/multisig/src/contract.rs
```diff
@@ -218,6 +229,41 @@ pub mod execute {
...
+    pub fn authorize_caller(
+        deps: DepsMut,
+        contract_address: Addr,
+    ) -> Result<Response, ContractError> {
+        AUTHORIZED_CALLERS.save(deps.storage, contract_address.clone(), &())?;
+
+        Ok(Response::new().add_event(Event::CallerAuthorized { contract_address }.into()))
+    }
+
+    pub fn remove_caller(deps: DepsMut, contract_address: Addr) -> Result<Response, ContractError> {
```
cjcobb23, line 252: should probably be called `unauthorize_caller` to match the message

contracts/multisig/src/contract.rs
```diff
@@ -218,6 +229,41 @@ pub mod execute {
             .into(),
         ))
     }
+
+    pub fn require_valid_caller(deps: &DepsMut, info: MessageInfo) -> Result<(), ContractError> {
+        AUTHORIZED_CALLERS
+            .load(deps.storage, info.sender.clone())
+            .map_err(|_| ContractError::Unauthorized {
+                contract_address: info.sender,
+            })?;
```
cjcobb23, line 238: can we use `change_context` here?

contracts/multisig/src/contract.rs
cjcobb23, line 233:
```suggestion
    pub fn require_authorized_caller(deps: &DepsMut, info: MessageInfo) -> Result<(), ContractError> {
```

contracts/multisig/src/error.rs
```diff
@@ -48,4 +48,7 @@ pub enum ContractError {
 
     #[error("key type mismatch")]
     KeyTypeMismatch,
+
+    #[error("caller is not authorized")]
+    Unauthorized { contract_address: Addr },
```
cjcobb23, line 53: I don't think we need the contract address here, since that address is the caller

contracts/multisig/src/contract.rs
```diff
@@ -218,6 +229,41 @@ pub mod execute {
...
+        Ok(Response::new().add_event(Event::CallerAuthorized { contract_address }.into()))
+    }
+
+    pub fn remove_caller(deps: DepsMut, contract_address: Addr) -> Result<Response, ContractError> {
+        AUTHORIZED_CALLERS.remove(deps.storage, contract_address.clone());
+
+        Ok(Response::new().add_event(Event::CallerUnauthorized { contract_address }.into()))
+    }
+
+    pub fn require_governance(deps: &DepsMut, info: MessageInfo) -> Result<(), ContractError> {
```
cgorenflo, line 258: pass the sender instead of info

contracts/multisig/src/error.rs
fahimahmedx: I found it helpful to see what address was trying to call the function when errors were thrown. Do you mean to rename `contract_address` to `caller`?

contracts/multisig/src/msg.rs
```diff
@@ -33,6 +33,14 @@ pub enum ExecuteMsg {
     RegisterPublicKey {
         public_key: PublicKey,
     },
+    AuthorizeCaller {
+        // Adds an address to AUTHORIZED_CALLERS, representing the addresses that can start a signing session.
```
cjcobb23, line 37: move this comment to above the message type `AuthorizeCaller` and don't mention the internal data structure. This is the API, so implementation shouldn't be mentioned here. Say something like "Authorizes a contract to call StartSigningSession". Also, probably add a comment on `StartSigningSession` saying something like "Can only be called by an authorized contract"

contracts/multisig/src/msg.rs
```diff
@@ -33,6 +33,14 @@ pub enum ExecuteMsg {
...
+        contract_address: Addr,
+    },
+    UnauthorizeCaller {
+        // Removes an address from AUTHORIZED_CALLERS
```
cjcobb23, line 41: Don't mention the internal data structure `AUTHORIZED_CALLERS` and move this comment above the message type

contracts/multisig/src/contract.rs
```diff
@@ -218,6 +229,41 @@ pub mod execute {
...
+    ) -> Result<Response, ContractError> {
+        AUTHORIZED_CALLERS.remove(deps.storage, &contract_address);
+
+        Ok(Response::new().add_event(Event::CallerUnauthorized { contract_address }.into()))
+    }
+
+    pub fn require_governance(deps: &DepsMut, contract_address: Addr) -> Result<(), ContractError> {
+        let config = CONFIG.load(deps.storage)?;
+        if config.governance != contract_address {
+            return Err(ContractError::Unauthorized);
```
cjcobb23, line 263: should test this case too

contracts/multisig/src/contract.rs
```diff
@@ -218,6 +229,41 @@ pub mod execute {
...
+    pub fn unauthorize_caller(
+        deps: DepsMut,
+        contract_address: Addr,
+    ) -> Result<Response, ContractError> {
+        AUTHORIZED_CALLERS.remove(deps.storage, &contract_address);
+
+        Ok(Response::new().add_event(Event::CallerUnauthorized { contract_address }.into()))
+    }
+
+    pub fn require_governance(deps: &DepsMut, contract_address: Addr) -> Result<(), ContractError> {
```
cgorenflo, line 260:
```suggestion
    pub fn require_governance(deps: &DepsMut, sender: Addr) -> Result<(), ContractError> {
```

cgorenflo (approved):

cjcobb23 (approved):

PR #162, opened 2023-10-18
feat(rewards): implement add rewards

cjcobb23:
https://axelarnetwork.atlassian.net/browse/AXE-2352


contracts/rewards/src/contract/execute.rs
```diff
@@ -156,18 +173,125 @@ mod test {
...
+        let epoch_duration = 100u64;
+        let rewards_store = Arc::new(RwLock::new(HashMap::new()));
+
+        let mut contract = setup_with_stores(
+            cur_epoch_num,
+            block_height_started,
+            epoch_duration,
+            rewards_store,
+        );
+        let test_data = vec![
```
cgorenflo, line 236: add a comment what the vec means

contracts/rewards/src/contract/execute.rs
```diff
@@ -156,18 +173,125 @@ mod test {
         }
     }
 
-    fn create_contract(stored_params: StoredParams) -> Contract<state::MockStore> {
+    /// Tests that rewards are added correctly to a single contract
```
cgorenflo, line 176: the test look good overall, but I'd like to further remove them from implementation details (e.g. the important thing is that the pool tracks the correct amount, not if it's initialized or not)
I think we need a query to check reward pool levels anyway, so that could be a good way to further remove the test from the exact data model (in a follow-up?). So that means instead of checking the store directly you could query for the amount

cgorenflo (approved):

contracts/rewards/src/contract/execute.rs
cjcobb23: We can add a query, and possibly other queries, in a follow up

PR #163, opened 2023-10-19
feat(rewards): update params

cjcobb23:
https://axelarnetwork.atlassian.net/browse/AXE-2372

### DIFF
--- a/contracts/rewards/src/contract/execute.rs
+++ b/contracts/rewards/src/contract/execute.rs
@@ -31,19 +31,18 @@ where
         if cur_block_height < epoch.block_height_started {
             return Err(ContractError::BlockHeightPassed.into());
         }
-        if cur_block_height >= epoch.block_height_started + epoch_duration {
-            let epochs_elapsed = (cur_block_height - epoch.block_height_started) / epoch_duration;
-            let epoch_num = epochs_elapsed + epoch.epoch_num;
-            let block_height_started = epoch.block_height_started + epochs_elapsed * epoch_duration;
-
-            let new_epoch = Epoch {
-                epoch_num,
-                block_height_started,
-                rewards: epoch.rewards,
-            };
-            return Ok(new_epoch);
+        if cur_block_height < epoch.block_height_started + epoch_duration {
+            return Ok(epoch);
         }
-        Ok(epoch)
+        let epochs_elapsed = (cur_block_height - epoch.block_height_started) / epoch_duration;
+        let epoch_num = epochs_elapsed + epoch.epoch_num;
+        let block_height_started = epoch.block_height_started + epochs_elapsed * epoch_duration;
+
+        Ok(Epoch {
+            epoch_num,
+            block_height_started,
+            rewards: epoch.rewards,
+        })
     }
 
     pub fn record_participation(
--- a/contracts/rewards/src/contract/execute.rs
+++ b/contracts/rewards/src/contract/execute.rs
@@ -1,8 +1,7 @@
-use error_stack::Result;
-use std::collections::HashMap;
-
 use axelar_wasm_std::nonempty::Uint256;
 use cosmwasm_std::Addr;
+use error_stack::Result;
+use std::collections::HashMap;
 
 use crate::{
     error::ContractError,
@@ -29,6 +28,9 @@ where
         let stored_params = self.store.load_params();
         let epoch_duration: u64 = stored_params.params.epoch_duration.into();
         let epoch = stored_params.last_updated;
+        if cur_block_height < epoch.block_height_started {
+            return Err(ContractError::BlockHeightPassed.into());
+        }
         if cur_block_height >= epoch.block_height_started + epoch_duration {
             let epochs_elapsed = (cur_block_height - epoch.block_height_started) / epoch_duration;
             let epoch_num = epochs_elapsed + epoch.epoch_num;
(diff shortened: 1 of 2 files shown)


contracts/rewards/src/contract/execute.rs
```diff
@@ -64,10 +65,30 @@ where
...
+        // epochs 1-4, and so all events prior to the start of epoch 1 have an epoch number of 0)
+        if cur_epoch.block_height_started + u64::from(new_params.epoch_duration) < block_height {
+            cur_epoch = Epoch {
+                block_height_started: block_height,
+                epoch_num: cur_epoch.epoch_num + 1,
+                rewards: new_params.rewards_per_epoch,
+            };
+        } else {
+            cur_epoch.rewards = new_params.rewards_per_epoch;
+        }
```
cgorenflo, line 86:
```suggestion
        let should_end = cur_epoch.block_height_started + u64::from(new_params.epoch_duration) < block_height;
        let cur_epoch =
            if  should_end{
                Epoch {
                    block_height_started: block_height,
                    epoch_num: cur_epoch.epoch_num + 1,
                    rewards: new_params.rewards_per_epoch,
                }
            } else {
                Epoch {                    
                    rewards: new_params.rewards_per_epoch,
                    ..cur_epoch
                }
            };
```
this way both branches are on equal footing, the whole thing is an expression and you don't need to mutate state

contracts/rewards/src/contract/execute.rs
```diff
@@ -156,20 +179,221 @@ mod test {
...
+        let new_params = RewardsParams {
+            rewards_per_epoch: cosmwasm_std::Uint256::from(initial_rewards_per_epoch + 100)
+                .try_into()
+                .unwrap(),
+            participation_threshold: (Uint64::new(2), Uint64::new(3)).try_into().unwrap(),
+            epoch_duration: epoch_duration.try_into().unwrap(), // keep this the same to not affect epoch computation
+        };
+
+        // the epoch shouldn't change when the params are updated, since we are not changing the epoch duration
+        let expected_epoch = contract.get_current_epoch(cur_height).unwrap();
```
cgorenflo, line 211: a few things I realized:
- get_current_epoch should return an error if the cur_height is in the past (otherwise the returned value is nonsensical)
- get_current_epoch is a private function and should remain that, it's kind of weird signature doesn't work well for a public interface of the contract
- in general the tests in the rewards module are tied very close to the actual implementation instead of using only public interfaces. I think we'll need to rework them when all contract interfaces are implemented

cjcobb23: added the error. I think it will be easier to write tests using only public interfaces once all of the methods are implemented (and probably some queries too).

contracts/rewards/src/contract/execute.rs
```diff
@@ -30,6 +28,9 @@ where
         let stored_params = self.store.load_params();
         let epoch_duration: u64 = stored_params.params.epoch_duration.into();
         let epoch = stored_params.last_updated;
+        if cur_block_height < epoch.block_height_started {
```
fish-sammy, line 31: Shouldn't this be a panic situation?

contracts/rewards/src/contract/execute.rs
```diff
@@ -30,6 +28,9 @@ where
...
+            return Err(ContractError::BlockHeightPassed.into());
+        }
         if cur_block_height >= epoch.block_height_started + epoch_duration {
```
fish-sammy, line 34:
```suggestion
        if cur_block_height < epoch.block_height_started + epoch_duration {
          return Ok(epoch)
        }

        ...
```
I always code to look cleaner when exit early and have the complicated code reside in where the least amount of indentations are.

contracts/rewards/src/contract/execute.rs
```diff
@@ -66,10 +67,35 @@ where
...
+                block_height_started: block_height,
+                epoch_num: cur_epoch.epoch_num + 1,
+                rewards: new_params.rewards_per_epoch,
+            }
+        } else {
+            Epoch {
+                rewards: new_params.rewards_per_epoch,
+                ..cur_epoch
+            }
+        };
```
fish-sammy, line 93:
```suggestion
        let cur_epoch = Epoch {
            epoch_num: cur_epoch.epoch_num + should_end as u64,
            block_height_started: if should_end {
                block_height
            } else {
                cur_epoch.block_height_started
            },
            rewards: new_params.rewards_per_epoch,
        };
```
This results in fewer lines of code but I would NOT say it's cleaner. I'll leave it up to you.

fish-sammy: Also, does it really hurt to always reset `block_height_started` to be the current block and always increment the `epoch_num` by 1? I feel it doesn't make any difference and would result in a simpler solution, no?

contracts/rewards/src/contract/execute.rs
cjcobb23: I feel like it's up to the caller to decide whether to panic or handle the error. But I don't feel strongly

contracts/rewards/src/contract/execute.rs
cjcobb23: if only rust had a ternary operator.

I don't think we can always increment the epoch num by 1. There are a variety of cases where the params update doesn't end the current epoch, in which case I don't think it makes sense to increment the epoch number (for example, the epoch duration doesn't change, the epoch duration is increasing, etc). I think it would be surprising and confusing to users, validators and developers to end the epoch in these scenarios.

contracts/rewards/src/contract/execute.rs
```diff
@@ -30,19 +28,21 @@ where
         let stored_params = self.store.load_params();
```
cgorenflo, line 28: I would just call the function `current_epoch` without the `get_`

contracts/rewards/src/error.rs
```diff
@@ -32,4 +32,7 @@ pub enum ContractError {
 
     #[error("invalid event id")]
     InvalidEventId,
+
+    #[error("specified block has already passed")]
+    BlockHeightPassed,
```
cgorenflo, line 37:
```suggestion
    BlockHeightInPast,
```

contracts/rewards/src/contract/execute.rs
fish-sammy: This one to me is the caller to blame, and later on may not be the only error that can occur in this method. I'd still prefer panic inside.

cgorenflo (approved):

PR #164, opened 2023-10-23
feat(multisig): integrate signing with rewards contract

eguajardo:
## Description
Calls reward contract to record participation during signature submission. Implements grace period to record participation for late signing

https://axelarnetwork.atlassian.net/browse/AXE-2097

## Todos

In future PR the prover will be updated to optimize the minimum amount of signatures required to construct the proof

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [x] Connect epics/issues

### DIFF
--- a/contracts/multisig/src/signing.rs
+++ b/contracts/multisig/src/signing.rs
@@ -29,18 +29,24 @@ impl SigningSession {
         }
     }
 
+    /// returns true if session state was changed
     pub fn recalculate_session_state(
         &mut self,
         signatures: &HashMap<String, Signature>,
         snapshot: &Snapshot,
         block_height: u64,
-    ) {
+    ) -> bool {
         let weight = signers_weight(signatures, snapshot);
 
-        if self.state == MultisigState::Pending && weight >= snapshot.quorum.into() {
-            self.state = MultisigState::Completed {
-                completed_at: block_height,
-            };
+        match self.state {
+            MultisigState::Pending if weight >= snapshot.quorum.into() => {
+                self.state = MultisigState::Completed {
+                    completed_at: block_height,
+                };
+
+                true
+            }
+            _ => false,
         }
     }
 }
@@ -196,16 +202,19 @@ mod tests {
             let signatures = config.signatures;
             let block_height = 12345;
 
-            session.recalculate_session_state(&HashMap::new(), &key.snapshot, block_height);
+            let res =
+                session.recalculate_session_state(&HashMap::new(), &key.snapshot, block_height);
             assert_eq!(session.state, MultisigState::Pending);
+            assert!(!res);
 
-            session.recalculate_session_state(&signatures, &key.snapshot, block_height);
+            let res = session.recalculate_session_state(&signatures, &key.snapshot, block_height);
             assert_eq!(
                 session.state,
                 MultisigState::Completed {
                     completed_at: block_height
                 }
             );
+            assert!(res);
         }
     }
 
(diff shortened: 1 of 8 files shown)


contracts/multisig/src/signing.rs
```diff
@@ -29,47 +29,64 @@ impl SigningSession {
...
         });
     }
 
+    if !signature.verify(&session.msg, pub_key)? {
+        return Err(ContractError::InvalidSignature {
+            session_id: session.id,
+            signer: signer.into(),
+        });
+    }
+
```
cjcobb23, line 75: I feel like this function should check whether the signature is a duplicate for this session. I know that check is being done elsewhere, but I feel like it might be cleaner to have all of the validity checks here.

contracts/multisig/src/signing.rs
```diff
@@ -29,47 +29,64 @@ impl SigningSession {
...
     }
 
+    /// returns true if session state was changed
     pub fn recalculate_session_state(
         &mut self,
         signatures: &HashMap<String, Signature>,
         snapshot: &Snapshot,
         block_height: u64,
-    ) {
+    ) -> bool {
```
cjcobb23, line 38: I wonder if it would be cleaner to just return an event here. I think we are only using this return value to determine whether or not to add an event into the response later on, so might be cleaner to just return the event itself here.

contracts/multisig/src/state.rs
```diff
@@ -14,10 +14,11 @@ use crate::{
 #[cw_serde]
 pub struct Config {
     pub governance: Addr,
+    pub rewards_contract: Addr,
+    pub grace_period: u64,
```
cjcobb23, line 18: We probably eventually want some mechanism to modify this grace period (through governance), but we can solve that later on. Maybe leave a TODO comment saying we need to add a way to modify the grace period after instantiation?

contracts/multisig/src/signing.rs
eguajardo: this was my first approach, but I don't know how to do it without it looking even weirder, because to do that check here I'll need to first load outside the `Option<Signature>` and pass it here as a parameter just to check if it contains something. For me that looks weirder, thoughts?

contracts/multisig/src/signing.rs
eguajardo: I'm not sure I like the idea of mixing events with logic, I'll prefer to have them separated the same way we are trying to keep storage separated from logic

contracts/multisig/src/signing.rs
cjcobb23: Well you could just load the signature here, and then store the signature later on, instead of doing an update. It's up to you though, I can see the benefit of having the storage access occur in one place. I just generally view the duplicate signature check as a validation check, and it would be nice to have all of the validation checks in one place. It would also be nice to have the check for whether the signer is a participant in here too. Then you have one function that does all of the checks.

contracts/multisig/src/signing.rs
cjcobb23: That's true. I just don't like this `bool` return value here. I feel like another option is to just return the new state, and compare that to the old state where the function is being called. Or don't even return the new state, just mutate the state, and then compare old state to new state where it's being called. Having the bool return value here just feels like how this function is being used is "bleeding" into the implementation here. But up to you, I don't feel super strongly about it.

cjcobb23 (approved):

PR #165, opened 2023-10-24
feat: create build part of multistage dockerfile

fahimahmedx:
## Description
https://axelarnetwork.atlassian.net/jira/software/c/projects/AXE/boards/18?selectedIssue=AXE-2371

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Steps to Test
run `docker build -t axelar-amplifier -f ampd/Dockerfile .`
## Expected Behaviour
build succeeds


Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM rust:latest as build
+WORKDIR /app
+
+RUN apt-get update && apt-get install -y clang protobuf-compiler
+COPY . .
+RUN cargo build
```
fish-sammy, line 6:
```suggestion
RUN cargo build --release
```

Dockerfile
```diff
@@ -0,0 +1,8 @@
...
+
```
fish-sammy, line 7: After build, do another `FROM` image and copy the binary over to run. Pick something lighter, such as alpine for that purpose.

.dockerignore
```diff
@@ -0,0 +1,32 @@
...
+**/.vscode
+**/*.*proj.user
+**/*.dbmdl
+**/*.jfm
+**/charts
+**/docker-compose*
+**/compose*
+**/Dockerfile*
+**/node_modules
+**/npm-debug.log
```
fish-sammy, line 26: These are npm stuff. Shouldn't be here.

.dockerignore
```diff
@@ -0,0 +1,32 @@
+# Include any files or directories that you don't want to be copied to your
```
fish-sammy, line 1: Try cleaning the ignore file to only contain stuff that we are concerned about.

fish-sammy (changes requested):

Dockerfile
fahimahmedx: Will create another PR to do that!

Cargo.toml
```diff
@@ -33,6 +33,5 @@ rpath = false
 lto = true
 debug-assertions = false
 codegen-units = 1
-panic = 'abort'
```
cgorenflo, line 36: what ramification does this have?

ampd/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM rust:latest as build
+WORKDIR /app
+
+RUN apt-get update && apt-get install -y clang protobuf-compiler
+COPY . .
+RUN cargo build --release
+
+CMD ["./target/release/ampd"]
```
cgorenflo, line 8: if this is supposed to be just the build stage then there is no need to set the `CMD`

Cargo.toml
fahimahmedx: The default strategy is `unwind` which is safer than abort because it allows the application to shut down in a clean way, instead of crashing. Also the crates we're using in `ampd` require `unwind`.

fish-sammy (approved):

PR #166, opened 2023-10-24
feat(multisig-prover): optimize to construct proof using the minimum amount of signatures required

eguajardo:
## Description

https://axelarnetwork.atlassian.net/browse/AXE-2384

## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [x] Connect epics/issues

### DIFF
--- a/contracts/multisig-prover/src/query.rs
+++ b/contracts/multisig-prover/src/query.rs
@@ -1,6 +1,12 @@
-use cosmwasm_std::{to_binary, Deps, QueryRequest, StdError, StdResult, Uint64, WasmQuery};
+use cosmwasm_std::{
+    to_binary, Deps, QueryRequest, StdError, StdResult, Uint256, Uint64, WasmQuery,
+};
 
-use multisig::{msg::Multisig, types::MultisigState};
+use multisig::{
+    key::Signature,
+    msg::{Multisig, Signer},
+    types::MultisigState,
+};
 
 use crate::{
     msg::{GetProofResponse, ProofStatus},
@@ -28,7 +34,10 @@ pub fn get_proof(deps: Deps, multisig_session_id: Uint64) -> StdResult<GetProofR
         MultisigState::Pending => ProofStatus::Pending,
         MultisigState::Completed { .. } => {
             let execute_data = batch
-                .encode_execute_data(multisig.quorum, multisig.signers)
+                .encode_execute_data(
+                    multisig.quorum,
+                    optimize_signers(multisig.quorum, multisig.signers),
+                )
                 .map_err(|err| {
                     StdError::generic_err(format!("failed to encode execute data: {}", err))
                 })?;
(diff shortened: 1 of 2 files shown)


contracts/multisig-prover/src/query.rs
```diff
@@ -28,7 +34,10 @@ pub fn get_proof(deps: Deps, multisig_session_id: Uint64) -> StdResult<GetProofR
         MultisigState::Pending => ProofStatus::Pending,
         MultisigState::Completed { .. } => {
             let execute_data = batch
-                .encode_execute_data(multisig.quorum, multisig.signers)
+                .encode_execute_data(
+                    multisig.quorum,
+                    optimize_signers(multisig.quorum, multisig.signers),
```
cgorenflo, line 39:
```suggestion
                    optimize_signers(multisig.signers, multisig.quorum),
```
makes more sense to me this way around

contracts/multisig-prover/src/query.rs
```diff
@@ -48,3 +57,89 @@ pub fn get_proof(deps: Deps, multisig_session_id: Uint64) -> StdResult<GetProofR
 pub fn get_worker_set(deps: Deps) -> StdResult<WorkerSet> {
     CURRENT_WORKER_SET.load(deps.storage)
 }
+
+/// Returns a new vector of signers with signatures containing the minimum amount of signatures required to reach quorum.
+/// The vector is sorted by weight and then is iterated over, removing signatures after the quorum is reached.
```
cgorenflo, line 62:
```suggestion
/// Returns the minimum amount of signatures to satisfy the quorum, sorted by weight
```

contracts/multisig-prover/src/query.rs
```diff
@@ -48,3 +57,89 @@ pub fn get_proof(deps: Deps, multisig_session_id: Uint64) -> StdResult<GetProofR
...
+            |acc, (signer, signature)| match signature {
+                Some(sig) if *acc < quorum => {
+                    *acc += signer.weight;
+                    Some((signer, Some(sig)))
+                }
+                _ => Some((signer, None)),
+            },
+        )
+        .collect()
+}
```
cgorenflo, line 82: Do we still need the signers without signatures? I suggest something along these lines. If signers are still needed at least use `sorted_by` so you can chain everything together

```suggestion
) -> Vec<(Signer, Signature)> {
        signers
            .into_iter()
            .filter_map(|(signer, sig)| {
                if let Some(sig) = sig {
                    Some((signer, sig))
                } else {
                    None
                }
            })
            .sorted_by(|(a, _), (b, _)| b.weight.cmp(&a.weight))
            .scan(Uint256::zero(), |acc, (signer, signature)| {
                if *acc < quorum {
                    *acc += signer.weight;
                    Some((signer, signature))
                } else {
                    None
                }
            })
            .collect()
    }
```

contracts/service-registry/src/state.rs
```diff
@@ -44,7 +44,11 @@ impl TryFrom<Worker> for Participant {
         match worker.bonding_state {
             BondingState::Bonded { amount: _ } => Ok(Self {
                 address: worker.address,
-                weight: Uint256::one() // Weight is set to one to ensure all workers have same weight. In future it should be derived from amount bonded
+                // Weight is set to one to ensure all workers have same weight. In future it should be derived from amount bonded
+                // Changing the weight to a non-constant value may impact optimization when constructing the proof because multisig does not
+                // wait for late signatures before emitting the session completed event. Is possible for a more optimal proof to be constructed if
+                // late signatures with more weigth would be considered.
```
cgorenflo, line 50: I don't understand these comments

contracts/multisig-prover/src/query.rs
eguajardo: thanks! I was looking for something like `.sorted_by` but couldn't find it. Signers without signatures are required though, the proof contains all possible participants.

contracts/service-registry/src/state.rs
eguajardo: does this reads better?

"If the weight is changed to a non-constant value, the signing session completed event from multisig and the signature optimization during proof construction may require re-evaluation, so that relayers could take advantage of late signatures to get a more optimized version of the proof."

fish-sammy (approved):

PR #168, opened 2023-10-26
feat(rewards): distribute rewards for specified epoch

cjcobb23:
https://axelarnetwork.atlassian.net/browse/AXE-2419

### DIFF
--- a/contracts/rewards/src/contract/execute.rs
+++ b/contracts/rewards/src/contract/execute.rs
@@ -9,7 +9,6 @@ use crate::{
     state::{Epoch, EpochTally, Event, RewardsPool, Store, StoredParams},
 };
 
-
 pub struct Contract<S>
 where
     S: Store,
--- a/contracts/rewards/src/contract/execute.rs
+++ b/contracts/rewards/src/contract/execute.rs
@@ -1,14 +1,14 @@
+use axelar_wasm_std::nonempty;
+use cosmwasm_std::{Addr, Fraction, Uint256};
 use error_stack::Result;
 use std::collections::HashMap;
-
-use axelar_wasm_std::nonempty;
-use cosmwasm_std::{Addr, Uint256};
 
 use crate::{
     error::ContractError,
     msg::RewardsParams,
     state::{Epoch, EpochTally, Event, RewardsPool, Store, StoredParams},
 };
+
 
 pub struct Contract<S>
 where
@@ -25,7 +25,7 @@ where
     /// Returns the current epoch. The current epoch is computed dynamically based on the current
     /// block height and the epoch duration. If the epoch duration is updated, we store the epoch
     /// in which the update occurs as the last checkpoint
-    fn get_current_epoch(&self, cur_block_height: u64) -> Result<Epoch, ContractError> {
+    fn current_epoch(&self, cur_block_height: u64) -> Result<Epoch, ContractError> {
         let stored_params = self.store.load_params();
         let epoch_duration: u64 = stored_params.params.epoch_duration.into();
         let epoch = stored_params.last_updated;
@@ -42,7 +42,6 @@ where
         Ok(Epoch {
             epoch_num,
             block_height_started,
-            rewards: epoch.rewards,
         })
     }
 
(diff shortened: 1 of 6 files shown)


contracts/rewards/src/contract/execute.rs
```diff
@@ -94,12 +96,50 @@ where
...
+        }
+    }
+
+    fn process_epoch_tally(
+        &mut self,
+        tally: EpochTally,
+    ) -> Result<HashMap<Addr, Uint256>, ContractError> {
+        match get_workers_to_reward(&tally) {
+            Ok(workers_to_reward) => {
+                let pool_balance = self.get_pool_balance(tally.contract.clone())?;
```
fish-sammy, line 116:
```suggestion
                let pool = self.store.load_rewards_pool(contract_addr.clone())?.unwrap_or( ... )
                let rewards_per_epoch = tally.rewards_params.rewards_per_epoch;
                let rewards_per_worker =
                    get_rewards_per_worker(&workers_to_reward, rewards_per_epoch, pool.balance)?;

                if rewards_per_worker.is_zero() {
                  return Ok(HashMap::new());
                }

                pool.distribute_rewards(workers_to_reward.len(), rewards_per_worker)
                self.store.save_rewards_pool(pool)?;

                Ok(Vec::<Addr>::from(workers_to_reward)
                    .into_iter()
                    .map(|worker| (worker, rewards_per_worker))
                    .collect())

```

fish-sammy (changes requested):

contracts/rewards/src/contract/execute.rs
```diff
@@ -149,6 +185,67 @@ where
...
+            contract,
+            balance: old_balance
+                - (rewards_per_worker * Uint256::from(workers_to_reward.len() as u128)),
+        })
+    }
+}
+
+fn get_workers_to_reward(
+    tally: &EpochTally,
+) -> Result<nonempty::Vec<Addr>, axelar_wasm_std::nonempty::Error> {
```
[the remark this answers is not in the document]
cjcobb23: you're right that it's an ok scenario for no workers to be rewarded. I just found the logic a bit simpler when this function returns a `nonempty::Vec`, since then we don't have to check for zero length when processing the rewards.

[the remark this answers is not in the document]
fish-sammy: We shouldn't have to check that either way.

packages/axelar-wasm-std/src/threshold.rs
```diff
@@ -40,6 +40,14 @@ impl Fraction<Uint64> for Threshold {
     }
 }
 
+impl PartialOrd for Threshold {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        let self_normalized = self.numerator().full_mul(other.denominator());
+        let other_normalized = other.numerator().full_mul(self.denominator());
+        self_normalized.partial_cmp(&other_normalized)
```
fish-sammy, line 47:
```suggestion
        
        self_normalized.partial_cmp(&other_normalized)
```

fish-sammy: Have some empty line. Rust doesn't use `return` for most of the cases, and an empty line before makes it look better.

contracts/rewards/src/state.rs
```diff
@@ -21,15 +21,17 @@ pub struct EpochTally {
     pub event_count: u64,
     pub participation: HashMap<Addr, u64>,
     pub epoch: Epoch,
+    pub params: Params,
 }
 
 impl EpochTally {
-    pub fn new(contract: Addr, epoch: Epoch) -> Self {
+    pub fn new(contract: Addr, epoch: Epoch, rewards_params: Params) -> Self {
```
fish-sammy, line 28:
```suggestion
    pub fn new(contract: Addr, epoch: Epoch, params: Params) -> Self {
```

contracts/rewards/src/state.rs
```diff
@@ -45,6 +47,18 @@ impl EpochTally {
         self.event_count += 1;
         self
     }
+
+    pub fn workers_to_reward(&self) -> Vec<Addr> {
```
fish-sammy, line 51: so much cleaner!

fish-sammy (approved):

fish-sammy (approved):

PR #171, opened 2023-10-27
feat(rewards): implement instantiate and require governance

cjcobb23:
* store governance and rewards denom in config
* require sender to be governance in update_params

https://axelarnetwork.atlassian.net/browse/AXE-2424

### DIFF
--- a/contracts/rewards/src/error.rs
+++ b/contracts/rewards/src/error.rs
@@ -36,6 +36,6 @@ pub enum ContractError {
     #[error("specified block has already passed")]
     BlockHeightInPast,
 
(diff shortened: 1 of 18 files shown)


contracts/rewards/src/error.rs
```diff
@@ -35,4 +35,7 @@ pub enum ContractError {
 
     #[error("specified block has already passed")]
     BlockHeightInPast,
+
+    #[error("called is not authorized")]
```
fish-sammy, line 39:
```suggestion
    #[error("caller is not authorized")]
```

fish-sammy (approved):

PR #173, opened 2023-10-27
feat(voting-verifier): allow end poll to complete before expiration if a result has reached quorum

eguajardo:
## Description
Prerequisite to integrate voting with rewards. Before this change a poll is completed only when the poll is expired or when everyone voted. In order to integrate with rewards, it should be allowed to be completed when the quorum is satisfied to allow for a grace period between expiration and poll completion

https://axelarnetwork.atlassian.net/browse/AXE-2201

## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [x] Connect epics/issues

### DIFF
--- a/packages/axelar-wasm-std/src/voting.rs
+++ b/packages/axelar-wasm-std/src/voting.rs
@@ -207,9 +207,17 @@ impl Poll for WeightedPoll {
             .iter()
             .all(|(_, participation)| participation.voted);
 
+        let results: Vec<bool> = self
+            .votes
+            .iter()
+            .map(|tally| *tally >= self.quorum.into())
+            .collect();
+
+        let quorum_satisfied = results.iter().all(|quorum| *quorum);
+
         if block_height < self.expires_at
-            // can tally early if all participants voted
-            && !everyone_voted
+            // can tally early if all participants voted or quorum was satisfied for all votes in poll
+            && !everyone_voted && !quorum_satisfied
         {
             return Err(Error::PollNotEnded);
         }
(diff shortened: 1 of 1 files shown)


cjcobb23 (approved):

packages/axelar-wasm-std/src/voting.rs
```diff
@@ -207,9 +207,17 @@ impl Poll for WeightedPoll {
             .iter()
             .all(|(_, participation)| participation.voted);
 
+        let results: Vec<bool> = self
+            .votes
```
cjcobb23, line 211: I feel like this field should be called `tallies`

packages/axelar-wasm-std/src/voting.rs
```diff
@@ -222,11 +230,7 @@ impl Poll for WeightedPoll {
 
         Ok(PollResult {
             poll_id: self.poll_id,
-            results: self
-                .votes
-                .iter()
-                .map(|tally| *tally >= self.quorum.into())
-                .collect(),
+            results,
```
haiyizxx, line 233: in line 229 `self.status = PollStatus::Finished;` means no more votes on the partial tally

haiyizxx (changes requested):

packages/axelar-wasm-std/src/voting.rs
```diff
@@ -207,9 +207,17 @@ impl Poll for WeightedPoll {
             .iter()
             .all(|(_, participation)| participation.voted);
 
+        let results: Vec<bool> = self
+            .votes
+            .iter()
+            .map(|tally| *tally >= self.quorum.into())
+            .collect();
```
fish-sammy, line 214:
```suggestion
        let quorum: Uint256 = self.quorum.into();
        let results: Vec<bool> = self.votes.iter().map(|tally| *tally >= quorum).collect();
```

fish-sammy (approved):

haiyizxx (approved):

PR #176, opened 2023-10-31
feat(voting-verifier): integrate voting with rewards contract

eguajardo:
## Description
Calls reward contract to record participation when casting a vote. Rewards gets distributed after poll completion to all participants that voted for the consensus result

https://axelarnetwork.atlassian.net/browse/AXE-2411

## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [x] Connect epics/issues

### DIFF
--- a/contracts/voting-verifier/src/execute.rs
+++ b/contracts/voting-verifier/src/execute.rs
@@ -1,6 +1,6 @@
 use axelar_wasm_std::operators::Operators;
 use cosmwasm_std::{
-    to_binary, Deps, DepsMut, Env, MessageInfo, QueryRequest, Response, Storage, WasmQuery,
+    to_binary, Deps, DepsMut, Env, MessageInfo, QueryRequest, Response, Storage, WasmMsg, WasmQuery,
 };
 
 use axelar_wasm_std::voting::{PollID, PollResult};
@@ -180,7 +180,16 @@ pub fn vote(
 
     POLLS.save(deps.storage, poll_id, &poll)?;
 
-    Ok(Response::new().add_event(
+    let rewards_msg = WasmMsg::Execute {
+        contract_addr: config.rewards_contract.into_string(),
+        msg: to_binary(&rewards::msg::ExecuteMsg::RecordParticipation {
+            event_id: poll_id.into(),
+            worker_address: info.sender.to_string(),
+        })?,
+        funds: vec![],
+    };
+
+    Ok(Response::new().add_message(rewards_msg).add_event(
         Voted {
             poll_id,
             voter: info.sender,
--- a/contracts/voting-verifier/src/execute.rs
+++ b/contracts/voting-verifier/src/execute.rs
@@ -274,7 +274,8 @@ pub fn end_poll(deps: DepsMut, env: Env, poll_id: PollID) -> Result<Response, Co
                 msg: to_binary(&rewards::msg::ExecuteMsg::RecordParticipation {
                     event_id: poll_id.into(),
                     worker_address: address.to_string(),
-                })?,
+                })
+                .expect("failed to serialize message for rewards contract"),
                 funds: vec![],
             })
         })
--- a/contracts/voting-verifier/src/execute.rs
+++ b/contracts/voting-verifier/src/execute.rs
@@ -259,7 +259,7 @@ pub fn end_poll(deps: DepsMut, env: Env, poll_id: PollID) -> Result<Response, Co
 
     let poll_result = match &mut poll {
         state::Poll::Messages(poll) | state::Poll::ConfirmWorkerSet(poll) => {
-            poll.tally(env.block.height)?
+            poll.complete(env.block.height)?
         }
     };
     POLLS.save(deps.storage, poll_id, &poll)?;
--- a/contracts/voting-verifier/src/execute.rs
+++ b/contracts/voting-verifier/src/execute.rs
@@ -167,12 +167,14 @@ pub fn vote(
     poll_id: PollID,
     votes: Vec<bool>,
 ) -> Result<Response, ContractError> {
+    let config = CONFIG.load(deps.storage)?;
+
     let mut poll = POLLS
         .may_load(deps.storage, poll_id)?
         .ok_or(ContractError::PollNotFound)?;
     match &mut poll {
         state::Poll::Messages(poll) | state::Poll::ConfirmWorkerSet(poll) => {
-            poll.cast_vote(env.block.height, &info.sender, votes)?
+            poll.cast_vote(env.block.height, config.grace_period, &info.sender, votes)?
         }
     };
 
--- a/contracts/voting-verifier/src/execute.rs
+++ b/contracts/voting-verifier/src/execute.rs
@@ -167,29 +167,20 @@ pub fn vote(
     poll_id: PollID,
     votes: Vec<bool>,
 ) -> Result<Response, ContractError> {
-    let config = CONFIG.load(deps.storage)?;
-
     let mut poll = POLLS
         .may_load(deps.storage, poll_id)?
         .ok_or(ContractError::PollNotFound)?;
     match &mut poll {
         state::Poll::Messages(poll) | state::Poll::ConfirmWorkerSet(poll) => {
-            poll.cast_vote(env.block.height, config.grace_period, &info.sender, votes)?
+            poll.cast_vote(env.block.height, &info.sender, votes)?
         }
     };
 
     POLLS.save(deps.storage, poll_id, &poll)?;
 
-    let rewards_msg = WasmMsg::Execute {
-        contract_addr: config.rewards_contract.into_string(),
-        msg: to_binary(&rewards::msg::ExecuteMsg::RecordParticipation {
-            event_id: poll_id.into(),
-            worker_address: info.sender.to_string(),
-        })?,
-        funds: vec![],
-    };
+    // TODO: add rewards for late voters
 
-    Ok(Response::new().add_message(rewards_msg).add_event(
+    Ok(Response::new().add_event(
         Voted {
             poll_id,
             voter: info.sender,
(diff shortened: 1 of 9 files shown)


contracts/voting-verifier/src/state.rs
```diff
@@ -19,6 +19,8 @@ pub struct Config {
     pub block_expiry: u64, // number of blocks after which a poll expires
     pub confirmation_height: u64,
     pub source_chain: ChainName,
+    pub rewards_contract: Addr,
+    pub grace_period: u64, // TODO: add update mechanism to change this after instantiation
```
cjcobb23, line 23: Do we want to just use the block_expiry as the grace period? All votes that are cast before expiration count towards rewards, and all votes after do not?

contracts/voting-verifier/src/execute.rs
```diff
@@ -167,18 +167,29 @@ pub fn vote(
...
+    let rewards_msg = WasmMsg::Execute {
+        contract_addr: config.rewards_contract.into_string(),
+        msg: to_binary(&rewards::msg::ExecuteMsg::RecordParticipation {
+            event_id: poll_id.into(),
+            worker_address: info.sender.to_string(),
+        })?,
+        funds: vec![],
+    };
+
+    Ok(Response::new().add_message(rewards_msg).add_event(
```
cjcobb23, line 192: We shouldn't do this blindly here. Workers are only supposed to be rewarded if they vote according to consensus. So if consensus is true, we only reward workers who voted true.

cjcobb23 (changes requested):

contracts/voting-verifier/src/execute.rs
cgorenflo: yeah I think this needs to be moved to the poll completion call, so we know who to reward

contracts/voting-verifier/tests/mock.rs
```diff
@@ -81,3 +81,44 @@ pub fn make_mock_service_registry(app: &mut App) -> Addr {
...
+            code_id,
+            Addr::unchecked("sender"),
+            &MockRewardsInstantiateMsg,
+            &[],
+            "Contract",
+            None,
+        )
+        .unwrap();
+    contract_address
+}
```
cgorenflo, line 124: what do we need this for?

cgorenflo (dismissed):

cgorenflo:
on top of the review, as discussed we should remove the grace period parameter again

fish-sammy:
Can you update the description to reflect the fact that grace period is no longer implemented?

packages/axelar-wasm-std/src/voting.rs
```diff
@@ -251,25 +268,24 @@ impl Poll for WeightedPoll {
...
-            .iter_mut()
-            .zip(votes.into_iter())
-            .filter(|(_, vote)| *vote)
-            .for_each(|(tally, _)| {
-                *tally += Uint256::from(participation.weight);
-            });
+        participation.vote = Some(votes);
 
-        Ok(PollStatus::InProgress)
+        Ok(())
```
fish-sammy, line 288:
```suggestion
        // late votes are not tallied
        if self.status != PollStatus::InProgress {
          return Ok(())
        }

        self.tallies
            .iter_mut()
            .zip(votes.iter())
            .filter(|(_, vote)| **vote)
            .for_each(|(tally, _)| {
                *tally += Uint256::from(participation.weight);
            });

        participation.vote = Some(votes);

        Ok(())
```
Exit early and leave the complicated code at the least indented place.

fish-sammy (changes requested):

contracts/voting-verifier/tests/mock.rs
eguajardo: the way tests are done right now will fail without this. Integration tests need to call an existing contract when the execution adds a message in the response addressed to the rewards contract

packages/axelar-wasm-std/src/voting.rs
```diff
@@ -220,15 +222,30 @@ impl Poll for WeightedPoll {
             return Err(Error::PollNotEnded);
         }
 
-        if self.status == PollStatus::Finished {
+        if matches!(self.status, PollStatus::Finished { .. }) {
```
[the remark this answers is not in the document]
eguajardo: moved this one, but can't move `block_height < self.expires_at` check above because it depends on everything else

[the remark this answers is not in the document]
fish-sammy: oh ur right. Didn't realize it.

contracts/voting-verifier/src/execute.rs
```diff
@@ -260,7 +264,25 @@ pub fn end_poll(deps: DepsMut, env: Env, poll_id: PollID) -> Result<Response, Co
         state::Poll::ConfirmWorkerSet(_) => end_poll_worker_set(deps, poll_id, &poll_result)?,
     };
 
+    // TODO: change rewards contract interface to accept a list of addresses to avoid creating multiple wasm messages
+    let rewards_msgs = poll_result
+        .consensus_participants
+        .iter()
+        .map(|address| -> Result<WasmMsg, ContractError> {
```
fish-sammy, line 271:
```suggestion
        .map(|address| -> WasmMsg {
```
doesn't work?

contracts/voting-verifier/src/execute.rs
```diff
@@ -260,7 +264,25 @@ pub fn end_poll(deps: DepsMut, env: Env, poll_id: PollID) -> Result<Response, Co
...
+                contract_addr: config.rewards_contract.to_string(),
+                msg: to_binary(&rewards::msg::ExecuteMsg::RecordParticipation {
+                    event_id: poll_id.into(),
+                    worker_address: address.to_string(),
+                })
+                .expect("failed to serialize message for rewards contract"),
+                funds: vec![],
+            })
+        })
+        .collect::<Result<Vec<WasmMsg>, ContractError>>()?;
```
fish-sammy, line 282:
```suggestion
        .collect::<Vec<_>>();
```

contracts/voting-verifier/src/execute.rs
```diff
@@ -260,7 +264,23 @@ pub fn end_poll(deps: DepsMut, env: Env, poll_id: PollID) -> Result<Response, Co
...
+        .map(|address| WasmMsg::Execute {
+            contract_addr: config.rewards_contract.to_string(),
+            msg: to_binary(&rewards::msg::ExecuteMsg::RecordParticipation {
+                event_id: poll_id.into(),
+                worker_address: address.to_string(),
+            })
+            .expect("failed to serialize message for rewards contract"),
+            funds: vec![],
+        })
+        .collect::<Vec<_>>();
```
fish-sammy, line 280:
```suggestion
        });
```
Don't even need to collect into a `Vec` cuz `add_messsages` takes `msgs: impl IntoIterator<Item = M>`

fish-sammy (approved):

cjcobb23 (approved):

packages/axelar-wasm-std/src/voting.rs
```diff
@@ -156,7 +158,7 @@ pub enum PollStatus {
 #[cw_serde]
 pub struct Participation {
     pub weight: nonempty::Uint256,
-    pub voted: bool,
+    pub vote: Option<Vec<bool>>,
```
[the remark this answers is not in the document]
eguajardo: yes I was feeling the same way, but decided to leave it for whenever it's refactored to use the enum to keep this PR simpler

packages/axelar-wasm-std/src/voting.rs
```diff
@@ -220,15 +226,26 @@ impl Poll for WeightedPoll {
...
+            .filter_map(|(address, participation)| {
+                participation.vote.as_ref().and_then(|vote| {
+                    if *vote == results {
+                        Some(address.to_owned())
+                    } else {
+                        None
+                    }
+                })
+            })
+            .collect();
```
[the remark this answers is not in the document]
eguajardo: I don't know, I like the idea that `PollResult` returns everything needed by the `execute` module to know how to react, I don't see much of an issue since the computation its coupled with the votes result. Computing it a the callsite will need access to some of the fields in `WeightedPoll` instead of accessing the trait `Poll`, might want to add a new function in the trait and compute there instead, but for simplicity I would like to keep it as it is unless someone has strong opinion about an alternative

packages/axelar-wasm-std/src/voting.rs
```diff
@@ -130,21 +130,23 @@ impl fmt::Display for PollID {
...
+    fn complete(&mut self, block_height: u64) -> Result<PollResult, Error>;
     // errors if sender is not a participant, if sender already voted, if the poll is finished or
     // if the number of votes doesn't match the poll size
     fn cast_vote(
         &mut self,
         block_height: u64,
         sender: &Addr,
         votes: Vec<bool>,
-    ) -> Result<PollStatus, Error>;
+    ) -> Result<(), Error>;
```
cgorenflo, line 141: I would prefer a signature like this:
`fn cast_vote(self, block_height: u64, sender: &Addr, votes: Vec<bool>) -> Result<Self, Error>`
Because the function owns self, it can still do mutation internally, but this way we get a cleaner immutable interface and don't have these awkward void returns

packages/axelar-wasm-std/src/voting.rs
```diff
@@ -193,22 +195,26 @@ impl WeightedPoll {
...
+            tallies: vec![Uint256::zero(); poll_size],
             status: PollStatus::InProgress,
             participation,
         }
     }
 }
 
 impl Poll for WeightedPoll {
-    fn tally(&mut self, block_height: u64) -> Result<PollResult, Error> {
+    fn complete(&mut self, block_height: u64) -> Result<PollResult, Error> {
```
cgorenflo, line 206: similar to my comment on cast_vote, I'd prefer to minimize side effects. This function sets the poll to finished (why the discrepancy in naming for complete <-> finished?) but mostly it computes the poll result. And then presumably the poll gets stored somewhere outside this context. I suggest splitting this of:
```
let completed_poll = poll.complete(...)?;
store.save(completed_poll);
return cumulate_poll_result(completed_poll);
```

PR #180, opened 2023-11-02
refactor: move WorkerSet struct from multisig-prover into multisig

fahimahmedx:
## Description
[AXE-2449]

Refactors `WorkerSet` to be in `multisig` to prevent cyclical dependancy when using `WorkerSet` in both `multisig` and `multisig-prover`.

[AXE-2449]: https://axelarnetwork.atlassian.net/browse/AXE-2449?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/multisig/src/workerset.rs
```diff
@@ -0,0 +1,50 @@
...
+    // and we need to be able to distinguish between the two
+    pub created_at: u64,
+}
+
+impl WorkerSet {
+    pub fn new(
+        participants: Vec<(Participant, PublicKey)>,
+        threshold: Uint256,
+        block_height: u64,
+    ) -> Result<Self, StdError> {
```
cjcobb23, line 24: Why do we need to return Result here? What's the possible error?

cgorenflo: Can't figure that out either

fahimahmedx: The `Result` was already there, but the line `weight: participant.weight.into()` may cause an issue because it's turning `nonempty::Uint256` into a `cosmwasm_std::Uint256`. 

In `nonempty::Uint256`, I dont see any implementations of `Into` except for `#[serde(into = "cosmwasm_std::Uint64")]` which doesn't seem equivalent. I think this means the `.into()` would cause an error for numbers bigger than `Uint64`.

fish-sammy: `.into()` can never cause an error but `.try_into()`. Get rid of the `Result`.

contracts/multisig/src/workerset.rs
```diff
@@ -0,0 +1,50 @@
...
+use cosmwasm_std::{HexBinary, StdError, Uint256};
+use sha3::{Digest, Keccak256};
+
+#[cw_serde]
+pub struct WorkerSet {
+    pub signers: BTreeSet<Signer>,
+    pub threshold: Uint256,
+    // for hash uniqueness. The same exact worker set could be in use at two different times,
+    // and we need to be able to distinguish between the two
+    pub created_at: u64,
```
fish-sammy, line 16: How can this be used to differentiate worker sets that are created in the same block?

fish-sammy (changes requested):

contracts/multisig/src/workerset.rs
fahimahmedx: I didn't think of that, thank you! Just made a commit that addresses it and have a few questions below.

1. I added a `Timestamp` field for the time it was created at, called `created_at_time`. Does this make `created_at_block` redundant? I feel like having block information is more "native" and is still worth keeping.

2. Adding the `created_at_time` field broke the test case `encoding::abi::test::test_new_command_batch_with_operator_transfer` for me. While I was looking into how to fix it, I noticed the data it's checking if it's equal to is a predefined value stored in `multisig-prover/src/test/test_data.rs` that is returned in the function `encoded_data_with_operator_transfer()` (hex value is too long to copy and paste into this comment).
How did they come up with that hex output? I assume I'll need to create a new hex value to update the test case with to accomodate the new `created_at_time` field, and want to learn how to create such a value myself without just replacing the current value with the new outputted value.

fahimahmedx:
The test case `test_new_command_batch_with_operator_transfer` fails, am working on fixing it asap.

contracts/multisig/src/workerset.rs
fahimahmedx: @cjcobb23 I think my questions would be more relevant to you. Let me know what you think!

fahimahmedx: to be resolved in future PR

contracts/multisig/src/workerset.rs
```diff
@@ -0,0 +1,52 @@
...
+use cosmwasm_schema::cw_serde;
+use cosmwasm_std::{HexBinary, Uint256};
+use sha3::{Digest, Keccak256};
+
+#[cw_serde]
+pub struct WorkerSet {
+    pub signers: BTreeSet<Signer>,
+    pub threshold: Uint256,
+    // for hash uniqueness. The same exact worker set could be in use at two different times,
+    // and we need to be able to distinguish between the two
```
cjcobb23, line 15: Can you expand on this comment and write a description of why this can cause a problem? Something like "without a nonce, updating to a worker set that is the exact same as a worker set in the past will be immediately confirmed. TODO add nonce to the voting verifier and to the evm gateway"

contracts/multisig/src/workerset.rs
cgorenflo, line 1: rename to worker_set

cgorenflo (approved):

PR #181, opened 2023-11-06
feat(rewards): implement execute

cjcobb23:
Implements execute, handling each case of ExecuteMsg, and fixes some bugs in the rewards contract

https://axelarnetwork.atlassian.net/browse/AXE-2523

### DIFF
--- a/contracts/rewards/src/contract.rs
+++ b/contracts/rewards/src/contract.rs
@@ -10,6 +10,8 @@ use cosmwasm_std::entry_point;
 use cosmwasm_std::{BankMsg, Coin, DepsMut, Env, MessageInfo, Response};
 use error_stack::ResultExt;
 
+use itertools::Itertools;
+
 mod execute;
 
 #[cfg_attr(not(feature = "library"), entry_point)]
@@ -57,13 +59,7 @@ pub fn execute(
         } => {
             let worker_address = deps.api.addr_validate(&worker_address)?;
             Contract::new(deps)
-                .record_participation(
-                    nonempty::String::try_from(event_id)
-                        .change_context(ContractError::InvalidEventId)?,
-                    worker_address,
-                    info.sender,
-                    env.block.height,
-                )
+                .record_participation(event_id, worker_address, info.sender, env.block.height)
                 .map_err(axelar_wasm_std::ContractError::from)?;
 
             Ok(Response::new())
@@ -75,7 +71,7 @@ pub fn execute(
                 .funds
                 .iter()
                 .find(|coin| coin.denom == contract.config.rewards_denom)
-                .filter(|_| info.funds.len() == 1)
+                .filter(|_| info.funds.len() == 1) // filter here to make sure expected denom is the only one attached to this message, and other funds aren't silently swallowed
                 .ok_or(ContractError::WrongDenom)?
                 .amount;
 
@@ -96,16 +92,17 @@ pub fn execute(
                 .distribute_rewards(contract_address, env.block.height, epoch_count)
                 .map_err(axelar_wasm_std::ContractError::from)?;
 
-            let mut msgs = vec![];
-            for (addr, amount) in rewards {
-                msgs.push(BankMsg::Send {
+            let msgs: Vec<BankMsg> = rewards
+                .into_iter()
+                .sorted()
+                .map(|(addr, amount)| BankMsg::Send {
                     to_address: addr.into(),
                     amount: vec![Coin {
                         denom: contract.config.rewards_denom.clone(),
                         amount,
                     }],
                 })
-            }
+                .collect();
 
             Ok(Response::new().add_messages(msgs))
         }
--- a/contracts/rewards/src/contract.rs
+++ b/contracts/rewards/src/contract.rs
@@ -1,18 +1,21 @@
 use crate::{
     contract::execute::Contract,
+    error::ContractError,
     msg::{ExecuteMsg, InstantiateMsg},
-    state::{Config, CONFIG},
+    state::{Config, Epoch, StoredParams, CONFIG, PARAMS},
 };
+use axelar_wasm_std::{self, nonempty};
 #[cfg(not(feature = "library"))]
 use cosmwasm_std::entry_point;
-use cosmwasm_std::{DepsMut, Env, MessageInfo, Response};
+use cosmwasm_std::{BankMsg, Coin, DepsMut, Env, MessageInfo, Response};
+use error_stack::ResultExt;
 
 mod execute;
 
 #[cfg_attr(not(feature = "library"), entry_point)]
 pub fn instantiate(
     deps: DepsMut,
-    _env: Env,
+    env: Env,
     _info: MessageInfo,
     msg: InstantiateMsg,
 ) -> Result<Response, axelar_wasm_std::ContractError> {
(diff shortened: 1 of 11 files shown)


contracts/rewards/src/contract/execute.rs
```diff
@@ -116,7 +116,7 @@ where
         contract: Addr,
         block_height: u64,
         count: Option<u64>,
-    ) -> Result<HashMap<Addr, Uint256>, ContractError> {
+    ) -> Result<HashMap<Addr, Uint128>, ContractError> {
```
cgorenflo, line 119: this function (and process_rewards_for_epochs) seems overly complicated. suggestions (maybe follow-up PR)
- block_height -> cur_block_height
- count -> epoch_process_limit
- instead of passing in HashMaps and continuously folding I suggest iterating over the epoch range, and returning all HashMaps or vec((Addr, Amount)) and then merging all in one go, i.e. keep the code of how to process a single epoch separate from how to bring it together for the range call

contracts/rewards/src/state.rs
```diff
@@ -42,9 +42,10 @@ impl EpochTally {
         }
     }
 
+    /// IMPORTANT: worker must be a valid address
```
cgorenflo, line 45: what does that mean?

contracts/rewards/src/contract/execute.rs
cjcobb23: I'd rather address this in a future PR, since it's not really related to this change set

contracts/rewards/src/state.rs
cjcobb23: This just means the address should be validated before being passed in (usually via `deps.api.addr_validate(...)?`). It shouldn't be created via `Addr::unchecked()` or accepted blindly from user input.

contracts/rewards/src/contract/execute.rs
cjcobb23: Well I went ahead and made the suggested renaming, since those are simple changes

contracts/rewards/src/contract.rs
```diff
@@ -26,21 +29,90 @@ pub fn instantiate(
...
+            Ok(Response::new())
+        }
+        ExecuteMsg::AddRewards { contract_address } => {
+            let contract_address = deps.api.addr_validate(&contract_address)?;
+            let mut contract = Contract::new(deps);
+            let amount = info
+                .funds
+                .iter()
+                .find(|coin| coin.denom == contract.config.rewards_denom)
+                .filter(|_| info.funds.len() == 1)
```
cgorenflo, line 78: good idea. Since it's not obvious why this check exists and we (or I) will definitely forget what it's for, could you add an explanatory comment? something like "make sure the expected denomination is the only one attached to this message so other amounts don't get silently swallowed". No strong opinion on this, but for me it makes a bit more sense to first check that you only have one denomination and then check that it's the correct one.

cgorenflo (approved):

contracts/rewards/src/contract.rs
```diff
@@ -26,21 +29,90 @@ pub fn instantiate(
...
-        ExecuteMsg::DistributeRewards { .. } => todo!(),
-        ExecuteMsg::UpdateParams { .. } => todo!(),
+        ExecuteMsg::RecordParticipation {
+            event_id,
+            worker_address,
+        } => {
+            let worker_address = deps.api.addr_validate(&worker_address)?;
+            Contract::new(deps)
+                .record_participation(
+                    nonempty::String::try_from(event_id)
```
fish-sammy, line 61: Make it `nonempty::String` in the message struct `ExecuteMsg::RecordParticipation` with custom deserializer.

contracts/rewards/src/contract.rs
fish-sammy: Agree. This `.filter` in the end is confusing.

contracts/rewards/src/state.rs
```diff
@@ -26,7 +26,7 @@ pub struct StoredParams {
 pub struct EpochTally {
     pub contract: Addr,
     pub event_count: u64,
-    pub participation: HashMap<Addr, u64>,
+    pub participation: HashMap<String, u64>, // maps a worker address to participation count. Can't use Addr as key else deserialization will fail
```
fish-sammy, line 29: Why change this?

contracts/rewards/src/state.rs
```diff
@@ -42,9 +42,10 @@ impl EpochTally {
         }
     }
 
+    /// IMPORTANT: worker must be a valid address
     pub fn record_participation(mut self, worker: Addr) -> Self {
         self.participation
-            .entry(worker)
+            .entry(worker.to_string())
```
fish-sammy, line 48: If it's a must, I'd rather adding a check to this method that panics if the address isn't valid.

contracts/rewards/src/contract.rs
```diff
@@ -26,21 +29,90 @@ pub fn instantiate(
...
+            let mut msgs = vec![];
+            for (addr, amount) in rewards {
+                msgs.push(BankMsg::Send {
+                    to_address: addr.into(),
+                    amount: vec![Coin {
+                        denom: contract.config.rewards_denom.clone(),
+                        amount,
+                    }],
+                })
+            }
```
fish-sammy, line 108:
```suggestion
            let msgs = rewards.into_iter().map(|(addr, amount)| BankMsg::Send {
                to_address: addr.into(),
                amount: vec![Coin {
                    denom: contract.config.rewards_denom.clone(),
                    amount,
                }],
            });
```
Avoid mutation whenever possible. Now, I can be wrong but I don't think iterating a hash map is necessarily deterministic in terms of ordering. You may have to switch to a BTreeMap instead.

fish-sammy (changes requested):

contracts/rewards/src/contract.rs
fish-sammy: Actually, the deserializer is already there https://github.com/axelarnetwork/axelar-amplifier/blob/271d452c6ae92a9d021112fad986a942b55852cb/packages/axelar-wasm-std/src/nonempty/string.rs#L7. Use it.

contracts/rewards/src/state.rs
cjcobb23: Using Addr as a key in a HashMap will cause deserialization to fail. I don't really understand why, but it does. Without this, the tests crash.

contracts/rewards/src/state.rs
cjcobb23: It's a little clunky to validate that here, we need access to `Api` somehow: https://docs.rs/cosmwasm-std/latest/cosmwasm_std/trait.Api.html

contracts/rewards/src/contract.rs
```diff
@@ -26,21 +31,85 @@ pub fn instantiate(
...
+                .into_iter()
+                .sorted()
+                .map(|(addr, amount)| BankMsg::Send {
+                    to_address: addr.into(),
+                    amount: vec![Coin {
+                        denom: contract.config.rewards_denom.clone(),
+                        amount,
+                    }],
+                })
+                .collect();
```
fish-sammy, line 105:
```suggestion
                });
```

fish-sammy: u do not need to collect it here.

fish-sammy (approved):

PR #183, opened 2023-11-07
feat(multisig-prover): implement MultiversX encoding for multisig prover

raress96:
Add support for encoding used by MultiversX to multisig prover.

### DIFF
--- a/contracts/multisig-prover/src/encoding/mvx.rs
+++ b/contracts/multisig-prover/src/encoding/mvx.rs
@@ -1,0 +1,537 @@
+use axelar_wasm_std::operators::Operators;
+use bech32::{self, FromBase32};
+use cosmwasm_std::{HexBinary, Uint256};
+use itertools::Itertools;
+use multisig::key::{PublicKey, Signature};
+use multisig::msg::Signer;
+use multiversx_sc_codec::top_encode_to_vec_u8;
+use sha3::{Digest, Keccak256};
+use multisig::worker_set::WorkerSet;
+
+use crate::types::{CommandBatch, Operator};
+use crate::{error::ContractError};
+
+use super::Data;
+
+pub fn command_params(
+    source_chain: String,
+    source_address: String,
+    destination_address: String,
+    payload_hash: &[u8; 32],
+) -> Result<HexBinary, ContractError> {
+    let map_addr_err = |_| ContractError::InvalidMessage {
+        reason: format!(
+            "destination_address is not a valid Mvx address: {}",
+            destination_address
+        ),
+    };
+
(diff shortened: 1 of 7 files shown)


cgorenflo (commented):
lgtm, but would like to have someone else check it, too

contracts/multisig-prover/src/encoding/mvx.rs
```diff
@@ -0,0 +1,537 @@
...
+
+pub fn msg_digest(command_batch: &CommandBatch) -> HexBinary {
+    // MultiversX is just mimicking EVM here
+    let unsigned = [
+        "\x19MultiversX Signed Message:\n".as_bytes(), // Keccek256 hash length = 32
+        encode(&command_batch.data).as_slice(),
+    ]
+    .concat();
+
+    Keccak256::digest(unsigned).as_slice().into()
```
cjcobb23, line 83: This is different than evm. Notice that we hash twice for evm:
```
    let msg = Keccak256::digest(encode(&command_batch.data).as_slice());

    // Prefix for standard EVM signed data https://eips.ethereum.org/EIPS/eip-191
    let unsigned = [
        "\x19Ethereum Signed Message:\n32".as_bytes(), // Keccek256 hash length = 32
        msg.as_slice(),
    ]
    .concat();

    Keccak256::digest(unsigned).as_slice().into()
```
Is the difference intentional?

raress96: It is intentional, yes. I saw that `bcs` encoding does it like this so I decided on this approach.

At first I just had the encoded data being hashed without the prefix since on MultiversX there is no need for this prefix.
I added it since I saw that both `abi` and `bcs` encoding use one, I assume to make sure that signed data from a chain can not be used on another chain (although unlikely).
Is there a good reason to have this? Because we can remove it to make this contract and our MultiversX Gateway contract slightly more gas efficient.

milapsheth: @raress96 We plan to release a slightly updated gateway version that does the domain separation between chains in a cleaner and more efficient way. In the meantime, I recommend keeping it the same for consistency

raress96:
One thing to note here is that the `multiversx-sc-codec` package requires Rust nightly version. I see that the pipelines are failing because of this.
Also `rust-optimizer` failed when I tried to use it to compile an optimized version of the contract, so I had to create a custom Docker image from `rustlang/rust:nightly-alpine` for it.

Not sure how it is best we handle this.

raress96:
Closing this since a fork will need to be maintained because of Rust nightly version requirement.

PR #184, opened 2023-11-07
refactor(voting-verifier): update poll functions to remove mutable reference to self and separate side effects

eguajardo:
## Description
Context: 
- https://github.com/axelarnetwork/axelar-amplifier/pull/176#discussion_r1382182379
- https://github.com/axelarnetwork/axelar-amplifier/pull/176#discussion_r1382190150

https://axelarnetwork.atlassian.net/browse/AXE-2520

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

### DIFF
--- a/contracts/voting-verifier/src/state.rs
+++ b/contracts/voting-verifier/src/state.rs
@@ -31,7 +31,7 @@ pub enum Poll {
 }
 
 impl Poll {
-    pub fn map_poll<F, E>(self, func: F) -> Result<Self, E>
+    pub fn map<F, E>(self, func: F) -> Result<Self, E>
     where
         F: FnOnce(WeightedPoll) -> Result<WeightedPoll, E>,
         E: From<ContractError>,
--- a/contracts/voting-verifier/src/state.rs
+++ b/contracts/voting-verifier/src/state.rs
@@ -34,7 +34,7 @@ impl voting::Poll for Poll {
     type E = ContractError;
 
     fn finish(self, block_height: u64) -> Result<Self, ContractError> {
-        self.map(|poll| poll.finish(block_height).map_err(ContractError::from))
+        self.try_map(|poll| poll.finish(block_height).map_err(ContractError::from))
     }
 
     fn result(&self) -> voting::PollResult {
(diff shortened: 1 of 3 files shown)


packages/axelar-wasm-std/src/voting.rs
```diff
@@ -275,7 +284,7 @@ impl Poll for WeightedPoll {
         // late votes are not tallied
         if self.status != PollStatus::InProgress {
             participation.vote = Some(votes);
-            return Ok(());
+            return Ok(self);
         }
```
cgorenflo, line 288: do we actually need this?
Basically, what we decided was that the poll is open until it expires, you can just query for a result early

eguajardo: I didn't wanted to mix refactoring PR with logic that changes behavior, this will be dealt in another PR. I just added a TODO meanwhile

contracts/voting-verifier/src/state.rs
```diff
@@ -28,6 +30,19 @@ pub enum Poll {
...
+        F: FnOnce(WeightedPoll) -> Result<WeightedPoll, E>,
+        E: From<ContractError>,
+    {
+        match self {
+            Poll::Messages(poll) => Ok(Poll::Messages(func(poll)?)),
+            Poll::ConfirmWorkerSet(poll) => Ok(Poll::ConfirmWorkerSet(func(poll)?)),
+        }
+    }
+}
+
```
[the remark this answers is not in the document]
eguajardo: I like the idea of implementing the poll trait for the enum and delegate the calls to the WeightedPoll, the option to invert the control I was considering it before but I didn't liked it because it would introduce some logic into the voting module that seems more fit for the contract itself

cgorenflo (approved):

contracts/voting-verifier/src/state.rs
```diff
@@ -30,6 +30,32 @@ pub enum Poll {
...
+impl voting::Poll for Poll {
+    type E = ContractError;
+
+    fn finish(self, block_height: u64) -> Result<Self, ContractError> {
+        self.map(|poll| poll.finish(block_height).map_err(ContractError::from))
+    }
+
+    fn result(&self) -> voting::PollResult {
+        match self {
+            Poll::Messages(poll) | Poll::ConfirmWorkerSet(poll) => poll.result(),
```
cgorenflo, line 42: use map here as well

PR #186, opened 2023-11-09
refactor(multisig): update WorkerSet to use BTreeMap instead of BTreeSet for signers field

fahimahmedx:
## Description
[AXE-2552]

Refactors `WorkerSet.signers` to be a `BTreeMap<String, Signer>` instead of `BTreeSet<Signer>` because it allows faster access to a signer when given a string address. In another branch I am refactoring, we are typically given the signer's address when we want to access the signer.

We use a `BTreeMap` instead of `HashMap` because ordering of the `Signer`s matters.

## Steps to Test
`cargo test` passes

[AXE-2552]: https://axelarnetwork.atlassian.net/browse/AXE-2552?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/multisig-prover/src/contract.rs
```diff
@@ -250,10 +250,15 @@ mod tests {
...
                 .into_iter()
-                .map(|op| Signer {
-                    address: op.address,
-                    pub_key: op.pub_key,
-                    weight: op.weight,
+                .map(|op| {
+                    (
+                        op.address.clone().to_string(),
+                        Signer {
+                            address: op.address,
```
cjcobb23, line 257: Address is stored twice now though. Is that necessary?

fahimahmedx: It's not necessary, but it allows the map to be a `<String, Signer>` instead of `<String, (Uint256, PublicKey)>` which is more readable. I can also remove the `address` from the `Signer` struct, but it feels like it's more understandable to keep all properties of signers into that one struct for future code readability.

What's your thoughts on that?

fahimahmedx:
codecov addressed in PR #187

contracts/multisig-prover/src/contract.rs
cjcobb23: Readability is good, but we don't want to sacrifice performance and store things twice just for the sake of readability. `Signer` is used in a lot of places, so I don't know that removing `address` from `Signer` makes sense. I think just the tuple is okay, but if you want it to be more readable, you can define a new type that is just a (weight, pub key) pair.

contracts/multisig-prover/src/encoding/abi.rs
```diff
@@ -133,12 +133,11 @@ fn encode_proof(
 pub fn make_operators(worker_set: WorkerSet) -> Operators {
     let mut operators: Vec<(HexBinary, Uint256)> = worker_set
         .signers
-        .iter()
+        .values()
         .map(|signer| {
```
cjcobb23, line 137:
```suggestion
        .map(|(weight,pub_key)| {
```
so you can use weight and pub_key instead of signer.0 and signer.1

contracts/multisig-prover/src/encoding/abi.rs
```diff
@@ -177,12 +176,11 @@ fn add27(recovery_byte: u8) -> u8 {
 pub fn transfer_operatorship_params(worker_set: &WorkerSet) -> Result<HexBinary, ContractError> {
     let mut operators: Vec<(HexBinary, Uint256)> = worker_set
         .signers
-        .iter()
+        .values()
         .map(|signer| {
```
cjcobb23, line 180:
```suggestion
        .map(|(weight,pub_key)| {
```

contracts/multisig-prover/src/encoding/bcs.rs
```diff
@@ -20,8 +20,8 @@ use sha3::{Digest, Keccak256};
 pub fn make_operators(worker_set: WorkerSet) -> Operators {
     let mut operators: Vec<(HexBinary, Uint256)> = worker_set
         .signers
-        .iter()
-        .map(|signer| (signer.pub_key.clone().into(), signer.weight))
+        .values()
+        .map(|signer| (signer.1.clone().into(), signer.0))
```
cjcobb23, line 24:
```suggestion
        .map(|(weight,pub_key)| (pub_key.clone().into(), weight))
```

contracts/multisig-prover/src/encoding/bcs.rs
```diff
@@ -33,8 +33,8 @@ pub fn make_operators(worker_set: WorkerSet) -> Operators {
 pub fn transfer_operatorship_params(worker_set: &WorkerSet) -> Result<HexBinary, ContractError> {
     let mut operators: Vec<(HexBinary, Uint256)> = worker_set
         .signers
-        .iter()
-        .map(|s| (s.pub_key.clone().into(), s.weight))
+        .values()
+        .map(|s| (s.1.clone().into(), s.0))
```
cjcobb23, line 37:
```suggestion
        .map(|(weight,pub_key)| (...))
```

contracts/multisig-prover/src/encoding/bcs.rs
```diff
@@ -234,7 +234,7 @@ mod test {
         let mut expected: Vec<(Vec<u8>, u128)> = worker_set
             .signers
             .into_iter()
-            .map(|s| (s.pub_key.as_ref().to_vec(), u256_to_u128(s.weight)))
+            .map(|(_, s)| (s.1.as_ref().to_vec(), u256_to_u128(s.0)))
```
cjcobb23, line 237:
```suggestion
            .map(|(_, (weight,pub_key)| (pub_key.as_ref().to_vec(), u256_to_u128(weight)))
```

contracts/multisig-prover/src/execute.rs
```diff
@@ -216,13 +219,8 @@ fn make_keygen_msg(
...
             .signers
-            .into_iter()
-            .map(|signer| {
-                (
-                    signer.address.to_string(),
-                    (KeyType::Ecdsa, signer.pub_key.as_ref().into()),
-                )
-            })
+            .iter()
+            .map(|(address, signer)| (address.clone(), (KeyType::Ecdsa, signer.1.as_ref().into())))
```
cjcobb23:
```suggestion
            .map(|(address, (_,pub_key)| (address.clone(), (KeyType::Ecdsa, pub_key.as_ref().into())))
```

contracts/multisig-prover/src/execute.rs
```diff
@@ -299,13 +297,8 @@ pub fn confirm_worker_set(deps: DepsMut) -> Result<Response, ContractError> {
...
```
cjcobb23:
```suggestion
            .map(|(address, (_,pub_key)| (address.clone(), (KeyType::Ecdsa, pub_key.as_ref().into())))
```

contracts/multisig-prover/src/execute.rs
```diff
@@ -317,8 +310,28 @@ pub fn should_update_worker_set(
     cur_workers: &WorkerSet,
     max_diff: usize,
 ) -> bool {
-    new_workers.signers.difference(&cur_workers.signers).count()
-        + cur_workers.signers.difference(&new_workers.signers).count()
+    let new_workers_signers = new_workers
+        .signers
+        .iter()
+        .map(|(address, signer)| Signer {
```
cjcobb23, line 326:
```suggestion
        .map(|(address, (weight,pub_key)| Signer {
```

contracts/multisig-prover/src/execute.rs
```diff
@@ -317,8 +310,28 @@ pub fn should_update_worker_set(
...
+
+    let cur_workers_signers = cur_workers
+        .signers
+        .iter()
+        .map(|(address, signer)| Signer {
+            address: Addr::unchecked(address.clone()),
+            weight: signer.clone().0,
+            pub_key: signer.clone().1,
+        })
+        .collect::<BTreeSet<Signer>>();
```
cjcobb23, line 341: Hmm, this is a bit clunky now that signers is not a BTreeSet. I don't know how I feel about creating a BTreeSet just for this. I have a feeling it would be faster with just vectors, especially since the inputs are likely to be small. Can you try it with just vectors (i.e. convert signers to a vector, and then just compare the vectors), and see which is faster, vector comparison versus BTreeSet?

cjcobb23 (changes requested):

contracts/multisig-prover/src/encoding/bcs.rs
fahimahmedx: Why `...`?

contracts/multisig-prover/src/execute.rs
fahimahmedx: Using a Vector is much faster, I'll update the PR to use them instead. Below are the results and the code used to test.

> Time elapsed in BTreeSet is: 51.016µs
> Time elapsed in Vector is: 13.716µs
> test execute::tests::should_update_worker_set_no_change ... ok

> Time elapsed in BTreeSet is: 41.678µs
> Time elapsed in Vector is: 14.988µs
> test execute::tests::should_update_worker_set_one_more ... ok

> Time elapsed in BTreeSet is: 40.916µs
> Time elapsed in Vector is: 14.437µs
> test execute::tests::should_update_worker_set_one_less ... ok

> Time elapsed in BTreeSet is: 40.737µs
> Time elapsed in Vector is: 13.055µs
> test execute::tests::should_update_worker_set_one_more_higher_threshold ... ok

> Time elapsed in BTreeSet is: 33.583µs
> Time elapsed in Vector is: 13.856µs
> test execute::tests::should_update_worker_set_diff_pub_key ... ok

Code used for testing:
```
pub fn should_update_worker_set(
    new_workers: &WorkerSet,
    cur_workers: &WorkerSet,
    max_diff: usize,
) -> bool {
    let mut result = false;
    {
        let start = Instant::now();
        
        let new_workers_signers = new_workers
            .signers
            .iter()
            .map(|(address, (weight, pub_key))| Signer {
                address: Addr::unchecked(address.clone()),
                weight: weight.clone(),
                pub_key: pub_key.clone(),
            })
            .collect::<BTreeSet<Signer>>();

        let cur_workers_signers = cur_workers
            .signers
            .iter()
            .map(|(address, (weight, pub_key))| Signer {
                address: Addr::unchecked(address.clone()),
                weight: weight.clone(),
                pub_key: pub_key.clone(),
            })
            .collect::<BTreeSet<Signer>>();

        result = new_workers_signers.difference(&cur_workers_signers).count()
            + cur_workers_signers.difference(&new_workers_signers).count()
            > max_diff;

        let duration = start.elapsed();
        println!("Time elapsed in BTreeSet is: {:?}", duration);
    }

    {
        let start = Instant::now();
        
        let new_workers_signers = new_workers
            .signers
            .iter()
            .map(|(address, (weight, pub_key))| Signer {
                address: Addr::unchecked(address.clone()),
                weight: weight.clone(),
                pub_key: pub_key.clone(),
            })
            .collect::<Vec<Signer>>();

        let cur_workers_signers = cur_workers
            .signers
            .iter()
            .map(|(address, (weight, pub_key))| Signer {
                address: Addr::unchecked(address.clone()),
                weight: weight.clone(),
                pub_key: pub_key.clone(),
            })
            .collect::<Vec<Signer>>();



        result = new_workers_signers.iter().filter(|item| !cur_workers_signers.contains(item)).count()
            + cur_workers_signers.iter().filter(|item| !new_workers_signers.contains(item)).count()
            > max_diff;

        let duration = start.elapsed();
        println!("Time elapsed in Vector is: {:?}", duration);
    }

    result
}
```

contracts/multisig-prover/src/encoding/bcs.rs
cjcobb23: I just didn't feel like typing out the rest of it, it's not proper syntax

contracts/multisig-prover/src/execute.rs
cjcobb23: yeah I figured

contracts/multisig-prover/src/execute.rs
```diff
@@ -317,8 +318,34 @@ pub fn should_update_worker_set(
...
-    new_workers.signers.difference(&cur_workers.signers).count()
-        + cur_workers.signers.difference(&new_workers.signers).count()
+    let new_workers_signers = new_workers
+        .signers
+        .iter()
+        .map(|(address, (weight, pub_key))| Signer {
+            address: Addr::unchecked(address.clone()),
+            weight: *weight,
+            pub_key: pub_key.clone(),
+        })
```
cjcobb23, line 328: Do you need the `map` here? Can you just collect the tuples?

fahimahmedx: In my code I have to convert the `String` address into an `Addr`.
Is this possible without a `map()`?

cgorenflo: `BTreeSet` already provides a `contains` function, can't you use that directly?

cgorenflo: even better, have a look at the `BTreeSet` `intersection()` function https://doc.rust-lang.org/std/collections/struct.BTreeSet.html#method.intersection

cjcobb23: What's wrong with collecting into that datatype? Can't we just do the comparison with that datatype.

@cgorenflo It's not a `BTreeSet` though, it's a `BTreeMap`

cgorenflo: ah damn, github showed me the version before the change 😠

contracts/multisig/src/worker_set.rs
```diff
@@ -1,15 +1,15 @@
...
 use axelar_wasm_std::Participant;
 use cosmwasm_schema::cw_serde;
 use cosmwasm_std::{HexBinary, Uint256};
 use sha3::{Digest, Keccak256};
 
 #[cw_serde]
 pub struct WorkerSet {
-    pub signers: BTreeSet<Signer>,
+    // An ordered map with the signer's address as the key, and the signer's weight and public key as the value.
+    pub signers: BTreeMap<String, (Uint256, PublicKey)>,
```
cgorenflo, line 12: don't make the key part of the value, if you make it a `BTreeMap<String, Signer>` you don't have to do all the mappings

fahimahmedx: When you mention not making the key part of the value, are you referring to `PublicKey

fahimahmedx: @cgorenflo When you mention not making the key part of the value, are you referring to `PublicKey` being part of the value?

The reason `Signer` is no longer the value is because it would cause the `address` field to be in both the `BTreeMap` key, and the value in `Signer.address`.
https://github.com/axelarnetwork/axelar-amplifier/pull/186#discussion_r1388482431

fahimahmedx: `BTreeMap` is back to a `Signer` value for the following reason below as per @cgorenflo's suggestion.
When using maps, it's good practice to have the key only be used for looking up the value, rather than also being part of the value (and subsequently being used in the code's logic).
For example, in my code I often use the `address` (key) to rebuild the `Signer` struct, when it's instead better to directly store the `Signer` struct in the map for that key.
Another reason is because if I need to change the datatype of `address` later on, then at least the key still remains the same and does not need to be modified again.

fish-sammy (approved):

contracts/multisig-prover/src/execute.rs
fahimahmedx: `BTreeMap` is back to using `Signer` instead of tuples as its value, and so this no longer needed.

cgorenflo (approved):

PR #187, opened 2023-11-09
test: add additional test cases to uint.rs

fahimahmedx:
## Description
[AXE-2574]
Adds test cases to `uint.rs` that codecov identified were missing.

## Steps to Test
`cargo test` passes

[AXE-2574]: https://axelarnetwork.atlassian.net/browse/AXE-2574?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/axelar-wasm-std/src/nonempty/uint.rs
+++ b/packages/axelar-wasm-std/src/nonempty/uint.rs
@@ -121,58 +121,58 @@ mod tests {
     use super::*;
 
     #[test]
-    fn test_non_zero_uint64() {
-        assert!(Uint64::try_from(cosmwasm_std::Uint64::one()).is_ok());
-        assert!(Uint64::try_from(1u64).is_ok());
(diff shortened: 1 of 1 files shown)


fish-sammy (approved):

fahimahmedx:
leaving the PR open for @cgorenflo

cgorenflo (commented):
Since you are already adding to the tests this would be a good opportunity to clean them up a bit. In rust, starting tests with `test_` is redundant, since they are already annotated with a `#[test]` attribute. So I would suggest renaming them along the lines of `convert_from_x_to_y` and because the test cases are pretty simple I think it's a good idea to merge all tests that check the same conversion with zero and non-zero values where appropriate.

PR #188, opened 2023-11-10
refactor: clean up rewards contract

cgorenflo:


contracts/rewards/src/contract/execute.rs
```diff
@@ -77,113 +77,106 @@ where
...
-
-        Ok(workers_to_reward
-            .into_iter()
-            .map(|worker| (worker, rewards_per_worker))
-            .collect())
+        contract_addr: &Addr,
+        from: u64,
+        to: u64,
+    ) -> HashMap<Addr, Uint128> {
+        (from..=to)
```
fish-sammy, line 171: I am honestly not convinced that this is cleaner than the recursion.

contracts/rewards/src/contract/execute.rs
```diff
@@ -77,113 +77,106 @@ where
...
-            .map_or(Ok(HashMap::new()), |tally| self.process_epoch_tally(tally))
+            .load_rewards_pool(contract_addr.clone())?
+            .sub_reward(
+                self.cumulate_rewards(&contract_addr, from, to)
+                    .values()
+                    .sum(),
+            )?
+            .then(|pool| self.store.save_rewards_pool(&pool))?;
+
+        Ok(self.cumulate_rewards(&contract_addr, from, to))
```
cjcobb23, line 162: Am I misreading this, or are we calling `self.cumulate_rewards` twice with the exact same parameters? If so, I don't think we should do that, for gas and performance reasons. Can we just cache the cumulated value somehow?

contracts/rewards/src/contract/execute.rs
cjcobb23: Yeah I'm not sure that this cleaner. The recursive version seems a little easier to read and understand in my opinion, and is more concise:
```
        if from > to {
            return Ok(rewards);
        }
        let new_rewards = self.process_rewards_for_epoch(contract.clone(), from)?;
        self.process_rewards_for_epochs(contract, from + 1, to, merge_rewards(rewards, new_rewards))
```

contracts/rewards/src/contract/execute.rs
```diff
@@ -77,113 +77,106 @@ where
...
+                cur_epoch,
+                self.store.load_params().params,
+            ))
+            .record_participation(worker)
+            .then(|mut tally| {
+                if matches!(event, LoadState::New(_)) {
+                    tally = tally.increment_event_count()
+                }
+                self.store.save_epoch_tally(&tally)
+            })
```
cjcobb23, line 96: I like this refactor, much cleaner!

contracts/rewards/src/contract/execute.rs
```diff
@@ -77,113 +77,106 @@ where
...
+            .unwrap_or(EpochTally::new(
+                contract_addr,
+                cur_epoch,
+                self.store.load_params().params,
+            ))
+            .record_participation(worker)
+            .then(|mut tally| {
+                if matches!(event, LoadState::New(_)) {
+                    tally = tally.increment_event_count()
+                }
```
cjcobb23, line 94: I think we can get rid of this and just set the event count to 1 in the `unwrap_or` a few lines above.

contracts/rewards/src/contract/execute.rs
cgorenflo: oh, line 162 just shouldn't be there!

contracts/rewards/src/contract/execute.rs
cgorenflo: no, because the tally has to be incremented based on if the event is new, not if the tally is new

contracts/rewards/src/contract/execute.rs
cgorenflo: I'll give it another try, my issue with the recursion was that it went three layers deep

contracts/rewards/src/contract/execute.rs
cjcobb23: oh yes, ok nvm

contracts/rewards/src/contract/execute.rs
fish-sammy: to me, recursion is always the cleanest when the call stack has 3 layers lol.

contracts/rewards/src/contract/execute.rs
```diff
@@ -42,23 +43,28 @@ where
     /// block height and the epoch duration. If the epoch duration is updated, we store the epoch
     /// in which the update occurs as the last checkpoint
     fn current_epoch(&self, cur_block_height: u64) -> Result<Epoch, ContractError> {
+        assert!(
+            cur_block_height <= i64::MAX as u64,
```
fish-sammy, line 47: Why do all the calculations in `i64` when everything is `u64`?

contracts/rewards/src/contract/execute.rs
```diff
@@ -77,113 +83,110 @@ where
...
+        self.store
+            .load_epoch_tally(contract_addr.clone(), event.epoch_num)?
+            .unwrap_or(EpochTally::new(
+                contract_addr,
+                cur_epoch,
+                self.store.load_params().params,
+            ))
+            .record_participation(worker)
+            .then(|mut tally| {
+                if matches!(event, LoadState::New(_)) {
```
fish-sammy, line 98: The enum name is confusing. I'd prefer renaming `Loaded` to `Existing` and `LoadState` to just `State`.

contracts/rewards/src/contract/execute.rs
```diff
@@ -77,113 +83,110 @@ where
...
-
-        self.store.save_epoch_tally(&tally)?;
-
-        Ok(())
     }
 
     pub fn distribute_rewards(
         &mut self,
-        contract: Addr,
+        contract_addr: Addr,
```
fish-sammy, line 127: I find the `_addr` suffix is pretty redudant. The type is already `Addr`.

contracts/rewards/src/state.rs
```diff
@@ -108,17 +124,12 @@ impl RewardsPool {
...
-        let rewards = rewards_per_worker * Uint128::from(worker_count as u128);
-        if self.balance < rewards {
+    pub fn sub_reward(mut self, reward: Uint128) -> Result<Self, ContractError> {
+        if self.balance < reward {
             return Err(ContractError::PoolBalanceInsufficient.into());
         }
-        self.balance -= rewards;
-        Ok(())
+        self.balance -= reward;
+        Ok(self)
```
fish-sammy, line 132:
```suggestion
        
        Ok(self)
```
Prefer always having an empty line before return.

contracts/rewards/src/state.rs
```diff
@@ -251,23 +268,40 @@ impl Store for RewardsStore<'_> {
     }
 }
 
-#[cfg(test)]
-mod test {
-    use std::collections::HashMap;
+pub(crate) enum LoadState<T> {
+    Loaded(T),
```
fish-sammy, line 272:
```suggestion
    Existing(T),
```

contracts/rewards/src/state.rs
```diff
@@ -251,23 +268,40 @@ impl Store for RewardsStore<'_> {
     }
 }
 
-#[cfg(test)]
-mod test {
-    use std::collections::HashMap;
+pub(crate) enum LoadState<T> {
```
fish-sammy, line 271:
```suggestion
pub(crate) enum State<T> {
```

contracts/rewards/src/contract/execute.rs
```diff
@@ -42,23 +43,28 @@ where
...
-        if cur_block_height < epoch.block_height_started {
-            return Err(ContractError::BlockHeightInPast.into());
-        }
-        if cur_block_height < epoch.block_height_started + epoch_duration {
-            return Ok(epoch);
+        let last_updated_epoch = stored_params.last_updated;
+
+        let epochs_elapsed = (cur_block_height as i64)
+            .saturating_sub_unsigned(last_updated_epoch.block_height_started)
+            / (epoch_duration as i64); // epoch duration is always positive, so this is safe
```
cjcobb23, line 57: Does this compute to 0 if cur_block_height is in the past, but less than an epoch duration before last_updated_epoch.block_height_started? For example, cur_block_height is 9, last_updated_epoch.block_height_started is 10, and epoch_duration is 5. The subtraction computes -1, and then -1 / 10 = 0. Is that right?

contracts/rewards/src/contract/execute.rs
cjcobb23: I also have this question

contracts/rewards/src/contract/execute.rs
```diff
@@ -42,23 +43,28 @@ where
...
+
+        match epochs_elapsed {
+            ..=-1 => Err(ContractError::BlockHeightInPast.into()),
```
cjcobb23, line 60: I think we need a test for this case, unless I missed it

contracts/rewards/src/contract/execute.rs
cgorenflo: damn this stupid calculation!

contracts/rewards/src/contract/execute.rs
cgorenflo: coverage showed that it was covered, but I'll doublecheck

contracts/rewards/src/contract/execute.rs
cgorenflo: I renamed it to `target_contract`, I want to avoid confusion, because self is of type `Contract`

contracts/rewards/src/contract/execute.rs
cgorenflo: I tried to use this to have a nicely aligned match statement, but I changed it now

contracts/rewards/src/contract/execute.rs
```diff
@@ -44,20 +45,23 @@ where
...
+        epochs_elapsed.map(|epochs_elapsed| {
+            match epochs_elapsed {
+                0 => last_updated_epoch,
+                1.. => Epoch {
+                    epoch_num: last_updated_epoch.epoch_num + epochs_elapsed,
+                    block_height_started: last_updated_epoch.block_height_started
+                        + (epochs_elapsed * epoch_duration), // result is strictly less than cur_block_height, so multiplication is safe
+                },
+            }
         })
```
fish-sammy, line 65:
```suggestion
        match cur_block_height {
            _ if cur_block_height < last_updated_epoch.block_height_started => Err(ContractError::BlockHeightInPast.into())
            _ => {
                let epochs_elapsed = (cur_block_height - last_updated_epoch.block_height_started) / epoch_duration;

                Ok(Epoch {
                    epoch_num: last_updated_epoch.epoch_num + epochs_elapsed,
                    block_height_started: last_updated_epoch.block_height_started
                        + (epochs_elapsed * epoch_duration), // result is strictly less than cur_block_height, so multiplication is safe
                })
            }
        }
```
There's no need to differentiate epochs_elapsed=0 and epochs_elapsed>=1

fish-sammy (approved):

cjcobb23 (approved):

PR #190, opened 2023-11-14
feat(voting-verifier): move early verification logic to query

eguajardo:
## Description
Lets the poll finish until it expires, but allows early verification before it finishes if consensus has reached already

https://axelarnetwork.atlassian.net/browse/AXE-2513

## Todos

For future PRs
- Throw event when voting if consensus has reached
- Apply same logic to worker set confirmation
- Change votes to enums

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [x] Connect epics/issues


cgorenflo (changes requested):
test

contracts/voting-verifier/src/execute.rs
```diff
@@ -92,54 +83,46 @@ pub fn verify_messages(
...
-            Pending(message) => Some(message),
-            Verified(_) => None,
+        .map(|message| verification_status(deps.as_ref(), &message).map(|status| (status, message)))
+        .collect::<Result<Vec<(VerificationStatus, Message)>, ContractError>>()?;
+
+    let msgs_to_verify: Vec<Message> = messages
+        .into_iter()
+        .filter_map(|(status, message)| match status {
+            VerificationStatus::NotVerified | VerificationStatus::None => Some(message),
+            _ => None,
```
cgorenflo, line 99: discard is dangerous because the compiler won't complain when the enum changes. Since it's not a simple "filter for one specific case and do something else in all other cases", I'd prefer to make all cases explicit

contracts/voting-verifier/src/execute.rs
```diff
@@ -249,10 +200,9 @@ pub fn end_poll(deps: DepsMut, env: Env, poll_id: PollID) -> Result<Response, Co
 
     let poll_result = poll.result();
 
-    match poll {
-        state::Poll::Messages(_) => end_poll_messages(deps, poll_id, &poll_result)?,
-        state::Poll::ConfirmWorkerSet(_) => end_poll_worker_set(deps, poll_id, &poll_result)?,
-    };
+    if matches!(poll, state::Poll::ConfirmWorkerSet(_)) {
+        end_poll_worker_set(deps, poll_id, &poll_result)?;
```
cgorenflo, line 204: can we get rid of this in the future as well?

contracts/voting-verifier/src/query.rs
```diff
@@ -1,28 +1,256 @@
...
+
+pub fn is_verified(deps: Deps, messages: &[Message]) -> StdResult<Vec<(CrossChainId, bool)>> {
     messages
-        .into_iter()
+        .iter()
         .map(|message| {
-            is_message_verified(deps, &message).map(|verified| (message.cc_id, verified))
+            verification_status(deps, message).map(|status| match status {
+                VerificationStatus::Verified => (message.cc_id.to_owned(), true),
+                _ => (message.cc_id.to_owned(), false),
```
cgorenflo, line 23: this is a good example where the discard makes sense

packages/axelar-wasm-std/src/voting.rs
```diff
@@ -137,6 +140,8 @@ pub trait Poll {
         Self: Sized;
     // returns the cumulated poll result
     fn result(&self) -> PollResult;
+    // returns the consensus for the message at the given index
+    fn consensus(&self, idx: usize) -> Result<bool, Self::E>;
```
cgorenflo, line 144: I don't like this as part of the trait. Polls don't necessarily have messages. Additionally, the name is a bit weird

packages/axelar-wasm-std/src/voting.rs
```diff
@@ -137,6 +140,8 @@ pub trait Poll {
         Self: Sized;
     // returns the cumulated poll result
     fn result(&self) -> PollResult;
```
cgorenflo, line 142: should PollResult be an associated type? It doesn't really work well for WorkerSet polls. Also, I don't understand the field `consensus_participants`, shouldn't we have a list per message instead of one for the whole poll?

contracts/voting-verifier/src/state.rs
```diff
@@ -69,13 +77,28 @@ impl Poll {
     }
 }
 
+#[cw_serde]
+pub struct PollMessage {
+    pub msg: Message,
+    pub poll_id: PollID,
+    pub index_in_poll: usize,
+}
```
cgorenflo, line 85: why is this now necessary?

contracts/voting-verifier/src/query.rs
```diff
@@ -1,28 +1,256 @@
...
-    deps: Deps,
-    messages: Vec<Message>,
-) -> StdResult<Vec<(CrossChainId, bool)>> {
+#[cw_serde]
+pub enum VerificationStatus {
+    Verified,
+    NotVerified,
+    InProgress, // still in an open poll
+    None,       // not in a poll
+}
```
cgorenflo, line 15: why do we need all these states now?

cgorenflo (changes requested):

contracts/voting-verifier/src/state.rs
eguajardo: Before, we had in storage the `PENDING_MESSAGES` which were moved to the `VERIFIED_MESSAGES` when the poll was finished and if they were verified, but now messages verification is calculated on the fly when the query is called so we don’t need those storage constants anymore, instead since the verification is calculated on the fly regardless of if the poll is finished or not (for early verification) wee need a way to locate a given message inside a poll to check for it’s verification status. The query receives the message itself and not a poll, so I created POLL_MESSAGES which use this struct to link a message to its location in a poll. This is where it’s being used: https://github.com/axelarnetwork/axelar-amplifier/blob/692c64508eee9a4bb7de518cd37a2fdc4b43f3d6/contracts/voting-verifier/src/query.rs#L44

contracts/voting-verifier/src/query.rs
eguajardo: it's used by the execute function to check if the messages requested to verify were already verified or not. Before we just checked `VERIFIED_MESSAGES` storage, messages could be in multiple polls in progress but would end up in a single place in the `VERIFIED_MESSAGES`, but now the calculation is done on the fly and we need to check the poll where the message is being voted on to check it's verification status, so we can't allow the same message to be in multiple polls in progress at the same time. We only allow a message to be part of a poll if its not verified and is not part of a poll in progress

contracts/voting-verifier/src/query.rs
```diff
@@ -1,28 +1,256 @@
...
 
-pub fn is_message_verified(deps: Deps, message: &Message) -> Result<bool, ContractError> {
-    match VERIFIED_MESSAGES.may_load(deps.storage, &message.cc_id)? {
-        Some(stored) if stored != *message => {
+pub fn verification_status(
+    deps: Deps,
+    message: &Message,
+) -> Result<VerificationStatus, ContractError> {
+    match POLL_MESSAGES.may_load(deps.storage, &message.cc_id)? {
+        Some(stored) if stored.msg != *message => {
```
cjcobb23, line 35: I think this could be problematic. Invalid messages submitted to the verifier that have the same id but different contents than a valid message will block validation and routing for the valid message. We need to somehow allow for two different messages with the same id to be in polls at the same time.

cjcobb23 (changes requested):

contracts/connection-router/src/state.rs
```diff
@@ -101,6 +102,18 @@ pub struct Message {
     pub payload_hash: [u8; 32],
 }
 
+impl Message {
+    pub fn hash_id(&self) -> [u8; 32] {
```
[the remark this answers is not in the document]
eguajardo: created task as well:  https://axelarnetwork.atlassian.net/browse/AXE-2649

contracts/voting-verifier/src/state.rs
```diff
@@ -69,13 +77,29 @@ impl Poll {
...
+
 pub const POLL_ID: counter::Counter<PollID> = counter::Counter::new("poll_id");
 
 pub const POLLS: Map<PollID, Poll> = Map::new("polls");
 
-pub const PENDING_MESSAGES: Map<PollID, Vec<Message>> = Map::new("pending_messages");
-
-pub const VERIFIED_MESSAGES: Map<&CrossChainId, Message> = Map::new("verified_messages");
+// message hash to `PollMessage`
```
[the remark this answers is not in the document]
eguajardo: updated

contracts/voting-verifier/src/execute.rs
```diff
@@ -92,54 +83,46 @@ pub fn verify_messages(
...
-            Pending(message) => Some(message),
-            Verified(_) => None,
+        .map(|message| verification_status(deps.as_ref(), &message).map(|status| (status, message)))
+        .collect::<Result<Vec<(VerificationStatus, Message)>, ContractError>>()?;
+
+    let msgs_to_verify: Vec<Message> = messages
+        .into_iter()
+        .filter_map(|(status, message)| match status {
+            VerificationStatus::FailedToVerify | VerificationStatus::NotVerified => Some(message),
+            VerificationStatus::InProgress | VerificationStatus::Verified => None,
```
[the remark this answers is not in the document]
eguajardo: I did a collect first to make it fail in case of error, I wasn't sure how to chain it when the previous map returned results

contracts/connection-router/src/state.rs
```diff
@@ -101,6 +102,19 @@ pub struct Message {
     pub payload_hash: [u8; 32],
 }
 
+impl Message {
+    // TODO: pending to finalize the design of the message hash
+    pub fn hash_id(&self) -> [u8; 32] {
+        let mut hasher = Keccak256::new();
+        hasher.update(self.cc_id.to_string());
+        hasher.update(self.source_address.as_str());
```
cjcobb23, line 110: Need to be careful about things like casing and "0x" prefixing in addresses.

contracts/connection-router/src/state.rs
```diff
@@ -275,6 +289,12 @@ impl KeyDeserialize for &ChainName {
     }
 }
 
+impl ChainName {
+    pub fn as_str(&self) -> &str {
+        self.0.as_str()
+    }
+}
```
cjcobb23, line 296: Can we just implement `AsRef` on ChainName to get a reference to the underlying string, and then just call as_str() on the reference?

contracts/voting-verifier/src/execute.rs
```diff
@@ -243,17 +194,18 @@ pub fn end_poll(deps: DepsMut, env: Env, poll_id: PollID) -> Result<Response, Co
...
-        state::Poll::Messages(_) => end_poll_messages(deps, poll_id, &poll_result)?,
-        state::Poll::ConfirmWorkerSet(_) => end_poll_worker_set(deps, poll_id, &poll_result)?,
+    let poll_result = match &poll {
+        Poll::Messages(poll) | Poll::ConfirmWorkerSet(poll) => poll.state(),
     };
 
+    if matches!(poll, state::Poll::ConfirmWorkerSet(_)) {
+        end_poll_worker_set(deps, poll_id, &poll_result)?;
+    }
+
```
cjcobb23, line 208: We could probably do some cleanup here, and delete messages in the poll that didn't achieve quorum. Those messages could just be spam.

contracts/voting-verifier/src/state.rs
```diff
@@ -69,13 +43,29 @@ impl Poll {
...
+
 pub const POLL_ID: counter::Counter<PollID> = counter::Counter::new("poll_id");
 
 pub const POLLS: Map<PollID, Poll> = Map::new("polls");
 
-pub const PENDING_MESSAGES: Map<PollID, Vec<Message>> = Map::new("pending_messages");
-
-pub const VERIFIED_MESSAGES: Map<&CrossChainId, Message> = Map::new("verified_messages");
+// key is message `hash_id`
+pub const POLL_MESSAGES: Map<&[u8; 32], PollMessage> = Map::new("poll_messages");
```
cjcobb23, line 68: I feel like it might be cleaner to use a named type for the key.

contracts/voting-verifier/src/query.rs
```diff
@@ -1,28 +1,263 @@
...
+use crate::state::{self, Poll, POLLS, POLL_MESSAGES};
 
-pub fn verification_statuses(
-    deps: Deps,
-    messages: Vec<Message>,
-) -> StdResult<Vec<(CrossChainId, bool)>> {
+#[cw_serde]
+pub enum VerificationStatus {
+    Verified,
+    FailedToVerify,
```
cjcobb23, line 12: Do we actually need this status? I feel like we could just use `NotVerified` for this.

cjcobb23 (approved):
Left some comments, but nothing blocking

contracts/connection-router/src/state.rs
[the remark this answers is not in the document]
fish-sammy: Why not just `#[derive(Hash)]` on `Message`?

contracts/voting-verifier/src/query.rs
fish-sammy: Also confused.

contracts/voting-verifier/src/query.rs
```diff
@@ -1,28 +1,263 @@
...
+pub fn is_verified(deps: Deps, messages: &[Message]) -> StdResult<Vec<(CrossChainId, bool)>> {
     messages
-        .into_iter()
+        .iter()
         .map(|message| {
-            is_message_verified(deps, &message).map(|verified| (message.cc_id, verified))
+            verification_status(deps, message).map(|status| match status {
+                VerificationStatus::Verified => (message.cc_id.to_owned(), true),
+                _ => (message.cc_id.to_owned(), false),
+            })
```
fish-sammy, line 24:
```suggestion
            verification_status(deps, message).map(|status| {
                (
                    message.cc_id.to_owned(),
                    matches!(status, VerificationStatus::Verified),
                )
            })
```

contracts/voting-verifier/src/query.rs
```diff
@@ -1,28 +1,263 @@
...
         })
         .collect::<Result<Vec<(_, _)>, _>>()
         .map_err(Into::into)
```
fish-sammy, line 27:
```suggestion
        .collect::<Result<Vec<_>, _>>()
```

contracts/voting-verifier/src/query.rs
```diff
@@ -1,28 +1,263 @@
...
+                .expect("invalid invariant: message poll not found");
+
+            let verified = match &poll {
+                Poll::Messages(poll) | Poll::ConfirmWorkerSet(poll) => poll
+                    .has_consensus(stored.index_in_poll)
+                    .expect("invalid invariant: message not found in poll"),
+            };
+
+            if verified {
+                if stored.msg != *message {
```
fish-sammy, line 47: How can it mismatch anymore if the messages is stored by its own hash? Panic should happen here.

contracts/voting-verifier/src/query.rs
```diff
@@ -1,28 +1,263 @@
...
+                    return Err(ContractError::MessageMismatch(message.cc_id.to_string()));
+                }
+                Ok(VerificationStatus::Verified)
+            } else if is_finished(&poll) {
+                Ok(VerificationStatus::FailedToVerify)
+            } else {
+                Ok(VerificationStatus::InProgress)
+            }
```
fish-sammy, line 55:
```suggestion
            if verified {
                assert_eq!(stored.msg, *message, "...");

                return Ok(VerificationStatus::Verified);
            }

            if is_finished(&poll) {
                return Ok(VerificationStatus::FailedToVerify);
            }

            Ok(VerificationStatus::InProgress)
```
Avoid nested-if's and else and else if at all cost.

contracts/voting-verifier/src/execute.rs
```diff
@@ -92,54 +80,46 @@ pub fn verify_messages(
...
 
-    let pending_messages: Vec<Message> = messages
+    let messages = messages
         .into_iter()
-        .filter_map(|status| match status {
-            Pending(message) => Some(message),
-            Verified(_) => None,
+        .map(|message| verification_status(deps.as_ref(), &message).map(|status| (status, message)))
+        .collect::<Result<Vec<(VerificationStatus, Message)>, ContractError>>()?;
+
```
fish-sammy, line 91:
```suggestion
        .collect::<Result<Vec<_>, _>>()?;
```
Don't provide what the compiler can infer.

contracts/voting-verifier/src/execute.rs
```diff
@@ -92,54 +80,46 @@ pub fn verify_messages(
...
+    for (idx, message) in msgs_to_verify.iter().enumerate() {
+        POLL_MESSAGES.save(
+            deps.storage,
+            &message.hash_id(),
+            &state::PollMessage::new(message.clone(), id, idx),
+        )?;
+    }
 
-    let evm_messages = pending_messages
+    let evm_messages = msgs_to_verify
```
fish-sammy, line 122: Why `evm_messages`?

contracts/connection-router/src/state.rs
eguajardo: lets discuss this in future PR

contracts/connection-router/src/state.rs
eguajardo: implemented `AsRef<str>` and used it directly

contracts/voting-verifier/src/query.rs
eguajardo: I tried to represent different status regardless of some being used same way for some things

contracts/voting-verifier/src/execute.rs
eguajardo: https://axelarnetwork.atlassian.net/browse/AXE-2650

fish-sammy (changes requested):

contracts/connection-router/src/state.rs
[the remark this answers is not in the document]
cgorenflo: for now I think it's ok to have a stable hashing function as a placeholder. We can address a finalized design in a followup PR

contracts/voting-verifier/src/execute.rs
cgorenflo: oh that's right, the voter shouldn't make any mentions of specific chains

contracts/voting-verifier/src/query.rs
cgorenflo: I had this discussion with @eguajardo, in the end we settled on "define states by what they represent, not what they are used for"

contracts/voting-verifier/src/query.rs
cgorenflo: nice suggestion!

contracts/voting-verifier/src/query.rs
cgorenflo: this is something we generally need to revisit (not just here). I don't like how the storage logic intermingles with other concerns and how we have these implicit expectations that specific parts can never break. I think we should move these panics down into the `store` layer (which the verifier doesn't have yet). That way the store can solely focus on which data always has to be consistent when stored and loaded, and the rest of the contract never has to think about if data is available or not.

contracts/voting-verifier/src/query.rs
cgorenflo: yes to avoid nesting, but I prefer `else if` when the cases are on the same hierarchical level, i.e. `if... return` is fine when it's guard clauses, but when it's effectively a switch between multiple similar cases I want the code to reflect that

contracts/voting-verifier/src/state.rs
cgorenflo: yes, please alias the key

contracts/connection-router/src/state.rs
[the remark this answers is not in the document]
eguajardo: was my first thought, but figured we would probably want to be in control of each individual field hash, specially if we change the message struct. Also I wasn't sure if it was a good idea to use the `Hash` trait since it's used by other things in Rust like HashMaps

contracts/voting-verifier/src/query.rs
eguajardo: agree with Chris, `else if` seem more readable in this case

cgorenflo (approved):

contracts/voting-verifier/src/query.rs
fish-sammy: I never find else and else if readable and clean, but you guys insist sure,

fish-sammy (approved):

contracts/connection-router/src/state.rs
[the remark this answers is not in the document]
cjcobb23: I am not sure that deriving Hash is safe for our use case: https://doc.rust-lang.org/std/hash/trait.Hash.html#portability

"Due to differences in endianness and type sizes, data fed by Hash to a Hasher should not be considered portable across platforms. Additionally the data passed by most standard library types should not be considered stable between compiler versions.

This means tests shouldn’t probe hard-coded hash values or data fed to a Hasher and instead should check consistency with Eq.

Serialization formats intended to be portable between platforms or compiler versions should either avoid encoding hashes or only rely on Hash and Hasher implementations that provide additional guarantees."

I worry that if we recompile the contract during an upgrade, the hashing might change. I do think it's very important to be absolutely sure the hashing function is stable.

PR #193, opened 2023-11-16
test: initial setup for integration tests

cjcobb23:
Code to instantiate all of the contracts in a simulation environment. Can execute contract calls and the contracts will talk to each other. No need for mock contracts.

### DIFF
--- a/integration-tests/src/test.rs
+++ b/integration-tests/src/test.rs
@@ -7,6 +7,7 @@ mod test {
     use cw_multi_test::{App, ContractWrapper, Executor};
 
     use multisig::key::PublicKey;
+    use tofn::ecdsa::KeyPair;
 
     const AXL_DENOMINATION: &str = "uaxl";
     #[test]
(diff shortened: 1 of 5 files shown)


integration-tests/src/test.rs
```diff
@@ -0,0 +1,472 @@
...
+    use axelar_wasm_std::nonempty;
+    use connection_router::state::ChainName;
+    use cosmwasm_std::{coins, Addr, Binary, Deps, Env, HexBinary, StdResult, Uint128, Uint256};
+    use cw_multi_test::{App, ContractWrapper, Executor};
+
+    use multisig::key::PublicKey;
+
+    const AXL_DENOMINATION: &str = "uaxl";
+    #[test]
+    fn test() {
```
fish-sammy, line 13: This is a bad name.

fish-sammy (changes requested):

integration-tests/src/test.rs
```diff
@@ -0,0 +1,494 @@
...
+        }
+    }
+
+    // return the all-zero array with the first bytes set to the bytes of `index`
+    fn generate_key(seed: usize) -> KeyPair {
+        let index_bytes = seed.to_be_bytes();
+        let mut result = [0; 64];
+        for (i, &b) in index_bytes.iter().enumerate() {
+            result[i] = b;
+        }
```
cgorenflo, line 142: is there no direct way to set this up without enumeration?

cgorenflo (approved):
looks good apart from the open comments

integration-tests/src/test.rs
cjcobb23: lol this is just a placeholder to show how the setup of the testing environment works. It's not really a test. But I can change it to something a little better for now anyways.

fish-sammy (approved):

PR #196, opened 2023-11-19
feat: nexus gateway RouteToRouter

fish-sammy:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

### DIFF
--- a/contracts/nexus-gateway/src/contract/execute.rs
+++ b/contracts/nexus-gateway/src/contract/execute.rs
@@ -20,24 +20,20 @@ where
     ) -> Result<Response<nexus::Message>> {
         match sender {
             sender if sender == self.config.nexus => self
-                .route_to_router(
-                    msgs.into_iter()
-                        .map(TryInto::try_into)
-                        .collect::<Result<Vec<_>>>()?,
-                )
+                .route_to_router(msgs)
                 .change_context(ContractError::RouteToRouter),
             sender if sender == self.config.router => self
-                .route_to_nexus(msgs.into_iter().map(Into::into))
+                .route_to_nexus(msgs)
                 .change_context(ContractError::RouteToNexus),
             _ => Err(report!(ContractError::Unauthorized)),
         }
     }
 
-    fn route_to_router(
-        self,
-        msgs: impl IntoIterator<Item = nexus::Message>,
-    ) -> Result<Response<nexus::Message>> {
-        let msgs: Vec<_> = msgs.into_iter().map(Into::into).collect();
+    fn route_to_router(self, msgs: Vec<msg::Message>) -> Result<Response<nexus::Message>> {
+        let msgs: Vec<_> = msgs
+            .into_iter()
+            .map(connection_router::Message::from)
+            .collect();
         if msgs.is_empty() {
             return Ok(Response::default());
         }
@@ -50,12 +46,10 @@ where
         }))
     }
 
-    fn route_to_nexus(
-        mut self,
-        msgs: impl IntoIterator<Item = connection_router::Message>,
-    ) -> Result<Response<nexus::Message>> {
+    fn route_to_nexus(mut self, msgs: Vec<msg::Message>) -> Result<Response<nexus::Message>> {
         let msgs = msgs
             .into_iter()
+            .map(connection_router::Message::from)
             .filter_map(|msg| match self.store.is_message_routed(&msg.cc_id) {
                 Ok(true) => None,
                 Ok(false) => Some(Ok(msg)),
@@ -66,10 +60,7 @@ where
         msgs.iter()
             .try_for_each(|msg| self.store.set_message_routed(&msg.cc_id))?;
 
-        let msgs = msgs
-            .into_iter()
-            .map(TryInto::try_into)
-            .collect::<Result<Vec<nexus::Message>>>()?;
+        let msgs: Vec<nexus::Message> = msgs.into_iter().map(Into::into).collect();
 
         Ok(Response::new().add_messages(msgs))
     }
@@ -176,14 +167,17 @@ mod test {
                     contract_addr,
                     msg,
                     funds,
-                }) => match from_binary(msg) {
-                    Ok(connection_router::msg::ExecuteMsg::RouteMessages(msgs)) => {
-                        *contract_addr == Addr::unchecked("router")
+                }) => {
+                    if let Ok(connection_router::msg::ExecuteMsg::RouteMessages(msgs)) =
+                        from_binary(msg)
+                    {
+                        return *contract_addr == Addr::unchecked("router")
                             && msgs.len() == 2
-                            && funds.is_empty()
+                            && funds.is_empty();
                     }
-                    _ => false,
-                },
+
+                    false
+                }
                 _ => false,
             }
         }));
--- a/contracts/nexus-gateway/src/contract/execute.rs
+++ b/contracts/nexus-gateway/src/contract/execute.rs
@@ -1,4 +1,3 @@
-use connection_router::state::Message;
 use cosmwasm_std::{to_binary, Addr, Response, WasmMsg};
 use error_stack::report;
 
@@ -40,7 +39,7 @@ where
     pub fn route_to_nexus(
         mut self,
         sender: Addr,
-        msgs: Vec<Message>,
+        msgs: Vec<connection_router::Message>,
     ) -> Result<Response<nexus::Message>> {
         if sender != self.config.router {
             return Err(report!(ContractError::Unauthorized));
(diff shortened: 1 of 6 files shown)


contracts/nexus-gateway/src/nexus.rs
```diff
@@ -1,8 +1,10 @@
 use axelar_wasm_std::nonempty;
-use connection_router::state::{Address, ChainName, Message as RouterMessage, ID_SEPARATOR};
+use connection_router::state::{
+    Address, ChainName, CrossChainId, Message as RouterMessage, ID_SEPARATOR,
```
cjcobb23, line 3: I don't like that we do this renaming: `Message as RouterMessage`. Just my take, but I feel like it makes the code confusing. Everywhere else it's referred to as `Message`, but in this crate we have a different type called `Message`, and I feel like that collision is a bit confusing when reading the code.

cjcobb23 (changes requested):
requesting changes because I'd prefer to adhere to the gateway interface unless there is some good reason not to

contracts/nexus-gateway/src/msg.rs
```diff
@@ -9,6 +11,6 @@ pub struct InstantiateMsg {
 
 #[cw_serde]
 pub enum ExecuteMsg {
-    VerifyMessages(Vec<Message>),
-    RouteMessages(Vec<Message>),
+    RouteMessagesToNexus(Vec<Message>),
+    RouteMessagesToRouter(Vec<nexus::Message>),
```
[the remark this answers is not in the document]
fish-sammy: There is a good reason here. We don't want the nexus module to have to receive and send different things from and to this nexus gateway. Gateway interface consistency doesn't matter when nothing external ever interacts with it.

contracts/nexus-gateway/src/nexus.rs
fish-sammy: What's your suggestion then?

contracts/nexus-gateway/src/msg.rs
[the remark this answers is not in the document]
cjcobb23: why can't the nexus module just receive and send `connection_router::state::Message` via `RouteMessages`?

contracts/nexus-gateway/src/nexus.rs
cjcobb23: I would just rename `Message` in this file to `NexusMessage`

contracts/nexus-gateway/src/msg.rs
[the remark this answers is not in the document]
fish-sammy: Cuz again, we need to avoid core having to make any change due to the wasm contracts, i.e. whatever change occurs to the `connection_router::state::Message` (rename of fields, field removal, new field added) should not affect core.

contracts/nexus-gateway/src/nexus.rs
fish-sammy: I would re-export that `connection_router::state::Message` from `connection_router`.

contracts/nexus-gateway/src/msg.rs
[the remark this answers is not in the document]
cjcobb23: Those things are protocol changes, that I don't think would happen very often, and would need to be backwards compatible anyways, else we will break the entire protocol.

At the very least, I think the connection router still should just call `RouteMessages` when passing a message to the nexus gateway, to keep it consistent with every other gateway. This really is just a gateway. So `RouteMessagesToNexus` should just stay as `RouteMessages`. And I would just rename `RouteMessagesToRouter` to something like `TransformAndRouteMessages`.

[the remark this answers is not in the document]
fish-sammy: Often or not, protocol changes in wash should have minimum impact on core. Sure I can keep it as RouteMessages for going to nexus. Don't see why the other one should have that "transform" though.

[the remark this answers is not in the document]
fish-sammy: Actually, I may have an idea. Let me try some.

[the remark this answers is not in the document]
fish-sammy: renamed

contracts/nexus-gateway/src/contract/execute.rs
```diff
@@ -15,20 +14,46 @@ where
...
+                .route_to_nexus(msgs.into_iter().map(Into::into))
                 .change_context(ContractError::RouteToNexus),
             _ => Err(report!(ContractError::Unauthorized)),
         }
     }
 
-    fn route_to_nexus(&mut self, msgs: Vec<Message>) -> Result<Response<nexus::Message>> {
+    fn route_to_router(
+        self,
+        msgs: impl IntoIterator<Item = nexus::Message>,
```
cgorenflo, line 38: this is used in a very specific context. Consider changing the type to the simpler `Vec`?

contracts/nexus-gateway/src/contract/execute.rs
```diff
@@ -15,20 +14,46 @@ where
...
+    ) -> Result<Response<nexus::Message>> {
+        let msgs: Vec<_> = msgs.into_iter().map(Into::into).collect();
```
cgorenflo, line 40: seems weird to have two type conversions split across two functions like this. Consider merging them either on the caller site or here

contracts/nexus-gateway/src/contract/execute.rs
```diff
@@ -15,20 +14,46 @@ where
...
+            sender if sender == self.config.nexus => self
+                .route_to_router(
+                    msgs.into_iter()
+                        .map(TryInto::try_into)
+                        .collect::<Result<Vec<_>>>()?,
+                )
+                .change_context(ContractError::RouteToRouter),
             sender if sender == self.config.router => self
-                .route_to_nexus(msgs)
+                .route_to_nexus(msgs.into_iter().map(Into::into))
```
cgorenflo, line 30: I prefer `X::from` so I can see which type it's supposed to be even without IDE

contracts/nexus-gateway/src/contract/execute.rs
cgorenflo, line 40: see previous `X::from` comment

contracts/nexus-gateway/src/contract/execute.rs
```diff
@@ -61,26 +87,119 @@ mod test {
...
+        assert!(res.is_ok_and(|res| {
+            if res.messages.len() != 1 {
+                return false;
+            }
+
+            match &res.messages[0].msg {
+                CosmosMsg::Wasm(WasmMsg::Execute {
+                    contract_addr,
+                    msg,
+                    funds,
```
cgorenflo, line 178: consider `let else` pattern to reduce nesting

contracts/nexus-gateway/src/nexus.rs
```diff
@@ -45,26 +45,47 @@ fn parse_message_id(message_id: &str) -> Result<(nonempty::Vec<u8>, u64), Contra
     Ok((tx_id, index))
 }
 
-impl TryFrom<RouterMessage> for Message {
+impl TryFrom<connection_router::Message> for Message {
     type Error = Report<ContractError>;
 
-    fn try_from(msg: RouterMessage) -> Result<Self, ContractError> {
+    fn try_from(msg: connection_router::Message) -> Result<Self, ContractError> {
```
cgorenflo, line 51: why is this a try_from?

contracts/nexus-gateway/src/nexus.rs
```diff
@@ -45,26 +45,47 @@ fn parse_message_id(message_id: &str) -> Result<(nonempty::Vec<u8>, u64), Contra
...
+impl From<Message> for connection_router::Message {
+    fn from(msg: Message) -> Self {
+        Self {
+            cc_id: CrossChainId {
+                chain: msg.source_chain,
+                id: format!(
+                    "{}:{}",
+                    msg.source_tx_id.as_ref().encode_hex::<String>(),
+                    msg.source_tx_index
+                )
```
cgorenflo, line 77: just for my understanding: this doesn't need to get parsed anywhere else right? It only needs to be unique but we could use any kind of format here

contracts/nexus-gateway/src/contract/execute.rs
cjcobb23: I also prefer `Vec<nexus::Message>` here, and doing the type conversion only once, at the call site.

cjcobb23 (approved):

contracts/nexus-gateway/src/nexus.rs
fish-sammy: good catch. It used to have to be a try_from but not anymore.

contracts/nexus-gateway/src/nexus.rs
fish-sammy: correct

cgorenflo (approved):

PR #197, opened 2023-11-20
feat: query worker address

haiyizxx:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


ampd/src/cli.rs
```diff
@@ -0,0 +1,34 @@
+use error_stack::{Report, ResultExt};
+
+use crate::state::{State, StateUpdater};
+use crate::tofnd::grpc::{MultisigClient, SharableEcdsaClient};
+use crate::tofnd::Config;
+use crate::Error;
+use crate::PREFIX;
+
+pub async fn worker_address(config: Config, state: State) -> Result<String, Report<Error>> {
```
cjcobb23, line 9: Does this function return the tendermint address, or the evm address?

ampd/src/main.rs
```diff
@@ -150,3 +168,27 @@ enum Error {
...
+
+async fn worker_address(args: &Args) -> ExitCode {
+    let config = init_config(&args.config);
+    let state_path = expand_home_dir(&args.state);
+
+    let state = match state::load(&state_path).change_context(Error::Fatal) {
+        Ok(state) => state,
+        Err(report) => {
+            eprintln!("{report:?}");
+            return ExitCode::FAILURE;
```
cgorenflo, line 180: reuse the setup for error logging and exit codes from the main() function, i.e. return the report instead of an error code and don't print the error here

cgorenflo: this goes for all spots this function does this

ampd/src/main.rs
```diff
@@ -38,26 +41,41 @@ enum Output {
     Json,
 }
 
+#[derive(Debug, Subcommand, Valuable)]
+enum SubCommand {
+    /// Run the ampd daemon process (default)
+    Daemon,
+    /// Query the worker address
```
cgorenflo, line 48: it's not just a query right? if the address hasn't been set yet it'll actually run a keygen

ampd/src/cli.rs
cgorenflo, line 1: I don't like cli.rs just for `worker_address`
- anything cli related is handled in main.rs, so this function could be called from anywhere
- it's on the same hierarchical level as the daemon, so I'd prefer to treat them equally

Suggestions:
- create a subfolder `commands` or `api` or somthing similar
- add `daemon.rs` and `worker_address.rs` files
- call them from `main.rs` as `commands::daemon::run()` or `api::worker_address::run()`

cgorenflo (changes requested):

ampd/src/commands/worker_address.rs
```diff
@@ -0,0 +1,39 @@
+use error_stack::{Report, ResultExt};
+use std::path::PathBuf;
+
+use crate::state::{load, StateUpdater};
+use crate::tofnd::grpc::{MultisigClient, SharableEcdsaClient};
+use crate::tofnd::Config;
+use crate::Error;
+use crate::PREFIX;
+
+pub async fn run(config: Config, state_path: PathBuf) -> Result<(), Report<Error>> {
```
cgorenflo, line 10: use either `AsRef<Path>` or `&Path`, `PathBuf` is only necessary to correctly parse the starting args. It's similar to `String` and `&str`, so `PathBuf` makes a full copy of the content

ampd/src/commands/worker_address.rs
```diff
@@ -0,0 +1,39 @@
+use error_stack::{Report, ResultExt};
+use std::path::PathBuf;
+
+use crate::state::{load, StateUpdater};
```
cgorenflo, line 4: don't import `load`, use `state::load` in the code to make it nicer to read

ampd/src/commands/daemon.rs
```diff
@@ -0,0 +1,26 @@
+use error_stack::{Report, ResultExt};
+use std::path::PathBuf;
+
+use crate::config::Config;
+use crate::state::{flush, load};
+use crate::Error;
+
+pub async fn run(config: Config, state_path: PathBuf) -> Result<(), Report<Error>> {
```
cgorenflo, line 8: use either `AsRef<Path>` or `&Path`, `PathBuf` is only necessary to correctly parse the starting args. It's similar to `String` and `&str`, so `PathBuf` makes a full copy of the content

cgorenflo (approved):

ampd/src/commands/worker_address.rs
```diff
@@ -0,0 +1,39 @@
...
+            Ok(pub_key)
+        }
+    }
+    .map(|pub_key| {
+        println!(
+            "worker address: {}",
+            pub_key
+                .account_id(PREFIX)
+                .expect("failed to convert to account identifier")
+                .to_string()
```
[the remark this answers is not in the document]
haiyizxx: we do check in tofnd client, and return error if it fails to parse

PR #198, opened 2023-11-22
feat: add json schema generators to every contract

cgorenflo:


contracts/voting-verifier/src/msg.rs
```diff
@@ -61,7 +61,7 @@ pub enum QueryMsg {
     #[returns(Poll)]
     GetPoll { poll_id: PollID },
 
-    #[returns(Vec<(CrossChainId, bool)>)]
+    #[returns(Vec<(connection_router::state::CrossChainId, bool)>)]
```
fish-sammy, line 64: What's the reason for this change?

cgorenflo: trying to build the json schema showed that the unqualified name couldn't be resolved

cjcobb23 (approved):
Looks good

PR #199, opened 2023-11-22
feat(ampd): implement MultiversX verify message & verify worker set

raress96:
Add verify message and verify worker set implementations for MultiversX to ampd module.


raress96:
Closing this since a fork will need to be maintained because of Rust nightly version requirement.

PR #201, opened 2023-11-23
feat(voting-verifier): move worker set verification logic to query

eguajardo:
## Description

Related to https://github.com/axelarnetwork/axelar-amplifier/pull/190, but applied to worker set confirmation

https://axelarnetwork.atlassian.net/browse/AXE-2677

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


contracts/voting-verifier/src/state.rs
```diff
@@ -45,7 +45,7 @@ impl Poll {
 
 #[cw_serde]
 pub struct PollMessage {
-    pub msg: Message,
+    pub msg: Message, // message is stored for migration purposes in case the hash changes
```
cgorenflo, line 48: nice

contracts/voting-verifier/src/state.rs
```diff
@@ -60,6 +60,18 @@ impl PollMessage {
     }
 }
 
+#[cw_serde]
+pub struct PollWorkerSet {
+    pub operators: Operators, // operators are stored for migration purposes in case the hash changes
+    pub poll_id: PollID,
+}
```
cgorenflo, line 67: Can we merge this with `PollMessage` to something like 
```
pub struct PollContent<T> {
    pub content: T<
    pub poll_id: PollId,
    pub index_in_poll: usize
}
```
that means we need to deal with fewer structs, both have the same structure and we don't need to put the explicit `0` in `Ok(verification_status(deps, stored.poll_id, 0))`, which is a bug waiting to happen, if WorkerSet can ever be confirmed together with other poll items.

As a sidenote, could you rename `PollID` to `PollId`? The rust naming conventions are quite clear that acronyms should also use camel case

packages/axelar-wasm-std/src/operators.rs
```diff
@@ -1,18 +1,47 @@
...
     pub weights_by_addresses: Vec<(HexBinary, Uint256)>,
     pub threshold: Uint256,
 }
 
 impl Operators {
-    pub fn hash(&self) -> Vec<u8> {
-        Keccak256::digest(to_vec(self).expect("could not convert Operators to vec"))
-            .as_slice()
-            .into()
+    pub fn hash_id(&self) -> OperatorsHash {
```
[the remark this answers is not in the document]
eguajardo: I used `hash_id` so it wouldn't be confused with the trait `Hash` which implements a `hash` function

contracts/voting-verifier/src/state.rs
eguajardo: done. Since we have a common structure now I also tried to remove some other duplicated code inside the verification status functions, take a look at this commit: https://github.com/axelarnetwork/axelar-amplifier/pull/201/commits/0123282503ab2b407f4bc00c35d465f52ee8a196

contracts/voting-verifier/src/contract.rs
```diff
@@ -57,10 +57,8 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> StdResult<Binary> {
         QueryMsg::GetPoll { poll_id: _ } => {
             todo!()
         }
-        QueryMsg::IsWorkerSetConfirmed { new_operators } => to_binary(
-            &CONFIRMED_WORKER_SETS
-                .may_load(deps.storage, new_operators.hash())?
-                .is_some(),
-        ),
+        QueryMsg::IsWorkerSetConfirmed { new_operators } => {
+            to_binary(&query::is_worker_set_confirmed(deps, &new_operators)?)
```
fish-sammy, line 61: 👍

packages/axelar-wasm-std/src/operators.rs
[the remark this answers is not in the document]
fish-sammy: Same, I'd also call it just `hash`. I don't see a need to avoid that confusion.

packages/axelar-wasm-std/src/operators.rs
```diff
@@ -1,18 +1,47 @@
...
+        let mut hasher = Keccak256::new();
+        hasher.update(
+            to_binary(&self.weights_by_addresses).expect("could not serialize serializable object"),
```
fish-sammy, line 18: Should we make sure that `weights_by_addresses` is sorted before hashing?

packages/axelar-wasm-std/src/operators.rs
```diff
@@ -1,18 +1,47 @@
...
+
+    // If this test fails, it means the operator hash has changed and therefore a migration is needed.
+    #[test]
+    fn hash_id_unchanged() {
+        let expected_operators_hash =
+            "a80417c50895c34c73dee98112e288a3b8dce0d9d77d42cbfbf6bbe29433770b";
+
+        let operators = Operators {
+            weights_by_addresses: vec![(
+                HexBinary::from_hex("6C51eec96bf0a8ec799cdD0Bbcb4512f8334Afe8").unwrap(),
```
fish-sammy, line 39: Add at least another one into the vec here. Otherwise, if order changes during the hasing process, this test wouldn't fail.

packages/axelar-wasm-std/src/operators.rs
```diff
@@ -1,18 +1,47 @@
 use cosmwasm_schema::cw_serde;
-use cosmwasm_std::{to_vec, HexBinary, Uint256};
+use cosmwasm_std::{to_binary, HexBinary, Uint256};
 
 use sha3::{Digest, Keccak256};
 
+pub type OperatorsHash = [u8; 32];
```
fish-sammy, line 6: Do we not have a Hash type already? There's nothing special about this "Operators" hash.

eguajardo: we only had `MessageHash`, I just created a type `Hash` to replace both of those

packages/axelar-wasm-std/src/operators.rs
cgorenflo: yeah that seems like a good use of a constructor

contracts/voting-verifier/src/state.rs
```diff
@@ -44,31 +46,28 @@ impl Poll {
...
+pub struct PollContent<T> {
+    pub content: T, // content is stored for migration purposes in case the hash changes
+    pub poll_id: PollId,
     pub index_in_poll: u32,
 }
 
-impl PollMessage {
-    pub fn new(msg: Message, poll_id: PollID, index_in_poll: usize) -> Self {
+impl<T> PollContent<T> {
+    pub fn new(content: T, poll_id: PollId, index_in_poll: usize) -> Self {
```
cgorenflo, line 56: this just moved the previously mentioned problem from the `verification_status` call to this constructor. I suggest creating two constructors, e.g. `new_for_msg` and `new_for_worker_set`, where `new_for_worker_set` doesn't take an index

contracts/voting-verifier/src/query.rs
```diff
@@ -48,19 +78,14 @@ pub fn verification_status(
             };
 
             if verified {
```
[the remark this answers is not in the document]
eguajardo: this should be a consideration for the next PR where the votes are changed to enums. Currently with booleans we only tally for `true`, so consensus can only mean it was verified

contracts/voting-verifier/src/query.rs
```diff
@@ -31,12 +32,41 @@ pub fn is_verified(
         .collect::<Result<Vec<_>, _>>()
 }
 
-pub fn verification_status(
+pub fn is_worker_set_confirmed(deps: Deps, operators: &Operators) -> Result<bool, ContractError> {
```
fish-sammy, line 35:
```suggestion
pub fn is_worker_set_verified(deps: Deps, operators: &Operators) -> Result<bool, ContractError> {
```
I dislike it that we mix the usage of verified and confirmed.

fish-sammy (approved):

PR #203, opened 2023-11-23
feat: ignore source chain validation when routed from the nexus gateway; route to the nexus gateway when the destination chain cannot be found

fish-sammy:
## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


contracts/connection-router/src/contract/execute.rs
```diff
@@ -103,18 +103,38 @@ pub fn unfreeze_chain(
...
+            return Ok(());
+        }
+
         let source_chain = self
             .store
-            .load_chain_by_gateway(&sender)?
+            .load_chain_by_gateway(sender)?
             .ok_or(ContractError::GatewayNotRegistered)?;
         if source_chain.incoming_frozen() {
             return Err(report!(ContractError::ChainFrozen {
```
cjcobb23, line 140: I feel like we should be able to freeze the nexus gateway

contracts/connection-router/src/contract/execute.rs
```diff
@@ -126,46 +146,461 @@ where
...
+            .group_by(|msg| msg.destination_chain.to_owned())
+            .into_iter()
+            .map(|(destination_chain, msgs)| {
+                let gateway = match self.store.load_chain_by_chain_name(&destination_chain)? {
+                    Some(destination_chain) => {
+                        if destination_chain.outgoing_frozen() {
+                            return Err(report!(ContractError::ChainFrozen {
+                                chain: destination_chain.name,
+                            }));
+                        }
```
cjcobb23, line 170: I feel like we should be able to freeze the nexus gateway

cjcobb23 (commented):
I wonder if it would be cleaner to just actually register the nexus gateway as a gateway for each chain registered in core. Then we can handle everything in a consistent way, and we can handle messages going to and from the nexus gateway the same as messages going to/from any other gateway. 

The router could actually be aware of what chains the nexus gateway is responsible for, instead of blindly forwarding messages for which it doesn't know the chain. Right now, messages with unknown destination chains are just passed to the nexus gateway. What error does that actually return if nexus also doesn't know the chain? It's a bit hard to find, because the code for that is in another repo, in another language.

We also can't freeze the nexus gateway in either direction, as far as I can tell at least. I am concerned about the safety of this. How exactly do we freeze the connection between nexus and core in the event of an emergency?

I think the only thing we'd need to do to enable this is just allow multiple chains to be registered to a single gateway. It requires a bit of a refactor, but not too much really. I also think this is worthwhile in the long run, because it would allow us to connect to other routers. The nexus is basically just another router, but blindly forwarding messages with unknown destination chains doesn't really work when there is more than one router connected to this router.

Thoughts?

contracts/connection-router/src/contract/execute.rs
fish-sammy: future PR

contracts/connection-router/src/msg.rs
```diff
@@ -7,6 +7,8 @@ pub struct InstantiateMsg {
     pub admin_address: String,
     // governance votes on chains being added or upgraded
     pub governance_address: String,
+    // the address of the nexus gateway
+    pub nexus_gateway: String,
```
cjcobb23, line 11: If the router needs to know the address of the nexus gateway, and the nexus gateway needs to know the address of the router, how do we instantiate the two contracts?

contracts/connection-router/src/contract/execute.rs
```diff
@@ -103,18 +103,38 @@ pub fn unfreeze_chain(
...
+    Ok(())
+}
+
+pub fn require_governance(deps: &DepsMut, info: MessageInfo) -> Result<(), ContractError> {
+    let config = CONFIG.load(deps.storage)?;
+    if config.governance != info.sender {
+        return Err(ContractError::Unauthorized);
+    }
+    Ok(())
+}
```
cgorenflo, line 120: take config and sender instead? the contract can already load config

contracts/connection-router/src/contract/execute.rs
```diff
@@ -103,18 +103,38 @@ pub fn unfreeze_chain(
...
-        self,
-        sender: Addr,
-        msgs: Vec<Message>,
-    ) -> error_stack::Result<Response, ContractError> {
+    fn validate_msgs(
+        &self,
+        sender: &Addr,
+        msgs: &[Message],
+    ) -> error_stack::Result<(), ContractError> {
+        if sender == self.config.nexus_gateway {
```
cgorenflo, line 131: add comment why this shortcuts the validation

contracts/connection-router/src/contract/execute.rs
```diff
@@ -103,18 +103,38 @@ pub fn unfreeze_chain(
...
-    pub fn route_messages(
-        self,
-        sender: Addr,
-        msgs: Vec<Message>,
-    ) -> error_stack::Result<Response, ContractError> {
+    fn validate_msgs(
+        &self,
+        sender: &Addr,
+        msgs: &[Message],
+    ) -> error_stack::Result<(), ContractError> {
```
cgorenflo, line 130: return messages instead of ()?

contracts/connection-router/src/contract/execute.rs
```diff
@@ -126,46 +146,461 @@ where
...
+                let gateway = match self.store.load_chain_by_chain_name(&destination_chain)? {
+                    Some(destination_chain) => {
+                        if destination_chain.outgoing_frozen() {
+                            return Err(report!(ContractError::ChainFrozen {
+                                chain: destination_chain.name,
+                            }));
+                        }
+
+                        destination_chain.gateway.address
+                    }
```
cgorenflo, line 173: make these two separate cases by moving the if clause into the pattern match?

contracts/connection-router/src/contract/execute.rs
```diff
@@ -126,46 +146,461 @@ where
...
+                    None if sender != self.config.nexus_gateway => {
+                        self.config.nexus_gateway.clone()
```
cgorenflo, line 175: add comment explaining that we want to route any unkown chains to nexus

contracts/connection-router/src/msg.rs
cgorenflo: probably better to register the nexus gateway after instantiation?

cjcobb23: yeah I think that's best. I'm not sure the best way to do this though. The two ways I can think of is to either add a new execute msg called `RegisterNexusGateway`, or to just handle it inside `RegisterChain`. I don't really want to add a new execute msg to the API, but I also don't want to complicate the implementation of `RegisterChain`.

contracts/connection-router/src/contract/execute.rs
fish-sammy: Like that

contracts/connection-router/src/contract/execute.rs
fish-sammy: I didn't make any change to these 2 functions but moved them up above the contract struct impl. I'd wanna avoid any change to them in this PR.

contracts/connection-router/src/contract.rs
```diff
@@ -58,176 +73,27 @@ pub fn execute(
             execute::require_admin(&deps, info)?;
             execute::unfreeze_chain(deps, chain, direction)
         }
-        ExecuteMsg::RouteMessages(msgs) => execute::route_message(deps, info, msgs),
+        ExecuteMsg::RouteMessages(msgs) => Ok(contract.route_messages(info.sender, msgs)?),
```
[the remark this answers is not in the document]
fish-sammy: Lol no. I did this only bcuz I didn't refactor the other functions, and as a result, they all return standard results but my method returns an error_stack report.

contracts/connection-router/src/contract/execute.rs
fish-sammy: I don't hate that.

cjcobb23 (approved):

PR #207, opened 2023-11-29
feat: dockerfile optimized build

RiceAndMeet:
## Description

- Building Dependencies and ampd separately in docker 
- Add axelard user to image
- add entrypoint.sh startup script 

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

### DIFF
--- a/ampd/Dockerfile
+++ b/ampd/Dockerfile
@@ -1,10 +1,32 @@
-FROM rust:bookworm as build
-WORKDIR /app
+FROM rust:1.74-bookworm as builder
 RUN apt-get update && apt-get install -y clang protobuf-compiler
-COPY . .
-RUN cargo build --release
 
-FROM debian:bookworm-slim AS final
-COPY --from=build /app/target/release/ampd /usr/local/bin/ampd
-WORKDIR /usr/local/bin
-ENTRYPOINT ["ampd"]
+WORKDIR /ampd
+COPY ./Cargo.toml .
+COPY ./ampd/Cargo.toml ./ampd/Cargo.toml
+COPY ./ampd/Cargo.lock ./ampd/Cargo.lock
+COPY ./packages ./packages
+COPY ./contracts ./contracts
+COPY ./integration-tests ./integration-tests
+COPY ./.cargo ./.cargo
+
+# build dependencies separately
+RUN mkdir ampd/src && echo 'fn main() {}' > ampd/src/main.rs
+RUN --mount=type=ssh cargo build --release 
+
+COPY ./ampd/proto ./ampd/proto
+COPY ./ampd/src ./ampd/src
+COPY ./ampd/build.rs ./ampd/build.rs
+
+RUN rustup component add rustfmt
+RUN --mount=type=ssh cargo install --locked --path ./ampd 
+
+FROM debian:bookworm-slim AS runner
+RUN addgroup --system --gid 1001 axelard && adduser --home /home/axelard --system --uid 1000 --ingroup axelard axelard
+WORKDIR /home/axelard
+RUN mkdir /.ampd && chown axelard /.ampd
+USER axelard
+COPY --from=builder /ampd/target/release/ampd /usr/local/bin
+COPY ./entrypoint.sh ./
+
+ENTRYPOINT ["./entrypoint.sh"]
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+set -e
+
+trap stop_gracefully TERM INT
+
+stop_gracefully(){
+  echo "stopping ampd process"
+  killall "ampd"
+  sleep 5
+  echo "ampd process stopped"
+}
+
+/usr/local/bin/ampd &
+wait
+exec "$@"


entrypoint.sh
```diff
@@ -0,0 +1,15 @@
...
+stop_gracefully(){
+  echo "stopping ampd process"
+  killall "ampd"
+  sleep 5
+  echo "ampd process stopped"
+}
+
+/usr/local/bin/ampd &
+wait
+exec "$@"
```
fahimahmedx, line 15: What are some examples of commands that could be executed in this context?

ampd/Dockerfile
```diff
@@ -1,10 +1,32 @@
...
+COPY ./ampd/Cargo.toml ./ampd/Cargo.toml
+COPY ./ampd/Cargo.lock ./ampd/Cargo.lock
+COPY ./packages ./packages
+COPY ./contracts ./contracts
+COPY ./integration-tests ./integration-tests
+COPY ./.cargo ./.cargo
+
+# build dependencies separately
+RUN mkdir ampd/src && echo 'fn main() {}' > ampd/src/main.rs
+RUN --mount=type=ssh cargo build --release
```
fahimahmedx, line 15: `--mount=type=ssh`
 This is a public github repo, do we still need this?

ampd/Dockerfile
```diff
@@ -1,10 +1,32 @@
...
+RUN rustup component add rustfmt
+RUN --mount=type=ssh cargo install --locked --path ./ampd 
+
+FROM debian:bookworm-slim AS runner
+RUN addgroup --system --gid 1001 axelard && adduser --home /home/axelard --system --uid 1000 --ingroup axelard axelard
+WORKDIR /home/axelard
+RUN mkdir /.ampd && chown axelard /.ampd
+USER axelard
+COPY --from=builder /ampd/target/release/ampd /usr/local/bin
+COPY ./entrypoint.sh ./
```
fahimahmedx, line 30: Is the line `RUN chmod +x ./entrypoint.sh` missing or required?

fahimahmedx (changes requested):

entrypoint.sh
RiceAndMeet: A common use case is `docker run -it ampd /bin/bash`
If ampd process exit with error, it will exec into container immediately, make it easier to debug within container environment.

ampd/Dockerfile
RiceAndMeet: Its not needed if `entrypoint.sh` file committed to git already has the execute permission.

ampd/Dockerfile
```diff
@@ -1,10 +1,32 @@
...
+
+COPY ./ampd/proto ./ampd/proto
+COPY ./ampd/src ./ampd/src
+COPY ./ampd/build.rs ./ampd/build.rs
+
+RUN rustup component add rustfmt
+RUN --mount=type=ssh cargo install --locked --path ./ampd 
+
+FROM debian:bookworm-slim AS runner
+RUN addgroup --system --gid 1001 axelard && adduser --home /home/axelard --system --uid 1000 --ingroup axelard axelard
```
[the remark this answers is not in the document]
RiceAndMeet: Good question, you got me thinking, I didn't think its possible to have a user without a group in most linux distro. But apparently, after some [googling](https://askubuntu.com/questions/853399/can-we-have-a-user-without-a-primary-group), there is such thing as `nogroup` in `/etc/passwd`. 

Regardless, its a common/good practice to always specify group when creating a user.

ampd/Dockerfile
```diff
@@ -1,10 +1,32 @@
...
 RUN apt-get update && apt-get install -y clang protobuf-compiler
-COPY . .
-RUN cargo build --release
 
-FROM debian:bookworm-slim AS final
-COPY --from=build /app/target/release/ampd /usr/local/bin/ampd
-WORKDIR /usr/local/bin
-ENTRYPOINT ["ampd"]
+WORKDIR /ampd
+COPY ./Cargo.toml .
```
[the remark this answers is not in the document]
RiceAndMeet: Good catch. Will change it to keep things consistent.

ampd/Dockerfile
RiceAndMeet: I don't think its needed in this case. Gotta double check.

ampd/Dockerfile
```diff
@@ -1,10 +1,32 @@
...
 
-FROM debian:bookworm-slim AS final
-COPY --from=build /app/target/release/ampd /usr/local/bin/ampd
-WORKDIR /usr/local/bin
-ENTRYPOINT ["ampd"]
+COPY ./ampd/proto ./ampd/proto
+COPY ./ampd/src ./ampd/src
+COPY ./ampd/build.rs ./ampd/build.rs
+
+RUN rustup component add rustfmt
```
talalashraf, line 21: why is the formatter needed for release ?

talalashraf (approved):

ampd/Dockerfile
RiceAndMeet: It doesn't, removed.  This was copied over from tnfnd dockerfile.

PR #208, opened 2023-11-29
feat: bond worker command

haiyizxx:
## Description
bond worker to service registry contract
- add ServiceRegistryConfig in ampd config file
- run `ampd bond-worker service_name amount denom`
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


ampd/src/commands/bond_worker.rs
```diff
@@ -0,0 +1,48 @@
...
+use crate::{Error, PREFIX};
+
+#[derive(clap::Args, Debug, Valuable)]
+pub struct Args {
+    pub amount: u128,
+    pub denom: String,
+}
+
+pub async fn run(config: Config, state_path: &Path, args: Args) -> Result<(), Error> {
+    let coin = Coin::new(args.amount, args.denom.as_str()).change_context(Error::Command)?;
```
cgorenflo, line 22: be more specific about why this fails. Seems weird that only the coin parsing fails with "command failed", while the rest shows some other error.

ampd/src/commands/bond_worker.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    let msg = serde_json::to_vec(&ExecuteMsg::BondWorker {
+        service_name: config.service_registry.service_name.clone(),
+    })
+    .expect("bond worker msg should serialize");
+
+    let tx = MsgExecuteContract {
+        sender: pub_key
+            .account_id(PREFIX)
+            .expect("failed to convert to account identifier"),
```
cgorenflo, line 34: I commented on this in the previous command PR already. I don't like this implicit coupling, relying on tofnd to provide the correct key. Two options: either return an error instead of panicking, or create type wrappers so it's guaranteed that account_id will be called with the correct key type

ampd/src/commands/mod.rs
```diff
@@ -1,13 +1,129 @@
...
+    BondWorker(bond_worker::Args),
     /// Query the worker address
     WorkerAddress,
 }
+
+#[derive(Debug, Deserialize, Serialize, PartialEq)]
+pub struct Config {
+    pub cosmwasm_contract: TMAddress,
+    pub service_name: String,
+}
```
cgorenflo, line 43: seeing the config like this I have no idea what it's supposed to represent 😅 
should cosmwasm_contract be `service_registry_contract`?

cgorenflo: Also, it seems like service_name should be an argument instead? A worker can join multiple services, right?

ampd/src/commands/worker_address.rs
```diff
@@ -1,33 +1,14 @@
-use error_stack::{Report, ResultExt};
 use std::path::Path;
 
-use crate::state::{self, StateUpdater};
-use crate::tofnd::grpc::{MultisigClient, SharableEcdsaClient};
+use error_stack::Result;
+
+use crate::commands::worker_pub_key;
 use crate::tofnd::Config;
```
cgorenflo, line 6: don't import the namespace for the `Config`, otherwise it's too easy to confuse with the root `Config` type

ampd/src/commands/worker_address.rs
```diff
@@ -1,33 +1,14 @@
...
-
-            state_updater.as_mut().pub_key = Some(pub_key);
-
-            Ok(pub_key)
-        }
-    }
-    .map(|pub_key| {
+pub async fn run(config: Config, state_path: &Path) -> Result<(), Error> {
+    worker_pub_key(state_path, config).await.map(|pub_key| {
         println!(
```
cgorenflo, line 12: same thing as for `bond-worker`, return the output instead of printing it here

ampd/src/commands/mod.rs
```diff
@@ -1,13 +1,129 @@
...
+
+    let ecdsa_client = SharableEcdsaClient::new(
+        MultisigClient::connect(tofnd_config.party_uid, tofnd_config.url)
+            .await
+            .change_context(Error::Connection)?,
+    );
+
+    broadcaster::BroadcastClientBuilder::default()
+        .client(service_client)
+        .signer(ecdsa_client.clone())
```
cgorenflo, line 103: clone unnecessary

ampd/src/commands/mod.rs
```diff
@@ -1,13 +1,129 @@
...
+        .acc_number(account.account_number)
+        .acc_sequence(account.sequence)
+        .pub_key((tofnd_config.key_uid, pub_key))
+        .config(broadcast.clone())
```
cgorenflo, line 107: clone unnecessary

ampd/src/commands/mod.rs
```diff
@@ -1,13 +1,129 @@
...
+    /// Bond the worker to the service registry contract
+    BondWorker(bond_worker::Args),
     /// Query the worker address
     WorkerAddress,
 }
+
+#[derive(Debug, Deserialize, Serialize, PartialEq)]
+pub struct Config {
+    pub cosmwasm_contract: TMAddress,
+    pub service_name: String,
```
cgorenflo, line 42: nonempty::String?

ampd/src/commands/bond_worker.rs
haiyizxx: will error out for now, prefer to create a type wrapper as we know tofn client returns an ecdsa key, error if it fails to parse

fish-sammy (approved):

ampd/src/lib.rs
```diff
@@ -61,6 +61,7 @@ async fn prepare_app(cfg: Config, state: State) -> Result<App<impl Broadcaster>,
         handlers,
         tofnd_config,
         event_buffer_cap,
+        service_registry: _service_registry,
```
cgorenflo, line 64: use `..` instead

cgorenflo (approved):

PR #209, opened 2023-11-30
update cosmwasm-check to use 1.3.3

fahimahmedx:
## Description
Matches the version of `cosmwasm-check` with `cosmwasm-std`, 1.3.3


haiyizxx:
we are using 1.3.0 in core, let's keep it the same

fish-sammy (approved):

PR #215, opened 2023-12-08
feat(multisig)!: support custom signature verification

cjcobb23:
* add optional callback address to start signing session
* add chain name to start signing session

TODO: still need to actually call into the callback address. Wanted to get this in sooner rather than later so that way we can stabilize the API.

https://axelarnetwork.atlassian.net/browse/AXE-2780

### DIFF
--- a/contracts/connection-router/src/contract.rs
+++ b/contracts/connection-router/src/contract.rs
@@ -1,12 +1,13 @@
 #[cfg(not(feature = "library"))]
 use cosmwasm_std::entry_point;
-use cosmwasm_std::{Binary, Deps, DepsMut, Env, MessageInfo, Response, StdResult};
+use cosmwasm_std::{to_binary, Binary, Deps, DepsMut, Env, MessageInfo, Response};
 
 use crate::events::RouterInstantiated;
 use crate::msg::{ExecuteMsg, InstantiateMsg, QueryMsg};
 use crate::state::{Config, RouterStore, Store};
 
 mod execute;
+mod query;
 
 #[cfg_attr(not(feature = "library"), entry_point)]
 pub fn instantiate(
@@ -99,8 +100,13 @@ where
 }
 
 #[cfg_attr(not(feature = "library"), entry_point)]
-pub fn query(_deps: Deps, _env: Env, _msg: QueryMsg) -> StdResult<Binary> {
-    todo!()
+pub fn query(
+    deps: Deps,
+    _env: Env,
+    msg: QueryMsg,
+) -> Result<Binary, axelar_wasm_std::ContractError> {
+    match msg {
+        QueryMsg::GetChainInfo(chain) => to_binary(&query::get_chain_info(deps, chain)?),
+    }
+    .map_err(axelar_wasm_std::ContractError::from)
 }
-
-pub mod query {}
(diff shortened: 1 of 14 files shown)


contracts/multisig/src/msg.rs
```diff
@@ -21,6 +22,13 @@ pub enum ExecuteMsg {
     StartSigningSession {
         worker_set_id: String,
         msg: HexBinary,
+        chain_name: ChainName,
+        /* Address of a contract responsible for signature verification.
+        The multisig contract verifies each submitted signature by default.
+        But some chains need custom verification beyond this, so the verification can be optionally overridden.
+        If a callback address is provided, signature verification is handled by the contract at that address
+        instead of the multisig contract. TODO: define interface for callback */
+        sig_verification_callback_address: Option<String>,
```
fish-sammy, line 31:
```suggestion
        sig_verifier: Option<String>,
```
This is what I would name it with.

contracts/multisig/src/contract.rs
```diff
@@ -42,13 +42,19 @@ pub fn execute(
...
 ) -> Result<Response, axelar_wasm_std::ContractError> {
     match msg {
-        ExecuteMsg::StartSigningSession { worker_set_id, msg } => {
+        ExecuteMsg::StartSigningSession {
+            worker_set_id,
+            msg,
+            chain_name,
+            sig_verification_callback_address: _, /*TODO: handle callback */
+        } => {
             execute::require_authorized_caller(&deps, info.sender)?;
```
fish-sammy, line 51:
```suggestion
                        sig_verification_callback_address,
        } => {
            execute::require_authorized_caller(&deps, info.sender)?;
            let _sig_verifier =
                match sig_verification_callback_address.map(|addr| deps.api.addr_validate(&addr)) {
                    Some(addr) => Some(addr?),
                    None => None,
                }; /*TODO: handle callback */
```

contracts/multisig/src/events.rs
```diff
@@ -15,6 +16,7 @@ pub enum Event {
         worker_set_id: String,
         pub_keys: HashMap<String, PublicKey>,
         msg: MsgToSign,
+        chain_name: ChainName,
```
fish-sammy, line 19: Why?

contracts/multisig/src/contract.rs
cgorenflo:
```suggestion
            sig_verification_callback_address,
       } => {
            execute::require_authorized_caller(&deps, info.sender)?;
            let _sig_verifier =
              sig_verification_callback_address.map(|addr| deps.api.addr_validate(&addr).ok()).flatten(); // TODO: handle callback
```

contracts/multisig/src/msg.rs
cgorenflo: do we just use string for addresses in other contexts or should this be an address type?

contracts/multisig/src/events.rs
cgorenflo: axelarons needs to know which chain a session was for. This is not necessarily the event they are going to listen to, but I think it's a good addition

cgorenflo (approved):

contracts/multisig/src/msg.rs
cjcobb23: We just use string and then convert to an address type, since addresses need to be validated via `deps.api.addr_validate()` and so we can't use serde to validate the address

contracts/multisig/src/events.rs
cjcobb23: also ampd needs to know, because singing needs to be done differently for certain chains

PR #216, opened 2023-12-08
refactor(multisig-prover)!: use CrossChainID in ConstructProof

cjcobb23:
https://axelarnetwork.atlassian.net/browse/AXE-2781

### DIFF
--- a/contracts/multisig-prover/src/execute.rs
+++ b/contracts/multisig-prover/src/execute.rs
@@ -20,10 +20,16 @@ use crate::{
 pub fn construct_proof(
     deps: DepsMut,
     env: Env,
-    message_ids: Vec<String>,
+    message_ids: Vec<CrossChainId>,
 ) -> Result<Response, ContractError> {
     let config = CONFIG.load(deps.storage)?;
-    let batch_id = BatchID::new(&message_ids, None);
+    let batch_id = BatchID::new(
+        &message_ids
+            .iter()
+            .map(|id| id.to_string())
+            .collect::<Vec<String>>(),
+        None,
+    );
 
     let messages = get_messages(
         deps.querier,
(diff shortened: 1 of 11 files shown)


contracts/multisig-prover/src/execute.rs
```diff
@@ -20,10 +20,16 @@ use crate::{
 pub fn construct_proof(
     deps: DepsMut,
     env: Env,
-    message_ids: Vec<String>,
+    message_ids: Vec<CrossChainId>,
 ) -> Result<Response, ContractError> {
     let config = CONFIG.load(deps.storage)?;
-    let batch_id = BatchID::new(&message_ids, None);
+    let batch_id = BatchID::new(
```
cgorenflo, line 26: can you rename it to `BatchId`?

cgorenflo, line 26: the constructor of BatchId is a bit strange. Why take a slice reference just to clone it into a new vector and sort it? I suggest changing the constructor to accept ownership of `Vec<CrossChainId>`

cgorenflo: furthermore, I don't understand why it accepts a worker set. That case is never used

cjcobb23: I changed the constructor to just accept a slice reference of CrossChainId, and removed the clone. We convert to string right away anyways, so we don't need to take ownership.

The worker set case is actually used by `CommandBatchBuilder::build`

contracts/multisig-prover/src/types.rs
```diff
@@ -35,42 +36,45 @@ pub struct Command {
...
 
-impl BatchID {
-    pub fn new(message_ids: &[String], new_worker_set: Option<WorkerSet>) -> BatchID {
-        let mut message_ids = message_ids.to_vec();
+impl BatchId {
+    pub fn new(message_ids: &[CrossChainId], new_worker_set: Option<WorkerSet>) -> BatchId {
+        let mut message_ids = message_ids
+            .iter()
+            .map(|id| id.to_string())
+            .collect::<Vec<String>>();
```
fish-sammy, line 77:
```suggestion
            .collect::<Vec<_>>();
```

fish-sammy (approved):

PR #218, opened 2023-12-11
test(multisig-prover): create integration test for updating worker set

fahimahmedx:
## Description
[AXE-2760]

Adds an integration test for initializing and manually updating a WorkerSet.

Additional tests for automatically updating a WorkerSet will be added in a separate PR.

[AXE-2760]: https://axelarnetwork.atlassian.net/browse/AXE-2760?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


integration-tests/tests/message_routing.rs
cgorenflo, line 1: can you add a comment that explains what setup_test_case does? It is getting a bit complicated to understand on first glance

integration-tests/tests/test_utils/mod.rs
```diff
@@ -64,7 +65,7 @@ pub fn route_messages(app: &mut App, gateway_address: &Addr, msgs: &[Message]) {
 pub fn vote_true_for_all(
     app: &mut App,
     voting_verifier_address: &Addr,
-    msgs: &Vec<Message>,
+    votes: &Vec<bool>,
```
cgorenflo, line 68: I don't understand this function. If it's supposed to vote true for all messages, why pass in a bool vector?

integration-tests/tests/message_routing.rs
```diff
@@ -140,6 +143,24 @@ fn setup_test_case() -> (Protocol, Chain, Chain, Vec<Worker>) {
...
+            min_num_workers: 0,
+            max_num_workers: Some(100),
+            min_worker_bond: min_worker_bond.clone(),
+            bond_denom: AXL_DENOMINATION.into(),
+            unbonding_period_days: 10,
+            description: "Some service".into(),
+        },
+        &[],
+    );
+    assert!(response.is_ok());
```
cgorenflo, line 162: I like the approach that's taken elsewhere to isolate these simple tasks into functions with easy to understand names (see test_utils::register_workers below). Do that for this call too?

integration-tests/tests/test_utils/mod.rs
fahimahmedx: This is because `update_worker_set` has no messages (its length is 0 as seen in `assert_eq!(proof.message_ids.len(), 0,);` ), and so if I were to keep it as a vector of messages, then previously it would have 0 votes when creating the `votes` vector inside of `vote_true_for_all`.

cjcobb23: you could just create a dummy `Message`. Or write a different function called `vote_true_for_workerset`. It is confusing that the function is called vote true for all, but doesn't actually vote true for all, depending on the value of the `votes` parameter.

integration-tests/tests/update_worker_set.rs
```diff
@@ -0,0 +1,157 @@
...
+
+    let ethereum = test_utils::setup_chain(&mut protocol, chains.get(0).unwrap().clone());
+    (protocol, ethereum, workers, min_worker_bond)
+}
+
+#[test]
+fn worker_set_can_be_initialized_and_then_updated() {
+    let chains: Vec<connection_router::state::ChainName> =
+        vec!["Ethereum".to_string().try_into().unwrap()];
+    let (mut protocol, ethereum, mut workers, min_worker_bond) = setup_test_case(chains.clone());
```
cjcobb23, line 56:
```suggestion
    let (mut protocol, ethereum, mut initial_workers, min_worker_bond) = setup_test_case(chains.clone());
```

cjcobb23: also, why does workers need to be `mut`?

integration-tests/tests/update_worker_set.rs
```diff
@@ -0,0 +1,157 @@
...
+        &mut protocol.app,
+        &ethereum.multisig_prover_address,
+        &session_id,
+    );
+    assert!(matches!(
+        proof.status,
+        multisig_prover::msg::ProofStatus::Completed { .. }
+    ));
+
+    assert_eq!(proof.message_ids.len(), 0,);
```
cjcobb23, line 124:
```suggestion
    assert_eq!(proof.message_ids.len(), 0);
```

integration-tests/tests/update_worker_set.rs
```diff
@@ -0,0 +1,157 @@
...
+
+    let simulated_new_worker_set = test_utils::workers_to_worker_set(&mut protocol, &new_workers); // test_utils::get_worker_set(&mut protocol.app, &ethereum.multisig_prover_address);
```
cjcobb23, line 126:
```suggestion
    let simulated_new_worker_set = test_utils::workers_to_worker_set(&mut protocol, &new_workers);
```

integration-tests/tests/update_worker_set.rs
```diff
@@ -0,0 +1,157 @@
...
+        &mut protocol.app,
+        &ethereum.voting_verifier_address,
+        poll_id,
+    );
+
+    test_utils::confirm_worker_set(
+        &mut protocol.app,
+        Addr::unchecked("relayer"),
+        ethereum.multisig_prover_address.clone(),
+    );
```
cjcobb23, line 156: Should query the prover contract for the current worker set now, and assert that it's what we expect

integration-tests/tests/update_worker_set.rs
```diff
@@ -0,0 +1,121 @@
...
+#[test]
+fn worker_set_can_be_initialized_and_then_updated() {
+    let chains = vec![
+        "Ethereum".to_string().try_into().unwrap(),
+        "Polygon".to_string().try_into().unwrap(),
+    ];
+    let (mut protocol, ethereum, _, initial_workers, min_worker_bond) =
+        test_utils::setup_test_case();
+
+    let simulated_worker_set = test_utils::workers_to_worker_set(&mut protocol, &initial_workers);
```
cjcobb23, line 18:
```suggestion
    let expected_new_worker_set = test_utils::workers_to_worker_set(&mut protocol, &initial_workers);
```

cjcobb23 (approved):

PR #221, opened 2023-12-13
feat: ampd Sui confirm worker set handler

haiyizxx:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


ampd/src/handlers/sui_verify_worker_set.rs
```diff
@@ -0,0 +1,227 @@
...
+    type Err = Error;
+
+    async fn handle(&self, event: &Event) -> error_stack::Result<(), Error> {
+        let PollStartedEvent {
+            contract_address,
+            poll_id,
+            source_gateway_address,
+            worker_set,
+            participants,
+            ..
```
fish-sammy, line 114:
```suggestion
```
All fields are used, no?

ampd/src/sui/verifier.rs
```diff
@@ -18,11 +16,33 @@ struct ContractCall {
...
 }
 
-// Event type is in the form of: <gateway_address>::gateway::ContractCall
-fn call_contract_type(gateway_address: &SuiAddress) -> StructTag {
-    format!("{}::{}", gateway_address, CONTRACT_CALL_EVENT)
-        .parse()
-        .expect("failed to parse struct tag")
+#[derive(Deserialize)]
+struct OperatorshipTransferred {
+    pub payload: Vec<u8>,
```
fish-sammy, line 21: Not related to this PR, but why does the Sui contract's event look like this? If the `OperatorshipTransferred` is done this way, why isn't `ContractCall` done the same? It could have just been a blob of json for literally anything.

fish-sammy (approved):

ampd/src/handlers/sui_verify_worker_set.rs
haiyizxx: nah, the source chain name was not used.

ampd/src/sui/verifier.rs
haiyizxx: yes, it a blob of bytes of contains new pub keys and threshold. idk, can definately be more structured.

PR #224, opened 2023-12-21
feat: add message ids to proof under construction event

cgorenflo:
## Summary by CodeRabbit

- **New Features**
  - Enhanced event tracking with additional fields for destination chain and message IDs.

- **Improvements**
  - Updated event serialization to support new event structure.
  - Improved multisig reply function to include chain name in event construction.

- **Tests**
  - Added tests for event serializability and attribute value serialization.

- **Documentation**
  - Created a new module documentation for event handling.


packages/axelar-wasm-std/src/event.rs
```diff
@@ -0,0 +1,38 @@
...
+use serde_json::to_string;
+
+pub fn attribute_value<T>(v: &T) -> Result<String, serde_json::error::Error>
+where
+    T: ?Sized + Serialize,
+{
+    let json = to_string(v)?;
+    // using strip_* instead of trim_matches because the latter trims any number of quotes instead of just one
+    let json = json.strip_prefix('"').unwrap_or(&json);
+    Ok(json.strip_suffix('"').unwrap_or(json).to_string())
```
fish-sammy, line 11:
```suggestion
    let json = json
        .strip_prefix('"')
        .unwrap_or(&json)
        .strip_suffix('"')
        .unwrap_or(&json);
    Ok(json.to_string())
```
I'm detinitely nitpicking but I feel this is clearer cuz each statement has a very explicit purpose.

contracts/multisig-prover/src/events.rs
```diff
@@ -1,32 +1,76 @@
...
         match other {
             Event::ProofUnderConstruction {
+                destination_chain,
                 command_batch_id,
                 multisig_session_id,
+                msg_ids,
             } => cosmwasm_std::Event::new("proof_under_construction")
+                .add_attribute(
+                    "destination_chain",
+                    event::attribute_value(&destination_chain)
```
fish-sammy, line 27: What I don't like about this is that, there's absolutely nothing there to require people to use it but nothing else.

contracts/multisig-prover/src/events.rs
```diff
@@ -1,32 +1,76 @@
+use crate::types::BatchId;
+use axelar_wasm_std::event;
+use connection_router::state::{ChainName, CrossChainId};
 use cosmwasm_std::Uint64;
 use serde_json::to_string;
```
fish-sammy, line 5:
```suggestion
```

contracts/multisig-prover/src/events.rs
cgorenflo: I'd prefer to catch this with something like dylint instead of requiring us wrapping their Event and basically reimplementing it. Also I'd like to tackle that separately from this PR

fish-sammy (approved):

PR #225, opened 2023-12-22
feat(ampd): add block height monitor

cjcobb23:
Add a struct to the `App` that keeps track of the current block height on chain

### DIFF
--- a/ampd/src/block_height_monitor.rs
+++ b/ampd/src/block_height_monitor.rs
@@ -1,0 +1,148 @@
+use error_stack::{Result, ResultExt};
+use std::{
+    sync::{atomic::AtomicU64, Arc},
+    time::Duration,
+};
+use thiserror::Error;
+use tokio::{select, time};
+use tokio_util::sync::CancellationToken;
+use tracing::info;
+
+use crate::tm_client::TmClient;
+
+#[derive(Clone)]
+pub struct BlockHeightMonitor<T: TmClient + Sync> {
+    latest_height: Arc<AtomicU64>,
+    client: T,
+    poll_interval: Duration,
+}
+
+#[derive(Error, Debug)]
+pub enum BlockHeightMonitorError {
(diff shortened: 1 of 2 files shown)


ampd/src/block_height_monitor.rs
```diff
@@ -0,0 +1,148 @@
...
+                    return Ok(())
+                },
+            }
+        }
+    }
+
+    #[allow(dead_code)]
+    pub fn latest_block_height(&self) -> u64 {
+        self.latest_height.load(std::sync::atomic::Ordering::SeqCst)
+    }
```
fish-sammy, line 66:
```suggestion
    pub fn latest_block_height(&self) -> Arc<AtomicU64> {
        self.latest_height.clone()
    }
```

ampd/src/block_height_monitor.rs
```diff
@@ -0,0 +1,148 @@
...
+    time::Duration,
+};
+use thiserror::Error;
+use tokio::{select, time};
+use tokio_util::sync::CancellationToken;
+use tracing::info;
+
+use crate::tm_client::TmClient;
+
+#[derive(Clone)]
```
fish-sammy, line 13:
```suggestion
```
It doesn't make sense to me to clone the entire monitor so that other threads can read. It doesn't make sense for them to be able to `.run()` again.

ampd/src/lib.rs
```diff
@@ -255,6 +265,7 @@ where
                 completed_height.increment(),
             )),
         };
+        // TODO: pass block_height_monitor to handler
```
fish-sammy, line 268: Just pass the `Arc<AtomicU64>` over, instead of the entire monitor. If you wanna make it nicer than just a `Arc<AtomicU64>`, you can create a wrapper around it.

fish-sammy (changes requested):

fish-sammy (approved):

PR #227, opened 2024-01-03
XRPL Multisig Prover

abresas:
Support for payments (native or with token) using tickets, and updating multisig signers.

- We do not yet support GMP
- Payment amounts are passed as cosmwasm coin amount, yet precision loss needs to be discussed
- We have (temporarily) added an `xrpl_voting_verifier` contract that allows us to differentiate between different message statuses. The generic `voting_verifier` has functionality for supporting this, but doesn't expose the actual status of a message yet.
- We haven't yet implemented rewards which is a new feature that wasn't part of the generic multisig prover when we started the project, and we are not sure if and how they should be implemented.
- This PR requires custom signature verification support from `multisig` contract, which isn't yet supported.
- We have a first implementation of integration tests. However, the integration tests fail, unless signature verification on multisig is disabled (commented out), until custom signature verification is implemented.
- No support for trustlines yet


cjcobb23:
Why do we need a separate voting verifier for XRPL? We added support for succeeded vs failed on chain to the voting verifier, I think that's the only thing you needed for the XRPL case, is that right? https://github.com/axelarnetwork/axelar-amplifier/blob/c0904e1c72c99a174a93b5f6c18b30f52d20b71f/packages/axelar-wasm-std/src/voting.rs#L140

k4m4:
> Why do we need a separate voting verifier for XRPL? We added support for succeeded vs failed on chain to the voting verifier, I think that's the only thing you needed for the XRPL case, is that right?
> 
> https://github.com/axelarnetwork/axelar-amplifier/blob/c0904e1c72c99a174a93b5f6c18b30f52d20b71f/packages/axelar-wasm-std/src/voting.rs#L140

@cjcobb23 We had added status confirmation logic to the voting verifier before the feature you mentioned was implemented, so we had forked it out as a separate project. We recently discussed switching back to the voting verifier, but it looked like the feature we need is still WIP, since we couldn't find a way to query for the message status (`Vote`) that reached consensus (just boolean: https://github.com/axelarnetwork/axelar-amplifier/blob/c0904e1c72c99a174a93b5f6c18b30f52d20b71f/contracts/voting-verifier/src/query.rs#L78).

cjcobb23:
> > Why do we need a separate voting verifier for XRPL? We added support for succeeded vs failed on chain to the voting verifier, I think that's the only thing you needed for the XRPL case, is that right?
> > https://github.com/axelarnetwork/axelar-amplifier/blob/c0904e1c72c99a174a93b5f6c18b30f52d20b71f/packages/axelar-wasm-std/src/voting.rs#L140
> 
> @cjcobb23 We had added status confirmation logic to the voting verifier before the feature you mentioned was implemented, so we had forked it out as a separate project. We recently discussed switching back to the voting verifier, but it looked like the feature we need is still WIP, since we couldn't find a way to query for the message status (`Vote`) that reached consensus (just boolean):
> 
> https://github.com/axelarnetwork/axelar-amplifier/blob/c0904e1c72c99a174a93b5f6c18b30f52d20b71f/contracts/voting-verifier/src/query.rs#L78

I see, yes the status is not yet exposed through query. I think it would be preferable to just expose the status through a query in the voting verifier, rather than write an entirely new contract. We would welcome a PR for the voting verifier to expose the status via query.

eguajardo:
> > > Why do we need a separate voting verifier for XRPL? We added support for succeeded vs failed on chain to the voting verifier, I think that's the only thing you needed for the XRPL case, is that right?
> > > https://github.com/axelarnetwork/axelar-amplifier/blob/c0904e1c72c99a174a93b5f6c18b30f52d20b71f/packages/axelar-wasm-std/src/voting.rs#L140
> > 
> > 
> > @cjcobb23 We had added status confirmation logic to the voting verifier before the feature you mentioned was implemented, so we had forked it out as a separate project. We recently discussed switching back to the voting verifier, but it looked like the feature we need is still WIP, since we couldn't find a way to query for the message status (`Vote`) that reached consensus (just boolean):
> > https://github.com/axelarnetwork/axelar-amplifier/blob/c0904e1c72c99a174a93b5f6c18b30f52d20b71f/contracts/voting-verifier/src/query.rs#L78
> 
> I see, yes the status is not yet exposed through query. I think it would be preferable to just expose the status through a query in the voting verifier, rather than write an entirely new contract. We would welcome a PR for the voting verifier to expose the status via query.

I just started taking a look at this, I'll be working on the query to expose the status

contracts/xrpl-multisig-prover/src/types.rs
```diff
@@ -0,0 +1,132 @@
...
+    FailedOnChain,
+    FailedOffChain,
+}
+
+// TODO: import from verifier
+pub fn parse_message_id(
+    message_id: &nonempty::String,
+) -> Result<(nonempty::String, u64), ContractError> {
+    // expected format: <tx_id>:<index>
+    let components = message_id.split(":").collect::<Vec<_>>();
```
cjcobb23, line 24: This is not a standard format for message ids. This is used by evm chains, but other chains could use different formats. What types of messages is this called on specifically?

k4m4:
Let's close this, since the XRPL-specific smart contracts will live in a Common Prefix fork. We will submit a separate PR with just the XRPL `ampd` components shortly.

PR #228, opened 2024-01-03
refactor: add ref to latest block height to handlers

cjcobb23:


cgorenflo (approved):

cgorenflo (commented):
What I don't like about `Arc` is that it provides shared _ownership_ of the value, i.e. anyone who calls `latest_block_height()` can also set it. I prefer to use the watch channel https://docs.rs/tokio/latest/tokio/sync/index.html#watch-channel. It does exactly what you want in this case

cgorenflo (changes requested):
If resolved before I'm back, feel free to discard my change request

ampd/src/block_height_monitor.rs
```diff
@@ -47,8 +54,8 @@ impl<T: TmClient + Sync> BlockHeightMonitor<T> {
         loop {
             select! {
                 _ = interval.tick() => {
-                    let latest_block = self.client.latest_block().await.change_context(BlockHeightMonitorError::LatestBlockError)?;
-                    self.latest_height.store(latest_block.block.header.height.into(), std::sync::atomic::Ordering::SeqCst);
+                    let latest_block = self.client.latest_block().await.change_context(BlockHeightMonitorError::LatestBlock)?;
+                    self.latest_height_tx.send(latest_block.block.header.height.into()).change_context(BlockHeightMonitorError::Publish)?;
```
fish-sammy, line 58: Does it make more sense to panic here? From the doc, it seems imposibble to fail here since `self` alwast holds an instance of the receiver.

ampd/src/lib.rs
```diff
@@ -238,6 +240,7 @@ where
                         cosmwasm_contract,
                         json_rpc::Client::new_http(&rpc_url).change_context(Error::Connection)?,
                         self.broadcaster.client(),
+                        self.block_height_monitor.latest_block_height(),
                     ),
                 ),
                 handlers::config::Config::SuiWorkerSetVerifier {
```
fish-sammy, line 246: What about this handler?

ampd/src/handlers/sui_verify_msg.rs
```diff
@@ -152,6 +156,8 @@ mod tests {
     use error_stack::{Report, Result};
     use ethers::providers::ProviderError;
     use sui_types::base_types::{SuiAddress, TransactionDigest};
+    use tendermint::abci;
```
fish-sammy, line 159:
```suggestion
```

ampd/src/block_height_monitor.rs
cjcobb23: yeah I think it makes more sense to panic, I changed it

ampd/src/lib.rs
cjcobb23: good catch, missed this in a merge

fish-sammy (approved):

ampd/src/block_height_monitor.rs
```diff
@@ -47,8 +52,8 @@ impl<T: TmClient + Sync> BlockHeightMonitor<T> {
         loop {
             select! {
                 _ = interval.tick() => {
-                    let latest_block = self.client.latest_block().await.change_context(BlockHeightMonitorError::LatestBlockError)?;
-                    self.latest_height.store(latest_block.block.header.height.into(), std::sync::atomic::Ordering::SeqCst);
+                    let latest_block = self.client.latest_block().await.change_context(BlockHeightMonitorError::LatestBlock)?;
+                    self.latest_height_tx.send(latest_block.block.header.height.into()).expect("failed to publish latest block height");
```
cgorenflo, line 56: Can you add a comment that the `expect` is fine because the receiver `latest_height_rx` is never closed and therefore the channel stays open?

PR #232, opened 2024-01-10
Flow verifier

thhachain:
## Description
This is the flow module that connects to Flow's gRPC Access API for getting the result of a transaction by it's transaction id. It then parses the events from the transaction and deserializes the event payload that is in Cadence-JSON into a proper struct for validation.
## Todos

- [X] Unit tests
- [X] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Steps to Test
```cargo test --package ampd --lib -- flow::verifier::tests --nocapture ```

### DIFF
--- a/ampd/build.rs
+++ b/ampd/build.rs
@@ -1,6 +1,9 @@
 fn main() -> Result<(), Box<dyn std::error::Error>> {
     tonic_build::configure()
         .build_server(false)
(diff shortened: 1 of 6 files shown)


cjcobb23:
Flow is integrating with evm layer, so closing this for now

PR #237, opened 2024-01-17
docs: update voting_verifier.md

maancham:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Summary by CodeRabbit

- **Documentation**
	- Enhanced the Voting Verifier documentation with the addition of a Verifier flowchart.
	- Included a new section detailing the reading order for contract code and explanation of Poll entities.
	- Added a Message Verification Sequence Diagram to improve comprehension of the contract's workflow.


doc/src/contracts/voting_verifier.md
```diff
@@ -16,29 +18,69 @@ subgraph Axelar
...
+   pub struct Poll {
+        poll_id: PollId,
+        messages: Vec<Message>,
+    }
+   ```
+
+
+
+
+
```
cjcobb23, line 70: I don't think comments about the code should be in this doc. I think this doc should be higher level, and the only code mentioned here should be code from the API. How it's actually implemented I think shouldn't be described here. I think comments like this are more appropriate for the source code.

cjcobb23 (changes requested):

doc/src/contracts/voting_verifier.md
maancham: Fixed here: https://github.com/axelarnetwork/axelar-amplifier/pull/237/commits/f0695dcaf709ab3c1327043adea7271fc7a10707

doc/src/contracts/voting_verifier.md
```diff
@@ -1,13 +1,22 @@
...
+Poll ID to the voting verifier. The voting verifier internally maps
+a Poll ID to the messages in the Poll, to be able to call back to
 the verifier and propagate the result back to the gateway.
 
+When dealing with gateways deployed on external chains,
+there is a requirement to inform them about the workers from whom they can accept signatures. While the `service-registry`
+manages this process for the Axelar chain, an alternative mechanism is needed for external chains.
+The internal Poll enum handles this need. It is employed whenever there is a necessity to update the set of workers—whether it involves adding a 
+new worker or removing an existing one. In such cases, a message is generated, and this message must be relayed to the external gateway. It is also 
+important to confirm that the external chain has received the message to finalize the process.
```
cjcobb23, line 14:
```suggestion
There are two types of polls: messages polls and worker set polls. Messages polls are used to verify incoming messages, while worker set polls are used to verify that the external gateway has updated it's stored worker set. Worker set polls are a necessary component of the worker set update flow. See [link to the worker set update diagram in the prover] for more details.
```

I feel like this is clearer. Most of the description of a worker set update already lives in the prover documentation, so I would just link there. Here you only really need to say that there are two types of polls. Also, I would not mention the internal enum, as that could change at any time, because it's an implementation detail.

cjcobb23 (approved):

PR #247, opened 2024-01-29
feat: axe-2791 first step, add denounce chain support to service registry

maancham:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

### DIFF
--- a/contracts/service-registry/src/contract.rs
+++ b/contracts/service-registry/src/contract.rs
@@ -94,14 +94,14 @@ pub fn execute(
                 AuthorizationState::NotAuthorized,
             )
         }
-        ExecuteMsg::DeclareChainSupport {
+        ExecuteMsg::RegisterChainSupport {
             service_name,
             chains,
-        } => execute::declare_chains_support(deps, info, service_name, chains),
-        ExecuteMsg::DenounceChainSupport {
+        } => execute::register_chains_support(deps, info, service_name, chains),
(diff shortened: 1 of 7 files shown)


contracts/service-registry/tests/tests.rs
cgorenflo, line 1: good coverage, I'd add two more cases though:
- try to denounce a chain for a worker that's not part of the service
- try to denounce a chain for a service that doesn't exist

It would also be good to add description to the tests of what you expect to happen. If that's too long for the test names add a comment that explains it. With just a bunch of `assert!(res.is_ok());` scattered about the test it's hard to know what you are actually testing.

As you mentioned in our previous discussion, the tests can get cleaned up a bit to make them more readable. Can be done in a separate PR though

maancham: Addressed in the last two commits

contracts/service-registry/src/contract.rs
```diff
@@ -98,6 +98,10 @@ pub fn execute(
             service_name,
             chains,
         } => execute::declare_chains_support(deps, info, service_name, chains),
+        ExecuteMsg::DenounceChainSupport {
```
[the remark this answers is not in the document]
maancham: Fixed in the last commit

PR #249, opened 2024-01-31
feat: add signature to public key registration

eguajardo:
## Description

Adds the signed worker address to the public key registration, to verify only the owner is registering the key

https://axelarnetwork.atlassian.net/browse/AXE-2637

## Todos

- [ ] Unit tests
- [x] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


contracts/multisig/src/error.rs
```diff
@@ -22,6 +22,9 @@ pub enum ContractError {
     #[error("{signer:?} submitted an invalid signature for signing session {session_id:?}")]
     InvalidSignature { session_id: Uint64, signer: String },
 
+    #[error("signed sender address could not be verified using submitted public key")]
+    InvalidPublicKeyRegistrationSignature,
```
cjcobb23, line 26: Do we need a new error type here? We can probably just reuse `InvalidSignature` above.

contracts/multisig/src/msg.rs
```diff
@@ -39,6 +39,7 @@ pub enum ExecuteMsg {
     },
     RegisterPublicKey {
         public_key: PublicKey,
+        signed_sender_address: HexBinary,
```
cjcobb23, line 42: Probably should add a comment about what this is and why it's needed.

cjcobb23 (approved):

contracts/multisig/src/error.rs
eguajardo: that was my original thought, but the error above requires session_id which is not available for public key registration

PR #251, opened 2024-02-01
feat(ampd): retry event handling with timeout and max attemmpts

fish-sammy:
## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


ampd/src/event_processor.rs
```diff
@@ -60,7 +60,15 @@ where
...
 
         if let StreamStatus::Active(event) = &stream_status {
-            handler.handle(event).await.change_context(Error::Handler)?;
+            future::with_retry(
+                || handler.handle(event),
+                // TODO: make timeout and max_attempts configurable
+                RetryPolicy::RepeatConstant(Duration::from_secs(1), 3),
+            )
+            .await
+            .change_context(Error::Handler)?;
```
cjcobb23, line 69: Don't we need to add this for the other handlers?

ampd/src/event_processor.rs
```diff
@@ -60,7 +60,15 @@ where
...
+
+            // handler.handle(event).await.change_context(Error::Handler)?;
```
cjcobb23, line 71: remove

ampd/src/event_processor.rs
cjcobb23: Also, I am not sure that we actually want this behavior for multisig handler, because the multisig handler is only talking to tofnd. Maybe we don't want to just ignore in this case?

cjcobb23: oh wait, disregard, we aren't ignoring, I missed the `?`

cjcobb23: Ok sorry, I realized I read this wrong initially. So this retry logic is applied to all handlers. But it doesn't like look we are ignoring failures in the case of the evm_verify_msg right? Don't we want to try 3 times, and if we still can't succeed, just continue on? Instead of failing the whole task.

ampd/src/asyncutil/future.rs
```diff
@@ -0,0 +1,175 @@
...
+    policy: RetryPolicy,
+    err_count: u64,
+}
+
+impl<F, Fut, R, Err> Unpin for FutureRetry<F, Fut, R, Err>
+where
+    F: Fn() -> Fut,
+    Fut: Future<Output = Result<R, Err>>,
+{
+}
```
cgorenflo, line 39: why is this necessary?

ampd/src/asyncutil/future.rs
```diff
@@ -0,0 +1,175 @@
...
+
+#[cfg(test)]
+mod tests {
+    use std::{sync::Mutex, time::Instant};
+
+    use super::*;
+
+    #[tokio::test]
+    async fn should_return_ok_when_the_inter_future_returns_ok_immediately() {
```
cgorenflo, line 106:
```suggestion
    async fn should_return_ok_when_the_internal_future_returns_ok_immediately() {
```

ampd/src/asyncutil/future.rs
```diff
@@ -0,0 +1,175 @@
...
+        let fut = with_retry(
+            || async { Ok::<(), ()>(()) },
```
cgorenflo, line 108:
```suggestion
            std::future::ready(Ok(())),
```

ampd/src/asyncutil/future.rs
```diff
@@ -0,0 +1,175 @@
...
+            RetryPolicy::RepeatConstant(Duration::from_secs(1), 3),
+        );
+        let start = Instant::now();
+
+        assert!(fut.await.is_ok());
+        assert!(start.elapsed().as_secs() < 1);
```
cgorenflo, line 114:
```suggestion
        assert!(start.elapsed() < Duration::from_secs(1));
```
it doesn't make much of a difference here, but I'm uncomfortable with the `as_secs()` rounding

ampd/src/asyncutil/future.rs
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    #[tokio::test]
+    async fn should_return_ok_when_the_inter_future_returns_ok_eventually() {
```
cgorenflo, line 118:
```suggestion
    async fn should_return_ok_when_the_internal_future_returns_ok_eventually() {
```

ampd/src/asyncutil/future.rs
```diff
@@ -0,0 +1,175 @@
...
+        let max_attempts = 3;
+        let count = Mutex::new(0);
+        let fut = with_retry(
+            || async {
+                *count.lock().unwrap() += 1;
+                sleep(Duration::from_secs(1)).await;
+
+                if *count.lock().unwrap() < max_attempts - 1 {
```
cgorenflo, line 126: won't this still be locked?

ampd/src/asyncutil/future.rs
```diff
@@ -0,0 +1,175 @@
...
+                }
+            },
+            RetryPolicy::RepeatConstant(Duration::from_secs(1), max_attempts),
+        );
+        let start = Instant::now();
+
+        assert!(fut.await.is_ok());
+        assert!(start.elapsed().as_secs() >= 3);
+        assert!(start.elapsed().as_secs() < 4);
+    }
```
cgorenflo, line 139: can you make this test a) deterministic in its execution order and b) not take multiple seconds?
Check out `std::sync::Condvar` for conditional blocking on mutexes

ampd/src/asyncutil/future.rs
```diff
@@ -0,0 +1,175 @@
...
+
+    #[tokio::test]
+    async fn should_return_error_when_the_inter_future_returns_error_after_max_attempts() {
```
cgorenflo, line 142:
```suggestion
    async fn should_return_error_when_the_internal_future_returns_error_after_max_attempts() {
```

ampd/src/asyncutil/future.rs
```diff
@@ -0,0 +1,175 @@
...
+        let fut = with_retry(
+            || async { Err::<(), ()>(()) },
```
cgorenflo, line 144:
```suggestion
         std::future::ready(Err(())),
```

ampd/src/asyncutil/future.rs
```diff
@@ -0,0 +1,175 @@
...
+            RetryPolicy::RepeatConstant(Duration::from_secs(1), 3),
+        );
+        let start = Instant::now();
+
+        assert!(fut.await.is_err());
+        assert!(start.elapsed().as_secs() >= 2);
```
cgorenflo, line 150:
```suggestion
        assert!(start.elapsed() > Duration::from_secs(2));
```

cgorenflo, line 150: make the times shorter and keep track of a counter variable to ensure the Future actually failed 3 times?

ampd/src/asyncutil/future.rs
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    #[tokio::test]
+    async fn should_return_ok_when_the_inter_future_returns_ok_within_max_attempts() {
```
cgorenflo, line 154:
```suggestion
    async fn should_return_ok_when_the_internal_future_returns_ok_within_max_attempts() {
```

cgorenflo: how is this different from `should_return_ok_when_the_inter_future_returns_ok_eventually`?

ampd/src/asyncutil/future.rs
cgorenflo: or is the lock immediately released in the same line?

ampd/src/asyncutil/future.rs
```diff
@@ -0,0 +1,175 @@
...
+                    Ok::<(), ()>(())
+                }
+            },
+            RetryPolicy::RepeatConstant(Duration::from_secs(1), max_attempts),
+        );
+        let start = Instant::now();
+
+        assert!(fut.await.is_ok());
+        assert!(start.elapsed().as_secs() >= 3);
+        assert!(start.elapsed().as_secs() < 4);
```
cgorenflo, line 138: same comment as before

ampd/src/asyncutil/future.rs
```diff
@@ -0,0 +1,175 @@
...
+                    Ok::<(), ()>(())
+                }
+            },
+            RetryPolicy::RepeatConstant(Duration::from_secs(1), max_attempts),
+        );
+        let start = Instant::now();
+
+        assert!(fut.await.is_ok());
+        assert!(start.elapsed().as_secs() >= 2);
+        assert!(start.elapsed().as_secs() < 3);
```
cgorenflo, line 173: same comment as before

ampd/src/event_processor.rs
cgorenflo: which failures are you talking about?

cjcobb23: any failure really, but specifically rpc error, Wasn't that the requirement? If we try to verify a msg, and encounter an rpc failure, try again. If we try 3 times and still don't succeed, just skip.

ampd/src/asyncutil/future.rs
```diff
@@ -0,0 +1,175 @@
+use futures::{Future, FutureExt};
+use std::pin::Pin;
+use std::task::{Context, Poll};
+use std::time::Duration;
+
+use tokio::time::sleep;
+
+pub fn with_retry<F, Fut, R, Err>(
+    get_future: F,
```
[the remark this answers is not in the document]
fish-sammy: `future_fn`? There's a `future` field in the struct already.

ampd/src/asyncutil/future.rs
fish-sammy: It's unlocked immediately when the `MutexGuard`, which is returned by `.lock()`, goes out of scope.

ampd/src/asyncutil/future.rs
fish-sammy: Let me try to explain this based. on my understanding. At lines [62](https://github.com/axelarnetwork/axelar-amplifier/blob/7e792b785eb713fd04eb761cadd4f964e0109dab/ampd/src/asyncutil/future.rs#L62), [73](https://github.com/axelarnetwork/axelar-amplifier/blob/7e792b785eb713fd04eb761cadd4f964e0109dab/ampd/src/asyncutil/future.rs#L73), and [94](https://github.com/axelarnetwork/axelar-amplifier/blob/7e792b785eb713fd04eb761cadd4f964e0109dab/ampd/src/asyncutil/future.rs#L94), I am trying to mutate the future which is of type `Pin<&mut Self>`.

ampd/src/asyncutil/future.rs
fish-sammy: good catch

ampd/src/asyncutil/future.rs
fish-sammy: Fixed it with `tokio::time::pause`

packages/events/src/event.rs
```diff
@@ -19,6 +21,24 @@ pub enum Event {
...
+            Event::BlockBegin(height) => write!(f, "BlockBegin({})", height),
+            Event::BlockEnd(height) => write!(f, "BlockEnd({})", height),
+            Event::Abci {
+                event_type,
+                attributes,
+            } => write!(
+                f,
+                "Abci {{ event_type: {}, attributes: {} }}",
+                event_type,
+                serde_json::to_string(attributes).expect("must serialize event attributes")
```
cgorenflo, line 36:
```suggestion
                serde_json::to_string(attributes).expect("event attributes must be serializable")
```

ampd/src/asyncutil/future.rs
```diff
@@ -0,0 +1,166 @@
...
+
+        assert!(fut.await.is_ok());
+        assert!(start.elapsed() < Duration::from_secs(1));
+    }
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn should_return_ok_when_the_internal_future_returns_ok_eventually() {
+        let max_attempts = 3;
+        let count = Mutex::new(0);
+        let fut = with_retry(
```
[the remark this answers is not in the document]
cgorenflo: not sure which is cleaner

ampd/src/event_processor.rs
```diff
@@ -60,7 +59,20 @@ where
...
+                    sleep: Duration::from_secs(1),
+                    max_attempts: 3,
+                },
+            )
+            .await
+            .or_else(|err| {
+                warn!("handler failed to process event {}: {}", event, err);
+
+                Ok::<(), ()>(())
+            });
```
cjcobb23, line 75: It occurred to me that there are probably some errors that we don't want to just ignore, like if signing fails. I'm happy with this for now, but can we add a TODO comment saying we need to find a way to ignore some errors but unwind on others?

ampd/src/asyncutil/future.rs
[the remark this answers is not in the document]
fish-sammy: ya I wouldn't say it makes much of a difference and I am ok with tests being a little tedious.

ampd/src/asyncutil/future.rs
```diff
@@ -0,0 +1,166 @@
...
+
+#[cfg(test)]
+mod tests {
+    use std::{future, sync::Mutex};
+
+    use tokio::time::Instant;
+
+    use super::*;
+
+    #[tokio::test(flavor = "current_thread")]
```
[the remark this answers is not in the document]
fish-sammy: ah I forgot that it's the default behaviour.

ampd/src/event_processor.rs
fish-sammy: What is that signing failure?

cjcobb23: Just any failure within the multisig handler

cjcobb23 (approved):

ampd/src/event_processor.rs
cgorenflo: since this is a new design decision and not a shortcoming of the code that might not even be handled any time soon (or ever), I'd prefer not to put a todo in the code. If you want to point attention to it, I'd rather have a comment on line 65 along the line of `// if handlers run into errors we log them and then move on to the next event`

cgorenflo (approved):

PR #255, opened 2024-02-02
feat(multisig): prevent duplicate public key registration

eguajardo:
## Description

https://axelarnetwork.atlassian.net/browse/AXE-2638

## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

### DIFF
--- a/contracts/multisig/src/contract/execute.rs
+++ b/contracts/multisig/src/contract/execute.rs
@@ -131,12 +131,7 @@ pub fn register_pub_key(
         return Err(ContractError::InvalidPublicKeyRegistrationSignature);
     }
 
-    save_pub_key(
-        deps.storage,
-        info.sender.clone(),
-        public_key.key_type(),
-        &public_key.clone().into(),
-    )?;
+    save_pub_key(deps.storage, info.sender.clone(), public_key.clone())?;
 
     Ok(Response::new().add_event(
         Event::PublicKeyRegistered {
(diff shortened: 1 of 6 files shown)


contracts/multisig/src/state.rs
```diff
@@ -72,6 +72,22 @@ pub fn get_worker_set(
 // key type is part of the key so signers can register multiple keys with different types
 pub const PUB_KEYS: Map<(Addr, KeyType), HexBinary> = Map::new("registered_pub_keys");
 
+pub const UNIQUE_PUB_KEYS: Map<&[u8], ()> = Map::new("unique_pub_keys");
```
cjcobb23, line 75: I think we can just use an index for this

contracts/multisig/src/contract/execute.rs
```diff
@@ -131,9 +131,10 @@ pub fn register_pub_key(
         return Err(ContractError::InvalidPublicKeyRegistrationSignature);
     }
 
-    PUB_KEYS.save(
+    save_pub_key(
         deps.storage,
-        (info.sender.clone(), public_key.key_type()),
+        info.sender.clone(),
+        public_key.key_type(),
         &public_key.clone().into(),
```
cjcobb23, line 138: Why don't we just pass the public key as one parameter, and extract the key_type and call into() within `save_pub_key`?

contracts/multisig/src/error.rs
```diff
@@ -28,6 +28,9 @@ pub enum ContractError {
     #[error("invalid public key format: {reason:?}")]
     InvalidPublicKeyFormat { reason: String },
 
+    #[error("public key is already registered")]
+    DuplicatePublicKey,
```
cjcobb23, line 32: Might be useful to return the worker address that registered the public key originally

cjcobb23 (approved):

PR #259, opened 2024-02-07
fix(ampd): skip events from unknown contracts

cjcobb23:
The handlers currently return an error when encountering an event with the same event_type as what they are looking for, but with different fields. This can happen due to contract upgrades, but also if any random contract emits an event with the same event_type as the event the handler cares about. This can cause either crashing the process, or unnecessary retries (and waiting).

This PR makes a change so that the handlers first check the "_contract_address" field of the event, and simply skip the event (returning `Ok(())` ) if the address doesn't match the contract address for which the handler is configured to handle events.

https://axelarnetwork.atlassian.net/browse/AXE-3080

### DIFF
--- a/ampd/src/handlers/multisig.rs
+++ b/ampd/src/handlers/multisig.rs
@@ -28,8 +28,6 @@ use crate::types::TMAddress;
 #[derive(Debug, Deserialize)]
 #[try_from("wasm-signing_started")]
 struct SigningStartedEvent {
-    #[serde(rename = "_contract_address")]
-    contract_address: TMAddress,
     session_id: u64,
     #[serde(deserialize_with = "deserialize_public_keys")]
     pub_keys: HashMap<TMAddress, PublicKey>,
@@ -132,8 +130,11 @@ where
     type Err = Error;
 
     async fn handle(&self, event: &events::Event) -> error_stack::Result<(), Error> {
+        if !events::event_is_from_contract(event, self.multisig.as_ref()) {
+            return Ok(());
+        }
+
         let SigningStartedEvent {
-            contract_address,
             session_id,
             pub_keys,
             msg,
@@ -144,10 +145,6 @@ where
             }
             result => result.change_context(DeserializeEvent)?,
         };
-
-        if self.multisig != contract_address {
-            return Ok(());
-        }
 
         info!(
             session_id = session_id,
@@ -278,6 +275,44 @@ mod test {
         .unwrap()
     }
 
+    // this returns an event that is named SigningStarted,
+    // but the fields are different than what the handler expects
+    fn wrong_signing_started_event() -> events::Event {
+        let pub_keys = (0..10)
+            .map(|_| (rand_account().to_string(), rand_public_key()))
+            .collect::<HashMap<String, PublicKey>>();
+
+        let poll_started = SigningStarted {
+            session_id: Uint64::one(),
+            worker_set_id: "worker_set_id".to_string(),
+            pub_keys,
+            msg: MsgToSign::unchecked(rand_message()),
+            chain_name: rand_chain_name(),
+            expires_at: 100u64,
+        };
+
+        let mut event: cosmwasm_std::Event = poll_started.into();
+        event.ty = format!("wasm-{}", event.ty);
+        event = event.add_attribute("_contract_address", MULTISIG_ADDRESS);
+        let idx = event
+            .attributes
+            .iter()
+            .position(|element| element.key == "expires_at")
+            .unwrap();
+        event.attributes.remove(idx);
+
+        events::Event::try_from(abci::Event::new(
+            event.ty,
+            event
+                .attributes
+                .into_iter()
+                .map(|cosmwasm_std::Attribute { key, value }| {
+                    (STANDARD.encode(key), STANDARD.encode(value))
+                }),
+        ))
+        .unwrap()
+    }
+
     fn get_handler(
         worker: TMAddress,
         multisig: TMAddress,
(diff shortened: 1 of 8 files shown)


ampd/src/handlers/multisig.rs
```diff
@@ -278,6 +275,44 @@ mod test {
         .unwrap()
     }
 
+    // this returns an event that is named SigningStarted,
+    // but the fields are different than what the handler expects
+    fn wrong_signing_started_event() -> events::Event {
```
cgorenflo, line 280:
```suggestion
    fn signing_started_event_with_wrong_fields() -> events::Event {
```

ampd/src/handlers/multisig.rs
```diff
@@ -278,6 +275,44 @@ mod test {
...
+            session_id: Uint64::one(),
+            worker_set_id: "worker_set_id".to_string(),
+            pub_keys,
+            msg: MsgToSign::unchecked(rand_message()),
+            chain_name: rand_chain_name(),
+            expires_at: 100u64,
+        };
+
+        let mut event: cosmwasm_std::Event = poll_started.into();
+        event.ty = format!("wasm-{}", event.ty);
```
cgorenflo, line 295: why is this not part of the type conversion?

ampd/src/handlers/multisig.rs
```diff
@@ -278,6 +275,44 @@ mod test {
...
+        event = event.add_attribute("_contract_address", MULTISIG_ADDRESS);
+        let idx = event
+            .attributes
+            .iter()
+            .position(|element| element.key == "expires_at")
+            .unwrap();
+        event.attributes.remove(idx);
```
cgorenflo, line 302: - there must be a simpler way to remove a specific element
- this doesn't look like an event with wrong fields, but an event with a missing field, or do I see that wrong?

packages/events/src/event.rs
```diff
@@ -71,3 +73,97 @@ fn decode_event_attribute(attribute: &EventAttribute) -> Result<(String, String)
...
+        } => {
+            if let Some(address) = attributes.get("_contract_address") {
+                return serde_json::from_value::<AccountId>(address.clone()).ok();
+            }
+            None
+        }
+        _ => None,
+    };
+    contract_address
+}
```
cgorenflo, line 91:
```suggestion
pub fn contract_address(event: &Event) -> Option<AccountId> {
    match event {
        Event::Abci {
            event_type: _,
            attributes,
        } => attributes
            .get("_contract_address")
            .and_then(|address| serde_json::from_value::<AccountId>(address.clone()).ok()),
        _ => None,
    }
}
}
```
rust convention is to not use `get_` for getters and instead just call out the thing you're getting

packages/events/src/event.rs
```diff
@@ -71,3 +73,97 @@ fn decode_event_attribute(attribute: &EventAttribute) -> Result<(String, String)
...
+    #[test]
+    fn should_get_contract_address_if_exists() {
+        let expected_contract_address =
+            AccountId::from_str("axelarvaloper1zh9wrak6ke4n6fclj5e8yk397czv430ygs5jz7").unwrap();
+
+        let event = make_event_with_contract_address(&expected_contract_address);
+
+        let contract_address = get_contract_address(&event);
+        assert!(contract_address.is_some());
+        assert_eq!(contract_address.unwrap(), expected_contract_address);
```
cgorenflo, line 129:
```suggestion
        assert_eq!(contract_address, Some(expected_contract_address));
```

packages/events/src/event.rs
```diff
@@ -71,3 +73,97 @@ fn decode_event_attribute(attribute: &EventAttribute) -> Result<(String, String)
 fn base64_to_utf8(base64_str: &str) -> std::result::Result<String, DecodingError> {
     Ok(STANDARD.decode(base64_str)?.then(String::from_utf8)?)
 }
+
+pub fn get_contract_address(event: &Event) -> Option<AccountId> {
```
cgorenflo, line 77: does this even need to be public?

ampd/src/handlers/multisig.rs
```diff
@@ -353,7 +388,44 @@ mod test {
...
+            TMAddress::from(MULTISIG_ADDRESS.parse::<AccountId>().unwrap()),
+            SharableEcdsaClient::new(client),
+            100u64,
+        );
+
+        assert!(handler
+            .handle(&wrong_signing_started_event())
+            .await
+            .is_err());
+    }
```
cgorenflo, line 425: looking at these tests I do not understand what is being tested if I don't drill into the code. For example, the client always returns an error, is that what the assert catches? With the `MULTISIG_ADDRESS` const call buried in `wrong_signing_started_event` I have no way of knowing that it's actually matching the address you give to the handler

ampd/src/handlers/multisig.rs
cjcobb23: yes it's just a missing field. I still consider this an event with wrong fields, but it can be renamed. This is the actual scenario I was running into: I deployed a new version of the contract that added a field, and events emitted from the old contract were crashing the handler.

ampd/src/handlers/multisig.rs
cjcobb23: Not sure. Somewhere along the way this is added automatically, but not totally sure at what step.

ampd/src/handlers/multisig.rs
cjcobb23: tried to clean this up and make it more clear

packages/events/src/event.rs
```diff
@@ -71,3 +73,92 @@ fn decode_event_attribute(attribute: &EventAttribute) -> Result<(String, String)
...
+        _ => None,
+    }
+}
+
+pub fn event_is_from_contract(event: &Event, from_address: &AccountId) -> bool {
+    match contract_address(event) {
+        Some(emitting_address) => &emitting_address == from_address,
+        _ => false,
+    }
+}
```
fish-sammy, line 94: I prefer having these two as methods on `Event`, i.e. `event.contract_address()` and `event.is_from_contract(address)`.

fish-sammy (approved):

ampd/src/handlers/multisig.rs
cgorenflo: yeah I'd prefer a more specific name

PR #260, opened 2024-02-08
refactor: isolate mutability and emit missing events

cgorenflo:


contracts/gateway/src/contract/execute.rs
```diff
@@ -1,143 +1,175 @@
...
-
-        let any_verified = !verified.is_empty();
+        match partitioned_msgs {
+            None => Ok(Response::new()),
+            Some((verified, unverified)) => self.route_verified_messages(verified, unverified),
+        }
+    }
 
-        let mut response = Response::new()
+    fn verify_unverified_messages(
```
eguajardo, line 104: could we extract the events part out from this function or at least rename the function? it seems weird to me that the function name references unverified messages and still accepts both verified and unverified as parameters. Same with `route_verified_messages`

contracts/gateway/src/router.rs
```diff
@@ -0,0 +1,20 @@
+use crate::error::ContractError;
+use cosmwasm_std::{to_binary, Addr, WasmMsg};
+use error_stack::ResultExt;
+
+pub struct RouterApi {
+    pub address: Addr,
+}
+
+impl RouterApi {
```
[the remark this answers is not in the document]
cgorenflo: if we start moving it in other contracts we can move it into a common area, for now I think it's fine to leave here

contracts/gateway/src/contract/execute.rs
cgorenflo: I'm open to suggestions for how to do it/alternative names 🙂

cgorenflo: let me know how you like this modification

packages/axelar-wasm-std/src/utils.rs
```diff
@@ -62,3 +62,13 @@ impl<T, E> InspectorResult<T, E> for Result<T, E> {
         })
     }
 }
+
+pub trait TryMapExt<T, B, E> {
```
fish-sammy, line 66: Why not just `OptionExt` so that more methods can be added later to the same trait?

contracts/gateway/src/contract.rs
```diff
@@ -1,54 +1,125 @@
...
+    SerializeResponse,
+    #[error("batch contains duplicate message ids")]
+    DuplicateMessageIds,
+    #[error("could not query the verifier contract")]
+    QueryVerifier,
+    #[error("invalid address")]
+    InvalidAddress,
+}
+
+mod internal {
```
fish-sammy, line 56: What's the benefit of having this `internal` mod?

contracts/gateway/src/contract.rs
```diff
@@ -1,54 +1,125 @@
...
+    }
+
+    pub(crate) fn execute(
+        deps: DepsMut,
+        _env: Env,
+        info: MessageInfo,
+        msg: ExecuteMsg,
+    ) -> Result<Response, Error> {
+        let config = state::load_config(deps.storage).change_context(Error::ConfigMissing)?;
+        let verifier = Verifier {
```
fish-sammy, line 97: I'd still prefer having a `new` function just for aesthetic.

contracts/gateway/src/verifier.rs
```diff
@@ -0,0 +1,109 @@
...
+            .change_context(Error::QueryVerifier)
+    }
+
+    fn execute(&self, msg: &aggregate_verifier::msg::ExecuteMsg) -> WasmMsg {
+        WasmMsg::Execute {
+            contract_addr: self.address.to_string(),
+            msg: to_binary(msg).expect("msg should always be serializable"),
+            funds: vec![],
+        }
+    }
```
fish-sammy, line 34: Wouldn't it be better to define these 2 methods in the `aggregate_verifier` itself?

contracts/gateway/src/verifier.rs
```diff
@@ -0,0 +1,109 @@
...
+                messages: msgs.clone(),
+            })?;
+
+        let status_by_id = response.into_iter().collect::<HashMap<_, _>>();
+
+        Ok(msgs.into_iter().map(move |msg| {
+            let status = status_by_id
+                .get(&msg.cc_id)
+                .copied()
+                .unwrap_or(VerificationStatus::None);
```
fish-sammy, line 51: What's the difference between `VerificationStatus::NotFound` and `VerificationStatus::None`?

contracts/gateway/src/contract/execute.rs
```diff
@@ -1,158 +1,144 @@
...
+use connection_router::state::Message;
+use cosmwasm_std::{Event, Response, Storage, WasmMsg};
 use error_stack::{Result, ResultExt};
 use itertools::Itertools;
 
-use crate::contract::query;
-use crate::contract::query::Verifier;
-use crate::error::ContractError;
-use connection_router::state::Message;
+pub fn verify_messages(verifier: &Verifier, msgs: Vec<Message>) -> Result<Response, Error> {
```
fish-sammy, line 13: I feel like there has been some back-and-forth regarding to the execute mod implementation with this PR now. Won't be able to mock store access anymore with this PR, which is the strength of the previous implementation?

contracts/gateway/src/contract/execute.rs
```diff
@@ -1,158 +1,144 @@
...
+        .map(|(msg, status)| (status, msg))
+        .into_group_map()
+        .into_iter()
+}
 
-        Ok(Response::new().add_events(
-            msgs.into_iter()
-                .map(|msg| GatewayEvent::MessageRouted { msg }.into()),
-        ))
+fn do_verification(
```
fish-sammy, line 72:
```suggestion
fn verify(
```

contracts/gateway/src/contract/execute.rs
```diff
@@ -1,158 +1,144 @@
...
-                    .change_context(ContractError::CreateRouterExecuteMsg)?,
-                funds: vec![],
-            })
-        }
+fn messages_to_events(msgs: Vec<Message>, transform: fn(Message) -> GatewayEvent) -> Vec<Event> {
+    msgs.into_iter().map(|msg| transform(msg).into()).collect()
+}
 
-        Ok(response)
+fn do_routing(
```
fish-sammy, line 103:
```suggestion
fn route(
```

packages/axelar-wasm-std/src/utils.rs
cgorenflo: I'm undecided if I'd rather keep those in separate traits or collect them all, so I'm deferring that decision to the future for now

contracts/gateway/src/contract.rs
cgorenflo: implicit conversion of `Error` to `ContractError` with the help of `?`

contracts/gateway/src/verifier.rs
cgorenflo: yep, moved

contracts/gateway/src/verifier.rs
cgorenflo: `None` means it's never been verified, `NotFound` means that it wasn't found on the source chain. We need to rename that in a separate PR

contracts/gateway/src/contract/execute.rs
cgorenflo: resolved during code walkthrough, mocking is still possible

haiyizxx (approved):

PR #262, opened 2024-02-09
chore: fix compiler warnings

cgorenflo:


ampd/Cargo.toml
```diff
@@ -1,7 +1,9 @@
 [package]
-edition = "2018"
+edition = "2021"
```
milapsheth, line 2: do we still need to specify edition if we're inheriting workspace setting? needed to publish/compile it individually?

milapsheth (approved):

ampd/Cargo.toml
cgorenflo: we can probably define it globally in the workspace and have them all inherit it. Feel free to make that change 🙂

PR #267, opened 2024-02-12
feat: add contract abstraction to connection router

maancham:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


integration-tests/src/contract.rs
```diff
@@ -0,0 +1,56 @@
+use cosmwasm_std::{Addr, Coin};
+use cw_multi_test::{App, AppResponse, Executor};
+use error_stack::{report, Result};
+use serde::de::DeserializeOwned;
+use serde::Serialize;
+
+pub trait Contract {
+    type QMsg;
+    type ExMsg;
+    type Err;
```
cgorenflo, line 10: can you check if all contracts use axelar_wasm_std::ContractError as output? In that case we can bake that type directly into the trait instead of making it an associate type

contracts/connection-router/tests/test.rs
```diff
@@ -62,18 +57,26 @@ fn make_chain(name: &str, config: &mut TestConfig) -> Chain {
...
-        )
-        .unwrap();
+    let _ = config.connection_router.execute(
+        &mut config.app,
+        config.governance_address.clone(),
+        &ExecuteMsg::RegisterChain {
+            chain: chain.chain_name.clone(),
+            gateway_address: chain.gateway.to_string(),
+        },
+    );
```
cgorenflo, line 67: be careful not to throw `Result<>` away. we might not be interested in the response value, but keep the `unwrap` so the tests fails if the registration fails

contracts/connection-router/tests/test.rs
```diff
@@ -62,18 +57,26 @@ fn make_chain(name: &str, config: &mut TestConfig) -> Chain {
...
+    //         &ExecuteMsg::RegisterChain {
+    //             chain: chain.chain_name.clone(),
+    //             gateway_address: chain.gateway.to_string(),
+    //         },
+    //         &[],
+    //     )
+    //     .unwrap();
 }
 
 fn generate_messages(
```
cgorenflo, line 82: not in the scope of this PR, I just wanted to point it out to you: I think this function would be much cleaner like this:
```
fn generate_messages(
    src_chain: &Chain,
    dest_chain: &Chain,
    nonce: usize,
    count: usize,
) -> Vec<Message> {
    (nonce..(nonce+count))
    .map(|nonce| 
        Message {
            cc_id: CrossChainId {
                id: format!("tx_id:{}", nonce).parse().unwrap(),
                chain: src_chain.chain_name.clone(),
            },
            destination_address: "idc".parse().unwrap(),
            destination_chain: dest_chain.chain_name.clone(),
            source_address: "idc".parse().unwrap(),
            payload_hash: [nonce as u8; 32],
        }
    ).collect()
}
```
that way you don't leak any mutable state and you don't have to worry about the details of the vector creation

contracts/connection-router/tests/test.rs
```diff
@@ -131,1335 +133,1335 @@ fn route() {
     // try to route twice
     let res = config.app.execute_contract(
         eth.gateway.clone(),
-        config.contract_address.clone(),
+        config.connection_router.contract_addr.clone(),
         &ExecuteMsg::RouteMessages(msgs.clone()),
         &[],
     );
```
cgorenflo, line 139: why did you not replace this with a call to the router itself?

integration-tests/src/contract.rs
maancham: yes, all of them use `use axelar_wasm_std_derive::IntoContractError`.

contracts/connection-router/tests/test.rs
maancham: `fn register_chain(config: &mut TestConfig, chain: &Chain) {
    config.connection_router.execute(
        &mut config.app,
        config.governance_address.clone(),
        &ExecuteMsg::RegisterChain {
            chain: chain.chain_name.clone(),
            gateway_address: chain.gateway.to_string(),
        },
    ).unwrap();
}`

contracts/connection-router/tests/test.rs
maancham: Thanks, updated!

contracts/connection-router/tests/test.rs
maancham: weird. because I already have the same piece of code in this part:

```
let res = config.app.execute_contract(
        eth.gateway.clone(),
        config.connection_router.contract_addr.clone(),
        &ExecuteMsg::RouteMessages(msgs.clone()),
        &[],
    );
```
marking this as resolved..

integration-tests/src/contract.rs
cgorenflo: replace `Self::Err` with `axelar_wasm_std::ContractError` and remove `type Err`

contracts/connection-router/tests/test.rs
cgorenflo: it's still good to keep the `let _ =`, so it is obvious that it's discarded on purpose and not just forgotten

integration-tests/src/contract.rs
maancham: ok.

contracts/connection-router/tests/test.rs
```diff
@@ -107,1359 +101,1358 @@ fn route() {
...
             eth.gateway.clone(),
-            config.contract_address.clone(),
-            &ExecuteMsg::RouteMessages(msgs.clone()),
-            &[],
+            &ExecuteMsg::RouteMessages(messages.clone()),
         )
         .unwrap();
 
-    let msgs_ret = mock::get_gateway_messages(&mut config.app, polygon.gateway, &msgs);
+    let messages_ret = mock::get_gateway_messages(&mut config.app, polygon.gateway, &messages);
```
cgorenflo, line 119:
```suggestion
    let outgoing_messages = mock::get_gateway_messages(&mut config.app, polygon.gateway, &messages);
```

contracts/connection-router/tests/test.rs
```diff
@@ -107,1359 +101,1358 @@ fn route() {
...
 
-    assert_eq!(msgs.len(), msgs_ret.len());
-    assert_eq!(msgs, msgs_ret);
+    assert_eq!(messages.len(), messages_ret.len());
+    assert_eq!(messages, messages_ret);
 
     // try to route twice
     let res = config.app.execute_contract(
```
cgorenflo, line 125: why are you not using `connection_router.execute` here?

contracts/connection-router/tests/test.rs
maancham: Done

contracts/connection-router/tests/test.rs
maancham: updated

cgorenflo (approved):

PR #269, opened 2024-02-13
chore(infra): add a seed job for building releases for ampd and fix naming for building main branch ecr image

talalashraf:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Other Notes

### DIFF
--- a/.github/workflows/build-ampd-main.yaml
+++ b/.github/workflows/build-ampd-main.yaml
@@ -1,17 +1,14 @@
-name: Build and push latest docker image
+name: Amplifier - Build main branch
 
 on:
   push:
     branches:
       - main
   workflow_dispatch:
+
 jobs:
   build-and-push-latest-docker-image:
-    strategy:
-      matrix:
-        os:
-          - ubuntu-20.04
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-22.04
     permissions:
       id-token: write
       contents: read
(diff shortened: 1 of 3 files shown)


.github/workflows/build-ampd-main.yaml
```diff
@@ -1,17 +1,14 @@
-name: Build and push latest docker image
+name: Amplifier - Build main branch
```
milapsheth, line 1:
```suggestion
name: Amplifier - main branch auto-builder
```
perhaps to better distinguish auto build actions vs manually triggered actions?

.github/workflows/build-ampd-release.yaml
```diff
@@ -0,0 +1,30 @@
...
+        description: Github tag to release binaries for (reusing an existing tag will make the pipeline fail)
+        required: true
+        default: latest
+
+jobs:
+  release-binaries:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
```
milapsheth, line 16: for release builds, we should fix the OS version
```suggestion
        os: [ubuntu-22.04, macos-12]
```

.github/workflows/build-ampd-main.yaml
talalashraf: its both auto build and can be manually triggered.

milapsheth: Fair but the main purpose is auto build anyways

talalashraf: yes but then ill get questions about how to trigger the auto builder manually.

cgorenflo: I'm with Talal on this. If I see auto-builder I will ignore it in the search for the correct workflow

milapsheth (approved):

PR #270, opened 2024-02-14
feat(multisig): add optional signature validation callback

eguajardo:
## Description

Support for delegating signature verification to another contract

https://axelarnetwork.atlassian.net/browse/AXE-2777

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


contracts/multisig/src/signing.rs
```diff
@@ -61,14 +65,32 @@ pub fn validate_session_signature(
...
+    let valid = sig_verifier.map_or_else(
+        || signature.verify(&session.msg, pub_key),
+        |verifier| {
+            verifier
+                .verify_signature(
+                    signature.as_ref().into(),
+                    session.msg.as_ref().into(),
+                    pub_key.as_ref().into(),
+                    signer.to_string(),
+                    session.id,
```
fish-sammy, line 85: Why are `signer_address` and `session_id` relevant?

packages/signature-verifier-api/src/client.rs
```diff
@@ -0,0 +1,44 @@
...
+
+    pub fn verify_signature(
+        &self,
+        signature: HexBinary,
+        message: HexBinary,
+        public_key: HexBinary,
+        signer_address: String,
+        session_id: Uint64,
+    ) -> Result<bool, Error> {
+        self.query(&QueryMsg::VerifySignature {
```
fish-sammy, line 30: What happens when the query message in JSON has more fields than the quering contract expects? Does it force deserialize all bytes and therefore would fail the query? Asking cuz we need to understand what kind of impact it would be to add new fieds to this query msg when some callbacks already exist.

eguajardo: good point, the deserialization will fail. I'll think a bit more about this

contracts/multisig/src/signing.rs
cjcobb23: the common prefix people specifically asked for these fields. I think passing more data here is preferable, to make this as generic and flexible as possible. I don't want to leave some stuff out, and then it turns out in the future, some logic needs more information, and then we need to change things and pass more information. I'd rather just pass everything now, and not have to touch this again.

packages/signature-verifier-api/src/msg.rs
```diff
@@ -0,0 +1,15 @@
...
+pub enum QueryMsg {
+    #[returns(bool)]
+    VerifySignature {
+        signature: HexBinary,
+        message: HexBinary,
+        public_key: HexBinary,
+        signer_address: String,
+        session_id: Uint64,
+    },
+}
```
cjcobb23, line 15: I would actually rather have this live in the msg.rs file of the multisig contract, so that way it's clearer to readers the interface of the callback. Right now, when you read the msg.rs file of multisig, there is mention of a callback, but no indication about what the interface to that callback is. We could leave a comment, but I feel just having this type right there would be preferable.

contracts/multisig/src/contract.rs
```diff
@@ -52,17 +52,14 @@ pub fn execute(
             sig_verifier,
         } => {
             execute::require_authorized_caller(&deps, info.sender)?;
-
-            let _sig_verifier = sig_verifier
-                .map(|addr| deps.api.addr_validate(&addr))
-                .transpose()?; // TODO: handle callback
```
cjcobb23, line 58: I'd actually prefer to keep this validation logic here actually, and pass an `Addr` into `start_signing_session`

cjcobb23 (commented):
This functionality seems like a perfect candidate for a proper integration test, since it's contract to contract interaction. That can be done as a separate task though.

packages/signature-verifier-api/src/msg.rs
eguajardo: I don't know, I think the documentation should show the proper package instead of having everything in the multisig. Seems more cleaner to me if the implementations just import the interface package

contracts/multisig/src/signing.rs
fish-sammy: Why do they ask for these to verify signatures? I am ok with the idea of sending as much data as possible but still find it weird to include these. If that's the goal, why not send the entire session over?

eguajardo: from the SigningSession fields, I think the only one that might make sense to expose is the `chain_name`, I wouldn't like to pass the whole struct as it is because then there is more risk of breaking changes if something changes in the struct

cjcobb23: the chain name is not necessary, the prover should know the chain name. But if you really want to send it, I'm not opposed. They need the session id to do some internal mapping. They are not actually validating the message, it's a different message actually that's being signed, and I'm pretty sure they need the session id to know what message is being signed. I think the bare minimum to send is session_id, signature and either pubkey or worker address. But again, I'd err on the side of sending everything.

cjcobb23 (approved):

PR #276, opened 2024-02-21
refactor: contract trait voting verifier

maancham:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


contracts/voting-verifier/tests/tests.rs
```diff
@@ -1,57 +1,58 @@
...
 
 const SENDER: &str = "sender";
 const POLL_BLOCK_EXPIRY: u64 = 100;
 fn source_chain() -> ChainName {
     "source_chain".parse().unwrap()
 }
 
-fn initialize_contract(app: &mut App, service_registry_address: nonempty::String) -> Addr {
-    let rewards_address = make_mock_rewards(app).into();
+struct TestConfig {
```
cgorenflo, line 26: don't call it config. Config(uration) usually means a set of parameters you use to set up the environment. Alternative suggestions: TestSetup, TestFixure, TestEnv

cgorenflo: when you change it make sure you change the variable name in the tests, too

cgorenflo (approved):

contracts/voting-verifier/tests/tests.rs
maancham: Thanks, fixed in latest commit.

PR #280, opened 2024-02-23
refactor: remove mock contracts from voting verifier tests and use mock querier `update_wasm()` instead

eguajardo:
## Description

https://axelarnetwork.atlassian.net/browse/AXE-3236

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


contracts/voting-verifier/src/contract.rs
```diff
@@ -64,3 +65,781 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> StdResult<Binary> {
...
+                    VerificationStatus::SucceededOnChain
+                ),
+                (messages[1].cc_id.clone(), VerificationStatus::FailedOnChain),
+                (messages[2].cc_id.clone(), VerificationStatus::NotFound),
+                (
+                    messages[3].cc_id.clone(),
+                    VerificationStatus::FailedToVerify
+                )
+            ]
+        );
```
cgorenflo, line 422: this setup could use some simplification/explanations

contracts/voting-verifier/src/contract.rs
```diff
@@ -64,3 +65,781 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> StdResult<Binary> {
...
+        let res = execute(deps.as_mut(), mock_env(), mock_info(SENDER, &[]), msg).unwrap();
+
+        let reply: VerifyMessagesResponse = from_binary(&res.data.unwrap()).unwrap();
+
+        assert_eq!(reply.verification_statuses.len(), messages.len());
+        assert_eq!(
+            reply.verification_statuses,
+            messages
+                .iter()
+                .map(|message| (message.cc_id.clone(), VerificationStatus::None))
```
cgorenflo, line 475: do you think it's confusing that an unknown message return `None` instead of `InProgress` after verification has started?

contracts/voting-verifier/src/contract.rs
```diff
@@ -64,3 +65,781 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> StdResult<Binary> {
...
+                ),
+                (messages[1].cc_id.clone(), VerificationStatus::FailedOnChain),
+                (messages[2].cc_id.clone(), VerificationStatus::InProgress),
+                (messages[3].cc_id.clone(), VerificationStatus::InProgress)
+            ]
+        );
+    }
+
+    #[test]
+    fn should_query_message_statuses() {
```
cgorenflo, line 458: this test is also a bit complicated

cgorenflo: the test name is a bit too non-descriptive

contracts/voting-verifier/src/contract.rs
```diff
@@ -64,3 +65,781 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> StdResult<Binary> {
...
+                deps.as_ref(),
+                mock_env(),
+                QueryMsg::GetWorkerSetStatus {
+                    new_operators: operators.clone(),
+                },
+            )
+            .unwrap(),
+        )
+        .unwrap();
+        assert_eq!(res, VerificationStatus::InProgress);
```
cgorenflo, line 585: why does worker set verification return `InProgress` and messages return `None`? Is that an inconsistency?

contracts/voting-verifier/src/contract.rs
```diff
@@ -64,3 +65,781 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> StdResult<Binary> {
...
+            mock_env_expired(),
+            mock_info(SENDER, &[]),
+            ExecuteMsg::VerifyWorkerSet {
+                message_id: message_id("id", 0),
+                new_operators: operators.clone(),
+            },
+        )
+        .unwrap_err();
+        are_contract_err_strings_equal(err, ContractError::WorkerSetAlreadyConfirmed);
+    }
```
cgorenflo, line 844: this also seems inconsistent between messages and worker sets, messages will just return the confirmed state without triggering a new verification, do we need to resolve this?

cgorenflo (approved):

contracts/voting-verifier/src/contract.rs
eguajardo: to be honest I would remove the whole response `data`, it is being set to the status before the verification starts, I think the original purpose was to do some logic with it in the aggregate verifier to combine multiple verifiers, I'm not sure it makes sense to use the previous status before the verification starts, but anyway I think the real combining logic would end up in the query not the execute since that's what the gateway uses.

contracts/voting-verifier/src/contract.rs
eguajardo: see my comment here: https://github.com/axelarnetwork/axelar-amplifier/pull/280#discussion_r1503443770

contracts/voting-verifier/src/contract.rs
cgorenflo: yeah I was wandering about that too

PR #281, opened 2024-02-26
refactor: change gateway get messages query name

maancham:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

### DIFF
--- a/packages/gateway-api/src/msg.rs
+++ b/packages/gateway-api/src/msg.rs
@@ -18,5 +18,5 @@ pub enum ExecuteMsg {
 #[derive(QueryResponses)]
 pub enum QueryMsg {
     #[returns(Vec<Message>)]
-    GetMessages { message_ids: Vec<CrossChainId> },
+    GetOutgoingMessages { message_ids: Vec<CrossChainId> },
 }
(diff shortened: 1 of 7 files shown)


packages/gateway-api/src/msg.rs
```diff
@@ -18,5 +18,5 @@ pub enum ExecuteMsg {
 #[derive(QueryResponses)]
 pub enum QueryMsg {
     #[returns(Vec<Message>)]
-    GetMessages { message_ids: Vec<CrossChainId> },
+    GetOutgoingMessages { message_ids: Vec<CrossChainId> },
```
cgorenflo, line 21: add a doc comment explaining what this query returns, i.e. messages that can be relayed to the chain corresponding to this gateway, in contrast to messages that came from there.

cgorenflo (approved):

packages/gateway-api/src/msg.rs
maancham: Added in latest commit

fish-sammy (approved):

PR #284, opened 2024-02-29
fix(multisig): normalize compressed and uncompressed public keys for proper duplicate check

eguajardo:
## Description

https://axelarnetwork.atlassian.net/browse/AXE-3263

## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

### DIFF
--- a/contracts/multisig/Cargo.toml
+++ b/contracts/multisig/Cargo.toml
@@ -46,6 +46,7 @@ cosmwasm-schema = { workspace = true }
 cosmwasm-std = { workspace = true }
 cosmwasm-storage = { workspace = true }
 cw-storage-plus = { workspace = true }
+ed25519-dalek = { version = "2.1.1", default-features = false }
 enum-display-derive = "0.1.1"
 error-stack = { workspace = true }
 getrandom = { version = "0.2", default-features = false, features = ["custom"] }
--- a/contracts/multisig/src/error.rs
+++ b/contracts/multisig/src/error.rs
@@ -28,9 +28,6 @@ pub enum ContractError {
     #[error("invalid public key")]
     InvalidPublicKey,
 
-    #[error("ECDSA public key must be in compressed format (33 bytes)")]
-    InvalidEcdsaPublicKeyFormat,
-
     #[error("public key is already registered")]
     DuplicatePublicKey,
 
--- a/contracts/multisig/src/error.rs
+++ b/contracts/multisig/src/error.rs
@@ -28,6 +28,9 @@ pub enum ContractError {
     #[error("invalid public key")]
     InvalidPublicKey,
 
+    #[error("ECDSA public key must be in compressed format (33 bytes)")]
+    InvalidEcdsaPublicKeyFormat,
+
     #[error("public key is already registered")]
     DuplicatePublicKey,
 
--- a/contracts/multisig/src/error.rs
+++ b/contracts/multisig/src/error.rs
@@ -25,8 +25,8 @@ pub enum ContractError {
     #[error("signed sender address could not be verified using submitted public key")]
     InvalidPublicKeyRegistrationSignature,
 
-    #[error("invalid public key format: {reason:?}")]
-    InvalidPublicKeyFormat { reason: String },
+    #[error("invalid public key")]
+    InvalidPublicKey,
 
     #[error("public key is already registered")]
     DuplicatePublicKey,
--- a/contracts/multisig/src/key.rs
+++ b/contracts/multisig/src/key.rs
@@ -10,6 +10,8 @@ use enum_display_derive::Display;
 use serde::{de::Error, Deserialize, Deserializer};
 use std::fmt::Display;
 
+const ECDSA_COMPRESSED_PUBKEY_LEN: usize = 33;
+
 #[cw_serde]
 #[derive(Copy, Display)]
 pub enum KeyType {
@@ -114,6 +116,7 @@ impl TryFrom<HexBinary> for Recoverable {
 #[cw_serde]
 #[derive(Ord, PartialOrd, Eq)]
 pub enum PublicKey {
+    /// ECDSA public key must be in compressed format (33 bytes)
     #[serde(deserialize_with = "deserialize_ecdsa_key")]
     Ecdsa(HexBinary),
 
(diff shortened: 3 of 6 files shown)


contracts/multisig/src/key.rs
```diff
@@ -339,12 +344,24 @@ mod ecdsa_tests {
...
+    fn should_normalize_compressed_and_uncompressed_public_keys() {
+        let uncompressed = HexBinary::from_hex("049bb8e80670371f45508b5f8f59946a7c4dea4b3a23a036cf24c1f40993f4a1daad1716de8bd664ecb4596648d722a4685293de208c1d2da9361b9cba74c3d1ec").unwrap();
+        let compressed = HexBinary::from_hex(
+            "029bb8e80670371f45508b5f8f59946a7c4dea4b3a23a036cf24c1f40993f4a1da",
+        )
+        .unwrap();
+
+        assert_eq!(
+            PublicKey::try_from((KeyType::Ecdsa, uncompressed.clone())),
+            PublicKey::try_from((KeyType::Ecdsa, compressed.clone()))
```
fish-sammy, line 361:
```suggestion
            PublicKey::try_from((KeyType::Ecdsa, uncompressed.clone())).unwrap(),
            PublicKey::try_from((KeyType::Ecdsa, compressed.clone())).unwrap()
```
If both `try_from` return `Err` instead of `Ok`, the test would still pass but that's no good.

fish-sammy (changes requested):

contracts/multisig/Cargo.toml
```diff
@@ -46,6 +46,7 @@ cosmwasm-schema = { workspace = true }
 cosmwasm-std = { workspace = true }
 cosmwasm-storage = { workspace = true }
 cw-storage-plus = { workspace = true }
+ed25519-dalek = { version = "2.1.1", default-features = false }
```
fish-sammy, line 49: Where's this used?

cgorenflo (approved):

contracts/multisig/Cargo.toml
eguajardo: https://github.com/axelarnetwork/axelar-amplifier/pull/284/files#diff-806ddeb05a6a238a6b153a52040776eef585c519227925ec480d7e91d42a10e7R227

contracts/multisig/src/key.rs
```diff
@@ -211,37 +215,40 @@ impl KeyDeserialize for KeyType {
...
+// an ECDSA public key can be represented as compressed or uncompressed. This function
+// validates and normalizes the public key to a single format.
+fn validate_and_normalize_public_key(
+    key_type: KeyType,
+    pub_key: HexBinary,
+) -> Result<HexBinary, ContractError> {
+    match key_type {
+        KeyType::Ecdsa => Ok(k256::PublicKey::from_sec1_bytes(pub_key.as_slice())
+            .map_err(|_| ContractError::InvalidPublicKey)?
+            .to_sec1_bytes()
```
milapsheth, line 228: Can you confirm if this using the compressed representation? tofn does [this](https://github.com/axelarnetwork/tofn/blob/main/src/ecdsa/mod.rs#L47) instead

contracts/multisig/src/key.rs
```diff
@@ -211,37 +215,40 @@ impl KeyDeserialize for KeyType {
...
 
-const ECDSA_COMPRESSED_PUBKEY_LEN: usize = 33;
-const ECDSA_UNCOMPRESSED_PUBKEY_LEN: usize = 65;
-
-const ED25519_PUBKEY_LEN: usize = 32;
-const ED25519_SIGNATURE_LEN: usize = 64;
+// It is possible for the same public key to be represented in different formats. For example
+// an ECDSA public key can be represented as compressed or uncompressed. This function
+// validates and normalizes the public key to a single format.
+fn validate_and_normalize_public_key(
```
milapsheth, line 221: We should just restrict what's allowed to be submitted instead, i.e it should already be in the normalized form and fail if it isn't. And we can control this in `ampd` anyways. This is generally a better cryptographic practice. Also simplifies the effort for any indexers.

contracts/multisig/src/key.rs
```diff
@@ -211,37 +215,40 @@ impl KeyDeserialize for KeyType {
...
+    key_type: KeyType,
+    pub_key: HexBinary,
+) -> Result<HexBinary, ContractError> {
+    match key_type {
+        KeyType::Ecdsa => Ok(k256::PublicKey::from_sec1_bytes(pub_key.as_slice())
+            .map_err(|_| ContractError::InvalidPublicKey)?
+            .to_sec1_bytes()
+            .as_ref()
+            .into()),
+        KeyType::Ed25519 => Ok(ed25519_dalek::VerifyingKey::from_bytes(
```
milapsheth, line 231: From the crate's [docs](https://docs.rs/ed25519-dalek/latest/ed25519_dalek/struct.VerifyingKey.html#method.from_bytes)
> Warning
>The caller is responsible for ensuring that the bytes passed into this method actually represent a >curve25519_dalek::curve::CompressedEdwardsY and that said compressed point is actually a point on the curve.

Looks like `from_bytes` doesn't do sufficient validation. It's also not clear to me which encoding scheme this crate is using (whether it's a custom one or another). If it's custom than it makes it harder for us to swap the ed25519 signer backend in the future. You can try investigating if there's some standard format and validation we can use. Otherwise, add a TODO and create a JIRA ticket and I'll take a look later.

contracts/multisig/src/key.rs
eguajardo: yes I tested and doesn't matter the input it throws the compressed representation. Internally it seems to be getting the bytes from the encoded point as well:

```
pub fn to_sec1_bytes(&self) -> Box<[u8]>
    where
        C: PointCompression,
        AffinePoint<C>: FromEncodedPoint<C> + ToEncodedPoint<C>,
        FieldBytesSize<C>: ModulusSize,
    {
        EncodedPoint::<C>::from(self).to_bytes()
    }
```

contracts/multisig/src/key.rs
fish-sammy: I am OK with that, too.

contracts/multisig/src/key.rs
milapsheth: Cool

contracts/multisig/src/key.rs
eguajardo: created this two Jira tasks to tackle this in a future PR:

https://axelarnetwork.atlassian.net/browse/AXE-3312
https://axelarnetwork.atlassian.net/browse/AXE-3313

contracts/multisig/src/error.rs
```diff
@@ -25,8 +25,11 @@ pub enum ContractError {
     #[error("signed sender address could not be verified using submitted public key")]
     InvalidPublicKeyRegistrationSignature,
 
-    #[error("invalid public key format: {reason:?}")]
-    InvalidPublicKeyFormat { reason: String },
+    #[error("invalid public key")]
+    InvalidPublicKey,
+
+    #[error("ECDSA public key must be in compressed format (33 bytes)")]
+    InvalidEcdsaPublicKeyFormat,
```
fish-sammy, line 32: I feel it's a bit silly to have 2 errors here with one being the superset of the other.

fish-sammy (approved):

PR #287, opened 2024-03-01
refactor: simplify complexity in verify() signature implementation

eguajardo:
## Description

https://axelarnetwork.atlassian.net/browse/AXE-3276

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


contracts/multisig/src/key.rs
```diff
@@ -174,18 +174,22 @@ impl KeyTyped for Signature {
...
             KeyType::Ecdsa => ecdsa_verify(msg.as_ref(), self, pub_key.as_ref()),
             KeyType::Ed25519 => ed25519_verify(msg.as_ref(), self.as_ref(), pub_key.as_ref()),
+        }?;
+
+        if !res {
+            Err(ContractError::SignatureVerificationFailed {
+                reason: "unable to verify signature".into(),
+            })
+        } else {
+            Ok(())
```
cgorenflo, line 192: I'd invert the check

contracts/multisig/src/key.rs
```diff
@@ -391,12 +395,12 @@ mod ecdsa_tests {
             .unwrap();
         let message = MsgToSign::try_from(ecdsa_test_data::message()).unwrap();
         let public_key = PublicKey::try_from((KeyType::Ecdsa, ecdsa_test_data::pub_key())).unwrap();
-        let result = signature.verify(&message, &public_key).unwrap();
-        assert_eq!(result, true);
+        let result = signature.verify(&message, &public_key);
+        assert_eq!(result.unwrap(), ());
```
cgorenflo, line 399:
```suggestion
        assert!(result.is_ok());
```

contracts/multisig/src/key.rs
```diff
@@ -391,12 +395,12 @@ mod ecdsa_tests {
...
     }
 
     #[test]
-    fn test_verify_signature_invalid_signature() {
+    fn should_fail_sig_verification_when_using_different_valid_sig() {
```
cgorenflo, line 403: nice

contracts/multisig/src/key.rs
```diff
@@ -405,8 +409,32 @@ mod ecdsa_tests {
...
         let message = MsgToSign::try_from(ecdsa_test_data::message()).unwrap();
         let public_key = PublicKey::try_from((KeyType::Ecdsa, ecdsa_test_data::pub_key())).unwrap();
-        let result = signature.verify(&message, &public_key).unwrap();
-        assert_eq!(result, false);
+        let result = signature.verify(&message, &public_key);
+        assert_eq!(
+            result.unwrap_err(),
+            ContractError::SignatureVerificationFailed {
+                reason: "unable to verify signature".into(),
+            }
```
cgorenflo, line 417: I'd would like to see the reason to be replaced with an error_stack context switch, but that's probably beyond the scope of this PR

contracts/multisig/src/key.rs
```diff
@@ -503,7 +536,7 @@ mod ed25519_tests {
         let public_key =
             PublicKey::try_from((KeyType::Ed25519, ed25519_test_data::pub_key())).unwrap();
         let result = signature.verify(&message, &public_key).unwrap();
-        assert_eq!(result, true);
+        assert_eq!(result, ());
```
cgorenflo, line 539:
```suggestion
        assert!(result.is_ok());
```

contracts/multisig/src/signing.rs
```diff
@@ -73,33 +73,51 @@ pub fn validate_session_signature(
...
-    let valid = sig_verifier.map_or_else(
-        || signature.verify(&session.msg, pub_key),
-        |verifier| {
-            verifier
-                .verify_signature(
+    sig_verifier
+        .map_or_else(
+            || signature.verify(&session.msg, pub_key),
+            |sig_verifier| {
+                call_sig_verifier(
```
cgorenflo, line 80: is there a specific reason why you extracted this? I think in this case the extra overhead is not worth it with so many parameters, just to convert the bool to an error

contracts/multisig/src/key.rs
eguajardo: I'm not fan of `is_ok()` because when the test fails it doesn't say anything about the error, it just say the assertion is not true. I would rather remove the assert and add the unwrap to the function call like this: `signature.verify(&message, &public_key).unwrap();`. Thoughts?

contracts/multisig/src/key.rs
eguajardo: agree, I thought the same, but I preferred to leave it for another PR

contracts/multisig/src/signing.rs
eguajardo: I thought the `map_or_else` would be easier to read without too much clutter in there

cgorenflo: ok good point!

cgorenflo (approved):

PR #288, opened 2024-03-02
fix: address risky outdated packages

maancham:
## Description

part one of [AXE-3262]

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Other Notes

[AXE-3262]: https://axelarnetwork.atlassian.net/browse/AXE-3262?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/periodic-audit.yaml
```diff
@@ -0,0 +1,12 @@
...
+  schedule:
+    - cron: '0 0 * * 0'
+jobs:
+  audit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: rustsec/audit-check@v1.4.1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
milapsheth, line 12: can we just have github dependabot do this? I'd imagine it supports this

Cargo.toml
```diff
@@ -16,11 +16,13 @@ events = { version = "^0.1.0", path = "packages/events" }
 events-derive = { version = "^0.1.0", path = "packages/events-derive" }
 axelar-wasm-std = { version = "^0.1.0", path = "packages/axelar-wasm-std" }
 axelar-wasm-std-derive = { version = "^0.1.0", path = "packages/axelar-wasm-std-derive" }
+h2 = "0.3.24"
```
cjcobb23, line 19: I'm confused about why we need to add these packages here. I think they are dependencies of our dependencies, which are already specified in this file, so can we just upgrade our dependencies? Like something in this file is already including `h2`, can we just upgrade whatever that is?

.github/workflows/periodic-audit.yaml
fish-sammy, line 12:
```suggestion
          token: ${{ secrets.GITHUB_TOKEN }}

```

maancham: will address in a future PR

Cargo.toml
cgorenflo: maybe I misunderstood you, but, having them here should mean that in the per package Cargo.toml files they should change to `{ workspace = true }`

cgorenflo: as it stands right now I'm also a bit confused why the only change here is to add new dependencies

maancham: so, what to do instead? The reason we put it here was for all other cargo.tomls to be using it rather than changing the version over multiple files

cjcobb23: which of our dependencies is using h2? Is it not possible to just update that dependency? h2 is not actually a dependency that we specify in any Cargo.toml file right now, I'd prefer not to add it, but rather just update the dependency that is including it

maancham: yeah I'm taking this path now.

cgorenflo (approved):

ampd/Cargo.toml
```diff
@@ -15,7 +15,7 @@ config = "0.13.2"
 connection-router-api = { workspace = true }
 cosmos-sdk-proto = "0.16.0"
 cosmrs = { version = "0.14.0", features = ["cosmwasm"] }
-cosmwasm-std = { version = "1.2.1", features = ["stargate"] }
+cosmwasm-std = { version = "1.5.3", features = ["stargate"] }
```
cgorenflo, line 18: does this work?
```suggestion
cosmwasm-std = { workspace = true, features = ["stargate"] }
```

cgorenflo (approved):

PR #291, opened 2024-03-05
refactor(ampd)!: decouple finalizer type from chain name

cjcobb23:
* default to Ethereum finalizer
* remove ChainName struct from ampd. Use struct from router

https://axelarnetwork.atlassian.net/browse/AXE-3272


ampd/src/evm/finalizer.rs
```diff
@@ -161,4 +171,36 @@ mod tests {
...
+        assert_eq!(
+            finalizer.latest_finalized_block_height().await.unwrap(),
+            block_number
+        );
+
+        let finalizer = pick(&FinalizerType::PoW, &rpc_client, pow_confirmation_height);
+        assert_eq!(
+            finalizer.latest_finalized_block_height().await.unwrap(),
+            block_number - U64::from(pow_confirmation_height - 1)
+        );
```
fish-sammy, line 204: I'd separate these into different tests to prevent this test from growing in the future, i.e. `pick_should_work_for_ethereum_finalizer` and `pick_should_work_for_pow_finalizer`.

ampd/src/evm/mod.rs
```diff
@@ -1,50 +1,4 @@
...
-#[derive(Debug, Deserialize, Serialize, PartialEq, Hash, Eq, Clone, Display)]
-pub enum ChainName {
-    Ethereum,
-    #[serde(untagged)]
-    Other(String),
-}
-
-impl PartialEq<connection_router_api::ChainName> for ChainName {
-    fn eq(&self, other: &connection_router_api::ChainName) -> bool {
-        self.to_string().eq_ignore_ascii_case(other.as_ref())
```
fish-sammy, line 20: Does the other definition of `ChainName` also compares ignoring cases?

cjcobb23: yes

cgorenflo (approved):
the handler is becoming quite complicated and has a lot of parameters for setup. Maybe we need to look into it a bit more, but out of scope for this PR. For now just pay attention how difficult it gets changing this code, so we can intervene before it becomes unmanageable

ampd/src/evm/finalizer.rs
```diff
@@ -15,18 +15,25 @@ pub trait Finalizer: Send + Sync {
     async fn latest_finalized_block_height(&self) -> Result<U64>;
 }
 
+#[derive(Debug, Deserialize, Serialize, PartialEq, Default)]
+pub enum FinalizerType {
+    #[default]
+    Ethereum,
+    PoW,
+}
```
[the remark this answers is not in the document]
cjcobb23: But then again, even within a specific class of finality, the code might need some different mechanism of actually interacting with the chain to determine finality. So I actually would just prefer to be specific with something like:
```
pub enum Finalization {
    #[default]
    FinalizedBlock,
    ConfimationHeight
}
```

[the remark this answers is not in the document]
fish-sammy: I am OK with `RPCFinalizedBlock` and `ConfimationHeight`.

fish-sammy (approved):

PR #294, opened 2024-03-06
fix(service_registry)!: correctly handle minimum workerset size

cjcobb23:
get_active_workers returns an error now when the number of active workers is less than the minimum required by the service.

https://axelarnetwork.atlassian.net/browse/AXE-3319

### DIFF
--- a/contracts/service-registry/src/contract.rs
+++ b/contracts/service-registry/src/contract.rs
@@ -385,7 +385,7 @@ pub mod query {
             .may_load(deps.storage, &service_name)?
             .ok_or(ContractError::ServiceNotFound)?;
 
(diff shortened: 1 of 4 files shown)


contracts/service-registry/src/contract.rs
```diff
@@ -385,7 +385,7 @@ pub mod query {
             .may_load(deps.storage, &service_name)?
             .ok_or(ContractError::ServiceNotFound)?;
 
-        let workers = WORKERS_PER_CHAIN
+        let workers: Vec<Worker> = WORKERS_PER_CHAIN
```
fish-sammy, line 388:
```suggestion
        let workers: Vec<_> = WORKERS_PER_CHAIN
```

fish-sammy (approved):

PR #296, opened 2024-03-07
refactor: update arithmetic operations for checked ones

eguajardo:
## Description

https://axelarnetwork.atlassian.net/browse/AXE-3274

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


contracts/rewards/src/contract/execute.rs
```diff
@@ -192,12 +200,14 @@ where
         // (i.e. we are in epoch 0, which started at block 0 and epoch duration is 1000. At epoch 500, the params
         // are updated to shorten the epoch duration to 100 blocks. We set the epoch number to 1, to prevent skipping
         // epochs 1-4, and so all events prior to the start of epoch 1 have an epoch number of 0)
-        let should_end =
-            cur_epoch.block_height_started + u64::from(new_params.epoch_duration) < block_height;
+        let should_end = cur_epoch
+            .block_height_started
+            .saturating_add(u64::from(new_params.epoch_duration))
```
cgorenflo, line 205: while the event is highly unlikely, if it happens the current epoch is suddenly shorter than what we expect based on the epoch duration. I don't think we should break that invariant, so I'd prefer a `checked_add()?` here

contracts/rewards/src/contract/execute.rs
```diff
@@ -192,12 +200,14 @@ where
...
+            < block_height;
         let cur_epoch = if should_end {
             Epoch {
                 block_height_started: block_height,
-                epoch_num: cur_epoch.epoch_num + 1,
+                epoch_num: cur_epoch.epoch_num.saturating_add(1),
```
cgorenflo, line 210: `checked_add().expect()`
the epoch number can never be larger than block_height

contracts/rewards/src/contract/execute.rs
```diff
@@ -215,7 +225,11 @@ where
         amount: nonempty::Uint128,
     ) -> Result<(), ContractError> {
         let mut pool = self.store.load_rewards_pool(pool_id)?;
-        pool.balance += Uint128::from(amount);
+        pool.balance = pool
+            .balance
+            .checked_add(Uint128::from(amount))
+            .map_err(Into::<ContractError>::into)
```
cgorenflo, line 231: does `into()` not work here?

contracts/rewards/src/contract/execute.rs
```diff
@@ -231,12 +245,17 @@ where
...
     rewards_2
         .into_iter()
-        .fold(rewards_1, |mut rewards, (addr, amt)| {
-            *rewards.entry(addr).or_default() += amt;
-            rewards
+        .try_fold(rewards_1, |mut rewards, (addr, amt)| {
+            let r = rewards.entry(addr).or_default();
+            *r = r
+                .checked_add(amt)
+                .map_err(Into::<ContractError>::into)
```
cgorenflo, line 255: does `into()` not work here?

contracts/rewards/src/contract/execute.rs
```diff
@@ -231,12 +245,17 @@ where
...
+                .map_err(Report::from)?;
+
+            Ok(rewards)
```
cgorenflo, line 258:
```suggestion
            Ok(r)
```

cgorenflo: this not failing a unit test indicates that we are missing coverage. Can you ensure we properly check the rewards?

contracts/rewards/src/error.rs
```diff
@@ -36,6 +37,9 @@ pub enum ContractError {
     #[error("specified block has already passed")]
     BlockHeightInPast,
 
+    #[error("{0}")]
```
cgorenflo, line 40:
```suggestion
    #[error(transparent)]
```

contracts/rewards/src/state.rs
```diff
@@ -200,7 +200,7 @@ impl RewardsPool {
         if self.balance < reward {
             return Err(ContractError::PoolBalanceInsufficient.into());
         }
-        self.balance -= reward;
+        self.balance = self.balance.saturating_sub(reward);
```
cgorenflo, line 203:
```
self.balance = self.balance.checked_sub(reward).map_err(|_| ContractError::PoolBalanceInsufficient.into())
```

contracts/service-registry/src/state.rs
```diff
@@ -89,7 +89,7 @@ impl BondingState {
             | BondingState::Unbonding {
                 amount,
                 unbonded_at: _,
-            } => amount + to_add,
+            } => amount.saturating_add(to_add),
```
cgorenflo, line 92: very dangerous, because we are taking one amount for staking that's being sent through the bank module and potentially set a different amount here. This must be `checked_add()?`

contracts/voting-verifier/src/execute.rs
```diff
@@ -58,7 +58,7 @@ pub fn verify_worker_set(
                 source_chain: config.source_chain,
                 source_gateway_address: config.source_gateway_address,
                 confirmation_height: config.confirmation_height,
-                expires_at: env.block.height + config.block_expiry,
+                expires_at: env.block.height.saturating_add(config.block_expiry),
```
cgorenflo, line 61: again, I'd rather use the checked version here because otherwise the expiry interval could have an unexpected length

contracts/voting-verifier/src/execute.rs
```diff
@@ -145,7 +145,7 @@ pub fn verify_messages(
...
```
cgorenflo, line 148: same here

contracts/voting-verifier/src/execute.rs
```diff
@@ -259,7 +259,7 @@ fn create_worker_set_poll(
 ) -> Result<PollId, ContractError> {
     let id = POLL_ID.incr(store)?;
 
-    let poll = WeightedPoll::new(id, snapshot, block_height + expiry, 1);
+    let poll = WeightedPoll::new(id, snapshot, block_height.saturating_add(expiry), 1);
```
cgorenflo, line 262: same here

contracts/voting-verifier/src/execute.rs
```diff
@@ -274,7 +274,7 @@ fn create_messages_poll(
 ) -> Result<PollId, ContractError> {
     let id = POLL_ID.incr(store)?;
 
-    let poll = WeightedPoll::new(id, snapshot, block_height + expiry, poll_size);
+    let poll = WeightedPoll::new(id, snapshot, block_height.saturating_add(expiry), poll_size);
```
cgorenflo, line 277: same here

packages/axelar-wasm-std/src/counter.rs
```diff
@@ -23,7 +21,7 @@ impl<'a, T: AddAssign + Copy + Default + One + Serialize + DeserializeOwned> Cou
 
     pub fn incr(&self, store: &mut dyn Storage) -> StdResult<T> {
         let mut value = self.cur(store);
-        value += T::one();
+        value = value.saturating_add(&T::one());
```
cgorenflo, line 24: since `incr` already returns a result, I think it makes more sense to use a checked_add here, so it doesn't behave in an unexpected manner

packages/axelar-wasm-std/src/threshold.rs
```diff
@@ -127,7 +129,7 @@ impl TryFrom<Threshold> for MajorityThreshold {
     type Error = Error;
 
     fn try_from(value: Threshold) -> Result<Self, Error> {
-        if value.numerator() <= value.denominator() / Uint64::from(2u64) {
+        if value.numerator() <= value.denominator().div(Uint64::from(2u64)) {
```
cgorenflo, line 132: what's the reason for this change? Does this get around the clippy warning?

packages/axelar-wasm-std/src/voting.rs
cgorenflo, line 1: why are arithmetic functions implemented on PollId in the first place? That doesn't make any sense

packages/axelar-wasm-std/src/voting.rs
```diff
@@ -179,10 +188,10 @@ impl Tallies {
     }
 
     pub fn tally(&mut self, vote: &Vote, weight: &Uint256) {
-        *self
-            .0
-            .get_mut(&vote.to_string())
-            .unwrap_or(&mut Uint256::zero()) += weight;
+        let mut binding = Uint256::zero();
```
cgorenflo, line 191:
```suggestion
        let mut zero = Uint256::zero();
```

cgorenflo (changes requested):

ampd/src/broadcaster.rs
```diff
@@ -165,14 +167,16 @@ where
             .change_context(Error::TxBuilding)?;
 
         self.estimate_gas(sim_tx).await.map(|gas| {
-            let gas_adj = (gas as f64 * self.config.gas_adjustment) as u64;
+            let gas_adj = gas as f64 * self.config.gas_adjustment;
 
             Fee::from_amount_and_gas(
                 Coin {
-                    amount: (gas_adj as f64 * self.config.gas_price.amount).ceil() as u128,
+                    amount: cast((gas_adj.mul(self.config.gas_price.amount)).ceil())
```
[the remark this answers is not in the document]
eguajardo: yeah neither `into()` nor `try_into()` work. The `mul` was indeed to get around clippy lint, clippy doesn't really flags multiplication of two floats, because any "overflowing" value is set to `f64::INFINITY` which is a valid value. In this case it's flagged because `amount` is a custom struct even though it behaves like a float `pub struct FiniteAmount(f64)`, so I just basically delegated the lint checking to the `mul` implementation

contracts/multisig-prover/src/execute.rs
```diff
@@ -283,10 +283,12 @@ pub fn should_update_worker_set(
...
         .count()
-        + cur_workers_signers
-            .iter()
-            .filter(|item| !new_workers_signers.contains(item))
-            .count()
+        .saturating_add(
+            cur_workers_signers
+                .iter()
+                .filter(|item| !new_workers_signers.contains(item))
+                .count(),
```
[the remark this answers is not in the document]
eguajardo: I ended up doing something a bit different to avoid unnecessary insertion to an intermediary structure

contracts/rewards/src/contract/execute.rs
eguajardo: tried, but it doesn't work

contracts/rewards/src/contract/execute.rs
eguajardo: tried, but it doesn't work

contracts/rewards/src/contract/execute.rs
eguajardo: `r` it's a mutable reference to the rewards entry, so it mutates rewards. This is why tests don't fail. I have changed the code to make it less confusing

packages/axelar-wasm-std/src/voting.rs
eguajardo: was wondering the same thing, but turns out it's a requirement for the `Counter` struct, it require `One` trait which in turn requires `Mul` trait, also it needs the addition trait to be able to increase the counter

ampd/src/queue/msg_queue.rs
```diff
@@ -9,7 +9,10 @@ pub struct MsgQueue {
 impl MsgQueue {
     pub fn push(&mut self, msg: Any, gas_cost: Gas) {
         self.msgs.push(msg);
-        self.gas_cost += gas_cost;
+        self.gas_cost = self
+            .gas_cost
+            .checked_add(gas_cost)
+            .expect("arithmetic operation overflows");
     }
```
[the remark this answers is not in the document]
eguajardo: I changed it to throw an error instead of panic, however I don't really like changing the signature to consume and return self, I think it's counterintuitive since usually other structs with a function similar to `push()` like Vec and HashMap are expected to be mutable to add new elements. I also think it's not necessary since we are not chaining push to some other functions

contracts/multisig/src/contract/execute.rs
```diff
@@ -29,12 +29,14 @@ pub fn start_signing_session(
...
-            counter += Uint64::one();
+            counter = counter
+                .checked_add(Uint64::one())
+                .map_err(StdError::overflow)?;
             Ok(counter)
         },
     )?;
 
-    let expires_at = env.block.height + config.block_expiry;
+    let expires_at = env.block.height.saturating_add(config.block_expiry);
```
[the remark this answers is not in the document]
eguajardo: in general, when the overflow was not caused by user input from the current transaction, then I lean towards saturation, the reason is that I prefer to let the protocol continue with some value rather than bricking the protocol. In this particular case I think it's more likely that the overflow is caused due to a misconfigured block_expiry (or a bug in the network?) and if that's the case then new signing sessions won't be produced until something is fixed

contracts/rewards/src/state.rs
eguajardo: yeah, at first I was thinking that it doesn't really matter since there was a check before, but this actually can replace that previous check

contracts/voting-verifier/src/execute.rs
eguajardo: I kind of feel the same way as here: https://github.com/axelarnetwork/axelar-amplifier/pull/296#discussion_r1525694763

contracts/voting-verifier/src/execute.rs
eguajardo: I kind of feel the same way as here: https://github.com/axelarnetwork/axelar-amplifier/pull/296#discussion_r1525694763

contracts/voting-verifier/src/execute.rs
eguajardo: I kind of feel the same way as here: https://github.com/axelarnetwork/axelar-amplifier/pull/296#discussion_r1525694763

contracts/voting-verifier/src/execute.rs
eguajardo: I kind of feel the same way as here: https://github.com/axelarnetwork/axelar-amplifier/pull/296#discussion_r1525694763

packages/axelar-wasm-std/src/threshold.rs
eguajardo: yes, I wanted to delegate the handling to the underlying implementation like I did with `FiniteAmount`, in this case is Uint64 which panics when there is an issue during the div, but now that I think about it, I think this is not a good idea with structs external to us like Uint64 because clippy won't be able to detect an issue if implementation changes

ampd/src/queue/msg_queue.rs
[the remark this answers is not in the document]
cgorenflo: even with a fluent interface the struct is mutable, we wouldn't create an immutable copy. Checking the current code, we use loops over pushes in multiple places, the reason there is no chaining is that we currently can't chain because it's interface doesn't allow it. A fluent interface would also allow to use it with `fold()`-like functions

contracts/rewards/src/contract/execute.rs
```diff
@@ -192,12 +205,24 @@ where
...
+            .block_height_started
+            .checked_add(u64::from(new_params.epoch_duration))
+            .ok_or_else(|| {
+                StdError::overflow(OverflowError::new(
+                    OverflowOperation::Add,
+                    cur_epoch.block_height_started,
+                    new_params.epoch_duration,
+                ))
+            })
+            .map_err(ContractError::from)?
```
cgorenflo, line 218: why is this double conversion necessary when there is already a `ContractError::Overflow`?

contracts/rewards/src/error.rs
```diff
@@ -1,8 +1,12 @@
 use axelar_wasm_std_derive::IntoContractError;
+use cosmwasm_std::{OverflowError, StdError};
 use thiserror::Error;
 
 #[derive(Error, Debug, PartialEq, IntoContractError)]
 pub enum ContractError {
+    #[error(transparent)]
+    Std(#[from] StdError),
```
cgorenflo, line 8: if at all possible, I'd like to avoid a catchall Std error below the root level. This makes it too easy to default wrapping everything in this which contradicts the purpose of error-stack to add context to the errror

contracts/multisig/src/contract/execute.rs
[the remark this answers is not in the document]
cgorenflo: a wrong configuration will lead to potentially wrong votes from this ampd instance, so the verifier could get penalized and worst case thrown out without being aware of what the problem was

contracts/voting-verifier/src/execute.rs
cgorenflo: see my response there

contracts/voting-verifier/src/execute.rs
cgorenflo: see my response there

contracts/voting-verifier/src/execute.rs
cgorenflo: see my response there

contracts/voting-verifier/src/execute.rs
cgorenflo: see my response there

packages/axelar-wasm-std/src/voting.rs
cgorenflo: omg, that sucks. Could you add comments on those trait implementations to explain that they are necessary to satisfy the `One` trait?

ampd/src/queue/msg_queue.rs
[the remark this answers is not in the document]
fish-sammy: I am OK with `push` consuming self and therefore enabling chaining, but I am not sure what difference panic v.s. returning error here makes?

contracts/multisig-prover/src/execute.rs
```diff
@@ -274,27 +277,13 @@ pub fn should_update_worker_set(
...
-            .filter(|item| !new_workers_signers.contains(item))
-            .count()
-        > max_diff
+    signers_difference_count(&new_workers.signers, &cur_workers.signers).saturating_add(
+        signers_difference_count(&cur_workers.signers, &new_workers.signers),
+    ) > max_diff
+}
+
+fn signers_difference_count(s1: &BTreeMap<String, Signer>, s2: &BTreeMap<String, Signer>) -> usize {
+    s1.values().filter(|v| !s2.values().contains(v)).count()
```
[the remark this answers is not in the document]
eguajardo: what do you mean is symetric? it's not. https://en.wikipedia.org/wiki/Complement_(set_theory)#Relative_complement. Here is how is described for BTreeSet: https://doc.rust-lang.org/std/collections/struct.BTreeSet.html#method.difference

ampd/src/queue/msg_queue.rs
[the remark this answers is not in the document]
eguajardo: could we leave the refactoring to a fluent interface struct out of scope of this PR?

contracts/multisig-prover/src/execute.rs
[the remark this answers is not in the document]
cgorenflo: you are right, my bad. I could've sworn that what is defined as symmetric difference is just called difference

contracts/multisig-prover/src/execute.rs
```diff
@@ -274,27 +277,13 @@ pub fn should_update_worker_set(
...
-        .iter()
-        .filter(|item| !cur_workers_signers.contains(item))
-        .count()
-        + cur_workers_signers
-            .iter()
-            .filter(|item| !new_workers_signers.contains(item))
-            .count()
-        > max_diff
+    signers_difference_count(&new_workers.signers, &cur_workers.signers).saturating_add(
+        signers_difference_count(&cur_workers.signers, &new_workers.signers),
```
cgorenflo, line 281: extract this into `signers_symmetric_difference_count` to make it easier to understand the condition

cgorenflo (approved):

PR #297, opened 2024-03-07
refactor: contract trait for integration tests

maancham:
## Description

AXE-3318

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


integration-tests/tests/test_utils/trait_mod.rs
```diff
@@ -0,0 +1,347 @@
...
+use multisig::key::KeyType;
+use multisig_prover::encoding::Encoder;
+use service_registry::contract::{execute, instantiate, query};
+
+#[derive(Clone)]
+pub struct ServiceRegistryContract {
+    pub contract_addr: Addr,
+}
+
+impl ServiceRegistryContract {
```
cjcobb23, line 16: Do we want to actually add the execution msgs and query msgs here as methods? So someone can just call something like `contract.get_active_workers(service_name, chain_name)` as a function?

integration-tests/tests/test_utils/trait_mod.rs
```diff
@@ -0,0 +1,347 @@
+use axelar_wasm_std::nonempty;
```
cjcobb23, line 1: I don't how I feel about all of the trait implementations for every contract existing in this one file. Can we move each to it's own file?

cjcobb23: Also, should the trait implementations live under the src folder? With the trait definition itself?

maancham: I will move each implementation to its own contract.
About the second question, there were other options but I talked with Chris and decided to put it there. We can change of course.

integration-tests/tests/test_utils/trait_mod.rs
```diff
@@ -0,0 +1,7 @@
+pub mod connection_router_trait;
+pub mod gateway_trait;
+pub mod multisig_prover_trait;
+pub mod multisig_trait;
+pub mod rewards_trait;
+pub mod service_registry_trait;
+pub mod voting_verifier_trait;
```
cjcobb23, line 7: Why do we need a file just for these? Why not just put these in mod.rs?

integration-tests/tests/test_utils/trait_mod/connection_router_trait.rs
```diff
@@ -0,0 +1,45 @@
+use cosmwasm_std::Addr;
```
cjcobb23, line 1: I'd prefer a different naming convention for these files. How about just calling it connection_router_contract.rs? And can we just put them all under src, next to the contract.rs file?

integration-tests/tests/test_utils/trait_mod.rs
maancham: i was trying not to over populate the original mod file. Will refactor.

integration-tests/tests/test_utils/trait_mod/connection_router_trait.rs
maancham: ok sure.

cjcobb23 (approved):

PR #308, opened 2024-03-14
fix: use cosmwasm-std 1.3

milapsheth:
## Description

AXE-3409

Old amplifier version also used a newer dependency than what's specified in [Cargo.toml](https://github.com/axelarnetwork/axelar-amplifier/blob/0be4b7d0d33303d4eecb7405a894b6304ddcecc2/Cargo.lock#L1677)

Need to look into how we can pin it so it doesn't auto use a new version

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -50,9 +50,9 @@ dependencies = [
 
 [[package]]
 name = "aes"
-version = "0.8.3"
+version = "0.8.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ac1f845298e95f983ff1944b728ae08b8cebab80d684f0a832ed0fc74dfa27e2"
+checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
 dependencies = [
  "cfg-if",
  "cipher",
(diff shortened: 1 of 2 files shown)


Cargo.lock
```diff
@@ -1666,22 +1675,22 @@ dependencies = [
...
- "quote 1.0.33",
+ "proc-macro2 1.0.79",
+ "quote 1.0.35",
  "syn 1.0.109",
 ]
 
 [[package]]
 name = "cosmwasm-std"
-version = "1.5.3"
+version = "1.5.2"
```
milapsheth, line 1689: it's not using the pinned version for some reason. it's not clear which dependency is causing resolving to the newer version yet (as per cargo tree it should use the workspace version)

Cargo.toml
```diff
@@ -7,9 +7,9 @@ rust-version = "1.73.0"
 
 [workspace.dependencies]
 connection-router = { version = "^0.1.0", path = "contracts/connection-router" }
-cosmwasm-std = "1.5.3"
-cosmwasm-schema = "1.3.3"
-cosmwasm-storage = "1.3.3"
+cosmwasm-std = "1.3.4"
+cosmwasm-schema = "1.3.4"
+cosmwasm-storage = "1.3.4"
```
eguajardo, line 12: should we add a note here explaining we should be careful updating these dependencies and why?

cgorenflo:
fixed [here](https://github.com/axelarnetwork/axelar-amplifier/pull/343)

PR #311, opened 2024-03-15
feat(minor-multisig-prover)!: allow governance to confirm workerset

cjcobb23:
https://axelarnetwork.atlassian.net/browse/AXE-2958


contracts/multisig-prover/src/execute.rs
```diff
@@ -241,22 +241,25 @@ pub fn update_worker_set(deps: DepsMut, env: Env) -> Result<Response, ContractEr
...
+            deps.querier.query(&QueryRequest::Wasm(WasmQuery::Smart {
+                contract_addr: config.voting_verifier.to_string(),
+                msg: to_json_binary(&query)?,
+            }))?;
 
-    if status != VerificationStatus::SucceededOnChain {
-        return Err(ContractError::WorkerSetNotConfirmed);
+        if status != VerificationStatus::SucceededOnChain {
+            return Err(ContractError::WorkerSetNotConfirmed);
+        }
```
cgorenflo, line 262: extract this out into something like `ensure_verification()`

contracts/multisig-prover/src/execute.rs
```diff
@@ -241,22 +241,25 @@ pub fn update_worker_set(deps: DepsMut, env: Env) -> Result<Response, ContractEr
     }
 }
 
-pub fn confirm_worker_set(deps: DepsMut) -> Result<Response, ContractError> {
+pub fn confirm_worker_set(deps: DepsMut, info: MessageInfo) -> Result<Response, ContractError> {
```
cgorenflo, line 244: only the sender is necessary, so I would prefer just the sender as the function argument instead of `MessageInfo`

fish-sammy: Agree with @cgorenflo

contracts/multisig-prover/src/execute.rs
fish-sammy: `verify_worker_set_eligibility`

contracts/multisig-prover/src/state.rs
```diff
@@ -12,6 +12,7 @@ use crate::types::{BatchId, CommandBatch};
 #[cw_serde]
 pub struct Config {
     pub admin: Addr,
+    pub governance: Addr,
```
fish-sammy, line 15: Now that we have the contracts deployed already on the testnet, doesn't this require some migration?

cjcobb23: yeah if we want to upgrade, as oppose to redeploy. The prover can pretty easily just be redeployed.

contracts/multisig-prover/src/msg.rs
```diff
@@ -42,6 +42,11 @@ pub enum QueryMsg {
     GetWorkerSet,
 }
 
+#[cw_serde]
+pub enum MigrateMsg {
+    UpdateConfig { new_config: InstantiateMsg },
+}
```
cgorenflo, line 48: I don't think enum is a good fit, because there is never going to be more than one allowed migration msg at a time. The way I see it, enums should only be used if there are multiple equally valid variant or it's clear the type needs to be expanded in the future.

cjcobb23: That's generally true, but if we just allow arbitrary updates to the config via migration, then that is always allowed, even if other migrations are allowed too. Updating the config can happen any number of times.

contracts/multisig-prover/src/contract.rs
```diff
@@ -131,14 +133,14 @@ mod tests {
         )
     }
 
-    fn confirm_worker_set(test_case: &mut TestCaseConfig) -> Result<AppResponse, Error> {
+    fn confirm_worker_set(
+        test_case: &mut TestCaseConfig,
```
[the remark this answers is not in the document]
cjcobb23: would take a decent amount of refactoring to use that here, since the trait is coupled with the integration tests. The multisig prover tests need a lot of refactoring, and much of it should be moved to the integration tests package (there are still a lot of mocks here). I'd rather leave this as is for now, and tackle refactoring these tests as a separate task.

contracts/multisig-prover/src/contract.rs
```diff
@@ -498,6 +524,39 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_governance_should_confirm_worker_set_without_verification() {
```
cgorenflo, line 528: do we have a test for non-governance to ensure verification is done in that case?

cjcobb23: yeah the above two tests are for the non-governance case. `test_confirm_worker_set_unconfirmed` tries to confirm as non-governance without verification and expects a failure. And then `test_confirm_worker_set` executes the full happy path flow and confirms a worker set as non governance once the worker set is verified by the voting verifier.

cgorenflo (approved):

PR #314, opened 2024-03-18
feat(minor-voting-verifier)!: use transaction log index instead of block log index to identify EVM events

fish-sammy:
## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,7 @@ serde_json = "1.0.89"
 schemars = "0.8.10"
 sha3 = { version = "0.10.8", default-features = false, features = [] }
 signature-verifier-api = { version = "^0.1.0", path = "packages/signature-verifier-api" }
+cast = "0.3.0"
 
 [workspace.lints.clippy]
 arithmetic_side_effects = "deny"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,6 @@ serde_json = "1.0.89"
 schemars = "0.8.10"
 sha3 = { version = "0.10.8", default-features = false, features = [] }
 signature-verifier-api = { version = "^0.1.0", path = "packages/signature-verifier-api" }
-cast = "0.3.0"
 
 [workspace.lints.clippy]
 arithmetic_side_effects = "deny"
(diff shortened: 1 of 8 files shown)


ampd/src/evm/verifier.rs
```diff
@@ -1,4 +1,5 @@
 use axelar_wasm_std::voting::Vote;
+use cast::usize;
```
cgorenflo, line 2: where is it used?

contracts/nexus-gateway/src/nexus.rs
```diff
@@ -25,6 +25,8 @@ pub struct Message {
 
 impl CustomMsg for Message {}
 
+// it's parsed into u64 instead of u32 (https://github.com/axelarnetwork/axelar-amplifier/blob/bf0b3049c83e540989c7dad1c609c7e2ef6ed2e5/contracts/voting-verifier/src/events.rs#L162)
+// here in order to match the message type defined in the nexus module
```
cgorenflo, line 29:
```suggestion
// here in order to match the message type defined in the nexus module. Changing nexus to use u32 instead is not worth the effort.
```

Cargo.toml
```diff
@@ -35,6 +35,7 @@ serde_json = "1.0.89"
 schemars = "0.8.10"
 sha3 = { version = "0.10.8", default-features = false, features = [] }
 signature-verifier-api = { version = "^0.1.0", path = "packages/signature-verifier-api" }
+cast = "0.3.0"
```
cgorenflo, line 38: do we need both `cast` and `num-traits`?

cgorenflo (approved):

ampd/src/evm/verifier.rs
fish-sammy: https://github.com/axelarnetwork/axelar-amplifier/blob/34d4a541897c1a53a6f54c9b2f2be4487d619ebb/ampd/src/evm/verifier.rs#L75

Cargo.toml
fish-sammy: oh I didn't know we had `num-traits`. Let me check.

PR #318, opened 2024-03-20
fix(minor)!: change evm msg id and chain name to match core

cjcobb23:
use dash instead of colon to separate tx hash and event index
disallow underscore in chain names

https://axelarnetwork.atlassian.net/browse/AXE-3410


contracts/voting-verifier/src/events.rs
```diff
@@ -177,6 +181,10 @@ fn parse_message_id(
     ))
 }
 
+pub fn construct_message_id(tx_hash: String, event_index: u32) -> String {
```
cgorenflo, line 184: this should accept `Hash` instead of `String`, since you are doing the conversion for the index in here anyway

contracts/voting-verifier/src/events.rs
```diff
@@ -177,6 +181,10 @@ fn parse_message_id(
...
+    [tx_hash, event_index.to_string()].join(&TX_HASH_EVENT_INDEX_SEPARATOR.to_string())
+}
```
cgorenflo, line 186: why use this `join` operation instead of the `format!` macro?

contracts/voting-verifier/src/query.rs
```diff
@@ -197,11 +198,13 @@ mod tests {
...
-                chain: "source_chain".parse().unwrap(),
-                id: format!("id:{id}").parse().unwrap(),
+                chain: "source-chain".parse().unwrap(),
+                id: format!("id{TX_HASH_EVENT_INDEX_SEPARATOR}{id}")
+                    .parse()
+                    .unwrap(),
             },
             source_address: format!("source_address{id}").parse().unwrap(),
-            destination_chain: format!("destination_chain{id}").parse().unwrap(),
+            destination_chain: format!("destination-chain{id}").parse().unwrap(),
```
cgorenflo, line 207: why change this?

packages/connection-router-api/src/primitives.rs
```diff
@@ -162,7 +162,7 @@ impl FromStr for ChainName {
     type Err = Error;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        if s.contains(ID_SEPARATOR) || s.is_empty() {
+        if s.contains(CHAIN_NAME_DELIMITER) || s.is_empty() {
```
cgorenflo, line 165: just as confirmation of my assumption: it's fine to have "_" in the ID? More specifically, the only thing we restrict for the ID is that it's not empty, any other restriction is chain specific

cgorenflo (approved):

contracts/voting-verifier/src/query.rs
cjcobb23: we disallow underscore now, to match core

packages/connection-router-api/src/primitives.rs
cjcobb23: yes that's correct, we just restrict non-empty for the ID and everything else is chain specific

PR #321, opened 2024-03-21
feat(minor-rewards): add query to retrieve rewards pool details

eguajardo:
## Description

https://axelarnetwork.atlassian.net/browse/AXE-3388

## Todos

For another PR: `Store` trait needs refactoring, it requires mutable storage, however there are a few read functions that could be reused for queries where the contract only has access to immutable storage. 

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


contracts/rewards/src/msg.rs
```diff
@@ -60,4 +61,15 @@ pub enum ExecuteMsg {
...
+    #[returns(RewardsPoolResponse)]
+    GetRewardsPool { pool_id: PoolId },
+}
+
+#[cw_serde]
+pub struct RewardsPoolResponse {
+    pub balance: Uint128,
+    pub epoch_duration: Uint64,
+    pub rewards_per_epoch: Uint128,
+}
```
cjcobb23, line 75: Can we just return the `RewardsPool` object? We should also return the watermark. And for the params, can we just return the `Params` object?

eguajardo: I was wondering if we should really expose our inner structs since any future change could potentially create a breaking change without us knowing. What do you think?

contracts/rewards/src/contract/query.rs
```diff
@@ -0,0 +1,186 @@
+use cosmwasm_std::Storage;
+use error_stack::Result;
+
+use crate::{
+    error::ContractError,
+    msg::RewardsPoolResponse,
+    state::{self, Epoch, PoolId},
+};
+
+pub fn get_rewards_pool(
```
cgorenflo, line 10:
```suggestion
pub fn rewards_pool(
```
rust convention is to just use the name of the thing for getters

contracts/rewards/src/contract/query.rs
```diff
@@ -0,0 +1,186 @@
...
+    block_height: u64,
+) -> Result<RewardsPoolResponse, ContractError> {
+    let pool = state::load_rewards_pool(storage, pool_id.clone())?;
+    let stored_params = state::load_params(storage);
+    let cur_epoch = Epoch::current(&stored_params, block_height)?;
+
+    let params = match state::load_epoch_tally(storage, pool_id, cur_epoch.epoch_num)? {
+        Some(epoch_tally) => epoch_tally.params,
+        None => stored_params.params,
+    };
```
cgorenflo, line 22: this might be a contentious take and we can discuss it with the team, but I would prefer not to reuse the stored params to make this function easier to read:
- get the pool
- get the current epoch
- try to load the tally to get the params, if that fails, load the latest params

reusing the params can potentially save you a storage access, but it gets kind of confusing to understand why which params is used

contracts/rewards/src/msg.rs
cgorenflo: I generally agree that DTOs should be disconnected from internal types most of the time. In this case the query response aggregates information from many different types, so we need a separate type anyway. I like the addition of the watermark (give it a different name so people understand what it means when they query it). I don't like the `RewardsPoolResponse` name though. I would prefer just `RewardsPool` defined here, and I'm also open to better suggestions

contracts/rewards/src/state.rs
```diff
@@ -178,6 +178,39 @@ pub struct Epoch {
     pub block_height_started: u64,
 }
 
+impl Epoch {
+    /// Returns the current epoch. The current epoch is computed dynamically based on the current
+    /// block height and the epoch duration. If the epoch duration is updated, we store the epoch
+    /// in which the update occurs as the last checkpoint
+    pub fn current(
+        stored_params: &StoredParams,
```
cgorenflo, line 186: Could we rename `StoredParams` to `LatestParams`? I find that name very confusing and doesn't really tell you anything about it's content

contracts/rewards/src/state.rs
```diff
@@ -363,6 +394,30 @@ impl<T> Deref for StorageState<T> {
...
+pub(crate) fn load_epoch_tally(
+    storage: &dyn Storage,
+    pool_id: PoolId,
+    epoch_num: u64,
+) -> Result<Option<EpochTally>, ContractError> {
+    TALLIES
+        .may_load(storage, TallyId { pool_id, epoch_num })
+        .change_context(ContractError::LoadEpochTally)
+}
+
```
cgorenflo, line 420: good first step to make it easier to delete the Store in a (separate) refactoring

cgorenflo (commented):
rename the commit (important! not just the PR title) `feat(minor-rewards): add query to retrieve rewards pool details`
adding a new feature is a minor version bump

contracts/rewards/src/msg.rs
cjcobb23: that's a good point, internal structs might change, or contain information the user doesn't need/want. We should return the watermark though. It can just be called `last_distributed` or some such.

As to the name, either `RewardsPool` or `RewardsPoolInfo` is fine with me. Or just `PoolInfo`

contracts/rewards/src/contract/query.rs
eguajardo: I would rather add a comment here to do some clarification than having an extra storage access

contracts/rewards/src/state.rs
eguajardo: renamed to `ParamsSnapshot` and `last_updated` to `created_at` after discussion

contracts/rewards/src/msg.rs
```diff
@@ -60,4 +61,16 @@ pub enum ExecuteMsg {
...
+    #[returns(RewardsPool)]
+    RewardsPool { pool_id: PoolId },
+}
+
+#[cw_serde]
+pub struct RewardsPool {
+    pub balance: Uint128,
+    pub epoch_duration: Uint64,
+    pub rewards_per_epoch: Uint128,
+    pub last_distribution_epoch: Option<Uint64>,
```
cjcobb23, line 75: I think we need the current epoch in here too, so that way a caller knows whether rewards need to be distributed. Without knowing the current epoch, I don't think the last distributed epoch is very useful.

contracts/rewards/src/contract/query.rs
```diff
@@ -0,0 +1,240 @@
...
+    let pool = state::load_rewards_pool(storage, pool_id.clone())?;
+    let current_params = state::load_params(storage);
+    let cur_epoch = Epoch::current(&current_params, block_height)?;
+
+    // the params could have been updated since the tally was created. Therefore we use the params from the
+    // active tally if it exists, otherwise we use the latest stored params.
+    let params = match state::load_epoch_tally(storage, pool_id.clone(), cur_epoch.epoch_num)? {
+        Some(epoch_tally) => epoch_tally.params,
+        None => current_params.params,
+    };
```
cjcobb23, line 24: Why do we want this? Don't we just want the latest params? I feel like a caller would prefer the latest info, instead of anything stale

cjcobb23: Actually, the caller probably most likely wants to know info for the current epoch, so I think what you have here is good.

cjcobb23 (approved):

contracts/rewards/src/contract/query.rs
cgorenflo: yes, the "current_params" might still be inactive if they didn't overwrite the epoch duration, so the way it's done here is correct

cgorenflo (approved):

PR #322, opened 2024-03-21
feat: add base structure of monitoring contract

maancham:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

### DIFF
--- a/contracts/monitoring/.cargo/config
+++ b/contracts/monitoring/.cargo/config
@@ -1,0 +1,4 @@
+[alias]
+wasm = "build --release --lib --target wasm32-unknown-unknown"
+unit-test = "test --lib"
+schema = "run --bin schema"
--- a/contracts/monitoring/.cargo/config
+++ b/contracts/monitoring/.cargo/config
@@ -1,4 +1,0 @@
-[alias]
-wasm = "build --release --lib --target wasm32-unknown-unknown"
-unit-test = "test --lib"
-schema = "run --bin schema"
--- a/contracts/monitoring/Cargo.toml
+++ b/contracts/monitoring/Cargo.toml
@@ -1,0 +1,52 @@
+[package]
+name = "monitoring"
+version = "0.1.0"
+rust-version = { workspace = true }
+edition = "2021"
+description = "Contract to provide chains support and verifier statuses"
+
+exclude = [
+    # Those files are rust-optimizer artifacts. You might want to commit them for convenience, but they should not be part of the source code publication.
+    "contract.wasm",
+    "hash.txt",
+]
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[lib]
+crate-type = ["cdylib", "rlib"]
+
+[[bin]]
+name = "monitoring-schema"
+path = "src/bin/schema.rs"
+
+[features]
+# for more explicit tests, cargo test --features=backtraces
+backtraces = ["cosmwasm-std/backtraces"]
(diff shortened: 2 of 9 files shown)


contracts/monitoring/.cargo/config
fish-sammy, line 1: I do not think this is needed.

contracts/monitoring/src/msg.rs
```diff
@@ -0,0 +1,16 @@
...
+pub struct InstantiateMsg {}
+
+#[cw_serde]
+pub enum ExecuteMsg {}
+
+#[cw_serde]
+#[derive(QueryResponses)]
+pub enum QueryMsg {
+    #[returns(Vec<(ChainName, WorkerSet)>)]
+    GetChainActiveWorkerSet { chains: Vec<ChainName> },
```
fish-sammy, line 15:
```suggestion
    #[returns(WorkerSet)]
    GetChainActiveWorkerSet { chain: ChainName },
```
I prefer working with APIs that don't batch.

contracts/monitoring/src/query.rs
```diff
@@ -0,0 +1,10 @@
+use connection_router_api::ChainName;
+use cosmwasm_std::Deps;
+use multisig::worker_set::WorkerSet;
+
+pub fn chains_active_worker_sets(
+    _deps: Deps,
+    _chains: &[ChainName],
+) -> Vec<(ChainName, WorkerSet)> {
+    Vec::new()
```
fish-sammy, line 9:
```suggestion
    todo!()
```

contracts/monitoring/src/lib.rs
```diff
@@ -0,0 +1,3 @@
+mod contract;
+pub mod msg;
+pub mod query;
```
fish-sammy, line 3: Why `pub`?

contracts/monitoring/src/msg.rs
cjcobb23, line 15: I think we should change worker to verifiers here, since we are now calling them verifiers. We should do this across the codebase, but definitely should not add new code that still uses workers.
```suggestion
    GetChainActiveVerifiers { chains: Vec<ChainName> },
```

contracts/monitoring/.cargo/config
cgorenflo: correct, the definitions of the workspace should be sufficient

contracts/monitoring/Cargo.toml
```diff
@@ -0,0 +1,52 @@
+[package]
+name = "monitoring"
+version = "0.1.0"
+rust-version = { workspace = true }
+edition = "2021"
+description = "Contract to provide chains support and verifier statuses"
```
cgorenflo, line 6: that's a bit of an odd description 😄

contracts/monitoring/src/msg.rs
cgorenflo, line 15:
```suggestion
    GetActiveVerifiersForChain{ chain: ChainName },
```
we might want to add batching later, but for the beginning I'm with Sammy in keeping it simple

contracts/monitoring/Cargo.toml
maancham: I changed it

contracts/monitoring/src/lib.rs
maancham: following the convention of other contracts. Also, can not use rust files (like msg) in schema.rs if they are private.

cgorenflo (approved):

PR #326, opened 2024-03-28
feat(minor-connection-router): paginated chains query

haiyizxx:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


contracts/connection-router/src/contract/query.rs
```diff
@@ -1,17 +1,41 @@
...
 use connection_router_api::error::Error;
 use connection_router_api::{ChainEndpoint, ChainName};
-use cosmwasm_std::Deps;
 use error_stack::{Result, ResultExt};
 
 use crate::state::chain_endpoints;
 
+// Pagination limits
+const MAX_LIMIT: u32 = 50;
+const DEFAULT_LIMIT: u32 = 10;
```
fish-sammy, line 12: Why not default to max u32?

packages/connection-router-api/src/msg.rs
```diff
@@ -48,4 +48,10 @@ pub enum ExecuteMsg {
 pub enum QueryMsg {
     #[returns(ChainEndpoint)]
     GetChainInfo(ChainName),
+
+    #[returns(Vec<ChainEndpoint>)]
+    Chains {
+        start_after: Option<ChainName>,
+        limit: Option<u32>,
+    },
```
cgorenflo, line 56: add documentation about defaults

packages/connection-router-api/src/msg.rs
```diff
@@ -48,4 +48,10 @@ pub enum ExecuteMsg {
 pub enum QueryMsg {
     #[returns(ChainEndpoint)]
     GetChainInfo(ChainName),
+
+    #[returns(Vec<ChainEndpoint>)]
```
cjcobb23, line 52: Do we want to actually return what should be passed to start_after? Right now the caller needs to extract the last chain from the returned vector and pass it back in. Might be easier to just return a somewhat opaque marker that the caller can just blindly pass into the next call.

haiyizxx: that's what cosmwasm range query exposed, the api accepts a start key and end key. need to re-implement it otherwise.

contracts/connection-router/src/contract/query.rs
fish-sammy: @cgorenflo If the calling contract isn't aware of that pagination and the default is 10, it will not get all the results which may lead to faulty behaviour. If the default is max u32, the calling contract would run out of gas. Either way requires a re-deployment of the calling contract to fix the issue so I am not sure what is it to be avoided here. Former sounds worse to me.

cgorenflo: fair point, lets go with max u32 then

packages/connection-router-api/src/msg.rs
cgorenflo: yeah I understand why they implemented it this way, but I don't like it. I think we'll have to live with it though. And since the query can be called without pagination I don't think we should return a dedicated "bookmark" object. So I suggest leaving the return value as is

contracts/connection-router/src/contract/query.rs
cgorenflo: in that case I think both should be set to max u32

contracts/connection-router/src/contract/query.rs
```diff
@@ -51,4 +75,54 @@ mod test {
...
+        // with page
+        let result =
+            super::chains(deps.as_ref(), Some("c-chain".parse().unwrap()), Some(2)).unwrap();
+        assert_eq!(result.len(), 1);
+        assert_eq!(result, vec![endpoints[3].clone()]);
+
+        // start after last chain
+        let result =
+            super::chains(deps.as_ref(), Some("d-chain".parse().unwrap()), Some(2)).unwrap();
+        assert_eq!(result.len(), 0);
```
cgorenflo, line 126: can you add an "e-chain" test? just to be safe that it works with data that's completely out of scope

contracts/connection-router/src/contract/query.rs
haiyizxx: removed the max as default to max u32, can add back if needed in the future

cgorenflo (approved):

PR #331, opened 2024-04-01
chore(multisig): test PublicKey is validated to represent a valid point in the curve

eguajardo:
## Description

There is a warning in function `ed25519_dalek::VerifyingKey::from_bytes` that says:

> The caller is responsible for ensuring that the bytes passed into this method actually represent a curve25519_dalek::curve::CompressedEdwardsY and that said compressed point is actually a point on the curve.

However upon further inspection it doesn't seem to be possible to generate a `VerifyingKey` that represents an invalid point in the curve using `from_bytes` function, because it will throw an error when trying to decompress into an invalid point.

This PR adds unit tests to ensure our `validate_and_normalize_public_key` doesn't allow the creation of public keys that represents invalid points in the curves

https://axelarnetwork.atlassian.net/browse/AXE-3312

## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

### DIFF
--- a/contracts/multisig/src/key.rs
+++ b/contracts/multisig/src/key.rs
@@ -235,7 +235,6 @@ fn validate_and_normalize_public_key(
     key_type: KeyType,
     pub_key: HexBinary,
 ) -> Result<HexBinary, ContractError> {
-    // TODO: ensure proper validation, since external crates may only check format.
     match key_type {
         KeyType::Ecdsa => Ok(k256::PublicKey::from_sec1_bytes(
             check_ecdsa_format(pub_key)?.as_slice(),
@@ -245,6 +244,8 @@ fn validate_and_normalize_public_key(
         .as_ref()
         .into()),
         // TODO: verify encoding scheme is standard
+        // Function `from_bytes()` will internally decompress into an EdwardsPoint which can only represent a valid point in the curve
+        // See https://docs.rs/curve25519-dalek/latest/curve25519_dalek/edwards/index.html#validity-checking
         KeyType::Ed25519 => Ok(ed25519_dalek::VerifyingKey::from_bytes(
             pub_key
                 .as_slice()
@@ -320,8 +321,14 @@ impl From<PublicKey> for HexBinary {
 #[cfg(test)]
 mod ecdsa_tests {
     use cosmwasm_std::HexBinary;
+    use k256::{elliptic_curve::sec1::FromEncodedPoint, AffinePoint, EncodedPoint};
 
-    use crate::{key::Signature, test::common::ecdsa_test_data, types::MsgToSign, ContractError};
+    use crate::{
+        key::{validate_and_normalize_public_key, Signature},
+        test::common::ecdsa_test_data,
+        types::MsgToSign,
+        ContractError,
+    };
 
     use super::{KeyType, PublicKey};
 
@@ -366,6 +373,26 @@ mod ecdsa_tests {
             PublicKey::try_from((KeyType::Ecdsa, uncompressed_pub_key.clone())).unwrap_err(),
             ContractError::InvalidPublicKey
         );
+    }
+
+    #[test]
+    fn validate_ecdsa_public_key_not_in_curve() {
+        // Valid compressed point format, which should produce an invalid point when decompressed
+        let invalid_compressed_point = EncodedPoint::from_bytes([
+            3, 132, 180, 161, 194, 115, 211, 43, 90, 122, 205, 26, 76, 14, 117, 209, 243, 206, 192,
+            34, 107, 93, 142, 13, 50, 95, 115, 188, 140, 82, 194, 140, 235,
+        ])
+        .unwrap();
+
+        assert!(Into::<bool>::into(
+            AffinePoint::from_encoded_point(&invalid_compressed_point).is_none()
+        ));
+
+        let result = validate_and_normalize_public_key(
+            KeyType::Ecdsa,
+            HexBinary::from(invalid_compressed_point.as_bytes()),
+        );
+        assert_eq!(result.unwrap_err(), ContractError::InvalidPublicKey);
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


contracts/multisig/src/key.rs
```diff
@@ -245,6 +244,8 @@ fn validate_and_normalize_public_key(
         .as_ref()
         .into()),
         // TODO: verify encoding scheme is standard
+        // Function `from_bytes()` will internally decompress into an EdwardsPoint which can only represent a valid point in the curve
```
cgorenflo, line 247:
```suggestion
        // Function `from_bytes()` will internally decompress into an EdwardsPoint which can only represent a valid point on the curve
```

contracts/multisig/src/key.rs
```diff
@@ -368,6 +375,26 @@ mod ecdsa_tests {
         );
     }
 
+    #[test]
+    fn validate_ecdsa_public_key_not_in_curve() {
```
cgorenflo, line 379:
```suggestion
    fn validate_ecdsa_public_key_not_on_curve() {
```

contracts/multisig/src/key.rs
```diff
@@ -368,6 +375,26 @@ mod ecdsa_tests {
...
+        // Valid compressed point format, which should produce an invalid point when decompressed
+        let invalid_compressed_point = EncodedPoint::from_bytes([
+            3, 132, 180, 161, 194, 115, 211, 43, 90, 122, 205, 26, 76, 14, 117, 209, 243, 206, 192,
+            34, 107, 93, 142, 13, 50, 95, 115, 188, 140, 82, 194, 140, 235,
+        ])
+        .unwrap();
+
+        assert!(Into::<bool>::into(
+            AffinePoint::from_encoded_point(&invalid_compressed_point).is_none()
+        ));
```
cgorenflo, line 389: why this is this `into` necessary?

contracts/multisig/src/key.rs
```diff
@@ -245,6 +244,8 @@ fn validate_and_normalize_public_key(
         .as_ref()
         .into()),
         // TODO: verify encoding scheme is standard
```
cgorenflo, line 246: remove this todo?

contracts/multisig/src/key.rs
```diff
@@ -368,6 +375,26 @@ mod ecdsa_tests {
...
+    #[test]
+    fn validate_ecdsa_public_key_not_in_curve() {
+        // Valid compressed point format, which should produce an invalid point when decompressed
+        let invalid_compressed_point = EncodedPoint::from_bytes([
+            3, 132, 180, 161, 194, 115, 211, 43, 90, 122, 205, 26, 76, 14, 117, 209, 243, 206, 192,
+            34, 107, 93, 142, 13, 50, 95, 115, 188, 140, 82, 194, 140, 235,
+        ])
+        .unwrap();
+
+        assert!(Into::<bool>::into(
```
cgorenflo, line 387: what does this assertion do?

contracts/multisig/src/key.rs
```diff
@@ -499,14 +532,31 @@ mod ed25519_tests {
...
         let hex = HexBinary::from_hex("049b").unwrap();
         assert_eq!(
             PublicKey::try_from((KeyType::Ed25519, hex.clone())).unwrap_err(),
             ContractError::InvalidPublicKey
         );
     }
 
+    #[test]
+    fn validate_ed25519_public_key_not_in_curve() {
+        // Valid compressed point format, which should produce an invalid point when decompressed
```
cgorenflo, line 545: why is it valid compressed but not decompressed?

contracts/multisig/src/key.rs
eguajardo: in this case `is_none()` doesn't return a bool, it returns `Choice`, so I had to convert

contracts/multisig/src/key.rs
eguajardo: this TODO corresponds to another task: https://axelarnetwork.atlassian.net/browse/AXE-3313. cc @milapsheth do we still need this?

contracts/multisig/src/key.rs
eguajardo: it's not strictly necessary, it was more useful during the development of this test, but I wanted to make sure the crate does indeed catch a point is invalid, both asserts must match otherwise our function is not doing the same internally

contracts/multisig/src/key.rs
eguajardo: the compressed format is a single coordinate and ed25519 accepts any 32 bytes, doesn't really say it's a valid point or not on the curve, but once you substitute `y` in the curve equation, then `x` would get you an invalid number like square root of a negative number

contracts/multisig/src/key.rs
cgorenflo: would this work instead?
```suggestion
        assert!(AffinePoint::try_from(&invalid_compressed_point).is_err());
```

contracts/multisig/src/key.rs
cgorenflo: could you add a comment explaining the reasoning?

contracts/multisig/src/key.rs
cgorenflo: I don't think I get it. The decompressed format is completely derived from the compressed one without extra information. So how is it possible that the compressed version could be valid?

eguajardo: The only thing valid is the format, ed25519 spec says any 32 bytes array is accepted, the crate doesn't really do any extra validation when creating a `CompressedEdwardsY`, but it will fail trying to decompress.

contracts/multisig/src/key.rs
```diff
@@ -368,6 +375,26 @@ mod ecdsa_tests {
         );
     }
 
+    #[test]
+    fn validate_ecdsa_public_key_not_on_curve() {
+        // Valid compressed point format, which should produce an invalid point when decompressed
```
cgorenflo, line 380:
```suggestion
        // the compressed format is not validated and should produce an invalid point when decompressed
```

contracts/multisig/src/key.rs
cgorenflo, line 545:
```suggestion
        // the compressed format is not validated and should produce an invalid point when decompressed
```

cgorenflo (approved):

PR #333, opened 2024-04-03
feat(voting-verifier-minor): migration entry point

haiyizxx:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


haiyizxx:
close as this is meant for show case migration, we agreed on introduce migration after next release.  will have a pr to set contract version on instantiation.

PR #334, opened 2024-04-03
feat(service-registry): add chains_per_worker map to service registry

maancham:
## Todos

[AXE-3638]

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Other Notes

[AXE-3638]: https://axelarnetwork.atlassian.net/browse/AXE-3638?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/service-registry/src/contract.rs
+++ b/contracts/service-registry/src/contract.rs
@@ -115,9 +115,10 @@ pub fn execute(
 
 pub mod execute {
     use connection_router_api::ChainName;
-    use std::collections::HashSet;
 
-    use crate::state::{AuthorizationState, CHAINS_PER_WORKER, WORKERS, WORKERS_PER_CHAIN};
+    use crate::state::{
+        register_chain_support_utility, AuthorizationState, WORKERS, WORKERS_PER_CHAIN,
+    };
 
     use super::*;
 
@@ -256,21 +257,7 @@ pub mod execute {
             .may_load(deps.storage, (&service_name, &info.sender))?
             .ok_or(ContractError::WorkerNotFound)?;
 
-        let mut worker_current_chains = CHAINS_PER_WORKER
-            .may_load(deps.storage, (&service_name, &info.sender))?
-            .unwrap_or_else(HashSet::new);
-
-        for chain in chains {
-            WORKERS_PER_CHAIN.save(deps.storage, (&service_name, &chain, &info.sender), &())?;
-            worker_current_chains.insert(chain.clone());
-        }
-        CHAINS_PER_WORKER.save(
-            deps.storage,
-            (&service_name, &info.sender),
-            &(worker_current_chains),
-        )?;
-
-        Ok(Response::new())
+        register_chain_support_utility(deps.storage, service_name.clone(), chains, info.sender)
     }
 
     pub fn deregister_chains_support(
@@ -370,15 +357,6 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> Result<Binary, ContractErr
             chain_name,
         } => to_json_binary(&query::get_active_workers(deps, service_name, chain_name)?)
             .map_err(|err| err.into()),
-        QueryMsg::GetRegisteredChainsForWorker {
-            service_name,
-            worker_address,
-        } => to_json_binary(&query::chains_per_worker(
-            deps,
-            service_name,
-            worker_address,
-        )?)
-        .map_err(|err| err.into()),
         QueryMsg::GetWorker {
             service_name,
             worker,
@@ -392,11 +370,8 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> Result<Binary, ContractErr
 
 pub mod query {
     use connection_router_api::ChainName;
-    use std::collections::HashSet;
 
-    use crate::state::{
-        WeightedWorker, CHAINS_PER_WORKER, WORKERS, WORKERS_PER_CHAIN, WORKER_WEIGHT,
-    };
+    use crate::state::{WeightedWorker, WORKERS, WORKERS_PER_CHAIN, WORKER_WEIGHT};
 
     use super::*;
 
@@ -431,16 +406,6 @@ pub mod query {
         } else {
             Ok(workers)
         }
-    }
-
-    pub fn chains_per_worker(
-        deps: Deps,
-        service_name: String,
-        worker_address: Addr,
-    ) -> Result<HashSet<ChainName>, ContractError> {
-        CHAINS_PER_WORKER
-            .may_load(deps.storage, (&service_name, &worker_address))?
-            .ok_or(ContractError::WorkerNotFound)
     }
 
     pub fn get_worker(
--- a/contracts/service-registry/src/contract.rs
+++ b/contracts/service-registry/src/contract.rs
@@ -116,9 +116,7 @@ pub fn execute(
 pub mod execute {
     use connection_router_api::ChainName;
 
-    use crate::state::{
-        register_chain_support_utility, AuthorizationState, WORKERS, WORKERS_PER_CHAIN,
-    };
+    use crate::state::{register_chain_support, AuthorizationState, WORKERS, WORKERS_PER_CHAIN};
 
     use super::*;
 
@@ -257,7 +255,9 @@ pub mod execute {
             .may_load(deps.storage, (&service_name, &info.sender))?
             .ok_or(ContractError::WorkerNotFound)?;
 
-        register_chain_support_utility(deps.storage, service_name.clone(), chains, info.sender)
+        let _res = register_chain_support(deps.storage, service_name.clone(), chains, info.sender);
+
+        Ok(Response::new())
     }
 
     pub fn deregister_chains_support(
--- a/contracts/service-registry/src/contract.rs
+++ b/contracts/service-registry/src/contract.rs
@@ -114,9 +114,9 @@ pub fn execute(
 }
 
 pub mod execute {
+    use crate::state;
+    use crate::state::{AuthorizationState, WORKERS, WORKERS_PER_CHAIN};
     use connection_router_api::ChainName;
-
-    use crate::state::{register_chain_support, AuthorizationState, WORKERS, WORKERS_PER_CHAIN};
 
     use super::*;
 
@@ -255,7 +255,8 @@ pub mod execute {
             .may_load(deps.storage, (&service_name, &info.sender))?
             .ok_or(ContractError::WorkerNotFound)?;
 
-        let _res = register_chain_support(deps.storage, service_name.clone(), chains, info.sender);
+        let _res =
+            state::register_chains_support(deps.storage, service_name.clone(), chains, info.sender);
 
         Ok(Response::new())
     }
--- a/contracts/service-registry/src/contract.rs
+++ b/contracts/service-registry/src/contract.rs
@@ -115,8 +115,9 @@ pub fn execute(
 
 pub mod execute {
     use connection_router_api::ChainName;
+    use std::collections::HashSet;
 
-    use crate::state::{AuthorizationState, WORKERS, WORKERS_PER_CHAIN};
+    use crate::state::{AuthorizationState, CHAINS_PER_WORKER, WORKERS, WORKERS_PER_CHAIN};
 
     use super::*;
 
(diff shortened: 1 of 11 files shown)


contracts/service-registry/src/contract.rs
```diff
@@ -257,6 +258,12 @@ pub mod execute {
 
         for chain in chains {
             WORKERS_PER_CHAIN.save(deps.storage, (&service_name, &chain, &info.sender), &())?;
+
+            let existing_chains =
+                CHAINS_PER_WORKER.may_load(deps.storage, (&service_name, &info.sender))?;
+            let mut chains_names = existing_chains.unwrap_or_else(HashSet::new);
```
cgorenflo, line 264:
```suggestion
            let mut chains =
                CHAINS_PER_WORKER.may_load(deps.storage, (&service_name, &info.sender))?.unwrap_or_else(HashSet::new);
```

contracts/service-registry/src/contract.rs
```diff
@@ -257,6 +258,12 @@ pub mod execute {
...
+            chains_names.insert(chain.clone());
+            CHAINS_PER_WORKER.save(deps.storage, (&service_name, &info.sender), &(chains_names))?;
```
cgorenflo, line 266: this operation is incredibly expensive. Currently, for each new chain, all the chains of the worker are loaded, one chain is added and the whole thing is saved back. Split it up:
- load the stored value
- add all the chains
- save it back

contracts/service-registry/src/contract.rs
```diff
@@ -257,6 +258,12 @@ pub mod execute {
...
         }
```
cgorenflo, line 267: move this whole part into state so there is no danger of accidentally updating only one of the storage maps?

contracts/service-registry/src/contract.rs
```diff
@@ -410,6 +429,16 @@ pub mod query {
...
     }
 
+    pub fn get_worker_chains(
```
cgorenflo, line 432:
```suggestion
    pub fn chains_per_worker(
```

contracts/service-registry/src/msg.rs
```diff
@@ -65,6 +66,12 @@ pub enum QueryMsg {
         chain_name: ChainName,
     },
 
+    #[returns(HashSet<ChainName>)]
+    GetRegisteredWorkerChains {
```
cgorenflo, line 70:
```suggestion
    GetRegisteredChainsForWorker {
```

contracts/service-registry/src/state.rs
```diff
@@ -134,6 +135,10 @@ pub enum AuthorizationState {
 pub const SERVICES: Map<&str, Service> = Map::new("services");
 // maps (service_name, chain_name, worker_address) -> ()
 pub const WORKERS_PER_CHAIN: Map<(&str, &ChainName, &Addr), ()> = Map::new("workers_per_chain");
+type ChainNames = HashSet<ChainName>;
+// maps (service_name, worker_address) -> List of ChainNames
+pub const CHAINS_PER_WORKER: Map<(&str, &Addr), ChainNames> = Map::new("chains_per_worker");
```
cgorenflo, line 140: since you are already adding a new alias, I think it would make sense to add aliases for `&str` and `&Addr` as well instead of using a comment

contracts/service-registry/src/contract.rs
maancham: I think doing this would interfere with the new optimized code. As the `may_load` and `save` parts are now out of the loop and the `insert` section is still inside the loop. So moving all these to another function adds overhead I think.

maancham: @cjcobb23 any comments?

contracts/service-registry/src/contract.rs
```diff
@@ -255,9 +256,19 @@ pub mod execute {
...
+
         for chain in chains {
             WORKERS_PER_CHAIN.save(deps.storage, (&service_name, &chain, &info.sender), &())?;
+            worker_current_chains.insert(chain.clone());
         }
+        CHAINS_PER_WORKER.save(
+            deps.storage,
+            (&service_name, &info.sender),
+            &(worker_current_chains),
+        )?;
```
cjcobb23, line 271: I think you can just use update, instead of load and then store.

contracts/service-registry/src/msg.rs
```diff
@@ -65,6 +66,12 @@ pub enum QueryMsg {
         chain_name: ChainName,
     },
 
+    #[returns(HashSet<ChainName>)]
+    GetRegisteredChainsForWorker {
+        service_name: String,
+        worker_address: Addr,
+    },
```
cjcobb23, line 73: Why do we need this?

cjcobb23: also, we can't accept `Addr` in user input, because it's not validated. We should accept string, and then validate (`deps.api.addr_validate(...)`)

cjcobb23 (changes requested):

contracts/service-registry/src/msg.rs
maancham: I am using it for testing (test.rs on service registry does not have access to state.rs contents). Any other ideas?

contracts/service-registry/src/contract.rs
maancham: then we should move the code inside the loop, which i believe will severely affect performance. Here, we are only doing load/save pair once. But using update will call it with every loop.

cgorenflo: to can provide the `update` function with an action instead of a value. So no need to put it into the loop, you can append the chains inside the action.

contracts/service-registry/src/state.rs
```diff
@@ -130,16 +131,73 @@ pub enum AuthorizationState {
...
+
+    for chain in chains {
+        WORKERS_PER_CHAIN.save(storage, (&service_name, &chain, &caller), &())?;
+        worker_current_chains.insert(chain.clone());
+    }
+
+    CHAINS_PER_WORKER.save(storage, (&service_name, &caller), &(worker_current_chains))?;
+
+    Ok(Response::new())
+}
```
cgorenflo, line 163:
```suggestion
pub fn register_chain_support(
    storage: &mut dyn Storage,
    service_name: String,
    chains: Vec<ChainName>,
    caller: WorkerAddress,
) -> Result<(), ContractError> {
    CHAINS_PER_WORKER.update(storage, (&service_name, &caller), |current_chains| {
        let mut current_chains = current_chains.unwrap_or_else(HashSet::new);
        current_chains.extend(chains.iter());
        Ok(current_chains)
    })?;

    for chain in chains.iter() {
        WORKERS_PER_CHAIN.save(storage, (&service_name, chain, &caller), &())?;
    }

    Ok(())
}
```

cgorenflo: the state doesn't need to know about `Response`

contracts/service-registry/tests/tests.rs
```diff
@@ -224,15 +224,15 @@ fn register_chain_support() {
...
+                chain_name: chain_name.clone(),
             },
         )
         .unwrap();
     assert_eq!(
         workers,
         vec![WeightedWorker {
             worker_info: Worker {
-                address: worker,
+                address: worker.clone(),
```
cgorenflo, line 235: why are they cloned now?

contracts/service-registry/src/contract.rs
```diff
@@ -255,11 +257,7 @@ pub mod execute {
             .may_load(deps.storage, (&service_name, &info.sender))?
             .ok_or(ContractError::WorkerNotFound)?;
 
-        for chain in chains {
-            WORKERS_PER_CHAIN.save(deps.storage, (&service_name, &chain, &info.sender), &())?;
-        }
-
-        Ok(Response::new())
+        register_chain_support_utility(deps.storage, service_name.clone(), chains, info.sender)
```
cgorenflo, line 260:
```suggestion
        state::register_chain_support(deps.storage, service_name, chains, info.sender)
```

contracts/service-registry/src/state.rs
```diff
@@ -130,16 +131,73 @@ pub enum AuthorizationState {
...
+    Map::new("workers_per_chain");
+pub const CHAINS_PER_WORKER: Map<(&ServiceName, &WorkerAddress), ChainNames> =
+    Map::new("chains_per_worker");
+pub const WORKERS: Map<(&ServiceName, &WorkerAddress), Worker> = Map::new("workers");
+
+pub fn register_chain_support_utility(
+    storage: &mut dyn Storage,
+    service_name: String,
+    chains: Vec<ChainName>,
+    caller: WorkerAddress,
```
cgorenflo, line 149:
```suggestion
    worker: WorkerAddress,
```
"caller" doesn't really mean anything to the state. "verifier" would be better, but I guess we keep it consistent for now and rename in bulk

contracts/service-registry/src/state.rs
maancham: thank you very much for the code!

contracts/service-registry/src/state.rs
maancham: Good call!

contracts/service-registry/src/contract.rs
```diff
@@ -255,9 +255,7 @@ pub mod execute {
             .may_load(deps.storage, (&service_name, &info.sender))?
             .ok_or(ContractError::WorkerNotFound)?;
 
-        for chain in chains {
-            WORKERS_PER_CHAIN.save(deps.storage, (&service_name, &chain, &info.sender), &())?;
-        }
+        let _res = register_chain_support(deps.storage, service_name.clone(), chains, info.sender);
```
cgorenflo, line 258: my bad, I missed the "s" in `chains`. I'd keep the names identival, i.e. make the call to `state::register_chains_support`

contracts/service-registry/src/state.rs
```diff
@@ -130,16 +131,72 @@ pub enum AuthorizationState {
...
+pub fn register_chain_support(
+    storage: &mut dyn Storage,
+    service_name: String,
+    chains: Vec<ChainName>,
+    worker: WorkerAddress,
+) -> Result<(), ContractError> {
+    CHAINS_PER_WORKER.update(storage, (&service_name, &worker), |current_chains| {
+        let mut current_chains = current_chains.unwrap_or_default();
+        current_chains.extend(chains.iter().cloned());
+        Ok::<HashSet<ChainName>, ContractError>(current_chains)
```
cgorenflo, line 154: is rust not able to figure out the type here?

contracts/service-registry/src/state.rs
```diff
@@ -130,16 +131,72 @@ pub enum AuthorizationState {
...
+    })?;
+
+    for chain in chains.iter() {
+        WORKERS_PER_CHAIN.save(storage, (&service_name, chain, &worker), &())?;
```
cgorenflo, line 158: is the referencing of `&()` necessary?

cgorenflo (approved):

contracts/service-registry/src/state.rs
maancham: no it was giving build errors on my side.

contracts/service-registry/src/state.rs
```diff
@@ -130,16 +131,72 @@ pub enum AuthorizationState {
...
+            service_name.into(),
+            chains,
+            worker.clone()
+        )
+        .is_ok());
+
+        let worker_chains =
+            may_load_chains_per_worker(deps.as_mut().storage, service_name.into(), worker).unwrap();
+        assert!(worker_chains.contains(&chain_name));
+    }
```
cjcobb23, line 199: I think we should have more tests than just this one test. This test only tests registering a single chain, and it doesn't test registering multiple chains in one call, over multiple calls, etc. I'll approve, but can you add a few more tests before merging?

cjcobb23 (approved):

contracts/service-registry/src/state.rs
maancham: yes it is.

contracts/service-registry/src/contract.rs
maancham: Changed! Thanks.

contracts/service-registry/src/state.rs
maancham: Thanks, I added three more tests.

PR #337, opened 2024-04-05
feat: add trait for message id

cjcobb23:
* implement trait for evm message id

https://axelarnetwork.atlassian.net/browse/AXE-3651


packages/axelar-wasm-std/src/msg_id/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+/// Any message id format must implement this trait.
+/// The implementation must satisfy the following invariants:
+///
+/// * if m1 != m2 then from_str(m1) != from_str(m2) (two different strings cannot parse to the same message id)
+///
+/// * if t1 != t2 then to_string(t1) != to_string(t2) (two different message ids cannot serialize to the same string)
+///
+/// There should be only one string that can identify a given message.
+/// Take extra care to handle things like leading 0s, casing, etc.
+pub trait MessageId: FromStr + Display {}
```
fish-sammy, line 23: Where's this used or going to be used?

packages/axelar-wasm-std/src/msg_id/tx_hash_event_index.rs
```diff
@@ -0,0 +1,199 @@
+use core::fmt;
+use std::{fmt::Display, str::FromStr};
+
+use cosmwasm_std::HexBinary;
+use regex::Regex;
+
+use super::Error;
+use crate::{hash::Hash, nonempty};
+pub struct HexTxHashAndEventIndex {
```
fish-sammy, line 9:
```suggestion

pub struct HexTxHashAndEventIndex {
```

packages/axelar-wasm-std/src/msg_id/tx_hash_event_index.rs
```diff
@@ -0,0 +1,199 @@
...
+impl HexTxHashAndEventIndex {
+    pub fn tx_hash_as_hex(&self) -> nonempty::String {
+        format!("0x{}", HexBinary::from(self.tx_hash).to_hex())
+            .try_into()
+            .expect("failed to convert tx hash to non-empty string")
+    }
+}
+
+impl FromStr for HexTxHashAndEventIndex {
+    type Err = Error;
```
fish-sammy, line 23:
```suggestion
    type Err = Error;
    
```

packages/axelar-wasm-std/src/msg_id/tx_hash_event_index.rs
```diff
@@ -0,0 +1,199 @@
...
+                        expected_format: PATTERN.to_string(),
+                    }
+                })?,
+            })
+        } else {
+            Err(Error::InvalidMessageID {
+                id: message_id.to_string(),
+                expected_format: PATTERN.to_string(),
+            })
+        }
```
fish-sammy, line 54:
```suggestion
        let caps: Vec<_> = regex
            .captures(message_id)
            .into_iter()
            .flat_map(|caps| caps.iter().collect::<Vec<_>>())
            .collect();

        match caps.as_slice() {
            [Some(_), Some(tx_id), Some(event_index)] => {
                let tx_id = tx_id.as_str();
                let event_index = event_index.as_str();

                Ok(HexTxHashAndEventIndex {
                    tx_hash: HexBinary::from_hex(&tx_id[2..])
                        .expect("invalid hex")
                        .as_slice()
                        .try_into()
                        .expect("invalid length tx hash"),
                    event_index: event_index.parse().map_err(|_| {
                        // this error can happen if the integer overflows u32
                        Error::InvalidMessageID {
                            id: message_id.to_string(),
                            expected_format: PATTERN.to_string(),
                        }
                    })?,
                })
            }
            _ => Err(Error::InvalidMessageID {
                id: message_id.to_string(),
                expected_format: PATTERN.to_string(),
            }),
        }
```

fish-sammy (changes requested):

packages/axelar-wasm-std/src/msg_id/mod.rs
cjcobb23: the primary places this would be used is by the router (verifying the format matches) and by the verifier (parsing the string into proper structured data).

packages/axelar-wasm-std/Cargo.toml
```diff
@@ -37,6 +37,7 @@ error-stack = { workspace = true }
 flagset = { version = "0.4.3", features = ["serde"] }
 num-traits = { workspace = true }
 report = { workspace = true }
+regex = "1.10.0"
```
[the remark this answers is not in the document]
cjcobb23: we could probably disable unicode features?

packages/axelar-wasm-std/src/msg_id/tx_hash_event_index.rs
```diff
@@ -0,0 +1,199 @@
...
+    }
+}
+
+impl FromStr for HexTxHashAndEventIndex {
+    type Err = Error;
+    fn from_str(message_id: &str) -> Result<Self, Self::Err>
+    where
+        Self: Sized,
+    {
+        const PATTERN: &str = "^(0x[0-9a-f]{64})-(0|[1-9][0-9]*)$";
```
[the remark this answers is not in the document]
cjcobb23: I agree regex is probably not the most performant solution, but I think it's probably cleaner and safer compared to manually checking everything. Also, if we go with regex and then discover there is a gas issue, we can always change the implementation, since it would not break the protocol. @cgorenflo thoughts?

packages/axelar-wasm-std/Cargo.toml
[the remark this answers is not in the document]
cjcobb23: ok i disabled the default features and just enabled performance features

packages/axelar-wasm-std/src/msg_id/tx_hash_event_index.rs
```diff
@@ -0,0 +1,211 @@
...
+            [Some(_), Some(tx_id), Some(event_index)] => {
+                let tx_id = tx_id.as_str();
+                let event_index = event_index.as_str();
+
+                Ok(HexTxHashAndEventIndex {
+                    tx_hash: HexBinary::from_hex(&tx_id[2..])
+                        .expect("invalid hex")
+                        .as_slice()
+                        .try_into()
+                        .expect("invalid length tx hash"),
```
cgorenflo, line 52: is it 100% certain that this will never fail?

cgorenflo: since this function already returns a result, I don't think we gain anything by adding a potential panic instead of just returning an error

cgorenflo (commented):
love the use of regex pattern matching!

packages/axelar-wasm-std/src/msg_id/tx_hash_event_index.rs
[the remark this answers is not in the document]
cgorenflo: agreed. Let the people who created the regex library worry about implementing the validation. This way it's much easier to ensure no edge cases have been missed

packages/axelar-wasm-std/src/msg_id/tx_hash_event_index.rs
```diff
@@ -0,0 +1,211 @@
...
+                    })?,
+                })
+            }
+            _ => Err(Error::InvalidMessageID {
+                id: message_id.to_string(),
+                expected_format: PATTERN.to_string(),
+            }),
+        }
+    }
+}
```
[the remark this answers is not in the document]
cgorenflo: the `Err` needs to be changed to `Report<Error>`

packages/axelar-wasm-std/src/msg_id/tx_hash_event_index.rs
cjcobb23: it should never fail unless there is a bug in our code

fish-sammy (approved):

PR #339, opened 2024-04-08
feat(monitoring): connect monitoring contract to multisig prover

maancham:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

### DIFF
--- a/contracts/monitoring/src/query.rs
+++ b/contracts/monitoring/src/query.rs
@@ -1,8 +1,10 @@
 use crate::error::ContractError;
 use crate::state::PROVERS_PER_CHAIN;
 use connection_router_api::ChainName;
-use cosmwasm_std::{Addr, Deps};
+use cosmwasm_std::{to_json_binary, Addr, Deps, QueryRequest, WasmQuery};
 use multisig::worker_set::WorkerSet;
+use multisig_prover::msg::QueryMsg::GetWorkerSet;
+use std::collections::HashSet;
 
 pub fn chains_active_worker_sets(
     _deps: Deps,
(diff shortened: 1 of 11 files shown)


integration-tests/tests/unbond_worker.rs
```diff
@@ -0,0 +1,98 @@
+use cosmwasm_std::{Addr, StdError};
+use integration_tests::contract::Contract;
+use std::collections::HashSet;
+
+pub mod test_utils;
+
+#[test]
+fn worker_should_be_in_collective_active_worker_set_of_all_supported_chains() {
+    let (protocol, ethereum, polygon, initial_workers, _) = test_utils::setup_test_case();
```
cgorenflo, line 9: not in scope for this PR, but something to think about: `setup_test_case` hides too much detail. I can't really see what this test does without digging three layers deep. We should remodel that, maybe with a builder pattern like so:
```
let setup = set_up_test_case()
    .with_workers(...)
    .with_chains(...)

```

integration-tests/tests/unbond_worker.rs
cgorenflo, line 1: all the tests have the same structure:
- a set of chains
- a worker address
- an expected outcome

you could parameterize this instead of writing a separate test for each (see https://github.com/axelarnetwork/axelar-amplifier/blob/main/contracts/gateway/tests/contract.rs)

Additionally, because this is a critical query for unbonding, we need to catch all the edge cases:
- chain is supported, but worker is not part of the active set (but supports other chains)
- call only with some of the known chains
- worker was in a previous worker set for chain, but not the currently active one

This can become quite difficult to set up with integration tests (because it involved a lot of verification, rotation etc), so it might we easier to set it up as unit tests and mocks for the queries to other contracts. Definitely have some integration tests to make sure the contract endpoints work together though

cgorenflo (changes requested):

contracts/monitoring/src/query.rs
```diff
@@ -16,3 +18,30 @@ pub fn provers(deps: Deps, chain_name: ChainName) -> Result<Vec<Addr>, ContractE
...
 }
+
+pub fn check_worker_is_active(
+    deps: Deps,
+    chain_names: HashSet<ChainName>,
+    worker: Addr,
+) -> StdResult<bool> {
+    for chain_name in chain_names {
+        if let Ok(prover_addresses) = PROVERS_PER_CHAIN.load(deps.storage, chain_name) {
+            for prover_address in prover_addresses {
```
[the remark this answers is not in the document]
maancham: ok thanks for double checking!

integration-tests/tests/unbond_worker.rs
maancham: Ok sure. I will add a ticket for it to the backlog to later come back to this.

PR #344, opened 2024-04-11
feat: add initial healthcheck endpoint (ampd)

eloylp:
## Description

At [Eiger](https://github.com/eigerco) we are working on the `Axelar Solana` implementation.  During this work, we realised `ampd` doesn't have a defined way to check the `liveness` of the process from an external perspective. This is specially useful when deploying workloads in i.e Kubernetes, as the cluster can query the process in order to check if its alive or not.

This is a PR proposal for adding an `HTTP` `/status` endpoint to the `ampd` process, that can be used by external systems to determine the `liveness` of the process.

Closes https://github.com/eigerco/axelar-amplifier/issues/12

## Todos

- [x] Unit tests
- [x] Manual tests See this [comment](https://github.com/axelarnetwork/axelar-amplifier/pull/344#issuecomment-2049722752) .
- [x] Documentation
- [x] Connect epics/issues

### Manual
1. Start the `ampd` server.
2. ```bash
    $ curl localhost:3000/status
    ```
### Unit tests

```bash
$ cd ampd && cargo test
```

## Expected Behaviour

Once `ampd` starts its activity, an HTTP request to the `/status` endpoint should return a `200 OK` response in json format:

```bash
{
 ok: true
}
```

## Other Notes

This is just a dummy HTTP health check endpoint that only provides a probe of liveness of the process. It doesn't query any subsystem in order to check their internal health. In a future, such option could be revisited.

### DIFF
--- a/ampd/README.md
+++ b/ampd/README.md
@@ -52,6 +52,7 @@ type=[handler type. Could be EvmMsgWorkerSetVerifier | SuiWorkerSetVerifier]
 
 Below is an example config for connecting to a local axelard node and local tofnd process, and verifying transactions from Avalanche testnet and Sui testnet.
 ```
+health_check_bind_addr = "0.0.0.0:3000"
 tm_jsonrpc="http://localhost:26657"
 tm_grpc="tcp://localhost:9090"
 event_buffer_cap=10000
--- a/ampd/README.md
+++ b/ampd/README.md
@@ -11,6 +11,7 @@ Below is the config file format, with explanations for each entry:
 tm_jsonrpc=[JSON-RPC URL of Axelar node]
 tm_grpc=[gRPC URL of Axelar node]
 event_buffer_cap=[max blockchain events to queue. Will error if set too low]
+health_check_bind_addr =[the /status endpoint bind address i.e "0.0.0.0:3000"]
 
 [service_registry]
 cosmwasm_contract=[address of service registry]
--- a/ampd/README.md
+++ b/ampd/README.md
@@ -53,7 +53,7 @@ type=[handler type. Could be EvmMsgWorkerSetVerifier | SuiWorkerSetVerifier]
 
 Below is an example config for connecting to a local axelard node and local tofnd process, and verifying transactions from Avalanche testnet and Sui testnet.
 ```
-health_check_bind_addr ="0.0.0.0:3000"
+health_check_bind_addr="0.0.0.0:3000"
 tm_jsonrpc="http://localhost:26657"
 tm_grpc="tcp://localhost:9090"
 event_buffer_cap=10000
--- a/ampd/README.md
+++ b/ampd/README.md
@@ -11,7 +11,7 @@ Below is the config file format, with explanations for each entry:
 tm_jsonrpc=[JSON-RPC URL of Axelar node]
 tm_grpc=[gRPC URL of Axelar node]
 event_buffer_cap=[max blockchain events to queue. Will error if set too low]
-health_check_bind_addr =[the /status endpoint bind address i.e "0.0.0.0:3000"]
+health_check_bind_addr=[the /status endpoint bind address i.e "0.0.0.0:3000"]
 
 [service_registry]
 cosmwasm_contract=[address of service registry]
--- a/ampd/README.md
+++ b/ampd/README.md
@@ -53,7 +53,7 @@ type=[handler type. Could be EvmMsgWorkerSetVerifier | SuiWorkerSetVerifier]
 
 Below is an example config for connecting to a local axelard node and local tofnd process, and verifying transactions from Avalanche testnet and Sui testnet.
 ```
-health_check_bind_addr = "0.0.0.0:3000"
+health_check_bind_addr ="0.0.0.0:3000"
 tm_jsonrpc="http://localhost:26657"
 tm_grpc="tcp://localhost:9090"
 event_buffer_cap=10000
--- a/ampd/src/config.rs
+++ b/ampd/src/config.rs
@@ -1,5 +1,4 @@
-use std::net::SocketAddr;
-use std::str::FromStr;
+use std::net::{Ipv4Addr, SocketAddrV4};
 use std::time::Duration;
 
 use serde::{Deserialize, Serialize};
@@ -13,7 +12,7 @@ use crate::url::Url;
 #[derive(Deserialize, Serialize, Debug, PartialEq)]
 #[serde(default)]
 pub struct Config {
-    pub health_check_bind_addr: SocketAddr,
+    pub health_check_bind_addr: SocketAddrV4,
     pub tm_jsonrpc: Url,
     pub tm_grpc: Url,
     pub event_buffer_cap: usize,
@@ -37,7 +36,7 @@ impl Default for Config {
             event_buffer_cap: 100000,
             event_stream_timeout: Duration::from_secs(15),
             service_registry: ServiceRegistryConfig::default(),
-            health_check_bind_addr: SocketAddrV4::new(Ipv4Addr::UNSPECIFIED, 3000);
+            health_check_bind_addr: SocketAddrV4::new(Ipv4Addr::UNSPECIFIED, 3000),
         }
     }
 }
--- a/ampd/src/config.rs
+++ b/ampd/src/config.rs
@@ -1,3 +1,5 @@
+use std::net::SocketAddr;
+use std::str::FromStr;
 use std::time::Duration;
 
 use serde::{Deserialize, Serialize};
@@ -11,6 +13,7 @@ use crate::url::Url;
 #[derive(Deserialize, Serialize, Debug, PartialEq)]
 #[serde(default)]
 pub struct Config {
+    pub health_check_bind_addr: SocketAddr,
     pub tm_jsonrpc: Url,
     pub tm_grpc: Url,
     pub event_buffer_cap: usize,
@@ -34,6 +37,7 @@ impl Default for Config {
             event_buffer_cap: 100000,
             event_stream_timeout: Duration::from_secs(15),
             service_registry: ServiceRegistryConfig::default(),
+            health_check_bind_addr: SocketAddr::from_str("0.0.0.0:3000").unwrap(),
         }
     }
 }
--- a/ampd/src/config.rs
+++ b/ampd/src/config.rs
@@ -37,7 +37,7 @@ impl Default for Config {
             event_buffer_cap: 100000,
             event_stream_timeout: Duration::from_secs(15),
             service_registry: ServiceRegistryConfig::default(),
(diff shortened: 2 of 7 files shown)


ampd/src/health_check.rs
```diff
@@ -0,0 +1,105 @@
...
+#[derive(Error, Debug)]
+pub enum HealthCheckError {
+    #[error("Health check server error: {0}")]
+    Error(String),
+}
+
+impl Server {
+    pub async fn new(bind_addr: SocketAddr) -> Result<Self, HealthCheckError> {
+        Ok(Self {
+            listener: tokio::net::TcpListener::bind(bind_addr)
```
eloylp, line 23: Normally i do not like to use `new()` for doing things like binding sockets. Initially i consider this approach doable taking into account how the rest of the codebase interacts with this server.

eloylp:
### Manual testing

In order to provide certain confidence the healthcheck endpoint is not disturbing other components, the `ampd` daemon was tested by:

1. Bringing up locally the [tofnd](https://github.com/axelarnetwork/tofnd) (pre-requisite). We used the docker-compose way sharing the port with the host.

2. Configure the `~/.ampd/config.toml` file with the following content:

   ```toml
   tm_jsonrpc = "https://axelar-testnet-rpc.qubelabs.io:443"
   tm_grpc = "http://axelar-testnet-grpc.qubelabs.io:9090"
   event_buffer_capacity = 1000
   health_check_bind_addr = "0.0.0.0:3000"
   
   [service_registry]
   cosmwasm_contract="axelar1hrv2zf8xsfsey2umewng68c8pwsyktq4sm8u6k5sch4gqkn0qxrsg28xnx"
   
   [broadcast]
   batch_gas_limit="10000000"
   broadcast_interval="1s"
   chain_id="devnet-amplifier"
   gas_adjustment="2"
   gas_price="0.00005uamplifier"
   queue_cap="1000"
   tx_fetch_interval="600ms"
   tx_fetch_max_retries="10"
   
   [tofnd_config]
   key_uid="axelar"
   party_uid="ampd"
   url="http://127.0.0.1:50051"
   
   [[handlers]]
   type = 'MultisigSigner'
   cosmwasm_contract = 'axelar1lz2hevr93dwa3l86n7hnyy36age39jsk0e8nsj804xr3n8hh958q79m2sq'
   ```

3. Execute `cd ampd && cargo run` :

   ```bash
   ❯ cargo run
       Finished dev [unoptimized + debuginfo] target(s) in 1.19s
        Running `/home/eloylp/projects/eiger/axelar-amplifier/target/debug/ampd`
   2024-04-10T13:15:35.305195Z  INFO ampd: found config file /home/eloylp/.ampd/config.toml
   2024-04-10T13:15:35.308031Z  INFO ampd: starting daemon args=Args { config: ["~/.ampd/config.toml", "config.toml"], state: "~/.ampd/state.json", output: Output::Text, cmd: () }
   2024-04-10T13:15:35.308152Z  INFO ampd::state: loading state from disk
   2024-04-10T13:15:37.850389Z  INFO ampd::health_check: Starting health check server at: 0.0.0.0:3000
   2024-04-10T13:15:40.417094Z  INFO ampd::state: state updated handler="multisig-signer" height=12887365
   ```

   We can see the output `Starting health check server at: 0.0.0.0:3000` , which announces the healthcheck endpoint should be reachable at this point.

4. Reach the health endpoint:

   ```bash
   ❯ curl localhost:3000/status
   {"ok":true}
   ```

ampd/src/config.rs
```diff
@@ -34,6 +37,7 @@ impl Default for Config {
             event_buffer_cap: 100000,
             event_stream_timeout: Duration::from_secs(15),
             service_registry: ServiceRegistryConfig::default(),
+            health_check_bind_addr: SocketAddr::from_str("0.0.0.0:3000").unwrap(),
```
tilacog, line 40: We can remove the call to a failing method followed by an unwrap by creating the new value directly: 
```suggestion
            health_check_bind_addr: SocketAddrV4::new(Ipv4Addr::UNSPECIFIED, 3000);
```

ampd/src/health_check.rs
```diff
@@ -0,0 +1,105 @@
...
+
+pub struct Server {
+    listener: tokio::net::TcpListener,
+}
+
+#[derive(Error, Debug)]
+pub enum HealthCheckError {
+    #[error("Health check server error: {0}")]
+    Error(String),
+}
```
tilacog, line 18: Instead of using a single `Error(String)` variant to represent all errors, it would be better to define dedicated error variants for each possible error case. This allows errors to be handled statically rather than requiring string parsing at runtime.

ampd/src/health_check.rs
tilacog: Could you share why you think `new` may not be ideal for this case?

ampd/src/health_check.rs
```diff
@@ -0,0 +1,105 @@
...
+            listener: tokio::net::TcpListener::bind(bind_addr)
+                .await
+                .change_context(HealthCheckError::Error(format!(
+                    "Failed binding to addr: {}",
+                    bind_addr
+                )))?,
+        })
+    }
+
+    #[allow(dead_code)]
```
tilacog, line 32: The `#[cfg(test)]` attribute is preferred over `#[allow(dead_code)]` for functions only used in tests, as it ensures the item will only be compiled for test builds.

```suggestion
    #[cfg(test)]
```

tilacog (changes requested):

ampd/src/health_check.rs
eloylp: I was also thinking on that, but then i realised rest of the code doesn't need to react to this errors, thus, representing all of them as a single opaque one. Maybe there are other cases i am not seeing, let me know.

ampd/src/health_check.rs
eloylp: great idea.

eloylp:
> Could you share why you think new may not be ideal for this case?

In this scenario, I typically reserve the use of new exclusively for configuring structs, abstaining from creating any resources. Instead, within the run function, I manage all resource acquisition. This approach fosters clearer responsibility segregation and enhances intuition. I believe reserving the socket allocation in the operating system is more suitable for the run function.

ampd/src/health_check.rs
eloylp: @tilacog In this scenario, I typically reserve the use of new exclusively for configuring structs, abstaining from creating any resources. Instead, within the run function, I manage all resource acquisition. This approach fosters clearer responsibility segregation and enhances intuition. I believe reserving the socket allocation in the operating system is more suitable for the run function.

tilacog (approved):

tilacog (approved):

ampd/src/lib.rs
```diff
@@ -457,4 +473,6 @@ pub enum Error {
     BlockHeightMonitor,
     #[error("invalid finalizer type for chain {0}")]
     InvalidFinalizerType(ChainName),
+    #[error("Health check system error")]
+    HealthCheckError,
```
cgorenflo, line 477:
```suggestion
    #[error("health check is not working")]
    HealthCheck,
```

ampd/src/health_check.rs
cgorenflo: I agree with @eloylp, it would be preferable to just store the address in the `new` function and then start the listener in the `run` function, so there is only one place that can return errors

ampd/src/health_check.rs
```diff
@@ -0,0 +1,98 @@
...
+        Ok(Self {
+            listener: tokio::net::TcpListener::bind(bind_addr)
+                .await
+                .map_err(HealthCheckError::from)?,
+        })
+    }
+
+    pub async fn run(self, cancel: CancellationToken) -> Result<(), HealthCheckError> {
+        let app = Router::new().route("/status", get(status));
+        let bind_address = self.listener.local_addr().map_err(HealthCheckError::from)?;
```
cgorenflo, line 31:
```suggestion
        let bind_address = self.listener.local_addr().change_context(Error::Start)?;
```

cgorenflo (commented):
I think this is a good addition to ampd, just address the suggestions and move the listener binding from the `new` function to the `run` function and we can merge it

ampd/src/health_check.rs
```diff
@@ -0,0 +1,98 @@
...
+
+        info!("Starting health check server at: {}", bind_address);
```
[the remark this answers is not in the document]
cgorenflo: this helps with monitoring using structured logging

ampd/src/health_check.rs
eloylp: @cgorenflo thanks for your review. The following actions were performed for this PR:

1. Applied all your suggestions. 
2. Move the listener creation to the `run` function at 571eeaffb06c41b084de109819008ec7a7d92517 .
3. Clippy checks.
4. Execute both, unit tests and [manual tests](https://github.com/axelarnetwork/axelar-amplifier/pull/344#issuecomment-2049722752) again.

We hope these changes are the expected ones. Let us know if there's anything else you'd like us to tweak or if we are good to go.

ampd/README.md
```diff
@@ -52,6 +53,7 @@ type=[handler type. Could be EvmMsgWorkerSetVerifier | SuiWorkerSetVerifier]
 
 Below is an example config for connecting to a local axelard node and local tofnd process, and verifying transactions from Avalanche testnet and Sui testnet.
 ```
+health_check_bind_addr ="0.0.0.0:3000"
```
fish-sammy, line 56:
```suggestion
health_check_bind_addr="0.0.0.0:3000"
```
to be consistent with other fields.

ampd/src/health_check.rs
```diff
@@ -0,0 +1,101 @@
...
+            .with_graceful_shutdown(async move { cancel.cancelled().await })
+            .await
+            .change_context(Error::WhileRunning)
+    }
+}
+
+// basic handler that responds with a static string
+async fn status() -> (StatusCode, Json<Status>) {
+    (StatusCode::OK, Json(Status { ok: true }))
+}
```
cjcobb23, line 48: might be useful in the future to extend this to actually give more info

cgorenflo (approved):

cgorenflo:
@eloylp could you anotate the ampd constructor with `#[allow(clippy::too_many_arguments)]` for now so the lint check succeeds? I'll clean it up after it's merged then

eloylp:
@cgorenflo 

> @eloylp could you anotate the ampd constructor with `#[allow(clippy::too_many_arguments)]` for now so the lint check succeeds? I'll clean it up after it's merged then

@cgorenflo fixed at 7323f71b0dc2f9f56f647230c0e9fc7ebd6fc481 . Thanks for everything.

PR #345, opened 2024-04-11
feat(multisig-prover)!: add monitoring field to prover instantiation

maancham:
## Description

[AXE-3697]

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Other Notes

[AXE-3697]: https://axelarnetwork.atlassian.net/browse/AXE-3697?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/multisig-prover/src/contract.rs
cgorenflo, line 1: since we now have the monitoring address at instantiation, should we directly register this contract with the monitoring contract at the same time?

cgorenflo (approved):

contracts/multisig-prover/src/contract.rs
maancham: well the execute message to save a prover address and its chain name is present on monitoring. Right now, this thing is called from service registry, we can change it to be called during instantiation of prover. Do you want to discuss it with the team?

PR #350, opened 2024-04-15
fix(minor-nexus-gateway)!: make msg id consistent with evm chains

cjcobb23:
Nexus gateway was still using colon, instead of dash in message ids, and was not prefixing with 0x.

https://axelarnetwork.atlassian.net/browse/AXE-3719

### DIFF
--- a/contracts/nexus-gateway/src/contract/execute.rs
+++ b/contracts/nexus-gateway/src/contract/execute.rs
@@ -2,7 +2,7 @@ use cosmwasm_std::{to_binary, Addr, Response, WasmMsg};
 use error_stack::report;
 
 use crate::error::ContractError;
-use crate::nexus::{self};
+use crate::nexus;
 use crate::state::Store;
 
 use super::Contract;
@@ -25,7 +25,7 @@ where
         let msgs: Vec<_> = msgs
             .into_iter()
             .map(connection_router_api::Message::try_from)
-            .collect::<Result<Vec<connection_router_api::Message>>>()?;
+            .collect::<Result<Vec<_>>>()?;
         if msgs.is_empty() {
             return Ok(Response::default());
         }
--- a/contracts/nexus-gateway/src/contract/execute.rs
+++ b/contracts/nexus-gateway/src/contract/execute.rs
@@ -2,7 +2,7 @@ use cosmwasm_std::{to_binary, Addr, Response, WasmMsg};
 use error_stack::report;
 
 use crate::error::ContractError;
-use crate::nexus;
+use crate::nexus::{self};
 use crate::state::Store;
 
 use super::Contract;
@@ -24,8 +24,8 @@ where
 
         let msgs: Vec<_> = msgs
             .into_iter()
-            .map(connection_router_api::Message::from)
-            .collect();
+            .map(connection_router_api::Message::try_from)
+            .collect::<Result<Vec<connection_router_api::Message>>>()?;
         if msgs.is_empty() {
             return Ok(Response::default());
         }
@@ -134,7 +134,7 @@ mod test {
                 .unwrap()
                 .try_into()
                 .unwrap(),
-                source_tx_id: vec![0x2f, 0xe4].try_into().unwrap(),
+                source_tx_id: vec![0x2f; 32].try_into().unwrap(),
                 source_tx_index: 100,
             },
             nexus::Message {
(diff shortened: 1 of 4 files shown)


packages/axelar-wasm-std/src/msg_id/tx_hash_event_index.rs
```diff
@@ -22,7 +22,7 @@ impl HexTxHashAndEventIndex {
     }
 }
 
-const PATTERN: &str = "^(0x[0-9a-f]{64})-(0|[1-9][0-9]*)$";
+pub const PATTERN: &str = "^(0x[0-9a-f]{64})-(0|[1-9][0-9]*)$";
```
cgorenflo, line 25: why is it public now?

contracts/nexus-gateway/src/nexus.rs
```diff
@@ -68,24 +68,37 @@ impl From<connection_router_api::Message> for Message {
...
+                    event_index: u32::try_from(msg.source_tx_index)
+                        .map_err(|_| {
+                            axelar_wasm_std::msg_id::Error::EventIndexOverflow(msg_id.clone())
+                        })
+                        .change_context(ContractError::InvalidMessageId(msg_id))?,
+                }
+                .to_string()
                 .try_into()
-                .expect("cannot be empty"),
+                .expect("must serialize message id to non-empty string"),
```
cgorenflo, line 95: seems weird that this is a panic. If the underlying string was a constant I would agree, but here I'd prefer just to return an error if this fails (even if it never should, but it relies on multiple other parts to implicitly work as expected). You don't save anything by panicking here, there return type is a `Result` either way

contracts/nexus-gateway/src/nexus.rs
```diff
@@ -68,24 +68,37 @@ impl From<connection_router_api::Message> for Message {
...
             },
             source_address: msg.source_address,
             destination_chain: msg.destination_chain,
             destination_address: msg.destination_address,
             payload_hash: msg.payload_hash,
-        }
+        })
```
fish-sammy, line 101:
```suggestion
        let id = format!(
            "0x{}-{}",
            msg.source_tx_id.as_ref().encode_hex::<String>(),
            msg.source_tx_index
        );
        Ok(Self {
            cc_id: CrossChainId {
                chain: msg.source_chain,
                id: id
                    .parse::<HexTxHashAndEventIndex>()
                    .change_context(ContractError::InvalidMessageId(id))?
                    .to_string()
                    .try_into()
                    .expect("must serialize message id to non-empty string"),
            },
            source_address: msg.source_address,
            destination_chain: msg.destination_chain,
            destination_address: msg.destination_address,
            payload_hash: msg.payload_hash,
        })
```
Why not just?

contracts/nexus-gateway/src/contract/execute.rs
```diff
@@ -2,7 +2,7 @@ use cosmwasm_std::{to_binary, Addr, Response, WasmMsg};
 use error_stack::report;
 
 use crate::error::ContractError;
-use crate::nexus;
+use crate::nexus::{self};
```
fish-sammy, line 5: Why change this here?

contracts/nexus-gateway/src/contract/execute.rs
```diff
@@ -24,8 +24,8 @@ where
 
         let msgs: Vec<_> = msgs
             .into_iter()
-            .map(connection_router_api::Message::from)
-            .collect();
+            .map(connection_router_api::Message::try_from)
+            .collect::<Result<Vec<connection_router_api::Message>>>()?;
```
fish-sammy, line 28:
```suggestion
            .collect::<Result<_>>()?;
```

packages/axelar-wasm-std/src/msg_id/tx_hash_event_index.rs
fish-sammy: Same question.

fish-sammy (changes requested):

packages/axelar-wasm-std/src/msg_id/tx_hash_event_index.rs
cjcobb23: oh I was using it in nexus gateway, but I guess I'm not now, so I'll revert this

contracts/nexus-gateway/src/nexus.rs
cjcobb23: I don't want to hardcode the string format in nexus gateway like this. That's what led to the message id being inconsistent in the first place. The `HexTxHashAndEventIndex` defines the string format, and can do all the serialization/deserialization.

cjcobb23: I tried to clean this up and make it simpler though, lmk what you think

cgorenflo (approved):

fish-sammy (approved):

PR #351, opened 2024-04-15
feat(voting-verifier): client mod

fish-sammy:
## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


contracts/voting-verifier/src/client.rs
```diff
@@ -0,0 +1,212 @@
...
+    #[test]
+    fn query_messages_status() {
+        let (querier, _, addr) = setup();
+        let client: Client = client::Client::new(QuerierWrapper::new(&querier), addr).into();
+
+        let msg_1 = Message {
+            cc_id: format!("eth{}0x1234", CHAIN_NAME_DELIMITER)
+                .as_str()
+                .parse()
+                .unwrap(),
```
cjcobb23, line 119: Can we use the `HexTxHashAndEventIndex` struct to construct a properly formed message id? Hard coded strings like this are such a pain if they ever need to be changed. And it's likely this will need to be changed, because the tx hash is not the right length, and there's no event index.

contracts/voting-verifier/src/client.rs
```diff
@@ -0,0 +1,212 @@
...
+            source_address: "0x1234".parse().unwrap(),
+            destination_address: "0x5678".parse().unwrap(),
+            destination_chain: "eth".parse().unwrap(),
+            payload_hash: [0; 32],
+        };
+        let msg_2 = Message {
+            cc_id: format!("eth{}0x4321", CHAIN_NAME_DELIMITER)
```
cjcobb23, line 126: same comment as above about the message id construction

cjcobb23 (changes requested):

contracts/voting-verifier/src/client.rs
```diff
@@ -0,0 +1,212 @@
...
+use cosmwasm_std::{Addr, WasmMsg};
+use error_stack::ResultExt;
+
+use crate::msg::{ExecuteMsg, Poll, QueryMsg};
+
+type Result<T> = error_stack::Result<T, Error>;
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("failed quering the voting verifier contract at {0}")]
```
cgorenflo, line 17:
```suggestion
    #[error("failed to query the voting verifier contract at {0}")]
```

cgorenflo (approved):

cgorenflo:
my approval is conditioned on the fix for the message ids as CJ mentioned

cjcobb23 (approved):

PR #352, opened 2024-04-16
feat(minor-monitoring)!: register active worker set with monitoring

maancham:
## Description

Axe 3695

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

### DIFF
--- a/contracts/monitoring/src/contract.rs
+++ b/contracts/monitoring/src/contract.rs
@@ -1,7 +1,6 @@
+use cosmwasm_std::entry_point;
 use crate::msg::{ExecuteMsg, InstantiateMsg, QueryMsg};
 use crate::state::{Config, CONFIG};
-#[cfg(not(feature = "library"))]
-use cosmwasm_std::entry_point;
 use cosmwasm_std::{to_binary, Binary, Deps, DepsMut, Env, MessageInfo, Response, StdResult};
 
 use crate::execute;
@@ -63,7 +62,7 @@ mod tests {
     use connection_router_api::ChainName;
     use cosmwasm_std::testing::{mock_dependencies, mock_env, mock_info};
     use cosmwasm_std::Addr;
-    use multisig::worker_set::WorkerSet;
+    use workers_to_worker_set;
 
     use super::*;
 
--- a/contracts/monitoring/src/contract.rs
+++ b/contracts/monitoring/src/contract.rs
@@ -1,6 +1,6 @@
-use cosmwasm_std::entry_point;
 use crate::msg::{ExecuteMsg, InstantiateMsg, QueryMsg};
 use crate::state::{Config, CONFIG};
+use cosmwasm_std::entry_point;
 use cosmwasm_std::{to_binary, Binary, Deps, DepsMut, Env, MessageInfo, Response, StdResult};
 
 use crate::execute;
@@ -62,7 +62,6 @@ mod tests {
     use connection_router_api::ChainName;
     use cosmwasm_std::testing::{mock_dependencies, mock_env, mock_info};
     use cosmwasm_std::Addr;
-    use workers_to_worker_set;
 
     use super::*;
 
--- a/contracts/monitoring/src/contract.rs
+++ b/contracts/monitoring/src/contract.rs
@@ -39,11 +39,10 @@ pub fn execute(
             execute::register_prover(deps, chain_name, new_prover_addr)
         }
         ExecuteMsg::RegisterActiveWorkerSet {
-            chain_name,
             next_worker_set,
         } => {
             // TODO: add check_prover to make sure prover is part of the system
-            execute::register_active_worker_set(deps, info, chain_name, next_worker_set)
+            execute::register_active_worker_set(deps, info, next_worker_set)
         }
     }
     .map_err(axelar_wasm_std::ContractError::from)
@@ -54,11 +53,9 @@ pub fn execute(
 pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> StdResult<Binary> {
     match msg {
         QueryMsg::GetActiveWorkerSet {
-            chain_name,
             prover_address,
         } => to_binary(&query::get_active_worker_set(
             deps,
-            chain_name,
             prover_address,
         )?),
     }
--- a/contracts/monitoring/src/contract.rs
+++ b/contracts/monitoring/src/contract.rs
@@ -36,7 +36,13 @@ pub fn execute(
         } => {
             execute::check_governance(&deps, info)?;
             execute::register_prover(deps, chain_name, new_prover_addr)
-        }
+        },
+        ExecuteMsg::RegisterActiveWorkerSet {
+            chain_name,
+            next_worker_set,
+        } => {
+            execute::register_active_worker_set(deps, chain_name, next_worker_set)
+        },
     }
     .map_err(axelar_wasm_std::ContractError::from)
 }
--- a/contracts/monitoring/src/contract.rs
+++ b/contracts/monitoring/src/contract.rs
@@ -1,8 +1,9 @@
 use crate::msg::{ExecuteMsg, InstantiateMsg, QueryMsg};
 use crate::state::{Config, CONFIG};
 use cosmwasm_std::entry_point;
-use cosmwasm_std::{to_binary, Binary, Deps, DepsMut, Env, MessageInfo, Response, StdResult};
+use cosmwasm_std::{to_binary, Binary, Deps, DepsMut, Env, MessageInfo, Response};
 
+use crate::error::ContractError;
 use crate::execute;
 use crate::query;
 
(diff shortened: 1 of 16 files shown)


contracts/monitoring/src/msg.rs
```diff
@@ -15,11 +15,14 @@ pub enum ExecuteMsg {
...
+        next_worker_set: WorkerSet,
+    },
 }
 
 #[cw_serde]
 #[derive(QueryResponses)]
 pub enum QueryMsg {
     #[returns(WorkerSet)]
-    GetActiveVerifiersForChain { chain: ChainName },
+    GetActiveWorkerSet { prover_address: Addr },
```
cjcobb23, line 27: A couple of things here:

1. I think it makes more sense to just pass the chain name, instead of the prover address. The caller probably doesn't know the prover address (if they did, they could just query the prover directly). One of the main jobs of this contract is to keep track of the prover address.
2. If this query is not actually going to be used by the service registry or some other contract, I would just remove it.

contracts/monitoring/src/contract.rs
```diff
@@ -37,110 +38,216 @@ pub fn execute(
             execute::check_governance(&deps, info)?;
             execute::register_prover(deps, chain_name, new_prover_addr)
         }
+        ExecuteMsg::RegisterActiveWorkerSet { next_worker_set } => {
+            // TODO: add check_prover to make sure prover is part of the system
```
cjcobb23, line 42: I actually think it's not totally necessary to check that. Internally you are just mapping this worker set to the sender address. If some random address sends in a worker set, that's not really going to cause a problem, because the contract is just going to associate a worker set with that address, but the address will never be associated with a chain.

contracts/monitoring/src/msg.rs
```diff
@@ -15,11 +15,14 @@ pub enum ExecuteMsg {
         chain_name: ChainName,
         new_prover_addr: Addr,
     },
+    RegisterActiveWorkerSet {
```
cjcobb23, line 18: Let's use verifiers instead of workers. workers was the old term, before we started using verifiers.

contracts/multisig-prover/src/test/mocks/monitoring.rs
```diff
@@ -0,0 +1,50 @@
...
+
+pub fn execute(
+    deps: DepsMut,
+    _env: Env,
+    info: MessageInfo,
+    msg: ExecuteMsg,
+) -> Result<Response, StdError> {
+    match msg {
+        ExecuteMsg::RegisterActiveWorkerSet { next_worker_set } => {
+            register_active_worker_set(deps, info.sender, next_worker_set);
```
cjcobb23, line 30: I don't think you actually need to do any of this. I think you can just return `Ok(Response::new())`

contracts/monitoring/src/state.rs
```diff
@@ -10,6 +11,10 @@ pub struct Config {
 
 pub const CONFIG: Item<Config> = Item::new("config");
 
-type ProverAddresses = Vec<Addr>;
+type ProverAddress = Addr;
```
cjcobb23, line 14: I don't think this is really necessary. Up to you, but I would just use Addr directly.

contracts/multisig-prover/src/execute.rs
```diff
@@ -283,11 +284,23 @@ pub fn confirm_worker_set(deps: DepsMut, sender: Addr) -> Result<Response, Contr
...
+
+    response = response.add_message(wasm_execute(
+        config.monitoring,
+        &monitoring::msg::ExecuteMsg::RegisterActiveWorkerSet {
+            next_worker_set: worker_set,
+        },
+        vec![],
+    )?);
+
+    Ok(response)
```
cjcobb23, line 303: Can't you just do something like `Ok(Response::new().add_message(...).add_message(...)` ?

contracts/monitoring/src/msg.rs
```diff
@@ -15,11 +15,14 @@ pub enum ExecuteMsg {
...
+        next_worker_set: WorkerSet,
```
cjcobb23, line 19:
```suggestion
        worker_set: WorkerSet,
```
Also, do we really need the WorkerSet in monitoring, or can we just use the Addresses?

integration-tests/src/multisig_prover_contract.rs
```diff
@@ -28,6 +28,13 @@ impl MultisigProverContract {
         let app = &mut protocol.app;
         let code_id = app.store_code(Box::new(code));
 
+        let numerator: nonempty::Uint64 = Uint64::from(2u8).try_into().unwrap();
+        let denominator: nonempty::Uint64 = Uint64::from(3u8).try_into().unwrap();
+        let signing_threshold = Threshold::try_from((numerator, denominator))
+            .unwrap()
+            .try_into()
+            .unwrap();
```
cgorenflo, line 36:
```suggestion
        let signing_threshold = Threshold::try_from((2u64, 3u64)).unwrap().try_into().unwrap();
```

contracts/monitoring/src/msg.rs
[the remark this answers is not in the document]
maancham: ok changed!

maancham: Changed it to `SetActiveVerifiers`.

contracts/multisig-prover/src/test/mocks/monitoring.rs
maancham: I am doing the same thing inside the function:
```
pub fn set_active_worker_set(
    deps: DepsMut,
    info: MessageInfo,
    next_worker_set: WorkerSet,
    ACTIVE_WORKERSET_FOR_PROVER.save(deps.storage, info.sender, &(next_worker_set))?;
    Ok(Response::new())
}
```

contracts/monitoring/src/msg.rs
maancham: 1. Sure
2. It is going to be used. For now, we know for certain axelarscan is going to need it.

contracts/monitoring/src/msg.rs
maancham: I think we will need it in future...

cgorenflo (approved):

contracts/multisig-prover/src/test/mocks/monitoring.rs
cjcobb23: yes but why do you need to actually store anything here? This is the mock and I don't think the query is being called, unless I am mistaken. You just need the execute msg to not error, right?

contracts/monitoring/src/msg.rs
```diff
@@ -15,11 +15,14 @@ pub enum ExecuteMsg {
...
+    },
 }
 
 #[cw_serde]
 #[derive(QueryResponses)]
 pub enum QueryMsg {
     #[returns(WorkerSet)]
-    GetActiveVerifiersForChain { chain: ChainName },
+    GetActiveWorkerSet { chain_name: ChainName },
```
cjcobb23, line 27:
```suggestion
    GetActiveVerifiers { chain_name: ChainName },
```

contracts/multisig-prover/src/test/mocks/monitoring.rs
```diff
@@ -0,0 +1,50 @@
...
+
+type ProverAddress = Addr;
+const ACTIVE_WORKERSET_FOR_PROVER: Map<ProverAddress, WorkerSet> =
+    Map::new("active_prover_workerset");
+
+fn register_active_worker_set(deps: DepsMut, caller_address: Addr, worker_set: WorkerSet) {
+    ACTIVE_WORKERSET_FOR_PROVER
+        .save(deps.storage, caller_address, &(worker_set))
+        .expect("Saving Active WorkerSet with Monitoring Failed");
+}
```
cjcobb23, line 44: you also don't need any of this right?

maancham: I think we do, this is the only functionality from monitoring we need to mock in order for prover tests to work.

integration-tests/tests/update_worker_set.rs
```diff
@@ -283,3 +289,140 @@ fn worker_set_update_can_be_resigned() {
         multisig_prover::msg::ProofStatus::Completed { .. }
     ));
 }
+
+#[test]
+fn register_active_workerset_with_monitoring_success() {
```
cjcobb23, line 294: why does this need it's own test case? Isn't this the exact same test as the first test in this file (worker_set_can_be_initialized_and_then_manually_updated)? The only difference seems to be that we check monitoring has the correct worker set. If so, just add that line into the first test

contracts/multisig-prover/src/test/mocks/monitoring.rs
maancham: Its done.

cjcobb23 (approved):

PR #353, opened 2024-04-16
fix(minor-voting-verifier)!: remove unused response data

eguajardo:
## Description

Response data is used for contract to contract communication, but the only contract using that is the aggregate verifier which is unused, the parsing is broken and the aggregate itself is planned to be removed. Therefore the response is being removed

https://axelarnetwork.atlassian.net/browse/AXE-3289

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


contracts/voting-verifier/src/contract.rs
```diff
@@ -249,39 +248,6 @@ mod test {
...
-                (
-                    CrossChainId {
-                        id: message_id("id", 1),
-                        chain: source_chain()
-                    },
-                    VerificationStatus::None
-                ),
-            ]
-        );
-    }
```
eguajardo, line 283: removed because it was only checking response data and checking other stuff was already considered in other tests

cjcobb23: Can't we just change this test to call VerifyMessages and then query GetMessageStatuses?

eguajardo: this is what I mean with checking other stuff is already done in other tests, changing this test will basically be a duplicate of this one: https://github.com/axelarnetwork/axelar-amplifier/blob/58e98060d86f51e949c1fd7fb4ec2765f5b37e0c/contracts/voting-verifier/src/contract.rs#L529-L562

cjcobb23 (approved):

PR #354, opened 2024-04-17
feat(axelar-wasm-std): implement sui message id trait

cjcobb23:
https://axelarnetwork.atlassian.net/browse/AXE-3720


packages/axelar-wasm-std/src/msg_id/base_58_event_index.rs
cgorenflo, line 1: can you add a test where you start with a hex with leading zeros and then you either keep or trim leading 1s after conversion? Then decode it again and make sure the hexes don't match up?

cgorenflo (approved):

PR #357, opened 2024-04-18
feat(minor-connection-router)!: verify message IDs

cjcobb23:
* chain registration includes message id format
* verify msg ids conform to registered format when executing route_messages

https://axelarnetwork.atlassian.net/browse/AXE-3560
https://axelarnetwork.atlassian.net/browse/AXE-3716
https://axelarnetwork.atlassian.net/browse/AXE-3717


contracts/connection-router/src/contract/execute.rs
```diff
@@ -137,9 +144,16 @@ where
         sender: &Addr,
         msgs: Vec<Message>,
     ) -> error_stack::Result<Vec<Message>, Error> {
+        let verify_msg_ids = |msgs: &Vec<Message>, expected_format| {
+            msgs.iter()
+                .try_for_each(|msg| msg_id::verify_msg_id(&msg.cc_id.id, expected_format))
+                .change_context(Error::InvalidMessageId)
+        };
```
cgorenflo, line 151: this doesn't close over any variable, so you could also move it into an actual function

contracts/connection-router/src/contract/execute.rs
```diff
@@ -137,9 +144,16 @@ where
...
+
         // if sender is the nexus gateway, we cannot validate the source chain
         // because the source chain is registered in the core nexus module
```
cgorenflo, line 154: add to this comment that we only allow one specific message id format from core

contracts/connection-router/src/contract/execute.rs
cgorenflo, line 1: could you add one more test that doesn't just take a completely invalid message id, but one that would be valid in a different format, i.e. register two chains with different formats, assert they each accept an appropriate id and then assert that they don't accept the id for the other chain

PR #358, opened 2024-04-19
feat(minor-coordinator)!: add can unbond utility and connect service registry to coordinator

maancham:
## Description

[AXE-1389]

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Other Notes

We also want to add a multi index map to keep the active/next verifiers for each prover. This will greatly decrease the amount of computation required when coordinator is asked if a verifier is ready to unbond. However, it's not in the scope of this PR (as we need a new execute message and probably an IndexedMap for coordinator and changes to provers confirm and update workerset functions).

[AXE-1389]: https://axelarnetwork.atlassian.net/browse/AXE-1389?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/coordinator/src/msg.rs
```diff
@@ -25,4 +26,10 @@ pub enum ExecuteMsg {
 pub enum QueryMsg {
     #[returns(WorkerSet)]
     GetActiveVerifiers { chain_name: ChainName },
+
+    #[returns(bool)]
+    CheckWorkerCanUnbond {
+        worker_address: Addr,
+        chains: HashSet<ChainName>,
```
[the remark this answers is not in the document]
cjcobb23: actually I realized we are not tracking the chains associated with a given worker in this contract. But didn't we want to do that? Wasn't that the whole reason to move to a push model and actually store the active workersets in this contract? Otherwise, we already have the prover address, so why not just query the prover directly each time.

contracts/service-registry/src/contract/execute.rs
```diff
@@ -166,15 +166,26 @@ pub fn unbond_worker(
...
+    let service = SERVICES
         .may_load(deps.storage, &service_name)?
         .ok_or(ContractError::ServiceNotFound)?;
 
     let worker = WORKERS
         .may_load(deps.storage, (&service_name, &info.sender))?
         .ok_or(ContractError::WorkerNotFound)?;
 
-    let can_unbond = true; // TODO: actually query the service to determine this value
+    let chains = CHAINS_PER_WORKER
```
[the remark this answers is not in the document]
cjcobb23: I think it does not. I think if a worker calls deregister chain support, the contract will remove them from this map, and then we won't actually check whether the worker is still in the active set. Double check me here.

integration-tests/tests/bond_unbond.rs
```diff
@@ -14,8 +23,29 @@ fn workers_can_claim_stake() {
...
     let before_balances = test_utils::query_balances(&protocol.app, &workers);
 
+    let new_workers = test_utils::create_new_workers_vec(
+        chains.clone(),
+        vec![("worker3".to_string(), 3), ("worker4".to_string(), 4)],
+    );
+
+    test_utils::register_workers(&mut protocol, &new_workers, min_worker_bond);
+
     test_utils::deregister_workers(&mut protocol, &workers);
```
cjcobb23, line 33: I think this function just deauthorizes right? If so, can we have a test where the worker is not deauthorized, but simply deregisters chain support? So worker deregisters chain support, and then tries to unbond while still in the active workerset, which should fail.

cjcobb23 (changes requested):

cjcobb23:
Also, I realized we need to handle the case where the worker is not in the active set, but they are in the next worker set that is already in rotation (created at the prover and just awaiting confirmation).

contracts/coordinator/src/contract.rs
```diff
@@ -52,6 +52,15 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> Result<Binary, ContractErr
         QueryMsg::GetActiveVerifiers { chain_name } => {
             to_binary(&query::get_active_worker_set(deps, chain_name)?).map_err(|err| err.into())
         }
+        QueryMsg::CheckWorkerCanUnbond {
```
cgorenflo, line 55:
```suggestion
        QueryMsg::ReadyToUnbond {
```
I find "can unbond" a confusing term because it sounds like we're checking if the worker is allowed to unbond. From how the response is used that is not the case: If the result is `false`, the service registry starts the unbonding process for the worker, if it's `true` it completes it. So I advocate for a rename

contracts/coordinator/src/msg.rs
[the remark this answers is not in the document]
cgorenflo: agreed, this check should be very simple: "is this worker ready to unbond?" the coordinator should figure out all the necessary information

contracts/service-registry/src/contract/execute.rs
```diff
@@ -166,15 +166,26 @@ pub fn unbond_worker(
...
+        .may_load(deps.storage, (&service_name, &worker.address))?
+        .unwrap_or_default();
```
cgorenflo, line 179: the service registry shouldn't have to figure out which chains to call

contracts/service-registry/src/contract/execute.rs
```diff
@@ -166,15 +166,26 @@ pub fn unbond_worker(
...
+
+    let query = coordinator::msg::QueryMsg::CheckWorkerCanUnbond {
+        worker_address: worker.address.clone(),
+        chains: chains.clone(),
+    };
+    let can_unbond = deps.querier.query(&QueryRequest::Wasm(WasmQuery::Smart {
+        contract_addr: service.coordinator_contract.into(),
+        msg: to_binary(&query)?,
+    }))?;
```
cgorenflo, line 188:
```suggestion
    let ready_to_unbond = deps.querier.query(&QueryRequest::Wasm(WasmQuery::Smart {
        contract_addr: service.coordinator_contract.into(),
        msg: to_binary(&query)?,
    }))?;
```

contracts/service-registry/src/contract/execute.rs
```diff
@@ -166,15 +166,26 @@ pub fn unbond_worker(
...
 
     let worker = worker.unbond(can_unbond, env.block.time)?;
```
cgorenflo, line 190: change the argument name inside the `unbond` function to `ready_to_unbond` as well

contracts/service-registry/src/state.rs
```diff
@@ -21,7 +21,7 @@ use crate::ContractError;
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
 pub struct Service {
     pub name: String,
-    pub service_contract: Addr,
+    pub coordinator_contract: Addr,
```
cgorenflo, line 24: did the purpose of this field change? What was it and why is it not needed anymore?

integration-tests/tests/bond_unbond.rs
cgorenflo, line 1: do we have a test for unbonding while jailed?

cjcobb23:
I realized there is an edge case that we didn't consider in the design. Consider the case where a workerset update is started at the multisig prover, and a given worker is not in the currently active workerset, but will be in the next workerset (once the update is confirmed). Then, that worker tries to unbond. They will not be in the active workerset yet, so the unbonding is allowed. Then, the new workerset is confirmed at the multisig prover. So now the worker is in the active workerset, but is unbonding. I think we are not handling this case, and we should.

contracts/service-registry/src/state.rs
maancham: I was told that it was always meant for the coordinator. Since the name was not finalized when creating the instantiation of service registry, it was named as `service_contract`.

contracts/coordinator/src/msg.rs
```diff
@@ -25,4 +26,10 @@ pub enum ExecuteMsg {
 pub enum QueryMsg {
     #[returns(WorkerSet)]
     GetActiveVerifiers { chain_name: ChainName },
+
+    #[returns(bool)]
+    CheckWorkerCanUnbond {
```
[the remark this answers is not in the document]
maancham: I changed it to `ReadyToUnbond` based on this comment.

contracts/coordinator/src/contract.rs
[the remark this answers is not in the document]
maancham: I changed it to `ReadyToUnbond` based on this comment.

maancham: Makes sense. Thanks.

contracts/coordinator/src/msg.rs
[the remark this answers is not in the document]
maancham: Thanks for mentioning it. This is fixed now.

contracts/service-registry/src/contract/execute.rs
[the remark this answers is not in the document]
maancham: I have migrated the map to coordinator itself. There are not deregister support there and therefore, won't have this issue now.

integration-tests/tests/bond_unbond.rs
maancham: Added a test for it.

integration-tests/tests/bond_unbond.rs
maancham: Added!

contracts/multisig-prover/src/execute.rs
```diff
@@ -190,13 +190,21 @@ fn get_next_worker_set(
...
     NEXT_WORKER_SET.save(storage, new_worker_set)?;
-    Ok(())
+
+    Ok(Response::new().add_message(wasm_execute(
+        coordinator_address,
+        &coordinator::msg::ExecuteMsg::SetNextVerifiers {
+            next_worker_set: new_worker_set.clone(),
+        },
+        vec![],
+    )?))
```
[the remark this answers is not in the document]
maancham: I changed it and added the message to update_worker_set instead which is the one who is calling the save_next_worker_set.

contracts/coordinator/src/query.rs
```diff
@@ -18,3 +20,52 @@ pub fn get_active_worker_set(deps: Deps, chain_name: ChainName) -> StdResult<Opt
 
     Ok(active_worker_set)
 }
+
+pub fn get_next_worker_set(deps: Deps, chain_name: ChainName) -> StdResult<Option<WorkerSet>> {
```
cgorenflo, line 24:
```suggestion
pub fn next_worker_set(deps: Deps, chain_name: ChainName) -> StdResult<Option<WorkerSet>> {
```

contracts/service-registry/src/state.rs
```diff
@@ -171,10 +171,10 @@ pub fn register_chains_support(
...
 ) -> Result<(), ContractError> {
-    CHAINS_PER_WORKER.update(storage, (&service_name, &worker), |current_chains| {
-        let mut current_chains = current_chains.unwrap_or_default();
-        current_chains.extend(chains.iter().cloned());
-        Ok::<HashSet<ChainName>, ContractError>(current_chains)
+    CHAINS_PER_WORKER.update(storage, (&service_name, &worker), |existing_chains| {
+        let mut existing_chains = existing_chains.unwrap_or_default();
+        existing_chains.extend(chains.iter().cloned());
+        Ok::<HashSet<ChainName>, ContractError>(existing_chains)
     })?;
```
cjcobb23, line 178: do we need this data structure at all anymore?

contracts/coordinator/src/msg.rs
```diff
@@ -18,11 +18,21 @@ pub enum ExecuteMsg {
     SetActiveVerifiers {
         next_worker_set: WorkerSet,
     },
+    SetNextVerifiers {
+        next_worker_set: WorkerSet,
+    },
+    AddSupportedChainsForWorker {
+        chains: Vec<ChainName>,
+        worker: Addr,
+    },
```
cjcobb23, line 27: Why do we need this method? Can't this info just be inferred from which provers are calling `SetActiveVerifiers` and `SetNextVerifiers`?

contracts/coordinator/src/query.rs
```diff
@@ -18,3 +20,52 @@ pub fn get_active_worker_set(deps: Deps, chain_name: ChainName) -> StdResult<Opt
...
+            .signers
+            .values()
+            .any(|signer| signer.address == *worker_address)
+        {
+            return Ok(true);
+        }
+    }
+
+    Ok(false)
+}
```
cjcobb23, line 58: you could probably simplify this and write it as just one function:
get the prover address
get the active workerset
get the next workerset
join the sets
check if the worker is in the joined set

contracts/coordinator/src/query.rs
```diff
@@ -18,3 +20,52 @@ pub fn get_active_worker_set(deps: Deps, chain_name: ChainName) -> StdResult<Opt
 
     Ok(active_worker_set)
 }
+
+pub fn get_next_worker_set(deps: Deps, chain_name: ChainName) -> StdResult<Option<WorkerSet>> {
+    let prover_address = provers(deps, chain_name).unwrap();
```
cjcobb23, line 25: I don't think it's safe to unwrap here, the function can return an error

contracts/coordinator/src/contract.rs
```diff
@@ -59,6 +59,12 @@ pub fn execute(
         ExecuteMsg::SetActiveVerifiers { next_worker_set } => {
             execute::set_active_worker_set(deps, info, next_worker_set)
         }
+        ExecuteMsg::SetNextVerifiers { next_worker_set } => {
+            execute::set_next_worker_set(deps, info, next_worker_set)
+        }
+        ExecuteMsg::AddSupportedChainsForWorker { chains, worker } => {
```
cjcobb23, line 65: I think this needs to be access restricted to the service registry. Right now, anyone can call this with whatever input they like. I think this can be removed though and the info can just be inferred from the above two methods (`SetActiveVerifiers` and `SetNextVerifiers`)

contracts/service-registry/src/state.rs
maancham: Removed!

contracts/coordinator/src/query.rs
maancham: I have some unit tests that rely on the individual functions (get active and get next) so we can't remove the functions.

contracts/coordinator/src/contract.rs
maancham: Removed

cjcobb23 (approved):

ampd/Cargo.toml
```diff
@@ -35,7 +35,6 @@ move-core-types = { git = "https://github.com/mystenlabs/sui", tag = "mainnet-v1
 multisig = { workspace = true }
 num-traits = { workspace = true }
 prost = "0.11.9"
-prost-types = "0.11.9"
```
fish-sammy, line 38: Why remove this?

fish-sammy (changes requested):

ampd/Cargo.toml
maancham: There are a ton of conflicts before merge. I talked with Chris. I have all other files on local and will push when we decide on what to do with them. I have actually not removed this file. Marking PR as draft until then...

cgorenflo (approved):

contracts/coordinator/src/query.rs
```diff
@@ -1,23 +1,66 @@
...
+        Ok(prover_address) => {
+            let active_worker_set =
+                ACTIVE_VERIFIER_SET_FOR_PROVER.may_load(deps.storage, prover_address)?;
+            Ok(active_worker_set)
+        }
+        Err(_err) => Ok(None),
+    }
+}
+
+// TODO: Use the chain name to find out which provers to query.
```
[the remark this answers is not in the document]
maancham: Yes. But it is an optimization rather that breaking change.

contracts/service-registry/src/migrations/v_0_3.rs
```diff
@@ -62,7 +62,7 @@ fn migrate_services(store: &mut dyn Storage) -> Result<(), ContractError> {
                     service_name,
                     Service {
                         name: service.name,
-                        service_contract: service.service_contract,
+                        coordinator_contract: service.service_contract,
```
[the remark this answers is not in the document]
maancham: If I revert the changes in this file, everything fails (all instantiations of service registry). Should I revert all the changes made?

contracts/coordinator/src/query.rs
```diff
@@ -1,23 +1,68 @@
...
+
+    for chain_name in chain_names? {
+        if let Ok(prover_address) = PROVER_PER_CHAIN.load(deps.storage, chain_name) {
+            if let Ok(Some(verifier_set)) =
+                ACTIVE_VERIFIER_SET_FOR_PROVER.may_load(deps.storage, prover_address.clone())
+            {
+                if verifier_set
+                    .signers
+                    .values()
+                    .any(|signer| signer.address == *verifier_address)
```
fish-sammy, line 40: Create a `.includes` method for `VerifierSet`. Also, the comment says that the keys in `.signers` are the signer addresses so why are you loop to check existence here with a `BTreeMap`?

contracts/service-registry/src/migrations/v_0_3.rs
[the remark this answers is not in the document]
haiyizxx: remove the v0.3 migration so it won't block merge, and have a separate pr for state migration.

contracts/coordinator/src/query.rs
fish-sammy, line 40: Create a `.includes` method for `VerifierSet`. Also, the comment says that the keys in `.signers` are the signer addresses so why are you loop to check existence here with a `BTreeMap`?

contracts/coordinator/src/query.rs
```diff
@@ -1,23 +1,68 @@
...
+// TODO: Use the chain name to find out which provers to query.
+fn is_verifier_in_verifier_set(deps: Deps, verifier_address: &Addr) -> StdResult<bool> {
+    let chain_names: StdResult<Vec<ChainName>> = PROVER_PER_CHAIN
+        .keys(deps.storage, None, None, cosmwasm_std::Order::Ascending)
+        .collect();
+
+    for chain_name in chain_names? {
+        if let Ok(prover_address) = PROVER_PER_CHAIN.load(deps.storage, chain_name) {
+            if let Ok(Some(verifier_set)) =
+                ACTIVE_VERIFIER_SET_FOR_PROVER.may_load(deps.storage, prover_address.clone())
```
fish-sammy, line 35: Why not use the `active_verifier_set` function above here?

contracts/coordinator/src/query.rs
```diff
@@ -1,23 +1,68 @@
...
+                }
+            }
+
+            if let Ok(Some(verifier_set)) =
+                NEXT_VERIFIER_SET_FOR_PROVER.may_load(deps.storage, prover_address)
+            {
+                if verifier_set
+                    .signers
+                    .values()
+                    .any(|signer| signer.address == *verifier_address)
```
fish-sammy, line 52: Same here. Looping through the `BTreeMap` defeats the purpose of using a map.

contracts/coordinator/src/query.rs
```diff
@@ -1,23 +1,68 @@
...
+    }
+}
+
+// TODO: Use the chain name to find out which provers to query.
+fn is_verifier_in_verifier_set(deps: Deps, verifier_address: &Addr) -> StdResult<bool> {
+    let chain_names: StdResult<Vec<ChainName>> = PROVER_PER_CHAIN
+        .keys(deps.storage, None, None, cosmwasm_std::Order::Ascending)
+        .collect();
+
+    for chain_name in chain_names? {
```
fish-sammy, line 32:
```suggestion
    let chain_names = PROVER_PER_CHAIN
        .keys(deps.storage, None, None, cosmwasm_std::Order::Ascending)
        .collect::<StdResult<Vec<_>>>()?;

    for chain_name in chain_names {
```

contracts/multisig-prover/src/execute.rs
```diff
@@ -241,10 +241,18 @@ pub fn update_verifier_set(deps: DepsMut, env: Env) -> Result<Response, Contract
...
-                wasm_execute(config.multisig, &start_sig_msg, vec![])?,
-                START_MULTISIG_REPLY_ID,
-            )))
+            Ok(Response::new()
+                .add_submessage(SubMsg::reply_on_success(
+                    wasm_execute(config.multisig, &start_sig_msg, vec![])?,
+                    START_MULTISIG_REPLY_ID,
+                ))
+                .add_message(wasm_execute(
+                    config.coordinator.clone(),
```
fish-sammy, line 250: The coordinator contract should have its client mod. Check https://github.com/axelarnetwork/axelar-amplifier/blob/24080b4d91528dfab2d91f8ada5d6e0985262be0/contracts/voting-verifier/src/client.rs#L4. Can do in a separate PR.

fish-sammy (changes requested):

contracts/multisig-prover/src/execute.rs
maancham: Created a ticket for it.

contracts/coordinator/src/query.rs
maancham: Added.

contracts/service-registry/Cargo.toml
```diff
@@ -35,6 +35,7 @@ optimize = """docker run --rm -v "$(pwd)":/code \
 [dependencies]
 axelar-wasm-std = { workspace = true }
 axelar-wasm-std-derive = { workspace = true }
+coordinator = { workspace = true }
```
haiyizxx, line 38:
```suggestion
coordinator = { workspace = true,  features = ["library"] }
```

contracts/coordinator/src/query.rs
fish-sammy: What do you mean it's used for tests only? It's used here https://github.com/axelarnetwork/axelar-amplifier/blob/1558bedeb0659a1c8c8b1b88692efa2fd9305f59/contracts/coordinator/src/contract.rs#L75.

contracts/multisig/src/verifier_set.rs
```diff
@@ -59,4 +59,8 @@ impl VerifierSet {
             .map(|(address, signer)| (address.clone(), signer.pub_key.clone()))
             .collect()
     }
+
+    pub fn includes(&self, given_address: &str) -> bool {
+        self.signers.contains_key(given_address)
+    }
```
fish-sammy, line 65:
```suggestion
    pub fn includes(&self, signer: &Addr) -> bool {
        self.signers.contains_key(signer.as_str())
    }
```

fish-sammy (approved):

contracts/coordinator/src/query.rs
fish-sammy: Then using functions to avoid duplicate code should be the consistency to follow.

PR #361, opened 2024-04-22
feat(ampd): broadcaster queries account sequence at broadcast

fish-sammy:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


ampd/src/broadcaster/accounts.rs
```diff
@@ -16,7 +16,7 @@ pub enum Error {
     MalformedResponse,
 }
 
-pub async fn account<T>(mut client: T, address: &TMAddress) -> Result<BaseAccount, Error>
+pub async fn account<T>(client: &mut T, address: &TMAddress) -> Result<BaseAccount, Error>
```
cgorenflo, line 19: huh, that was a strange decision to have an owned parameter previously

cgorenflo (approved):

ampd/src/broadcaster.rs
```diff
@@ -150,16 +152,47 @@ where
...
+            self.acc_sequence.unwrap_or_default(),
+        ));
+
+        Ok((account.account_number, *acc_sequence))
+    }
+
+    async fn estimate_fee(
+        &mut self,
+        msgs: Vec<cosmrs::Any>,
+        acc_sequence: u64,
```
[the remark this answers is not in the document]
fish-sammy: But there's more useage of self in the method. The reason for making this change is that if the `broadcast` method makes the account query call twice in itself, it makes things unnecessarily complicated.

ampd/src/broadcaster/accounts.rs
fish-sammy: yup it was.

PR #366, opened 2024-04-23
feat(minor-multisig-prover): amplifier gateway rotate signers

haiyizxx:
## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [x] Connect epics/issues


contracts/multisig-prover/src/execute.rs
```diff
@@ -228,17 +233,37 @@ pub fn update_worker_set(deps: DepsMut, env: Env) -> Result<Response, ContractEr
...
-            let batch = builder.build()?;
+                let batch = builder.build()?;
 
-            COMMANDS_BATCH.save(deps.storage, &batch.id, &batch)?;
-            REPLY_BATCH.save(deps.storage, &batch.id)?;
+                COMMANDS_BATCH.save(deps.storage, &batch.id, &batch)?;
+                REPLY_BATCH.save(deps.storage, &batch.id)?;
+
+                msg_to_sign = batch.msg_digest();
+            }
```
cjcobb23, line 249: Can we extract this into a function? As well as the case where the feature flag is enabled?

contracts/multisig-prover/src/msg.rs
```diff
@@ -49,6 +49,7 @@ pub struct InstantiateMsg {
     /// deployed on the destination chain. The multisig contract supports multiple public keys per verifier (each a different type of key), and this
     /// parameter controls which registered public key to use for signing for each verifier registered to the destination chain.
     pub key_type: KeyType,
+    pub domain_separator: HexBinary,
```
cjcobb23, line 52: Can we use a type that has a length here, like Uint256, or just Hash?

contracts/multisig-prover/src/execute.rs
```diff
@@ -228,17 +233,37 @@ pub fn update_worker_set(deps: DepsMut, env: Env) -> Result<Response, ContractEr
...
+
+            #[cfg(feature = "amplifier-gateway")]
+            {
+                let message_to_sign = MessageToSign::new(Payload::WorkerSet(new_worker_set));
+                MESSAGE_TO_SIGN.save(deps.storage, &message_to_sign.id, &message_to_sign)?;
```
cjcobb23, line 254: Where is this stored value actually being used?

contracts/multisig-prover/src/state.rs
```diff
@@ -35,6 +36,7 @@ fn default_governance() -> Addr {
 
 pub const CONFIG: Item<Config> = Item::new("config");
 pub const COMMANDS_BATCH: Map<&BatchId, CommandBatch> = Map::new("command_batch");
+pub const MESSAGE_TO_SIGN: Map<&BatchId, MessageToSign> = Map::new("message_to_sign");
```
cjcobb23, line 39: This is only needed for the amplifier gateway right? If so, I would put this under the feature flag. Otherwise can be confusing.

cjcobb23 (changes requested):
I am not so sure about this approach of supporting both gateway types at the same time. I feel like this was our most complex contract already and needed refactoring, and now it is significantly more complex. I think it's very hard to understand when to use the old vs the new logic, what certain types are meant to be used for (there are now a bunch of types that are similar to existing types). I think the contract is not really maintainable in this state, and I have a hard time understanding what is for what (and I wrote a lot of this contract lol).

I propose that we just migrate to the new gateway logic, and delete the old. The release we just did will work for the old gateway. Going forward we can target the new gateway.

contracts/multisig-prover/src/encoding/abi2/solidity/AmplifierGatewayTypes.sol
```diff
@@ -0,0 +1,27 @@
...
+ * @param contractAddress The address of the contract that the message is intended for
+ * @param payloadHash The hash of the payload that is to be processed
+ */
+struct Message {
+    string messageId;
+    string sourceChain;
+    string sourceAddress;
+    address contractAddress;
+    bytes32 payloadHash;
+}
```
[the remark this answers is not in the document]
haiyizxx: I wish so, it accepts either solidity code, or  a path to a solidity file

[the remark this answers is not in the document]
fish-sammy: @haiyizxx Does it accept a URL?

contracts/multisig-prover/src/encoding/abi2/mod.rs
```diff
@@ -0,0 +1,202 @@
...
+            WeightedSigners::from(worker_set),
+        )
+            .abi_encode_sequence(),
+    }
+}
+
+fn evm_address(pub_key: &[u8]) -> Result<Address, ContractError> {
+    PublicKey::from_sec1_bytes(pub_key)
+        .map(|pub_key| pub_key.to_encoded_point(false))
+        .map(|pub_key| Address::from_raw_public_key(&pub_key.as_bytes()[1..]))
```
fish-sammy, line 102: Why exclude the first byte?

contracts/multisig-prover/src/state.rs
fish-sammy: I am confused with this, too. What is this v.s. `COMMANDS_BATCH`?

contracts/multisig-prover/src/types.rs
```diff
@@ -114,3 +114,65 @@ pub struct WorkersInfo {
...
+    }
+}
+
+#[cw_serde]
+pub struct MessageToSign {
+    pub id: BatchId,
+    pub payload: Payload,
+}
+
+impl MessageToSign {
```
fish-sammy, line 150: I don't like this name.

fish-sammy (changes requested):

contracts/multisig-prover/src/encoding/abi2/solidity/AmplifierGatewayTypes.sol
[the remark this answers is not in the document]
haiyizxx: don't think so, could you guys double check [doc](https://docs.rs/alloy-sol-macro/0.7.1/alloy_sol_macro/macro.sol.html#examples)

contracts/multisig-prover/src/encoding/abi2/mod.rs
cgorenflo, line 1: is the name of the file temporary or permanent? I prefer versioning to do `encoding/abi/v2`

contracts/multisig-prover/src/msg.rs
```diff
@@ -49,6 +49,8 @@ pub struct InstantiateMsg {
     /// deployed on the destination chain. The multisig contract supports multiple public keys per verifier (each a different type of key), and this
     /// parameter controls which registered public key to use for signing for each verifier registered to the destination chain.
     pub key_type: KeyType,
+    // an opaque value created to distinguish distinct chains, that the external gateway should be initialized with.
```
cgorenflo, line 52:
```suggestion
    // an opaque value created to distinguish distinct chains that the external gateway should be initialized with.
```

contracts/multisig-prover/src/encoding/abi2/solidity/AmplifierGatewayTypes.sol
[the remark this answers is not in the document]
cgorenflo: I tried it, it doesn't work. I also wouldn't want to be dependent on a working url

contracts/multisig-prover/src/types.rs
cgorenflo, line 1: mark old types as deprecated?

cgorenflo: this is a good reason to improve the file structure. Instead of adding to the confusing in the super generic `types.rs` put the new types into better fitting files.

cgorenflo (changes requested):

contracts/multisig-prover/src/encoding/abi2/solidity/AmplifierGatewayTypes.sol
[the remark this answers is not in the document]
cjcobb23: I wonder if we can add the gateway repo as a submodule to this repo, and then just include it via path.

contracts/multisig-prover/src/encoding/abi2/mod.rs
haiyizxx: it's a just temporary name, not versioning

contracts/multisig-prover/src/command.rs
```diff
@@ -0,0 +1,74 @@
...
+    pub fn id(&self) -> BatchId {
+        match &self {
+            Payload::Messages(msgs) => {
+                let mut message_ids = msgs
+                    .iter()
+                    .map(|msg| msg.cc_id.to_string())
+                    .collect::<Vec<_>>();
+                message_ids.sort();
+
+                Keccak256::digest(message_ids.join(",")).as_slice().into()
```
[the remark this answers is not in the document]
haiyizxx: just want to keep the current behaviour https://github.com/axelarnetwork/axelar-amplifier/blob/main/contracts/multisig-prover/src/types.rs#L74

contracts/multisig-prover/src/command.rs
```diff
@@ -0,0 +1,74 @@
...
+            Encoder::Bcs => todo!(),
+        }
+    }
+
+    pub fn message_ids(&self) -> Vec<CrossChainId> {
+        match &self.payload {
+            Payload::Messages(msgs) => msgs.iter().map(|msg| msg.cc_id.clone()).collect(),
+            Payload::WorkerSet(_) => vec![],
+        }
+    }
```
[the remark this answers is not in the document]
haiyizxx: not saying I like that, following the precedence https://github.com/axelarnetwork/axelar-amplifier/blob/main/contracts/multisig-prover/src/types.rs#L90, it returns empty vector if the batch contains update worker set. I can remove the function in this pr, it will be used GetProofResponse query. will improve in the follow up pr.

contracts/multisig-prover/src/command.rs
```diff
@@ -0,0 +1,74 @@
...
+            Payload::WorkerSet(worker_set) => worker_set.hash().into(),
+        }
+    }
+}
+
+#[cw_serde]
+pub struct Command {
+    pub id: BatchId,
+    pub payload: Payload,
+}
```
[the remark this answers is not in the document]
haiyizxx: it's just an id and payload mapping, a tradeoff compute and storage. 
It can be removed if we re-derive the payload id everytime.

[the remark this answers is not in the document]
cgorenflo: I'd rather not copy that if it's not necessary 🙂

contracts/multisig-prover/src/command.rs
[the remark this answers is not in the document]
cgorenflo: ok that's what I meant. If we're already doing it like that we should probably keep it

contracts/multisig-prover/src/msg.rs
```diff
@@ -49,6 +49,8 @@ pub struct InstantiateMsg {
     /// deployed on the destination chain. The multisig contract supports multiple public keys per verifier (each a different type of key), and this
     /// parameter controls which registered public key to use for signing for each verifier registered to the destination chain.
     pub key_type: KeyType,
+    // an opaque value created to distinguish distinct chains that the external gateway should be initialized with.
```
cgorenflo, line 52:
```suggestion
    /// an opaque value created to distinguish distinct chains that the external gateway should be initialized with.
```

cgorenflo (approved):

contracts/multisig-prover/src/encoding/abi2/mod.rs
```diff
@@ -0,0 +1,205 @@
...
+impl WeightedSigners {
+    pub fn hash(&self) -> Hash {
+        Keccak256::digest(self.abi_encode()).into()
+    }
+}
+
+impl From<&Signer> for WeightedSigner {
+    fn from(signer: &Signer) -> Self {
+        let weight: Uint128 = signer
+            .weight
```
cjcobb23, line 33: not necessarily for this PR, but we should just change the signer weight to uint128

contracts/multisig-prover/src/encoding/abi2/mod.rs
```diff
@@ -0,0 +1,205 @@
...
+    fn from(worker_set: &WorkerSet) -> Self {
+        let mut signers = worker_set
+            .signers
+            .values()
+            .map(WeightedSigner::from)
+            .collect::<Vec<_>>();
+
+        signers.sort_by_key(|weighted_signer| weighted_signer.signer);
+
+        let threshold: u128 = Uint128::try_from(worker_set.threshold)
```
cjcobb23, line 54: we should consider changing the threshold to uint128 in WorkerSet

contracts/multisig-prover/src/encoding/abi2/mod.rs
```diff
@@ -0,0 +1,205 @@
...
+    domain_separator: &Hash,
+    signer: &WorkerSet,
+    payload: &Payload,
+) -> HexBinary {
+    let signer = WeightedSigners::from(signer);
+    let data_to_sign = encode(payload);
+
+    // Prefix for standard EVM signed data https://eips.ethereum.org/EIPS/eip-191
+    let unsigned = [
+        "\x19Ethereum Signed Message:\n96".as_bytes(), // 96 is the length of the trailing bytes
```
cjcobb23, line 76: we should either compute the length of the trailing bytes, or at least have a test that 96 is the length

contracts/multisig-prover/src/payload.rs
```diff
@@ -0,0 +1,57 @@
...
+        match encoder {
+            Encoder::Abi => abi2::payload_hash_to_sign(domain_separator, curr_worker_set, self),
+            Encoder::Bcs => todo!(),
+        }
+    }
+
+    pub fn message_ids(&self) -> Vec<CrossChainId> {
+        match &self {
+            Payload::Messages(msgs) => msgs.iter().map(|msg| msg.cc_id.clone()).collect(),
+            Payload::WorkerSet(_) => vec![],
```
cjcobb23, line 54: Returning empty vec looks a little weird, I think I would prefer wrapping the return type in Option and returning None

contracts/multisig-prover/src/state.rs
```diff
@@ -26,6 +30,7 @@ pub struct Config {
     pub worker_set_diff_threshold: u32,
     pub encoder: Encoder,
     pub key_type: KeyType,
+    pub domain_separator: Hash,
```
cjcobb23, line 33: Are we planning to migrate the old contract to this contract, or are we just redeploying?

cjcobb23 (approved):

contracts/multisig-prover/src/encoding/abi2/mod.rs
haiyizxx: ya, have a [ticket](https://axelarnetwork.atlassian.net/browse/AXE-3882?atlOrigin=eyJpIjoiMjVjY2NlMWY0ZWI2NDQ4ZmJhMDAzOTYxNTM0ZDRlNjEiLCJwIjoiaiJ9) to track

contracts/multisig-prover/src/encoding/abi2/mod.rs
haiyizxx: keeping consistence with [solidity](https://github.com/axelarnetwork/axelar-gmp-sdk-solidity/blob/v5.8.0/contracts/governance/BaseWeightedMultisig.sol#L186), the function is tested in unit test. 

a bit hesitate to compute length because clippy complains for plain arithmetic operation, use wrapper type leads to unnecessary complexity

contracts/multisig-prover/src/state.rs
```diff
@@ -35,6 +40,7 @@ fn default_governance() -> Addr {
 
 pub const CONFIG: Item<Config> = Item::new("config");
 pub const COMMANDS_BATCH: Map<&BatchId, CommandBatch> = Map::new("command_batch");
```
fish-sammy, line 42: Also decrecate this.

fish-sammy (approved):

PR #369, opened 2024-04-23
chore(minor-gateway)!: switch use of aggregate-verifier for voting-verifier

eguajardo:
## Description

https://axelarnetwork.atlassian.net/browse/AXE-3666

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


contracts/voting-verifier/src/msg.rs
```diff
@@ -76,7 +76,7 @@ pub enum QueryMsg {
     #[returns(Poll)]
     GetPoll { poll_id: PollId },
 
-    #[returns(Vec<(CrossChainId, VerificationStatus)>)]
+    #[returns(Vec<(Message, VerificationStatus)>)]
```
cjcobb23, line 79: What is the reason for this change again? I know we talked about it in standup but I can't recall

eguajardo: besides what we just talked about in the standup today, I see another advantage of having this change. One of the issues I ran into when removing the aggregate-verifier was that I couldn't swap directly the aggregate-verifier client for the one in the voting-verifier because this transformation was done in one but not the other. Changing the return type directly in the API will allow us in the future to easily swap one verifier for another if we want to, or if we return to the aggregate-verifier in the future it would be easier to plug it in if all verifier clients behave the same and no mapping is needed.

contracts/voting-verifier/src/query.rs
```diff
@@ -16,12 +16,10 @@ pub fn voting_threshold(deps: Deps) -> Result<MajorityThreshold, ContractError>
...
     messages: &[Message],
-) -> Result<Vec<(CrossChainId, VerificationStatus)>, ContractError> {
+) -> Result<Vec<(Message, VerificationStatus)>, ContractError> {
     messages
         .iter()
-        .map(|message| {
-            message_status(deps, message).map(|status| (message.cc_id.to_owned(), status))
-        })
+        .map(|message| message_status(deps, message).map(|status| (message.to_owned(), status)))
         .collect::<Result<Vec<_>, _>>()
```
fish-sammy, line 23:
```suggestion
        .collect()
```

cgorenflo (approved):

RiceAndMeet (approved):

PR #370, opened 2024-04-23
refactor(ampd): use typed-builder instead of derive_builder

cgorenflo:
The typed-builder crate is compile-time safe instead of just returning an error at runtime if the build fails

### DIFF
--- a/ampd/src/broadcaster/tx.rs
+++ b/ampd/src/broadcaster/tx.rs
@@ -7,10 +7,10 @@ use cosmrs::{
     tx::{BodyBuilder, Fee, SignDoc, SignerInfo},
     Any, Coin,
 };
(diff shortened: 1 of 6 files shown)


ampd/src/broadcaster/tx.rs
```diff
@@ -25,30 +25,25 @@ pub enum Error {
...
-    pub fn zero_fee(&mut self) -> &mut Self {
-        self.fee(Fee::from_amount_and_gas(
-            Coin {
-                denom: "".parse().unwrap(),
-                amount: 0,
-            },
-            0u64,
-        ))
-    }
+fn zero_fee() -> Fee {
```
cgorenflo, line 39:
```suggestion

fn zero_fee() -> Fee {
```

PR #371, opened 2024-04-23
feat(minor-voting-verifier)!: verify msg ids

cjcobb23:
https://axelarnetwork.atlassian.net/browse/AXE-3836

deployment script change: https://github.com/axelarnetwork/axelar-contract-deployments/pull/219

### DIFF
--- a/contracts/voting-verifier/src/events.rs
+++ b/contracts/voting-verifier/src/events.rs
@@ -170,20 +170,20 @@ pub struct TxEventConfirmation {
     pub payload_hash: [u8; 32],
 }
 
-pub fn make_tx_event_confirmation(
-    other: Message,
-    msg_id_format: &MessageIdFormat,
-) -> Result<TxEventConfirmation, ContractError> {
-    let (tx_id, event_index) = parse_message_id(other.cc_id.id, msg_id_format)?;
+impl TryFrom<(Message, &MessageIdFormat)> for TxEventConfirmation {
+    type Error = ContractError;
+    fn try_from((msg, msg_id_format): (Message, &MessageIdFormat)) -> Result<Self, Self::Error> {
+        let (tx_id, event_index) = parse_message_id(msg.cc_id.id, msg_id_format)?;
 
-    Ok(TxEventConfirmation {
-        tx_id,
-        event_index,
-        destination_address: other.destination_address,
-        destination_chain: other.destination_chain,
-        source_address: other.source_address,
-        payload_hash: other.payload_hash,
-    })
+        Ok(TxEventConfirmation {
+            tx_id,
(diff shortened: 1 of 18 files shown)


contracts/voting-verifier/src/events.rs
```diff
@@ -119,9 +120,33 @@ pub struct WorkerSetConfirmation {
     pub operators: Operators,
 }
 
+/// If parsing is successful, returns (tx_id, event_index). Otherwise returns ContractError::InvalidMessageID
+fn parse_message_id(
```
fish-sammy, line 124: Why not define such a method for `MessageIdFormat`?

contracts/voting-verifier/src/events.rs
```diff
@@ -145,55 +170,20 @@ pub struct TxEventConfirmation {
...
-}
-
-pub fn construct_message_id(tx_hash: Hash, event_index: u32) -> String {
-    format!(
-        "{}{}{}",
-        HexBinary::from(tx_hash),
-        TX_HASH_EVENT_INDEX_SEPARATOR,
-        event_index
-    )
+pub fn make_tx_event_confirmation(
```
fish-sammy, line 173: Consider `impl TryFrom<(Message, &MessageIdFormat)> for TxEventConfirmation`

contracts/voting-verifier/src/events.rs
cjcobb23: this method returns tx_id and event_index. Not every message id format has those two components. They all implement FromStr though.

contracts/voting-verifier/src/events.rs
cjcobb23: I thought of that, it just felt a bit messy. But if you think that's cleaner, I can reconsider.

contracts/voting-verifier/src/events.rs
fish-sammy: Sure, but it's not like you can ignore those that don't contain those two components in this function. Rust forces you to have all the options in the match block. You'd have to put `unimplemented!()` here anyway. I feel a method on the `MessageIdFormat` is a better indicator that it won't be implemented ever.

fish-sammy (approved):

contracts/voting-verifier/src/events.rs
cjcobb23: I just don't like the idea of adding something to `MessageIdFormat` that is not generic. I think if we add formats that can't be parsed to this structure (string, uint), then we need to decide whether this voting verifier will support those formats. If it does, we change this function. If it doesn't, this function errors or panics. I'd rather address that if/when the time comes.

PR #377, opened 2024-05-03
fix(minor-voting-verifier): remove string escaping from poll events

cgorenflo:


cgorenflo:
this makes deserializing events more difficult

PR #380, opened 2024-05-06
fix(service-registry): deregister chains support updates CHAINS_PER_WORKER

maancham:
## Description

AXE-3920

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


contracts/service-registry/src/contract/execute.rs
```diff
@@ -156,9 +156,8 @@ pub fn deregister_chains_support(
         .may_load(deps.storage, (&service_name, &info.sender))?
         .ok_or(ContractError::WorkerNotFound)?;
 
-    for chain in chains {
-        WORKERS_PER_CHAIN.remove(deps.storage, (&service_name, &chain, &info.sender));
-    }
+    let _res =
+        state::deregister_chains_support(deps.storage, service_name.clone(), chains, info.sender);
```
cjcobb23, line 160:
```suggestion
        state::deregister_chains_support(deps.storage, service_name.clone(), chains, info.sender)?;
```
We shouldn't ignore results.

contracts/service-registry/src/state.rs
```diff
@@ -184,6 +184,27 @@ pub fn register_chains_support(
...
+) -> Result<(), ContractError> {
+    CHAINS_PER_WORKER.update(storage, (&service_name, &worker), |current_chains| {
+        let mut current_chains = current_chains.unwrap_or_default();
+        for chain in chains.iter() {
+            current_chains.remove(chain);
+        }
+        Ok::<HashSet<ChainName>, ContractError>(current_chains)
+    })?;
+
+    for chain in chains.iter() {
```
cjcobb23, line 201:
```suggestion
    for chain in chains {
```

contracts/service-registry/src/state.rs
```diff
@@ -184,6 +184,27 @@ pub fn register_chains_support(
...
+    service_name: String,
+    chains: Vec<ChainName>,
+    worker: WorkerAddress,
+) -> Result<(), ContractError> {
+    CHAINS_PER_WORKER.update(storage, (&service_name, &worker), |current_chains| {
+        let mut current_chains = current_chains.unwrap_or_default();
+        for chain in chains.iter() {
+            current_chains.remove(chain);
+        }
+        Ok::<HashSet<ChainName>, ContractError>(current_chains)
```
cjcobb23, line 198:
```suggestion
        Ok::<HashSet<ChainName>, ContractError>(
            current_chains
                .unwrap_or_default()
                .into_iter()
                .filter(|chain| !chains.contains(chain))
                .collect(),
        )
```

contracts/service-registry/src/state.rs
```diff
@@ -184,6 +184,27 @@ pub fn register_chains_support(
     Ok(())
 }
 
+pub fn deregister_chains_support(
+    storage: &mut dyn Storage,
+    service_name: String,
+    chains: Vec<ChainName>,
+    worker: WorkerAddress,
+) -> Result<(), ContractError> {
+    CHAINS_PER_WORKER.update(storage, (&service_name, &worker), |current_chains| {
```
cjcobb23, line 193: This function allows workers to deregister for chains that they are not registered for. Is that intended? One side effect is that if a worker is not registered for any chains, this function is going to write an empty hashset into CHAINS_PER_WORKER

maancham: yeah it would not be an error.

PR #381, opened 2024-05-06
refactor(multisig-prover): remove mock contracts from prover tests

eguajardo:
## Description

https://axelarnetwork.atlassian.net/browse/AXE-3783

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


contracts/multisig-prover/src/contract.rs
```diff
@@ -206,36 +235,57 @@ mod tests {
...
+
+    fn reply_construct_proof(deps: DepsMut) -> Result<Response, axelar_wasm_std::ContractError> {
+        let session_id = to_binary(&MULTISIG_SESSION_ID).unwrap();
+
+        let response = SubMsgResponse {
+            events: vec![],
+            // see https://github.com/CosmWasm/cw-minus/blob/11164624d1bbb8df8a89e5142a03e106e751dfc4/packages/cw-utils/src/parse_reply.rs#L245
+            data: Some(
+                [
+                    b"\x0a".as_ref(),
```
cgorenflo, line 249: can you explain this? why is the prefix 10?

eguajardo: while parsing there are some weird requirements here: https://github.com/CosmWasm/cw-minus/blob/main/packages/cw-utils/src/parse_reply.rs#L61-L75, but to be honest I don't know why it's designed that way

cgorenflo: I would prefer
```
prost::bytes::Bytes::from(session_id.as_ref()).encode_to_vec().into()
```

contracts/multisig-prover/src/contract.rs
```diff
@@ -206,36 +236,53 @@ mod tests {
...
-            &[],
+        execute(deps, mock_env(), mock_info(RELAYER, &[]), msg)
+    }
+
+    fn reply_construct_proof(deps: DepsMut) -> Result<Response, axelar_wasm_std::ContractError> {
+        let session_id = to_binary(&MULTISIG_SESSION_ID).unwrap();
+
+        let response = SubMsgResponse {
+            events: vec![],
+            // see https://github.com/CosmWasm/cw-minus/blob/11164624d1bbb8df8a89e5142a03e106e751dfc4/packages/cw-utils/src/parse_reply.rs#L245
```
cgorenflo, line 247:
```suggestion
            // the reply data gets protobuf encoded when moving through the wasm module. We need to emulate this behaviour in tests as well
```

cgorenflo (approved):

PR #390, opened 2024-05-08
refactor(ampd): move the broadcaster logic from the handlers to the event processor

fish-sammy:
## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


ampd/src/queue/queued_broadcaster.rs
```diff
@@ -55,15 +51,9 @@ pub struct QueuedBroadcasterClient {
 
 #[async_trait]
 impl BroadcasterClient for QueuedBroadcasterClient {
-    async fn broadcast<T>(&self, tx: T) -> Result
-    where
-        T: Msg + Send + Sync + 'static,
-    {
+    async fn broadcast(&self, msg: Any) -> Result {
```
cgorenflo, line 54: why did you decide to accept `Any` instead of `Msg`?

cgorenflo (approved):

cgorenflo:
seems like there is quite a bit of untested code in this PR, could you investigate that?

ampd/src/queue/queued_broadcaster.rs
fish-sammy: I don't think that the handlers should limited to only being able to return one type of msgs, and returning a `Vec<Box<dyn Msg>>` seems a bit insane to me. Returning `Vec<Any>` would also work better once we've transformed the handlers to run on their own and returning msgs via gPRC.

PR #394, opened 2024-05-10
refactor(ampd): change the event_sub so that subscribers can be added at runtime

fish-sammy:
## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

### DIFF
--- a/ampd/src/event_sub.rs
+++ b/ampd/src/event_sub.rs
@@ -1,11 +1,9 @@
-use std::convert::TryInto;
 use std::iter;
 use std::time::Duration;
 
 use error_stack::ResultExt;
 use error_stack::{FutureExt, Report, Result};
 use futures::TryStreamExt;
-use tendermint::abci;
 use tendermint::block;
 use thiserror::Error;
 use tokio::select;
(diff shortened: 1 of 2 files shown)


ampd/src/event_sub.rs
```diff
@@ -112,32 +118,31 @@ impl<T: TmClient + Sync> EventPublisher<T> {
...
-                    .collect::<Result<Vec<_>, _>>()
-                    .change_context(EventSubError::Publish)?,
-            )
+            .chain(self.events(height).await?.into_iter())
             .chain(iter::once(Event::BlockEnd(height)));
 
         for event in events {
-            self.tx.send(event).change_context(EventSubError::Publish)?;
+            // ignore the error if there are no subscribers
+            let _ = self.tx.send(event);
```
haiyizxx, line 132: it skips processing if no subscribers earlier, why suppress the error here?

fish-sammy: The previous check prevents the block from being downloaded if no subscriber exists. However, subscribers can still drop while the publisher is in this loop broadcasting events.

haiyizxx (approved):

PR #395, opened 2024-05-10
feat(minor-multisig-prover)!: update migration handler to migrate domain_separator

eguajardo:
## Description

https://axelarnetwork.atlassian.net/browse/AXE-3937


contracts/multisig-prover/src/migrate.rs
```diff
@@ -0,0 +1,68 @@
...
+    domain_separator: Hash,
+) -> Result<Response, axelar_wasm_std::ContractError> {
+    let old_config = CONFIG.load(deps.storage)?;
+    let new_config = Config {
+        domain_separator,
+        ..old_config
+    };
+    CONFIG.save(deps.storage, &new_config)?;
+
+    Ok(Response::default())
```
cjcobb23, line 17: Do our existing multisig prover instances have a domain separator? If not, I think this will fail, because it won't be able to deserialize the config.

haiyizxx: ya, need to provide the old config struct to deserialize

eguajardo: fixed, I manually created the old struct and migrated instead of something like default deserialization of the missing field. I feel the default deserialization could lead to unintended side effects

contracts/multisig-prover/src/migrate.rs
```diff
@@ -0,0 +1,143 @@
...
+    pub service_registry: Addr,
+    pub voting_verifier: Addr,
+    pub destination_chain_id: Uint256,
+    pub signing_threshold: MajorityThreshold,
+    pub service_name: String,
+    pub chain_name: ChainName,
+    pub worker_set_diff_threshold: u32,
+    pub encoder: Encoder,
+    pub key_type: KeyType,
+}
```
cjcobb23, line 28: We should probably put this in some sort of folder or namespace or something to denote this struct represents the config from version 0.2. Or just call it Config0.2 or something. Otherwise what will we do when we need to add a new field to the config and perform a similar migration again?

contracts/multisig-prover/src/migrate.rs
```diff
@@ -0,0 +1,143 @@
...
+            coordinator: self.coordinator,
+            service_registry: self.service_registry,
+            voting_verifier: self.voting_verifier,
+            destination_chain_id: self.destination_chain_id,
+            signing_threshold: self.signing_threshold,
+            service_name: self.service_name,
+            chain_name: self.chain_name,
+            worker_set_diff_threshold: self.worker_set_diff_threshold,
+            encoder: self.encoder,
+            key_type: self.key_type,
```
cjcobb23, line 47: Can't you just write `..self` for all of this?

cjcobb23 (approved):
approved but see my comment about the `OldConfig` struct

contracts/multisig-prover/src/migrate.rs
eguajardo: only works for exactly the same struct

contracts/multisig-prover/src/migrate.rs
eguajardo: instead of renaming I moved the migration to migrate::v_0_3, let me know what you think

contracts/multisig-prover/src/lib.rs
```diff
@@ -3,6 +3,7 @@ pub mod encoding;
 pub mod error;
 pub mod events;
 mod execute;
+mod migrate;
```
haiyizxx, line 6: prefer `migrations`

haiyizxx (approved):

contracts/multisig-prover/src/migrate.rs
cjcobb23: ah ok

cjcobb23 (approved):

PR #400, opened 2024-05-13
feat(minor-multisig-prover): use big endian to encode nonce

haiyizxx:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


contracts/multisig-prover/src/encoding/abi/mod.rs
```diff
@@ -63,7 +63,7 @@ impl From<&WorkerSet> for WeightedSigners {
         WeightedSigners {
             signers,
             threshold,
-            nonce: Uint256::from(worker_set.created_at).to_le_bytes().into(),
+            nonce: Uint256::from(worker_set.created_at).to_be_bytes().into(),
```
milapsheth, line 66: can you test coverage for this so changes are caught?

milapsheth (approved):

contracts/multisig-prover/src/encoding/abi/mod.rs
haiyizxx: tracked [here](https://axelarnetwork.atlassian.net/browse/AXE-3994)

PR #401, opened 2024-05-13
fix(voting-verifier): migrate missing field msg_id_format

eguajardo:
## Description

https://axelarnetwork.atlassian.net/browse/AXE-3995


contracts/voting-verifier/src/migrations/v_0_3.rs
```diff
@@ -0,0 +1,114 @@
+use axelar_wasm_std::{msg_id::MessageIdFormat, nonempty, FnExt, MajorityThreshold};
```
haiyizxx, line 1: add a comment indicates what has been migrated

haiyizxx (approved):

PR #411, opened 2024-05-22
chore: upgrade to cosmwasm-std 1.5.5

cgorenflo:
AXE-4031

### DIFF
--- a/contracts/multisig/src/contract.rs
+++ b/contracts/multisig/src/contract.rs
@@ -369,11 +369,11 @@ mod tests {
 
         let res = query_worker_set(&worker_set_1.id(), deps.as_ref());
         assert!(res.is_ok());
-        assert_eq!(worker_set_1, from_json(&res.unwrap()).unwrap());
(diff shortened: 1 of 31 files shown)


contracts/multisig/src/contract.rs
```diff
@@ -457,11 +459,10 @@ mod tests {
...
 
-            assert_eq!(
-                res.unwrap_err().to_string(),
-                axelar_wasm_std::ContractError::from(ContractError::Unauthorized).to_string()
-                    + ": () not found"
-            );
+            assert!(res
+                .unwrap_err()
+                .to_string()
+                .contains(&ContractError::Unauthorized.to_string()));
```
cgorenflo, line 465: cw-storage-plus returns a different error, so this broke

contracts/multisig/src/contract.rs
```diff
@@ -991,11 +992,10 @@ mod tests {
...
```
cgorenflo, line 998: cw-storage-plus returns a different error, so this broke

milapsheth (approved):

PR #413, opened 2024-05-22
refactor(ampd): remove unused Uint256

cgorenflo:
after the file only contained TMAddress it was renamed to better reflect that

### DIFF
--- a/ampd/src/tm_address.rs
+++ b/ampd/src/tm_address.rs
@@ -3,31 +3,12 @@ use std::hash::{Hash as StdHash, Hasher};
 
 use cosmrs::crypto;
 use cosmrs::AccountId;
-use cosmwasm_std::Uint256;
 use ethers::types::{Address, H256};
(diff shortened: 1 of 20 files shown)


ampd/src/tm_address.rs
```diff
@@ -3,32 +3,13 @@ use std::hash::{Hash as StdHash, Hasher};
 
 use cosmrs::crypto;
 use cosmrs::AccountId;
-use cosmwasm_std::Uint256;
 use ethers::types::{Address, H256};
 use serde::{Deserialize, Serialize};
 
 pub type EVMAddress = Address;
```
fish-sammy, line 9: But there are things like `EVMAddress` defined in this mod. Why call it `tm_address`?

cgorenflo: ok I missed that 🤦 
I'll think of a new name

fish-sammy (approved):

PR #415, opened 2024-05-23
feat(ampd): protobuf definition for the gRPC service between ampd and ampd plugins

fish-sammy:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

### DIFF
--- a/ampd/proto/ampd.proto
+++ b/ampd/proto/ampd.proto
@@ -1,0 +1,37 @@
+syntax = "proto3";
+
+import "google/protobuf/any.proto";
+
+package ampd;
+
+message SubscribeRequest {
+  string event_type = 1;
+  map<string, string> event_attributes = 2;
+  bool include_block_begin_end = 3;
+}
+
+message EventBlockBegin { uint64 height = 1; }
+
+message EventBlockEnd { uint64 height = 1; }
+
+message Event {
+  string type = 1;
+  map<string, string> attributes = 2;
+}
+
+message SubscribeResponse {
+  oneof event {
+    EventBlockBegin block_begin = 1;
+    EventBlockEnd block_end = 2;
+    Event abci = 3;
+  }
+}
+
+message BroadcastRequest { google.protobuf.Any msg = 1; }
+
+message BroadcastResponse {}
+
+service Ampd {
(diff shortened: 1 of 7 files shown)


ampd/proto/ampd.proto
```diff
@@ -0,0 +1,37 @@
+syntax = "proto3";
+
+import "google/protobuf/any.proto";
+
+package ampd;
+
+message SubscribeRequest {
+  string event_type = 1;
+  map<string, string> event_attributes = 2;
```
cgorenflo, line 9:
```suggestion
  repeated Event events_filter = 1;
```

ampd/src/grpc/ampd.rs
```diff
@@ -0,0 +1,37 @@
...
+
+impl Ampd {
+    pub fn new() -> Self {
+        Self {}
+    }
+}
+
+#[async_trait]
+impl proto::ampd_server::Ampd for Ampd {
+    type SubscribeStream =
```
cgorenflo, line 21:
```suggestion
    type SubscriptionStream =
```

ampd/proto/ampd.proto
fish-sammy: Interesting thought, but we are not going to do a strict comparison though. The filter may only provide a subset of the matching event.

ampd/src/grpc/ampd.rs
fish-sammy: The trait is auto-generated code that I cannot change without changing the service method to `Subscription`.

cgorenflo: ah

ampd/proto/ampd.proto
```diff
@@ -0,0 +1,38 @@
+syntax = "proto3";
+
+import "google/protobuf/any.proto";
+
+package ampd;
+
+message EventFilter {
+  repeated Event filters = 1;
```
cgorenflo, line 8:
```suggestion
  // the subscription will return all events that match ANY of these filters
  repeated Event filters = 1;
```

cgorenflo (approved):

PR #420, opened 2024-05-23
feat(minor-multisig)!: migrate worker set to verifier set

cjcobb23:
migration for multisig and multisig prover to change worker set to verifier set


contracts/multisig-prover/src/contract.rs
```diff
@@ -126,12 +127,21 @@ pub fn query(
...
 #[cfg_attr(not(feature = "library"), entry_point)]
 pub fn migrate(
-    _deps: DepsMut,
+    deps: DepsMut,
     _env: Env,
     _msg: Empty,
 ) -> Result<Response, axelar_wasm_std::ContractError> {
-    // TODO migrate
-    Ok(Response::default())
+    // any version checks should be done before here
```
cgorenflo, line 134: what does this mean, what's missing?

contracts/multisig/src/contract.rs
```diff
@@ -23,11 +25,22 @@ const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
 
 #[cfg_attr(not(feature = "library"), entry_point)]
 pub fn migrate(
-    _deps: DepsMut,
+    deps: DepsMut,
     _env: Env,
     _msg: Empty,
 ) -> Result<Response, axelar_wasm_std::ContractError> {
-    // TODO migrate
+    // any version checks should be done before here
```
cgorenflo, line 32: same question here

cgorenflo (approved):

contracts/multisig-prover/src/contract.rs
cjcobb23: this is an old comment, I can remove

contracts/multisig/src/contract.rs
cjcobb23: old comment, will remove

PR #425, opened 2024-05-24
refactor(minor-voting-verifier): remove operator struct

cjcobb23:


packages/evm-gateway/src/lib.rs
```diff
@@ -22,38 +23,37 @@ abigen!(
 pub enum Error {
     #[error("address is invalid: {reason}")]
     InvalidAddress { reason: String },
+    #[error("public key is invalid: {reason}")]
+    InvalidPublicKey { reason: String },
```
cgorenflo, line 27: please don't add more errors with reasons. Use `change_context` instead

cgorenflo (approved):

contracts/voting-verifier/src/execute.rs
```diff
@@ -63,16 +62,16 @@ pub fn verify_verifier_set(
 
     POLL_VERIFIER_SETS.save(
         deps.storage,
-        &new_operators.hash(),
-        &PollContent::<Operators>::new(new_operators.clone(), poll_id),
+        &new_verifier_set.hash().as_slice().try_into().unwrap(),
```
milapsheth, line 65: we wanted to change the hash implementation since it's just serializing to JSON rn and hashing that. This would be a breaking change later

contracts/voting-verifier/src/state.rs
```diff
@@ -65,10 +65,10 @@ impl PollContent<Message> {
     }
 }
 
-impl PollContent<Operators> {
-    pub fn new(operators: Operators, poll_id: PollId) -> Self {
+impl PollContent<VerifierSet> {
+    pub fn new(workerset: VerifierSet, poll_id: PollId) -> Self {
```
milapsheth, line 69: needs to be `verifier_set`

contracts/voting-verifier/src/state.rs
```diff
@@ -65,10 +65,10 @@ impl PollContent<Message> {
...
         Self {
-            content: operators,
+            content: workerset,
```
milapsheth, line 71: same

PR #426, opened 2024-05-24
feat(ampd): implement the gRPC services

fish-sammy:
## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


ampd/src/grpc/ampd.rs
```diff
@@ -1,37 +1,402 @@
...
     async fn broadcast(
         &self,
-        _req: Request<proto::BroadcastRequest>,
+        req: Request<proto::BroadcastRequest>,
     ) -> Result<Response<proto::BroadcastResponse>, Status> {
-        todo!()
+        let req = req.into_inner();
+        let msg = req.msg.ok_or(Status::invalid_argument("missing msg"))?;
+
+        // TODO: validate msg by estimating its gas cost before broadcasting?
```
cgorenflo, line 118: is the broadcaster not doing that?

ampd/src/grpc/ampd.rs
```diff
@@ -1,37 +1,402 @@
...
+                event_attributes: attributes
+                    .into_iter()
+                    .map(|(key, value)| (key, value.to_string()))
+                    .collect(),
+            }),
+        }
+    }
+}
+
+impl proto::SubscribeRequest {
```
cgorenflo, line 37: flatten SubscribeRequest to make working with it here simpler

cgorenflo (approved):

ampd/src/grpc/ampd.rs
fish-sammy: It does, but not at the point when `BroadcasterClient` pushes it through. Esstially, the `BroadcasterClient` wouldn't get an error for that even if it happens.

PR #439, opened 2024-05-29
feat(minor-rewards): add query for verifier participation

cjcobb23:
https://axelarnetwork.atlassian.net/browse/AXE-4100


contracts/rewards/src/contract.rs
```diff
@@ -157,6 +157,12 @@ pub fn query(
                 .change_context(ContractError::SerializeResponse)
                 .map_err(axelar_wasm_std::ContractError::from)
         }
+        QueryMsg::VerifierParticipation { pool_id, epoch_num } => {
```
fish-sammy, line 160: How does the querier get to know the `epoch_num`? I thought it's internal and an implementation detail, no?

cjcobb23: It's optional, so normally won't be specified and will just default to current. But the query itself returns the current epoch, so you could query prior epochs if you want.

contracts/rewards/src/contract/query.rs
```diff
@@ -35,9 +35,37 @@ pub fn rewards_pool(
...
+            let current_params = state::load_params(storage);
+            Epoch::current(&current_params, block_height)?.epoch_num
+        }
+    };
+
+    let tally = state::load_epoch_tally(storage, pool_id, epoch_num)?;
+
+    match tally {
+        None => Ok(None),
+        Some(tally) => Ok(Some(msg::Participation {
```
milapsheth, line 56: Looks good to me. I would still not recommend clients to compute actual rewards rate themselves from this response, since it's easy to introduce discrepancies between implementations (e.g. how thresholds are compared within the contract vs if floats are used in JS, missing error checks etc) and chances of client code becoming outdated.

milapsheth: Returning the rewards by verifier map would be nice to have as well (or a separate query for it) but we can get by with the participation rates for now to measure performance. Maybe we introduce a separate REST API layer later that avoids adding more query code to the contract itself.

contracts/rewards/src/msg.rs
```diff
@@ -65,6 +67,13 @@ pub enum QueryMsg {
     /// Gets the rewards pool details for the given `pool_id``
```
milapsheth, line 67:
```suggestion
    /// Gets the rewards pool details for the given `pool_id`
```
misc observation

milapsheth (approved):

PR #440, opened 2024-05-30
feat(minor-voting-verifier)!: emit event when quorum is reached

cjcobb23:
Will handle migration in a separate PR

https://axelarnetwork.atlassian.net/browse/AXE-3989


contracts/voting-verifier/src/execute.rs
```diff
@@ -171,21 +230,40 @@ pub fn vote(
...
-        Voted {
-            poll_id,
-            voter: info.sender,
-        }
-        .into(),
-    ))
+    let final_results = get_indices_with_quorum(&poll);
+
+    // we only want to emit a quorum reached event when quorum is first reached, and not subsequently
+    let reached_quorum: Vec<&(u32, Vote)> = final_results
```
fish-sammy, line 246: I am so confused looking at these variable names. `initial_results` and `final_results` coming from a function called `get_indices_with_quorum`, and then `reached_quorum` being a `Vec` with its name sounds liked a boolean.

contracts/voting-verifier/src/execute.rs
```diff
@@ -171,21 +230,40 @@ pub fn vote(
...
+        .iter()
+        .filter(|e| !initial_results.contains(e))
+        .collect();
+
+    let events = reached_quorum
```
fish-sammy, line 251: So `events` would be empty if this is not the first time when the quorum has been reached?

fish-sammy (changes requested):

contracts/voting-verifier/src/execute.rs
cjcobb23: yes

fish-sammy: I don't understand how it's achieved

contracts/voting-verifier/src/execute.rs
```diff
@@ -162,6 +165,62 @@ pub fn verify_messages(
     ))
 }
 
+/// Returns indices of items in the poll that have achieved quorum, along with their respective results
```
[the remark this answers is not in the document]
cjcobb23: On line 175, we get the results of the poll. This returns `Option<Vote>` for each item in the poll, where the value is Some for items that achieved quorum, and None otherwise.

contracts/voting-verifier/src/execute.rs
cjcobb23: yeah I really struggled with finding good names for these variables. Do you have any suggestions?

contracts/voting-verifier/src/execute.rs
cjcobb23: What do you mean? initial_results are the messages that have reached quorum before tallying this vote, final_results are the messages that have reached quorum after tallying this vote, and then reached_quorum is everything that is in final_results but not in initial_results, so that would be the messages for which this particular vote caused them to achieve quorum.

fish-sammy: How can there be anything in the initial_results that has reached quorum when we are trying to find out if this is the very first time quorum has been reached?

contracts/voting-verifier/src/execute.rs
fish-sammy: I need to understand what's going on here before I am able to suggest anything.

contracts/voting-verifier/src/execute.rs
```diff
@@ -161,7 +164,78 @@ pub fn verify_messages(
...
     ))
 }
+struct PollResults(Vec<Option<Vote>>);
+
+// would be better to implement the Sub trait, but clippy is configured to not allow arithmetic operators
+impl PollResults {
+    /// Returns the elements in self that are Some, but in rhs are None. All other elements are converted to None.
+    /// This is used to determine which elements have quorum in self, but do not have quorum in rhs.
+    /// Vectors must be equal length.
+    fn sub(self, rhs: Self) -> Result<PollResults, ContractError> {
```
cgorenflo, line 174:
```suggestion
    fn difference(self, rhs: Self) -> Result<PollResults, ContractError> {
```
see https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.difference for example

contracts/voting-verifier/src/execute.rs
```diff
@@ -161,7 +164,78 @@ pub fn verify_messages(
...
+                    if lhs.is_some() && rhs.is_none() {
+                        Some(lhs)
+                    } else {
+                        None
+                    }
+                })
+                .collect(),
+        ))
+    }
+}
```
fish-sammy, line 192: I am not sure if these should be defined in execute.rs.

fish-sammy (approved):

contracts/voting-verifier/src/execute.rs
```diff
@@ -171,21 +245,38 @@ pub fn vote(
...
     let poll = POLLS
         .may_load(deps.storage, poll_id)?
-        .ok_or(ContractError::PollNotFound)?
-        .try_map(|poll| {
-            poll.cast_vote(env.block.height, &info.sender, votes)
-                .map_err(ContractError::from)
-        })?;
+        .ok_or(ContractError::PollNotFound)?;
 
+    let before_results = get_poll_results(&poll);
```
cgorenflo, line 250:
```suggestion
    let results_before_voting = get_poll_results(&poll);
```

contracts/voting-verifier/src/execute.rs
```diff
@@ -171,21 +245,38 @@ pub fn vote(
...
     POLLS.save(deps.storage, poll_id, &poll)?;
 
-    Ok(Response::new().add_event(
-        Voted {
-            poll_id,
-            voter: info.sender,
-        }
-        .into(),
-    ))
+    let after_results = get_poll_results(&poll);
```
cgorenflo, line 258:
```suggestion
    let results_after_voting = get_poll_results(&poll);
```

PR #448, opened 2024-06-05
refactor(minor-coordinator)!: single map for active and next verifier sets

maancham:
## Description

AXE-4157

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


contracts/coordinator/src/msg.rs
```diff
@@ -15,20 +15,14 @@ pub enum ExecuteMsg {
         chain_name: ChainName,
         new_prover_addr: Addr,
     },
-    SetActiveVerifiers {
-        next_verifier_set: VerifierSet,
-    },
-    SetNextVerifiers {
-        next_verifier_set: VerifierSet,
+    UpdateVerifierUnionSet {
+        union_set: HashSet<Addr>,
```
cjcobb23, line 19:
```suggestion
    SetVerifiers {
        verifiers: HashSet<Addr>,
```

The fact that the set is a union of two sets is an implementation detail that may not always be true

contracts/coordinator/src/state.rs
```diff
@@ -12,17 +12,72 @@ pub struct Config {
...
+) -> Result<(), ContractError> {
+    let existing_verifiers = VERIFIER_PROVER_INDEXED_MAP
+        .prefix(prover_address.clone())
+        .keys(storage, None, None, Order::Ascending)
+        .filter_map(Result::ok)
+        .collect::<HashSet<VerifierAddress>>();
+
+    for verifier in existing_verifiers.difference(&new_union_set) {
+        let _result =
+            VERIFIER_PROVER_INDEXED_MAP.remove(storage, (prover_address.clone(), verifier.clone()));
```
cjcobb23, line 68: the result shouldn't be ignored

contracts/coordinator/src/state.rs
```diff
@@ -12,17 +12,72 @@ pub struct Config {
...
-type ChainNames = HashSet<ChainName>;
-type VerifierAddress = Addr;
+pub type ChainNames = HashSet<ChainName>;
+pub type VerifierAddress = Addr;
 pub const CHAINS_OF_VERIFIER: Map<VerifierAddress, ChainNames> = Map::new("chains_of_verifier");
 
-pub const NEXT_VERIFIER_SET_FOR_PROVER: Map<ProverAddress, VerifierSet> =
-    Map::new("next_prover_verifier");
+pub struct VerifierSetIndex<'a> {
+    pub by_verifier: MultiIndex<'a, Addr, VerifierProverRecord, (Addr, Addr)>,
```
cjcobb23, line 24: where is this index actually used? Every place I saw the code is just reading/writing to the map directly, or iterating based on prefix. Is the index used?

cjcobb23 (changes requested):

contracts/coordinator/src/state.rs
```diff
@@ -12,17 +12,79 @@ pub struct Config {
...
+    for verifier in existing_verifiers.difference(&new_verifiers) {
+        if VERIFIER_PROVER_INDEXED_MAP
+            .remove(storage, (prover_address.clone(), verifier.clone()))
+            .is_err()
+        {
+            return Err(ContractError::NoSuchVerifierRegisteredForProver(
+                verifier.clone(),
+                prover_address,
+            ));
+        }
```
cgorenflo, line 75: that looks like it should be possible to solve with `change_context`/`map_err` and `?`

contracts/multisig-prover/src/execute.rs
```diff
@@ -300,13 +304,37 @@ pub fn confirm_verifier_set(deps: DepsMut, sender: Addr) -> Result<Response, Con
...
-            &coordinator::msg::ExecuteMsg::SetActiveVerifiers {
-                next_verifier_set: verifier_set,
+            &coordinator::msg::ExecuteMsg::SetVerifiers {
+                verifiers: verifier_union_set,
             },
             vec![],
         )?))
 }
 
+fn compute_union_of_verifier_sets(deps: &DepsMut) -> Result<HashSet<Addr>, ContractError> {
```
cgorenflo, line 314:
```suggestion
fn all_active_verifiers(deps: &DepsMut) -> Result<HashSet<Addr>, ContractError> {
```

contracts/multisig-prover/src/execute.rs
```diff
@@ -300,13 +304,37 @@ pub fn confirm_verifier_set(deps: DepsMut, sender: Addr) -> Result<Response, Con
...
+    }
+
+    if let Some(next_verifier_set) = NEXT_VERIFIER_SET.may_load(deps.storage)? {
+        union_set.extend(
+            next_verifier_set
+                .signers
+                .values()
+                .map(|signer| signer.address.clone()),
+        );
+    }
```
cgorenflo, line 333: do something like this (pseudo code):
```
let verifiers = CURRENT_VERIFIER_SET.may_load(deps.storage)?.then(to_verifiers)
   .concatenate(NEXT_VERIFIER_SET.may_load(deps.storage)?.then(to_verifiers))
Ok(HashSet::from_iter(verifiers)
```

integration-tests/tests/test_utils/mod.rs
```diff
@@ -723,7 +710,7 @@ pub struct Chain {
 pub fn setup_chain(
     protocol: &mut Protocol,
     chain_name: ChainName,
-    verifiers: &Vec<Verifier>,
+    verifiers: &[Verifier],
```
cgorenflo, line 713: nice

contracts/coordinator/src/query.rs
```diff
@@ -12,47 +12,29 @@ pub fn prover(deps: Deps, chain_name: ChainName) -> Result<Addr, ContractError>
...
-                    return Ok(true);
-                }
-            }
-        }
-    }
+    let verifier_present_in_sets = match verifier_entries {
+        Ok(records) => records
+            .iter()
+            .map(|((prover_address, _), _)| prover_address)
+            .any(|prover_address| !prover_address.as_str().is_empty()),
```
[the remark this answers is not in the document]
maancham: Yeah, it makes much more sense. We are doing an extra check here. Will change.

contracts/coordinator/src/state.rs
```diff
@@ -12,17 +12,79 @@ pub struct Config {
...
+        let v: Vec<&dyn Index<VerifierProverRecord>> = vec![&self.by_verifier];
+        Box::new(v.into_iter())
+    }
+}
+
+#[cw_serde]
+pub struct VerifierProverRecord {
+    pub prover: ProverAddress,
+    pub verifier: VerifierAddress,
+}
```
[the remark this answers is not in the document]
maancham: I'm using it to grab the `verifier` field as index function for the IndexedMap.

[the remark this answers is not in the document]
cjcobb23: I also think we don't need this

contracts/coordinator/src/contract.rs
```diff
@@ -57,11 +57,8 @@ pub fn execute(
             execute::check_governance(&deps, info)?;
             execute::register_prover(deps, chain_name, new_prover_addr)
         }
-        ExecuteMsg::SetActiveVerifiers { next_verifier_set } => {
-            execute::set_active_verifier_set(deps, info, next_verifier_set)
-        }
-        ExecuteMsg::SetNextVerifiers { next_verifier_set } => {
-            execute::set_next_verifier_set(deps, info, next_verifier_set)
+        ExecuteMsg::SetVerifiers { verifiers } => {
```
cgorenflo, line 60: This should be more descriptive, just looking at the interface I would have no idea what this is for. I'm fine still calling it `SetActiveVerifiers` (ask the rest of the team about it) or maybe `SetVerifiersInRotation`?

contracts/coordinator/src/query.rs
```diff
@@ -12,49 +9,24 @@ pub fn prover(deps: Deps, chain_name: ChainName) -> Result<Addr, ContractError>
...
-                if verifier_set.includes(verifier_address) {
-                    return Ok(true);
-                }
-            }
-        }
-    }
+fn is_verifier_in_any_verifier_set(
+    deps: Deps,
+    verifier_address: &VerifierAddress,
+) -> StdResult<bool> {
```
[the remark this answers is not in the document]
maancham: yes?

contracts/coordinator/src/state.rs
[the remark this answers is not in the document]
cgorenflo: it is needed for `IndexedMap` to work correctly

contracts/coordinator/src/contract.rs
cjcobb23: `SetActiveVerifiers` is fine with me

contracts/coordinator/src/state.rs
```diff
@@ -12,17 +12,78 @@ pub struct Config {
...
+    let existing_verifiers = VERIFIER_PROVER_INDEXED_MAP
+        .prefix(prover_address.clone())
+        .keys(storage, None, None, Order::Ascending)
+        .filter_map(Result::ok)
+        .collect::<HashSet<VerifierAddress>>();
+
+    for verifier in existing_verifiers.difference(&new_verifiers) {
+        VERIFIER_PROVER_INDEXED_MAP
+            .remove(storage, (prover_address.clone(), verifier.clone()))
+            .map_err(|_| {
```
cjcobb23, line 69: I think this error should just be an expect, since this should never happen.

contracts/coordinator/src/execute.rs
```diff
@@ -25,20 +23,11 @@ pub fn register_prover(
     Ok(Response::new())
 }
 
-pub fn set_active_verifier_set(
+pub fn update_prover_union_set(
```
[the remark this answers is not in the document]
cjcobb23:
```suggestion
pub fn set_active_verifier_set(
```

contracts/coordinator/src/query.rs
```diff
@@ -12,49 +9,24 @@ pub fn prover(deps: Deps, chain_name: ChainName) -> Result<Addr, ContractError>
...
+    let verifier_present_in_sets = VERIFIER_PROVER_INDEXED_MAP
+        .idx
+        .by_verifier
+        .prefix(verifier_address.clone())
+        .range(deps.storage, None, None, Order::Ascending)
+        .next()
+        .is_some();
 
-    Ok(false)
+    Ok(verifier_present_in_sets)
```
cjcobb23, line 24: Do we need ok here? Can this function just return bool? I don't see an error being returned anywhere, or `?` being used.

cjcobb23 (approved):

contracts/coordinator/src/error.rs
```diff
@@ -13,4 +14,7 @@ pub enum ContractError {
 
     #[error("no provers registered for chain {0}")]
     NoProversRegisteredForChain(ChainName),
+
+    #[error("verifier {0} not found for prover {1}")]
+    NoSuchVerifierRegisteredForProver(VerifierAddress, ProverAddress),
```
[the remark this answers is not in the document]
maancham: I will remove it.

contracts/coordinator/src/query.rs
maancham: I changed the setting.

contracts/coordinator/src/query.rs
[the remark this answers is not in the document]
maancham: Added `.any` instead.

PR #450, opened 2024-06-07
fix(minor-multisig-prover): filter out signers without pubkey

haiyizxx:
## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


contracts/multisig-prover/src/execute.rs
```diff
@@ -121,31 +123,50 @@ fn get_verifiers_info(deps: &DepsMut, config: &Config) -> Result<VerifiersInfo,
...
-    let participants = verifiers
-        .clone()
+    let min_num_verifiers = deps
+        .querier
+        .query::<Service>(&QueryRequest::Wasm(WasmQuery::Smart {
+            contract_addr: config.service_registry.to_string(),
+            msg: to_json_binary(&service_registry::msg::QueryMsg::GetService {
+                service_name: config.service_name.clone(),
+            })?,
+        }))?
```
cgorenflo, line 133: do we not have a client for the service_registry contract yet?

contracts/multisig-prover/src/execute.rs
cgorenflo, line 1: the `get_verifiers_info` function should be renamed, it's not just a simple query, it makes oppinonated checks and creates snapshots

cgorenflo: make `collect_` or `create_`

contracts/multisig-prover/src/execute.rs
```diff
@@ -129,31 +132,50 @@ fn get_verifiers_info(deps: &DepsMut, config: &Config) -> Result<VerifiersInfo,
...
+            .map(|(participant, _)| participant.clone())
+            .collect::<Vec<_>>()
+            .try_into()?,
+    );
+
     Ok(VerifiersInfo {
         snapshot,
-        pubkeys_by_participant: participants.into_iter().zip(pub_keys).collect(),
+        pubkeys_by_participant: participants_with_pubkeys,
     })
```
cjcobb23, line 179: I think we should get rid of this type, I don't see why we can't just directly create the `VerifierSet`. We have too many types that are basically the same thing, and it's quite confusing. I think this is the only place it's used.

contracts/multisig-prover/src/execute.rs
cjcobb23: I would just return a `VerifierSet` and call this `make_verifier_set`.

contracts/multisig-prover/src/execute.rs
```diff
@@ -129,31 +132,50 @@ fn get_verifiers_info(deps: &DepsMut, config: &Config) -> Result<VerifiersInfo,
             msg: to_json_binary(&active_verifiers_query)?,
         }))?;
 
-    let participants = verifiers
-        .clone()
+    let min_num_verifiers = deps
```
cjcobb23, line 135: Not necessarily for this PR, but I think this should be configurable per connection. We could store this in the config at instantiation, and provide some way to update it. It could default to what's stored in the service registry if unspecified.

contracts/multisig-prover/src/execute.rs
fish-sammy: sad

contracts/multisig-prover/src/execute.rs
haiyizxx: let me try to remove

contracts/multisig-prover/src/execute.rs
```diff
@@ -129,31 +132,50 @@ fn get_verifiers_info(deps: &DepsMut, config: &Config) -> Result<VerifiersInfo,
...
+        .querier
+        .query::<Service>(&QueryRequest::Wasm(WasmQuery::Smart {
+            contract_addr: config.service_registry.to_string(),
+            msg: to_json_binary(&service_registry::msg::QueryMsg::GetService {
+                service_name: config.service_name.clone(),
+            })?,
+        }))?
```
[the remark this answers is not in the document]
haiyizxx: it's a good point, we should have the context error after refactor to use client query other contracts. don't want to make a special case this pr, as the contract also talk to other contracts.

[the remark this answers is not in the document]
cgorenflo: we should build a client and wrap that there like for the voting verifier

cgorenflo (approved):

PR #451, opened 2024-06-10
feat(ampd): change the QueuedBroadcaster so that the BroadcasterClient returns error immediately on fee estimation failures

fish-sammy:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


ampd/src/queue/queued_broadcaster.rs
```diff
@@ -130,29 +125,26 @@ where
                 }
               },
               _ = interval.tick() => {
-                broadcast_all(&mut queue, &mut broadcaster).await?;
-                interval.reset();
-              },
-              _ = self.broadcast_rx.recv() => {
```
cjcobb23, line 136: when was this case being triggered?

ampd/src/queue/queued_broadcaster.rs
```diff
@@ -25,36 +29,36 @@ pub enum Error {
...
 #[async_trait]
-impl BroadcasterClient for QueuedBroadcasterClient {
+impl<F> BroadcasterClient for QueuedBroadcasterClient<F>
+where
+    F: FeeEstimator + Send,
+{
     async fn broadcast(&self, msg: Any) -> Result {
+        let fee = self
+            .fee_estimator
+            .lock()
```
cjcobb23, line 54: Why do we need to lock here? Also, is this going to contend with the lock acquired when broadcasting? Do we want to use a RwLock?

ampd/src/queue/queued_broadcaster.rs
fish-sammy: There was a functionality to force broadcast which was never used. Removed it for simplicity.

ampd/src/queue/queued_broadcaster.rs
fish-sammy: Both fee_estimate and broadcast methods require a mutable reference of the broadcaster. Both of them update the sequence number maintained in memory. Even if they don't, they still both require a mutable reference of the underlying gRPC client, which is just the nature of gRPC, i.e. sending any request requires a mutable reference of the client. Therefore RwLock doesn't make sense.

PR #454, opened 2024-06-12
feat(ampd): change the QueuedBroadcaster so that the BroadcasterClient returns error immediately on fee estimation failures

fish-sammy:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


ampd/src/queue/queued_broadcaster.rs
```diff
@@ -108,35 +91,40 @@ where
...
+                        }
+
+                        let message_type = msg.type_url.clone();
+                        queue.push(msg, fee.gas_limit).change_context(Error::Queue)?;
+                        info!(
+                            message_type,
+                            queue_size = queue.len(),
+                            queue_gas_cost = queue.gas_cost(),
+                            "pushed a new message into the queue"
+                        );
```
cgorenflo, line 121: I would move this into the `Ok(fee)` branch and get rid of the continue

ampd/src/queue/queued_broadcaster.rs
```diff
@@ -81,23 +70,17 @@ where
...
+            batch_gas_limit,
+            broadcast_interval,
+            channel: mpsc::channel(capacity),
+        }
     }
 
-    pub async fn run(mut self) -> Result {
+    pub async fn run(self) -> Result {
         let (tx, mut rx) = self.channel;
         drop(tx);
```
cgorenflo, line 85: how does this work with the client function? Does it only work because client() must be called first?

cgorenflo, line 85:
```suggestion
        let (_, mut rx) = self.channel;
```
is this equivalent?

ampd/src/queue/queued_broadcaster.rs
```diff
@@ -81,23 +70,17 @@ where
...
+            broadcaster,
+            queue: MsgQueue::default(),
+            batch_gas_limit,
+            broadcast_interval,
+            channel: mpsc::channel(capacity),
+        }
     }
 
-    pub async fn run(mut self) -> Result {
+    pub async fn run(self) -> Result {
```
cgorenflo, line 83: how do you shut this down?

ampd/src/queue/queued_broadcaster.rs
```diff
@@ -108,35 +91,40 @@ where
 
         loop {
             select! {
-              msg = rx.recv() => match msg {
-                None => break,
-                Some(msg) => {
-                  let fee = broadcaster.estimate_fee(vec![msg.clone()]).await.change_context(Error::EstimateFee)?;
```
cjcobb23, line 114: So it looks like before, this loop would exit if there was a fee estimation error. Whereas now, the loop continues. Why the change?

fish-sammy: It ignores tx that cannot be estimated and sends the error back to the client.

ampd/src/queue/queued_broadcaster.rs
fish-sammy: Unfortunately, it is not. The unit tests hang with it. I've tried.

fish-sammy: Yes

ampd/src/queue/queued_broadcaster.rs
```diff
@@ -152,6 +122,52 @@ where
...
+            let message_type = msg.type_url.clone();
+            queue
+                .push(msg, fee.gas_limit)
+                .change_context(Error::Queue)?;
+            info!(
+                message_type,
+                queue_size = queue.len(),
+                queue_gas_cost = queue.gas_cost(),
+                "pushed a new message into the queue"
+            );
```
cgorenflo, line 160: I'd keep the rest in the `run` function and only extract the part that deals with the queue and broadcast

cgorenflo (approved):

ampd/src/queue/queued_broadcaster.rs
```diff
@@ -78,113 +79,99 @@ where
...
-        0 => Ok(()),
-        n => {
-            info!(message_count = n, "ready to broadcast messages");
-
-            broadcaster
-                .broadcast(msgs)
-                .await
-                .map(|_| ())
-                .change_context(Error::Broadcast)
+    async fn handle_msg(
```
cgorenflo, line 136: looks much better now. Could we return `Result<Interval,_>` instead? I'm not a fan of silently modifying the interval as a side effect

ampd/src/queue/queued_broadcaster.rs
```diff
@@ -78,113 +79,99 @@ where
...
-        drop(tx);
-
-        let mut queue = self.queue;
-        let mut broadcaster = self.broadcaster;
-
+    pub async fn run(mut self) -> Result {
+        let (_, mut rx) = self
+            .channel
+            .take()
+            .expect("channel must be set before running");
```
cgorenflo, line 90: I would prefer an explanation why this shouldn't fail instead of a "regular" error message. Example:
"broadcast channel is expected to be set during initialization and must be available when running the broadcaster"

ampd/src/queue/queued_broadcaster.rs
```diff
@@ -78,113 +79,99 @@ where
...
         Ok(())
     }
 
     pub fn client(&self) -> QueuedBroadcasterClient {
         QueuedBroadcasterClient {
-            sender: self.channel.0.clone(),
+            sender: self
+                .channel
+                .as_ref()
+                .expect("channel must be set before running")
```
cgorenflo, line 113: see other expect

PR #466, opened 2024-06-21
feat(ampd): add support for multiversx blockchain

raress96:
## Description

Add support for MultiversX blockchain to ampd, both for verify messages and verify verifier set.


ampd/src/mvx/verifier.rs
```diff
@@ -0,0 +1,608 @@
...
+
+            let event_name = unwrap_or_continue!(topics.get(0));
+            let event_name = STANDARD.decode(event_name).unwrap_or(Vec::new());
+            if event_name.as_slice() == needed_event_name {
+                return Some(event);
+            }
+        }
+    }
+
+    return None;
```
cgorenflo, line 133: no plain `unwrap()` in production code. You can get around it by making use of the `?` operator. That also saves you the use of a macro.

```suggestion
    // We only support log index being 0, so one supported event per transaction because of MultiversX limitations
    if log_index != 0 {
        return None;
    }

    // Because of current relayer limitation, if log_index is 0, we will search through the logs and get the first log
    // which corresponds to the gateway address, hence only supporting one cross chain call per transaction
    transaction
        .logs
        .as_ref()?
        .events
        .iter()
        .filter(|events| {
            events.address.to_bytes() == gateway_address.to_bytes()
                && events.identifier == identifier
        })
        .filter_map(|events| {
            let event_name = events.topics.as_ref()?.get(0)?;
            let event_name = STANDARD.decode(event_name).ok()?;
            if event_name.as_slice() == needed_event_name {
                Some(events)
            } else {
                None
            }
        })
        .next()
```

ampd/src/mvx/verifier.rs
```diff
@@ -0,0 +1,608 @@
...
+        CONTRACT_CALL_EVENT.as_bytes(),
+    ) {
+        Some(event)
+        if transaction.hash.as_ref().unwrap() == &message.tx_id.encode_hex::<String>()
+            && message.eq_event(event).unwrap_or(false) =>
+            {
+                Vote::SucceededOnChain
+            }
+        _ => Vote::NotFound,
+    }
```
cgorenflo, line 155: no `unwrap()` in production code

```suggestion
   let hash = transaction
        .hash
        .as_ref()
        .map(String::as_str)
        .unwrap_or_default();

    if hash.is_empty() {
        return Vote::NotFound;
    }

    match find_event(
        transaction,
        gateway_address,
        message.event_index as usize,
        CONTRACT_CALL_IDENTIFIER,
        CONTRACT_CALL_EVENT.as_bytes(),
    ) {
        Some(event)
            if hash == message.tx_id.encode_hex::<String>().as_str()
                && message.eq_event(event).unwrap_or(false) =>
        {
            Vote::SucceededOnChain
        }
        _ => Vote::NotFound,
    }

```

ampd/src/mvx/verifier.rs
```diff
@@ -0,0 +1,608 @@
...
+        SIGNERS_ROTATED_EVENT.as_bytes(),
+    ) {
+        Some(event)
+        if transaction.hash.as_ref().unwrap() == &verifier_set.tx_id.encode_hex::<String>()
+            && verifier_set.eq_event(event).unwrap_or(false) =>
+            {
+                Vote::SucceededOnChain
+            }
+        _ => Vote::NotFound,
+    }
```
cgorenflo, line 177: no `unwrap()` in production code

```suggestion
let hash = transaction
        .hash
        .as_ref()
        .map(String::as_str)
        .unwrap_or_default();

    if hash.is_empty() {
        return Vote::NotFound;
    }

    match find_event(
        transaction,
        gateway_address,
        verifier_set.event_index as usize,
        ROTATE_SIGNERS_IDENTIFIER,
        SIGNERS_ROTATED_EVENT.as_bytes(),
    ) {
        Some(event)
            if hash == verifier_set.tx_id.encode_hex::<String>().as_str()
                && verifier_set.eq_event(event).unwrap_or(false) =>
        {
            Vote::SucceededOnChain
        }
        _ => Vote::NotFound,
    }
```

.gitignore
```diff
@@ -17,3 +17,5 @@
 *.iml
 .idea
 .vscode
+
+config.toml
```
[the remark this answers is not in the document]
raress96: This is the default config.toml file that is loaded when running `cargo run --bin ampd` command
I have such config file so I can easily debug and run the ampd service so I think it is best to ignore it so it is not committed to this repository.
Other teams might also test the ampd service using their own config file, so it can not be commited IMO.

ampd/src/mvx/verifier.rs
```diff
@@ -0,0 +1,608 @@
...
+        }
+
+        let destination_address = topics.get(3).ok_or("")?;
+        let destination_address = STANDARD.decode(destination_address)?;
+        let destination_address = String::from_utf8(destination_address)?;
+        if destination_address != self.destination_address {
+            return Ok(false);
+        }
+
+        let payload_hash = topics.get(4).ok_or("")?;
```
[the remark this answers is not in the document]
raress96: Added a custom error, although these errors are ignored, since it will error if the data we need is not available, hence the message verification will fail (vote not found).

ampd/src/mvx/verifier.rs
```diff
@@ -0,0 +1,608 @@
...
+
+        let topics = event.topics.as_ref().ok_or("")?;
+
+        let event_name = topics.get(0).ok_or("")?;
+        let event_name = STANDARD.decode(event_name)?;
+        if event_name.as_slice() != SIGNERS_ROTATED_EVENT.as_bytes() {
+            return Ok(false);
+        }
+
+        let signers_hash = topics.get(2).ok_or("")?;
```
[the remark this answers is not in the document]
raress96: Same as above.

ampd/src/handlers/mod.rs
```diff
@@ -1,12 +1,14 @@
 pub mod chain;
 pub mod config;
 pub mod end_block;
-mod errors;
+pub mod errors;
```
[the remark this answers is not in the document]
raress96: Done

ampd/src/mvx/proxy.rs
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    fn is_valid_transaction(tx: &TransactionOnNetwork) -> bool {
+        if tx.hash.is_none() || tx.logs.is_none() || tx.status != STATUS_SUCCESS.to_string() {
+            return false;
+        }
+
+        true
+    }
+}
```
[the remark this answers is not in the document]
raress96: Done.

ampd/src/mvx/verifier.rs
raress96: Done.

ampd/src/mvx/verifier.rs
raress96: Done.

ampd/src/mvx/verifier.rs
```diff
@@ -0,0 +1,620 @@
...
+    transaction: &'a TransactionOnNetwork,
+    gateway_address: &Address,
+    log_index: usize,
+    identifier: &str,
+    needed_event_name: &[u8],
+) -> Option<&'a Events> {
+    // We only support log index being 0, so one supported event per transaction because of MultiversX limitations
+    if log_index != 0 {
+        return None;
+    }
```
cjcobb23, line 103: Can you expand more on why this is the case? Not in the code necessarily, but just in a comment. Will this always be the case? If so, why not just use the transaction hash by itself as the message id?

ampd/src/mvx/verifier.rs
```diff
@@ -0,0 +1,620 @@
...
+
+    // Because of current relayer limitation, if log_index is 0, we will search through the logs and get the first log
+    // which corresponds to the gateway address, hence only supporting one cross chain call per transaction
+    transaction
+        .logs
+        .as_ref()?
+        .events
+        .iter()
```
cjcobb23, line 111: why can't you just add an index to these events, and then properly use the event index of the message id?

raress96: The index problem is not here (on amplifier side), but rather on MUltiversX Event Notifier service side which is used by the Relayer. When events come there, they don't have any log index (so no order from the transaction), hence we only know that a `callContract` event happened, we know the txHash but we don't know the actual log index of that event, and this is not going to be fixed in the near future as far as I know.

ampd/src/mvx/verifier.rs
raress96: Will add a comment. I want to keep the ability to modify this in the future, maybe with an activation epoch flag (eg: after epoch X, we consider only messages with log index, before that we consider only log index 0).

I wanted to also leave the message id like it is since then I won't have to add a custom message id type in the library and it is easier this way.

ampd/src/mvx/error.rs
```diff
@@ -0,0 +1,9 @@
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum Error {
+    #[error("Not ed25519 key error")]
+    NotEd25519Key,
+    #[error("Required property is empty error")]
+    PropertyEmpty,
```
cgorenflo, line 8:
```suggestion
    #[error("provided key is not ed25519")]
    NotEd25519Key,
    #[error("required property is empty")]
    PropertyEmpty,
```

cgorenflo (approved):

ampd/src/handlers/config.rs
```diff
@@ -47,6 +47,14 @@ pub enum Config {
         rpc_url: Url,
         rpc_timeout: Option<Duration>,
     },
+    MvxMsgVerifier {
+        cosmwasm_contract: TMAddress,
+        proxy_url: Url,
```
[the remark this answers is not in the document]
raress96: MultiversX doesn't use RPC like other blockchains, it uses normal HTTP calls. The internal naming for this service in MultiversX is also proxy, hence the `proxy_url` here to keep the same naming.
https://docs.multiversx.com/sdk-and-tools/proxy/

ampd/src/mvx/error.rs
raress96: Done

ampd/src/mvx/verifier.rs
cjcobb23: Does multiversx support multiple contract call events in the same transaction though? If so, what would happen if a single transaction calls the gateway more than once and thus generates more than one cross chain message? It looks like only the first will be routed.

raress96: Yes, only the first will be routed, it is a limitation that applications should keep in mind since there is no way to limit this on the MultiversX contract side.

cjcobb23: we need to find a way to either route them, or reject those later events at the gateway. The gateway should not accept msgs that cannot be routed

cjcobb23: also, is the order of events here deterministic? If so, can we just assign indexes in ampd? If not, I think that's a larger problem, as different ampd nodes might see a different order, and thus not agree on the first event in the transaction

milapsheth: Yeah, this would limit what apps can do severely. In fact, for ITS we allow a user to deploy multiple tokens within the same tx via a multicall (i.e multiple contract calls). Other apps might need to perform a broadcast as well. I would like to know what the limitation is on having unique event indexes for the contract call event. If it's not possible to uniquely identify indices, then I recommend using a counter in the external gateway and including that in the event for uniqueness.

raress96: @cjcobb23 @milapsheth TLDR: We will add support for event index in ampd.

The reason we didn't add this initially is because of limitations on the Relayer/Event Notifier side for MultiversX, since the order is not supported there.

However, the event order is supported on the endpoints we use in ampd. To answer your questions:

1. There is no way to limit the events on Mvx Gateway contract side, the Gateway supports multiple cross chain contract calls per transaction and there is no limitation here.
Furthermore, the Gateway contract code is finished and is very close to the Solidity implementation, hence we will not do any other changes there

2. The order of events is deterministic, so we can indeed assign indexes in ampd. This was one of the initial implementations we had, I will re-add the necessary code and support event indexes for MultiversX

3. I can see how having multiple cross chain calls from MultiversX is useful, and we will support these. 

The only issue remains on Relayer side, there is a workaround that is not that efficient but we will figure out another solution.

Our priority is to get the ampd code merged and then we can handle the Relayer. I will do the necessary changes ASAP.

cjcobb23 (approved):

PR #468, opened 2024-06-21
chore(service-registry): migrate service struct to rename service_contract field

maancham:
## Description

AXE-4251

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


contracts/service-registry/src/msg.rs
```diff
@@ -79,3 +79,8 @@ pub enum QueryMsg {
         verifier: String,
     },
 }
+
+#[cw_serde]
+pub struct MigrateMsg {
+    pub coordinator_contract: Addr,
```
cjcobb23, line 85: We just renamed the field, so you don't need to pass this address in. You don't even need to define a message here, you can just use `Empty`.

contracts/service-registry/src/migrations/v_0_4.rs
```diff
@@ -0,0 +1,119 @@
...
+) -> Result<Response, axelar_wasm_std::ContractError> {
+    let keys_and_services = v0_3_state::SERVICES
+        .range(store, None, None, Order::Ascending)
+        .map(|result| {
+            result.map(|(service_name, service)| {
+                (
+                    service_name,
+                    Service {
+                        name: service.name,
+                        coordinator_contract: coordinator_contract.clone(),
```
cjcobb23, line 40:
```suggestion
                        coordinator_contract: service.service_contract,
```

contracts/service-registry/src/msg.rs
maancham: oh Ok, I was using that Empty but Edwin told me he used message for the previous migration so I changed it. Will revert back to not passing in the coordinator address and remove the message as well.

contracts/service-registry/src/migrations/v_0_4.rs
```diff
@@ -0,0 +1,117 @@
...
+                        description: service.description,
+                    },
+                )
+            })
+        })
+        .collect::<Result<Vec<_>, _>>()?;
+
+    for (service_name, new_service) in keys_and_services {
+        SERVICES.save(store, &service_name, &new_service)?;
+    }
```
cgorenflo, line 54:
```suggestion
   v0_3_state::SERVICES
        .range(store, None, None, Order::Ascending)
        .collect::<Result<Vec<_>, _>>()?
        .into_iter()
        .map(|(service_name, service)| {
            let service = Service {
                name: service.name,
                coordinator_contract: service.service_contract,
                min_num_verifiers: service.min_num_verifiers,
                max_num_verifiers: service.max_num_verifiers,
                min_verifier_bond: service.min_verifier_bond,
                bond_denom: service.bond_denom,
                unbonding_period_days: service.unbonding_period_days,
                description: service.description,
            };
            SERVICES.save(store, &service_name, &service)
        })
        .collect::<Result<Vec<_>, _>>()?;
```

cgorenflo (approved):

PR #470, opened 2024-06-21
feat(ampd): remove the state updater

fish-sammy:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


cgorenflo (commented):
I'd rather not add exceptions for dead code everywhere and would prefer removing that code. If needed we can bring it back in the future

cgorenflo (approved):

haiyizxx (approved):

PR #472, opened 2024-06-24
feat(minor-router): add command to freeze all chains

cgorenflo:


milapsheth (approved):

packages/router-api/src/msg.rs
```diff
@@ -25,12 +25,23 @@ pub enum ExecuteMsg {
...
-    // Unfreezes a chain, in the specified direction.
+    /// If no filter is provided, freezes all chain connections in every direction. Should only be called in emergencies.
+    /// If the filter is set, only the specified chains will be frozen.
+    FreezeAllChains { chains: Option<Vec<ChainName>> },
+
+    /// Unfreezes the specified chains in the specified directions. In contrast to FreezeAllChains,
+    /// this does not allow unfreezing all chains, because that command is too dangerous.
+    /// It would make it too easy to accidentally unfreeze a chain that should not be unfrozen.
+    UnfreezeChains {
+        chains: Vec<(ChainName, GatewayDirection)>,
```
milapsheth, line 41:
```suggestion
        chains: Vec<ChainName>,
```
Since this batch cmd is really only needed when unfreezing multiple chains from a freeze all status, I think the direction can be treated as bidirectional implicitly to keep the command simpler. UnfreezeChain can be used in other cases

milapsheth: Based on discussion in Slack, we can have cases of unidirectional connections when deprecating a chain so allowing the direction to be configured when unfreezing makes sense then

cjcobb23, line 41: Why not remove `FreezeChain` and `UnfreezeChain` in favor of these versions that take a list of chains? You can always pass a list of length 1, and now I feel the API is larger than it needs to be

cjcobb23: Sorry, I mean why not have `FreezeChains`, `UnfreezeChains` and `FreezeAllChains` as the API?

packages/router-api/src/msg.rs
```diff
@@ -25,23 +26,25 @@ pub enum ExecuteMsg {
...
      */
-    // Freezes a chain, in the specified direction.
-    FreezeChain {
-        chain: ChainName,
-        direction: GatewayDirection,
+    /// If no filter is provided, freezes all chain connections in every direction. Should only be called in emergencies.
+    /// If the filter is set, only the specified chains will be frozen in the specified directions.
+    FreezeChains {
+        chains: Option<HashMap<ChainName, GatewayDirection>>,
     },
```
cjcobb23, line 33: I actually would prefer to have a specific `FreezeAllChains` msg, instead of wrapping the argument in an Option. Then `FreezeChains` and `UnfreezeChains` can be symmetrical, and have the same parameters. I am not sure that we actually gain anything by combining freeze all here, since it complicates the parameter.

cjcobb23:
Something occurred to me. What if we just had a global flag that, when set, disables all routing? The message would be something like `DisableRouting` and `EnableRouting`. I feel like this may be better than the current approach for a couple of reasons.

1. The flag is independent of freeze status, so if you disable routing and then enable routing, any chains that were frozen prior to routing being disabled are still frozen after routing is re-enabled.

2. The flag would cover new chains added after the flag is set. Right now, if you freeze all chains, and then register a new chain, that new chain is not frozen.

3. The flag would be able to prevent routing to and from core as well (right now the code doesn't cover this case as far as I can tell).

4. Setting and unsetting this flag would be cheaper gas wise, since you don't have to loop through many chains.

Thoughts?

milapsheth:
> Something occurred to me. What if we just had a global flag that, when set, disables all routing? The message would be something like `DisableRouting` and `EnableRouting`. I feel like this may be better than the current approach for a couple of reasons.

I like it!

> 1. The flag is independent of freeze status, so if you disable routing and then enable routing, any chains that were frozen prior to routing being disabled are still frozen after routing is re-enabled.
> 2. The flag would cover new chains added after the flag is set. Right now, if you freeze all chains, and then register a new chain, that new chain is not frozen.
> 3. The flag would be able to prevent routing to and from core as well (right now the code doesn't cover this case as far as I can tell).
> 4. Setting and unsetting this flag would be cheaper gas wise, since you don't have to loop through many chains.
> 
> Thoughts?

We also don't seem to check freeze status for key rotations. Those should be prevented for frozen chains or if routing is disabled (could rename RoutingDisabled to something more generic to encompass this, such as AmplifierDisabled etc.).

This approach does make chain status check a bit more complex (we need to check both frozen status and global routing status) for both on-chain and off-chain purposes. So we should consider simplifying this in the query.

cjcobb23:
> We also don't seem to check freeze status for key rotations. Those should be prevented for frozen chains or if routing is disabled (could rename RoutingDisabled to something more generic to encompass this, such as AmplifierDisabled etc.).
> 

Hmm the only way I can think to do that is to have the multisig contract disallow signing when a chain is frozen. Multisig could query the router.

contracts/router/src/contract/execute.rs
```diff
@@ -160,11 +223,21 @@ fn validate_msgs(
...
     msgs: Vec<Message>,
 ) -> error_stack::Result<Response, Error> {
-    let config = store.load_config()?;
+    STATE
+        .load(store.storage())
+        .change_context(Error::StoreFailure)
+        .map(|state| match state {
+            State::Disabled => Err(Error::RoutingDisabled),
+            State::Enabled => Ok(()),
+        })??;
```
cjcobb23, line 236:
```suggestion
     if STATE
        .load(store.storage())
        .change_context(Error::StoreFailure)? == State::Disabled {
            return Err(report!(Error::RoutingDisabled));
        }
```

I find this easier to read

contracts/router/src/contract/query.rs
```diff
@@ -35,6 +35,10 @@ pub fn chains(
         .collect()
 }
 
+pub fn is_enabled(deps: Deps) -> bool {
+    STATE.load(deps.storage).unwrap_or(State::Disabled) == State::Enabled
```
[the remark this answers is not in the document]
cgorenflo: this is not necessarily that it's not set, but that there is an issue with the storage access. So since I have to deal with a possible error anyway I thought it made sense to disable the contract if there are any issues

contracts/router/src/contract/execute.rs
```diff
@@ -74,47 +78,106 @@ pub fn upgrade_gateway(
...
+}
+
+pub fn enable_routing(deps: DepsMut) -> Result<Response, Error> {
+    let state = STATE.update(deps.storage, |state| match state {
+        State::Disabled => Ok(State::Enabled),
+        State::Enabled => Err(StateUpdateError::SameState),
+    });
+
+    state_toggle_response(state, events::RoutingEnabled)
+}
```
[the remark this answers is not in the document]
cgorenflo: The reason I did it this way is that it is treated slightly differently:
- toggle state? emit respective event 
- set same state? noop but without event
- actual error? return error

the `state_toggle_response` translates the same state into a noop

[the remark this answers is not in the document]
cgorenflo: The `StateUpdateError` is just an internal marker

PR #476, opened 2024-06-27
feat(minor-voting-verifier): consistent gas cost for voting

cjcobb23:
https://axelarnetwork.atlassian.net/browse/AXE-4318

Prior to this, the gas cost of each vote was more than prior votes in the same poll. Performance testing with 25 verifiers showed later votes costing over 100% more gas than earlier votes, which was causing out of gas errors. This PR makes the gas cost of each vote constant, except for the vote that achieves quorum. In testing with 25 verifiers and 20 messages in a batch, the vote that achieves quorum cost 6% more gas than other votes.


packages/axelar-wasm-std/src/voting.rs
```diff
@@ -320,18 +330,23 @@ impl WeightedPoll {
...
-                    });
-
-                    if voted_consensus {
-                        Some(address.to_owned())
-                    } else {
-                        None
-                    }
-                })
+            .filter_map(|(address, _)| {
+                voting_history
```
cgorenflo, line 334: voting_history is just used as a lookup, yet it's iterated for each address. convert it into a Hashmap first and do constant time lookup in the filter_map

cgorenflo (commented):
I just want to confirm that I understand the change correctly: it comes down to replacing a `Vec<Votes>` with a single `bool`, right? That means it still scales with the number of verifiers, but not with the number of votes anymore.

packages/axelar-wasm-std/src/voting.rs
cjcobb23: I feel like Hashmap will be more readable, so I'll switch to Hashmap. But I am not sure that it's actually more performant: oftentimes linear searching small vectors like this is actually faster than lookup in a hashmap (because all the memory is contiguous in a vector and so searching is very cache friendly), and the hashmap lookup only becomes faster once the dataset grows larger.

cgorenflo (approved):

PR #480, opened 2024-06-28
fix(ampd): directly pass interval to queued broadcaster and ignore first tick

maancham:
AXE-4240

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

### DIFF
--- a/ampd/src/queue/queued_broadcaster.rs
+++ b/ampd/src/queue/queued_broadcaster.rs
@@ -89,11 +89,9 @@ where
 
         loop {
             select! {
-                 msg = rx.recv() => match msg {
+                msg = rx.recv() => match msg {
                     None => break,
-                    Some(msg_and_res_chan) => {
-                        self.handle_msg(msg_and_res_chan).await?;
-                    },
+                    Some(msg_and_res_chan) => self.handle_msg(msg_and_res_chan).await?,
                 },
                 _ = self.broadcast_interval.tick() => {
                     self.broadcast_all().await?.then(|_| {
@@ -181,8 +179,8 @@ mod test {
     use cosmrs::Any;
     use cosmrs::{bank::MsgSend, tx::Msg, AccountId};
     use error_stack::Report;
-    use tokio::time::{advance, interval, sleep, Duration};
-    use tokio::{select, test};
+    use tokio::test;
+    use tokio::time::{interval, Duration};
 
     use super::{Error, QueuedBroadcaster};
     use crate::broadcaster::{self, MockBroadcaster};
@@ -195,8 +193,8 @@ mod test {
             .expect_estimate_fee()
             .return_once(|_| Err(Report::new(broadcaster::Error::FeeEstimation)));
 
-        let desired_interval = interval(Duration::from_secs(5));
-        let queued_broadcaster = QueuedBroadcaster::new(broadcaster, 100, 10, desired_interval);
+        let broadcast_interval = interval(Duration::from_secs(5));
+        let queued_broadcaster = QueuedBroadcaster::new(broadcaster, 100, 10, broadcast_interval);
         let client = queued_broadcaster.client();
         let handle = tokio::spawn(queued_broadcaster.run());
 
@@ -240,14 +238,12 @@ mod test {
                 Ok(TxResponse::default())
             });
 
-        let mut desired_interval = interval(Duration::from_secs(5));
-        select! {
-            _ = desired_interval.tick() => {
-            },
-        }
+        let mut broadcast_interval = interval(Duration::from_secs(5));
+        // get rid of tick on startup
+        broadcast_interval.tick().await;
 
         let queued_broadcaster =
-            QueuedBroadcaster::new(broadcaster, batch_gas_limit, tx_count, desired_interval);
+            QueuedBroadcaster::new(broadcaster, batch_gas_limit, tx_count, broadcast_interval);
         let client = queued_broadcaster.client();
         let handle = tokio::spawn(queued_broadcaster.run());
 
@@ -259,11 +255,10 @@ mod test {
         assert!(handle.await.unwrap().is_ok());
     }
 
-    #[test]
+    #[test(start_paused = true)]
     async fn should_broadcast_when_broadcast_interval_has_been_reached() {
         let tx_count = 9;
         let batch_gas_limit = 100;
-        let broadcast_interval = Duration::from_millis(100);
         let gas_limit = 10;
 
         let mut broadcaster = MockBroadcaster::new();
@@ -286,17 +281,18 @@ mod test {
 
                 Ok(TxResponse::default())
             });
-        let desired_interval = interval(Duration::from_millis(100));
+        let mut broadcast_interval = interval(Duration::from_millis(100));
+        // get rid of tick on startup
+        broadcast_interval.tick().await;
 
         let queued_broadcaster =
-            QueuedBroadcaster::new(broadcaster, batch_gas_limit, tx_count, desired_interval);
+            QueuedBroadcaster::new(broadcaster, batch_gas_limit, tx_count, broadcast_interval);
         let client = queued_broadcaster.client();
         let handle = tokio::spawn(queued_broadcaster.run());
 
         for _ in 0..tx_count {
             client.broadcast(dummy_msg()).await.unwrap();
         }
-        sleep(broadcast_interval).await;
         drop(client);
 
         assert!(handle.await.unwrap().is_ok());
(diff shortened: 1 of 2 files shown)


ampd/src/queue/queued_broadcaster.rs
```diff
@@ -88,15 +86,20 @@ where
             .channel
             .take()
             .expect("broadcast channel is expected to be set during initialization and must be available when running the broadcaster");
-        let mut interval = time::interval(self.broadcast_interval);
 
         loop {
             select! {
-                msg = rx.recv() => match msg {
+                 msg = rx.recv() => match msg {
```
cgorenflo, line 92:
```suggestion
                msg = rx.recv() => match msg {
```

ampd/src/queue/queued_broadcaster.rs
```diff
@@ -88,15 +86,20 @@ where
...
                     None => break,
-                    Some(msg_and_res_chan) => interval = self.handle_msg(interval, msg_and_res_chan).await?,
+                    Some(msg_and_res_chan) => {
+                        self.handle_msg(msg_and_res_chan).await?;
+                    },
+                },
+                _ = self.broadcast_interval.tick() => {
```
cgorenflo, line 98: the outer `{ }` are not necessary

ampd/src/queue/queued_broadcaster.rs
```diff
@@ -196,8 +195,8 @@ mod test {
             .expect_estimate_fee()
             .return_once(|_| Err(Report::new(broadcaster::Error::FeeEstimation)));
 
-        let queued_broadcaster =
-            QueuedBroadcaster::new(broadcaster, 100, 10, Duration::from_secs(5));
+        let desired_interval = interval(Duration::from_secs(5));
```
cgorenflo, line 198:
```suggestion
        let broadcast_interval = interval(Duration::from_secs(5));
```

ampd/src/queue/queued_broadcaster.rs
```diff
@@ -236,17 +235,19 @@ mod test {
...
                 Ok(TxResponse::default())
             });
 
-        let queued_broadcaster = QueuedBroadcaster::new(
-            broadcaster,
-            batch_gas_limit,
-            tx_count,
-            Duration::from_secs(5),
-        );
+        let mut desired_interval = interval(Duration::from_secs(5));
```
cgorenflo, line 243:
```suggestion
        let mut broadcast_interval = interval(Duration::from_secs(5));
```

ampd/src/queue/queued_broadcaster.rs
```diff
@@ -236,17 +235,19 @@ mod test {
...
+        select! {
+            _ = desired_interval.tick() => {
+            },
```
cgorenflo, line 246:
```suggestion
            _ = desired_interval.tick() => {}
```
or does the formatter choose this format?

ampd/src/queue/queued_broadcaster.rs
```diff
@@ -281,13 +282,14 @@ mod test {
             .expect_broadcast()
             .once()
             .returning(move |msgs| {
-                assert!(msgs.len() == tx_count);
+                assert_eq!(msgs.len(), tx_count);
 
                 Ok(TxResponse::default())
             });
+        let desired_interval = interval(Duration::from_millis(100));
```
cgorenflo, line 289:
```suggestion
        let broadcast_interval = interval(Duration::from_millis(100));
```

ampd/src/queue/queued_broadcaster.rs
```diff
@@ -281,13 +282,14 @@ mod test {
...
 
         let queued_broadcaster =
```
cgorenflo, line 291: we should do the same thing here as in the other cases: pause the test, get rid of the initial tick, queue msgs, then advance time instead of sleeping

ampd/src/queue/queued_broadcaster.rs
```diff
@@ -322,31 +324,36 @@ mod test {
...
                 Ok(TxResponse::default())
             });
 
-        let queued_broadcaster = QueuedBroadcaster::new(
-            broadcaster,
-            batch_gas_limit,
-            tx_count,
-            Duration::from_secs(5),
-        );
+        let mut desired_interval = interval(Duration::from_secs(5));
```
cgorenflo, line 339:
```suggestion
        let mut broadcast_interval = interval(Duration::from_secs(5));
```

ampd/src/queue/queued_broadcaster.rs
```diff
@@ -322,31 +324,36 @@ mod test {
...
+        select! {
+            _ = desired_interval.tick() => {
+            },
```
cgorenflo, line 342:
```suggestion
            _ = desired_interval.tick() => {}
```

ampd/src/queue/queued_broadcaster.rs
```diff
@@ -322,31 +324,36 @@ mod test {
...
+            QueuedBroadcaster::new(broadcaster, batch_gas_limit, tx_count, desired_interval);
         let client = queued_broadcaster.client();
         let handle = tokio::spawn(queued_broadcaster.run());
 
         for _ in 0..tx_count {
             client.broadcast(dummy_msg()).await.unwrap();
         }
+
+        // Advance time to cause the last message broadcast
+        advance(Duration::from_secs(7)).await;
```
cgorenflo, line 355: can you remove this? the broadcaster should actually do a final broadcast on shutdown (after `drop(client)`), so we don't have to rely on the ticker here at all

ampd/src/queue/queued_broadcaster.rs
```diff
@@ -300,7 +302,7 @@ mod test {
         assert!(handle.await.unwrap().is_ok());
     }
 
-    #[test]
+    #[test(start_paused = true)]
     async fn should_broadcast_when_gas_limit_has_been_reached() {
```
cgorenflo, line 306: did you run it in a 100+ loop to check that it is stable?

ampd/src/queue/queued_broadcaster.rs
```diff
@@ -322,31 +324,36 @@ mod test {
...
             });
 
-        let queued_broadcaster = QueuedBroadcaster::new(
-            broadcaster,
-            batch_gas_limit,
-            tx_count,
-            Duration::from_secs(5),
-        );
+        let mut desired_interval = interval(Duration::from_secs(5));
+        select! {
```
cgorenflo, line 340:
```suggestion
        // get rid of tick on startup
        select! {
```

ampd/src/queue/queued_broadcaster.rs
```diff
@@ -236,17 +235,19 @@ mod test {
...
```
cgorenflo, line 244:
```suggestion
        // get rid of tick on startup
        select! {
```

ampd/src/queue/queued_broadcaster.rs
```diff
@@ -214,7 +213,7 @@ mod test {
         assert!(handle.await.unwrap().is_ok());
     }
 
-    #[test]
+    #[test(start_paused = true)]
     async fn should_not_broadcast_when_gas_limit_has_not_been_reached() {
```
cgorenflo, line 217: this test doesn't actually test what it should:
- the MockBroadcaster broadcast function should not be called at all
- we should advance the interval after the messages are queued for broadcast to make sure the ticker doesn't trigger a broadcast
- no drop(client)
- no waiting for the handle to complete at the end
- do a short sleep (50ms or something like that) to make sure the broadcaster had every chance to be triggered before the test ends

ampd/src/queue/queued_broadcaster.rs
maancham: Yeah. This is how I tested it:

```
#[test(start_paused = true)]
async fn test_should_broadcast_when_gas_limit_has_been_reached() {
    for _ in 0..5000 {
        should_broadcast_when_gas_limit_has_been_reached().await;
    }
}

async fn should_broadcast_when_gas_limit_has_been_reached() {
    function code.....
}
```

ampd/src/queue/queued_broadcaster.rs
fish-sammy: I don't understand this `select!` at all. Why not just `broadcast_interval.tick().await;`?

ampd/src/queue/queued_broadcaster.rs
fish-sammy: Same here. What's with this `select!`?

ampd/src/queue/queued_broadcaster.rs
maancham: yeah I removed it and the test was fine.

ampd/src/queue/queued_broadcaster.rs
maancham: Sure. Also removed the sleep since we have a the `drop(client)`.

ampd/src/queue/queued_broadcaster.rs
maancham: Created a ticket for this to come back to later.

ampd/src/queue/queued_broadcaster.rs
```diff
@@ -88,15 +86,18 @@ where
...
             select! {
                 msg = rx.recv() => match msg {
                     None => break,
-                    Some(msg_and_res_chan) => interval = self.handle_msg(interval, msg_and_res_chan).await?,
+                    Some(msg_and_res_chan) => self.handle_msg(msg_and_res_chan).await?,
+                },
+                _ = self.broadcast_interval.tick() => {
+                    self.broadcast_all().await?.then(|_| {
+                        self.broadcast_interval.reset();
+                    });
```
fish-sammy, line 99: Either remove the outer `{ }` or 
```
                    self.broadcast_all().await?;
                    self.broadcast_interval.reset();
```

fish-sammy (approved):

PR #481, opened 2024-07-01
feat(minor-voting-verifier): allow restarting expired polls

cjcobb23:
https://axelarnetwork.atlassian.net/browse/AXE-4319

### DIFF
--- a/contracts/voting-verifier/src/query.rs
+++ b/contracts/voting-verifier/src/query.rs
@@ -1,7 +1,4 @@
-use axelar_wasm_std::{
-    voting::{PollStatus, Vote},
-    MajorityThreshold, VerificationStatus,
-};
+use axelar_wasm_std::{voting::Vote, MajorityThreshold, VerificationStatus};
 use cosmwasm_std::Deps;
 use multisig::verifier_set::VerifierSet;
 use router_api::Message;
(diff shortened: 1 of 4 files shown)


contracts/voting-verifier/src/query.rs
```diff
@@ -76,18 +90,18 @@ fn verification_status<T: PartialEq + std::fmt::Debug>(
...
         None => VerificationStatus::Unknown,
     }
 }
 
-fn is_finished(poll: &state::Poll) -> bool {
+fn is_finished(poll: &state::Poll, cur_block_height: u64) -> bool {
     match poll {
         state::Poll::Messages(poll) | state::Poll::ConfirmVerifierSet(poll) => {
-            poll.status == PollStatus::Finished
+            cur_block_height >= poll.expires_at
```
cgorenflo, line 104:
```suggestion
            poll.status == PollStatus::Finished | cur_block_height >= poll.expires_at
```
If a poll completes before expiry it should also count as finished

cjcobb23: it's actually not possible for a poll to be in the finished state prior to expiration at the moment

cgorenflo: that's fine, I would still like to make it explicit instead of relying on implicit state checks. That could potentially change in the future and then this code breaks without us knowing. I'm also fine if you want to extract it out into a function that is used across the code base. I just don't want multiple checks across the codebase that imply state from circumstantial evidence. They are too easy to miss when changing behaviour and we have an explicit state to use. Frankly, I would prefer it even more if we for example change `poll.status` into a function that takes the current block and returns the correct state on it's own. Then we could add a `PollStatus::Expired` and maybe rename `PollStatus::Finished` to make the difference clear, so all the status logic is directly tied to the poll instead of distributed across the code

cjcobb23: I changed the status into a function, lmk what you think. It was a bit weird because we still need to actually modify the struct somehow when `finish` is called, but whether a poll is in progress or expired is just determined dynamically.

contracts/voting-verifier/src/query.rs
```diff
@@ -90,18 +93,23 @@ fn verification_status<T: PartialEq + std::fmt::Debug>(
...
 }
 
-fn is_finished(poll: &state::Poll, cur_block_height: u64) -> bool {
+fn voting_completed(poll: &state::Poll, cur_block_height: u64) -> bool {
     match poll {
         state::Poll::Messages(poll) | state::Poll::ConfirmVerifierSet(poll) => {
-            cur_block_height >= poll.expires_at
+            matches!(
+                poll.status(cur_block_height),
+                PollStatus::Expired | PollStatus::Finished
```
cgorenflo, line 111: I didn't know this worked, cool!

cgorenflo (approved):

PR #482, opened 2024-07-02
feat: add interchain token service encoding/decoding api

milapsheth:
## Description

[AXE-4071]

## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Other Notes

[AXE-4071]: https://axelarnetwork.atlassian.net/browse/AXE-4071?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/interchain-token-service-api/src/msg.rs
```diff
@@ -0,0 +1,21 @@
...
+
+    // Returns a list of chains registered with the router
+    // The list is paginated by:
+    // - start_after: the chain name to start after, which the next page of results should start.
+    // - limit: limit the number of chains returned, default is u32::MAX.
+    #[returns(Vec<ChainEndpoint>)]
+    Chains {
+        start_after: Option<ChainName>,
+        limit: Option<u32>,
+    },
```
cjcobb23, line 20: What's the use case for this? Who/what will query this that can't just query the router?

packages/interchain-token-service-api/src/primitives.rs
```diff
@@ -0,0 +1,90 @@
...
+use cosmwasm_std::{HexBinary, Uint256};
+use error_stack::Report;
+
+#[cw_serde]
+#[derive(Eq)]
+pub struct TokenId {
+    #[serde(with = "axelar_wasm_std::hex")]
+    #[schemars(with = "String")]
+    pub id: [u8; 32],
+}
```
cjcobb23, line 12: consider just `pub type TokenId = [u8;32];`

packages/interchain-token-service-api/src/primitives.rs
```diff
@@ -0,0 +1,90 @@
...
+
+#[cw_serde]
+#[derive(Eq)]
+pub enum ITSMessage {
+    InterchainTransfer {
+        token_id: TokenId,
+        source_address: HexBinary,
+        destination_address: HexBinary,
+        amount: Uint256,
+        data: HexBinary,
```
cjcobb23, line 34: Is this the full payload or the hash?

packages/interchain-token-service-api/src/primitives.rs
```diff
@@ -0,0 +1,90 @@
...
+    Gateway = 5,
+}
+
+#[cw_serde]
+#[derive(Eq)]
+pub enum ITSMessage {
+    InterchainTransfer {
+        token_id: TokenId,
+        source_address: HexBinary,
+        destination_address: HexBinary,
```
cjcobb23, line 32: Should probably use the `Address` type that we use for `Message`. It's defined in `router_api::primitives`. Will make composability a bit cleaner.

packages/interchain-token-service-api/src/primitives.rs
```diff
@@ -0,0 +1,90 @@
...
+        amount: Uint256,
+        data: HexBinary,
+    },
+    DeployInterchainToken {
+        token_id: TokenId,
+        name: String,
+        symbol: String,
+        decimals: u8,
+        minter: HexBinary,
```
cjcobb23, line 41: If this is an address, I'd prefer to use `Address` in `router_api::primitives`

packages/interchain-token-service-api/src/client.rs
```diff
@@ -0,0 +1,31 @@
...
+            contract_addr: self.address.to_string(),
+            msg: to_json_binary(&msg).expect("msg should always be serializable"),
+            funds: vec![],
+        }
+    }
+
+    pub fn route(&self, msgs: Vec<Message>) -> Option<WasmMsg> {
+        ignore_empty(msgs).map(|msgs| self.execute(&ExecuteMsg::RouteMessages(msgs)))
+    }
+}
```
cjcobb23, line 22: This looks like a client for the router. Why is this here?

packages/interchain-token-service-api/src/primitives.rs
```diff
@@ -0,0 +1,90 @@
...
+#[cw_serde]
+#[derive(Eq)]
+pub enum ITSMessage {
+    InterchainTransfer {
+        token_id: TokenId,
+        source_address: HexBinary,
+        destination_address: HexBinary,
+        amount: Uint256,
+        data: HexBinary,
+    },
```
cjcobb23, line 35: Why is there no source or destination chain here?

packages/interchain-token-service-api/src/abi.rs
```diff
@@ -0,0 +1,427 @@
...
+    struct DeployTokenManager {
+        uint256 messageType;
+        bytes32 tokenId;
+        uint256 tokenManagerType;
+        bytes params;
+    }
+
+    struct RoutedCall {
+        uint256 messageType;
+        string chain;
```
cjcobb23, line 47: Is this the destination chain or the source chain?

packages/interchain-token-service-api/src/msg.rs
milapsheth: ah, checked in by mistake

packages/interchain-token-service-api/src/primitives.rs
milapsheth: Full payload. The whole struct gets encoded into the GMP payload that we approve over. But because ITS hub needs to receive the GMP payload to process, the ITS contract data also gets included. While we don't store this payload, it is going to affect p2p bandwidth and get included into the block data unfortunately. Not including this would require breaking changes on the relayers to support ITS GMP execute calls in a special way.

packages/interchain-token-service-api/src/primitives.rs
milapsheth: same comment as above

packages/interchain-token-service-api/src/primitives.rs
milapsheth: That info comes from the GMP Message struct. The true destination chain is included in `ITSRoutedMessage`, to be used for the next GMP call

packages/interchain-token-service-api/src/abi.rs
milapsheth: I'll change the name to remote_chain and add a comment. It's used for both. When ITS edge contract sends a RoutedCall, the remote_chain is the true destination, and when it receives the GMP call from ITS hub it's the true source_chain.

packages/interchain-token-service-api/src/client.rs
milapsheth: same, got checked in by mistake

packages/interchain-token-service-api/src/primitives.rs
milapsheth: since TokenId will be a public facing struct, we'd like to serialize it as hex. That's not doable if it's a type alias, right?

packages/interchain-token-service-api/src/abi.rs
```diff
@@ -0,0 +1,469 @@
...
+        uint256 tokenManagerType;
+        bytes params;
+    }
+
+    struct RoutedCall {
+        uint256 messageType;
+        /// Remote chain name.
+        /// ITS edge source contract -> ITS Hub GMP call: Set to the true destination chain name.
+        /// ITS Hub -> ITS edge destination contract: Set to the true source chain name.
+        string remote_chain;
```
cjcobb23, line 55: Is there a way we can not overload this field like this, and use some sort of enum? I find this confusing.

packages/interchain-token-service-api/src/primitives.rs
```diff
@@ -0,0 +1,87 @@
...
+            0 => TokenManagerType::NativeInterchainToken,
+            1 => TokenManagerType::MintBurnFrom,
+            2 => TokenManagerType::LockUnlock,
+            3 => TokenManagerType::LockUnlockFee,
+            4 => TokenManagerType::MintBurn,
+            5 => TokenManagerType::Gateway,
+            _ => return Err(Report::new(Error::InvalidEnum)),
+        })
+    }
+}
```
cgorenflo, line 87: use `strum_macros::FromRepr` to generate this instead

packages/interchain-token-service-api/src/primitives.rs
```diff
@@ -0,0 +1,87 @@
...
+#[cw_serde]
+#[derive(Eq, Copy)]
+#[repr(u8)]
+pub enum TokenManagerType {
+    NativeInterchainToken = 0,
+    MintBurnFrom = 1,
+    LockUnlock = 2,
+    LockUnlockFee = 3,
+    MintBurn = 4,
+    Gateway = 5,
```
cgorenflo, line 24: no need to number them all, rust will do the enumeration automatically

packages/interchain-token-service-api/src/primitives.rs
```diff
@@ -0,0 +1,87 @@
...
+    },
+}
+
+#[cw_serde]
+#[derive(Eq)]
+pub struct ITSRoutedMessage {
+    /// Remote chain name.
+    /// ITS edge source contract -> ITS Hub GMP call: Set to the true destination chain name.
+    /// ITS Hub -> ITS edge destination contract: Set to the true source chain name.
+    pub remote_chain: Address,
```
cgorenflo, line 57: as CJ mentioned, this is really weird behaviour

packages/interchain-token-service-api/src/abi.rs
```diff
@@ -0,0 +1,469 @@
...
+        bytes message;
+    }
+}
+
+impl ITSRoutedMessage {
```
cgorenflo, line 60: why is this defined in a separate module?

packages/interchain-token-service-api/src/abi.rs
```diff
@@ -0,0 +1,469 @@
...
+            ITSMessage::DeployTokenManager {
+                token_id,
+                token_manager_type,
+                params,
+            } => DeployTokenManager {
+                messageType: MessageType::DeployTokenManager.into(),
+                tokenId: FixedBytes::<32>::new(token_id.to_bytes()),
+                tokenManagerType: token_manager_type.into(),
+                params: Vec::<u8>::from(params).into(),
+            }
```
cgorenflo, line 102: seems to me that the mapping should be defined on `ItsMessage`,

packages/interchain-token-service-api/src/abi.rs
```diff
@@ -0,0 +1,469 @@
...
+        /// ITS edge source contract -> ITS Hub GMP call: Set to the true destination chain name.
+        /// ITS Hub -> ITS edge destination contract: Set to the true source chain name.
+        string remote_chain;
+        bytes message;
+    }
+}
+
+impl ITSRoutedMessage {
+    pub fn abi_encode(&self) -> HexBinary {
+        let message: Vec<u8> = match self.message.clone() {
```
cgorenflo, line 62: why is this defined on a reference only to then clone the message?

cgorenflo (changes requested):

Foivos (approved):

packages/interchain-token-service-api/src/abi.rs
milapsheth: Yeah, we can have two message types on the wire to distinguish it. The downside is that now you can have message types that are only valid in a certain direction (send to hub / receive from hub). Another option is to keep it the same on the wire but use send/receive type Rust types. But this makes decoding context sensitive which seems confusing. Discussed with SC team, and we decided to go with two separate message types

packages/interchain-token-service-api/src/testdata/deploy_interchain_token_encode_decode.golden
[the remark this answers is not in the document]
milapsheth: yeah, don't see a way to force a newline. Since this in a test file, I'm ok to not add it. Otherwise we can look at adding a script that does it post generation

packages/interchain-token-service-api/src/abi.rs
milapsheth: We might have other encoding types in the future potentially. This module is responsible for only adding ABI functionality

packages/interchain-token-service-api/src/primitives.rs
milapsheth: https://github.com/axelarnetwork/axelar-amplifier/pull/482#discussion_r1686782811

packages/interchain-token-service-api/src/primitives.rs
```diff
@@ -0,0 +1,87 @@
...
+impl TokenId {
+    #[inline]
+    pub const fn new(id: [u8; 32]) -> Self {
+        Self(id)
+    }
+
+    #[inline]
+    pub const fn to_bytes(&self) -> [u8; 32] {
+        self.0
+    }
```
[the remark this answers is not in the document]
milapsheth: Converted to `into`. The compiler will inline it so it shouldn't copy. To ensure that the inlining optimization works [cross-crate boundaries](https://matklad.github.io/2021/07/09/inline-in-rust.html), I added the `#[inline]` macro. I've seen this as a common pattern for other crates.

interchain-token-service/src/error.rs
```diff
@@ -0,0 +1,10 @@
+use axelar_wasm_std_derive::IntoContractError;
+use thiserror::Error;
+
+#[derive(Error, Debug, PartialEq, IntoContractError)]
+pub enum Error {
+    #[error("failed to decode ITS message: {0}")]
+    InvalidMessage(String),
```
cgorenflo, line 7: I'm not a fan of how the error conversions are handled. This is exactly what error-stack is for:
define it as 
```
#[error("failed to decode ITS message")]
```

and then use
```
result.change_context(Error::InvalidMessage)
```
instead of 
```
result.map_err(|e| Report::new(Error::InvalidMessage(e.to_string())));
```

cgorenflo (approved):

PR #483, opened 2024-07-02
feat(minor-multisig)!: add command to disable all signing

cjcobb23:
adds a generic killswitch module that can be used by any contract (currently router and multisig)

Followups:
- Allow authorize and unauthorize caller to take a list of addresses
- migration of multisig
- use callback in permission control for start signing session, submit signature, and register verifier set

https://axelarnetwork.atlassian.net/browse/AXE-4359

### DIFF
--- a/contracts/multisig/src/state.rs
+++ b/contracts/multisig/src/state.rs
@@ -13,7 +13,6 @@ use crate::{
 
 #[cw_serde]
 pub struct Config {
-    pub governance: Addr,
     pub rewards_contract: Addr,
     pub block_expiry: u64, // number of blocks after which a signing session expires
 }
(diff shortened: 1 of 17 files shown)


contracts/multisig/src/state.rs
```diff
@@ -13,7 +13,6 @@ use crate::{
 
 #[cw_serde]
 pub struct Config {
-    pub governance: Addr,
```
cgorenflo, line 16: your missing the migration for this, aren't you?

cgorenflo (commented):
I really like the generalized kill switch, I'm not sure about the naming though. At least to me, it isn't clear if the contract is supposed to be disabled if the switch is engaged or the other way around

cgorenflo (approved):

PR #486, opened 2024-07-05
fix(ampd): reconfiguration of queued_broadcaster tests

maancham:
## Description

AXE-4335

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


ampd/src/queue/queued_broadcaster.rs
```diff
@@ -253,6 +253,108 @@ mod test {
         assert!(handle.await.unwrap().is_ok());
     }
 
+    #[test(start_paused = true)]
+    async fn should_broadcast_after_interval_in_low_load() {
```
cgorenflo, line 257: the test is doing the right thing now, but can be improved to get rid of the sleep. see if you can figure out how to use a combination of a oneshot channel https://docs.rs/tokio/latest/tokio/sync/oneshot/ and a timeout https://docs.rs/tokio/latest/tokio/time/fn.timeout.html to remove it

ampd/src/queue/queued_broadcaster.rs
```diff
@@ -253,6 +253,108 @@ mod test {
...
+                payer: None,
+            })
+        });
+        broadcaster
+            .expect_broadcast()
+            .times(3) // Expect 3 broadcasts: 2 full batches and 1 partial
+            .returning(move |msgs| {
+                assert!(msgs.len() == 9 || msgs.len() == 2);
+                Ok(TxResponse::default())
+            });
```
cgorenflo, line 333: this would let the test succeed even if all batches where size two, that's not what we want

ampd/src/queue/queued_broadcaster.rs
```diff
@@ -253,6 +253,108 @@ mod test {
...
+
+        let elapsed = start_time.elapsed();
+
+        // Assert that broadcast happened after one interval but before two
+        assert!(elapsed > interval_duration);
+        assert!(elapsed < interval_duration * 2);
+    }
+
+    #[test(start_paused = true)]
+    async fn should_broadcast_full_batches_in_high_load() {
```
cgorenflo, line 311: similar to the other test, this is a big improvement. Can you figure out the same improvement here that I suggested for the other one? Hint: this test doesn't test that all messages in the pipe are sent, just that when it sends something, the batch is full.

ampd/src/queue/queued_broadcaster.rs
```diff
@@ -227,73 +233,102 @@ mod test {
                     payer: None,
                 })
             });
+
+        let tx_clone = Arc::clone(&tx);
         broadcaster
             .expect_broadcast()
-            .once()
+            .times(1)
```
haiyizxx, line 240: what does times do?

cgorenflo (approved):

PR #489, opened 2024-07-08
feat: axelarnet gateway api

milapsheth:
## Description

[AXE-4366]

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Other Notes

[AXE-4366]: https://axelarnetwork.atlassian.net/browse/AXE-4366?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/axelarnet-gateway-api/src/msg.rs
```diff
@@ -0,0 +1,39 @@
...
+        payload: HexBinary,
+    },
+
+    // Permissioned: Can only be called by the router
+    /// Receive messages from the router and mark them as approved.
+    RouteMessages(Vec<Message>),
+
+    // Permissioned: Can be called by the address receiving the message
+    /// Validate if the message was received for the caller contract and mark as executed. The receiving contract should call this before executing a message.
+    ValidateMessage(Message),
```
cjcobb23, line 22:
```suggestion
    ValidateMessageAndMarkAsExecuted(Message),
```
or just
```suggestion
    MarkAsExecuted(Message),
```

packages/axelarnet-gateway-api/src/msg.rs
```diff
@@ -0,0 +1,39 @@
...
+        destination_address: Address,
+        payload: HexBinary,
+    },
+
+    // Permissioned: Can only be called by the router
+    /// Receive messages from the router and mark them as approved.
+    RouteMessages(Vec<Message>),
+
+    // Permissioned: Can be called by the address receiving the message
+    /// Validate if the message was received for the caller contract and mark as executed. The receiving contract should call this before executing a message.
```
cjcobb23, line 21:
```suggestion
    /// Validate if the message was received for the caller contract and mark as executed.
```

I think it's more likely the contract will actually just add this message as a submessage, which will revert the whole execution on error.

packages/axelarnet-gateway-api/src/msg.rs
```diff
@@ -0,0 +1,39 @@
...
+    #[returns(Vec<Message>)]
+    GetOutgoingMessages { message_ids: Vec<CrossChainId> },
+
+    // check if message is approved
+    #[returns(bool)]
+    IsMessageApproved { message: Message },
+
+    // check if message is executed
+    #[returns(bool)]
+    IsMessageExecuted { message_id: CrossChainId },
```
cjcobb23, line 38: Might want to just combine all of these into one query that returns a message and status

packages/axelarnet-gateway-api/Cargo.toml
cgorenflo, line 1: does this need to be it's own package instead of just a contract?

packages/axelarnet-gateway-api/src/msg.rs
```diff
@@ -0,0 +1,39 @@
+use cosmwasm_schema::{cw_serde, QueryResponses};
+use cosmwasm_std::HexBinary;
+use router_api::{Address, ChainName, CrossChainId, Message};
+
+#[cw_serde]
+pub enum ExecuteMsg {
```
cgorenflo, line 6: add permission controls

packages/interchain-token-service-api/src/abi.rs
cgorenflo, line 1: are there any changes compared to the other PR?

packages/axelarnet-gateway-api/Cargo.toml
haiyizxx: other contracts can use this client to send message?

packages/axelarnet-gateway-api/src/msg.rs
haiyizxx: what if the caller contract forgot to mark as executed? the message stays in approved and can be executed many times?

packages/axelarnet-gateway-api/Cargo.toml
cjcobb23: other contracts don't need this as a package though to use it. Package is only necessary for circular dependencies, or if there will be multiple implementations of the same API

contracts/axelarnet-gateway/src/msg.rs
```diff
@@ -0,0 +1,44 @@
...
+    #[returns(Vec<Message>)]
+    GetOutgoingMessages { message_ids: Vec<CrossChainId> },
+
+    // check if message is approved
+    #[returns(bool)]
+    IsMessageApproved { message: Message },
+
+    // check if message is executed
+    #[returns(bool)]
+    IsMessageExecuted { message_id: CrossChainId },
```
cjcobb23, line 43: I would just change this to `GetMessageStatus` and return an enum (which can be approved or executed)

PR #497, opened 2024-07-09
feat(minor-multisig)!: authorize and unauthorize vector of addresses

cjcobb23:
https://axelarnetwork.atlassian.net/browse/AXE-4360


contracts/multisig/src/msg.rs
```diff
@@ -56,12 +57,12 @@ pub enum ExecuteMsg {
...
-    #[permission(Governance)]
-    AuthorizeCaller { contract_address: Addr },
-    // Unauthorizes a contract, so it can no longer call StartSigningSession. Callable only by governance
+    // Authorizes a set of contracts to call StartSigningSession.
     #[permission(Governance)]
-    UnauthorizeCaller { contract_address: Addr },
+    AuthorizeCallers { contracts: Vec<(String, ChainName)> },
+    // Unauthorizes a set of contracts, so they can no longer call StartSigningSession.
+    #[permission(Elevated)]
+    UnauthorizeCallers { contracts: Vec<(String, ChainName)> },
```
cgorenflo, line 65: make these hashmaps instead? That means we don't have to deal with duplicates and non-rust users will have a bad time constructing these tuples again

contracts/multisig/src/msg.rs
```diff
@@ -56,12 +57,12 @@ pub enum ExecuteMsg {
         to sign their own address using the private key */
         signed_sender_address: HexBinary,
     },
-    // Authorizes a contract to call StartSigningSession. Callable only by governance
-    #[permission(Governance)]
-    AuthorizeCaller { contract_address: Addr },
-    // Unauthorizes a contract, so it can no longer call StartSigningSession. Callable only by governance
+    // Authorizes a set of contracts to call StartSigningSession.
```
cgorenflo, line 60: change all the documentation comments to `///` for the file

contracts/multisig/src/events.rs
cgorenflo, line 1: not related to this PR directly, but is there any benefit in defining the Event as an enum instead of separate structs?

cjcobb23: One thing I can think of is it forces all of the events to be defined in one spot in the file, which is easier to read. Contrast this with the events in the voting verifier: https://github.com/axelarnetwork/axelar-amplifier/blob/main/contracts/voting-verifier/src/events.rs

cgorenflo: yeah fair enough

cgorenflo (approved):

PR #498, opened 2024-07-10
feat(minor): add legacy (uppercase) chain name support

cgorenflo:


contracts/router/src/contract/execute.rs
```diff
@@ -208,19 +208,20 @@ pub fn route_messages(
...
-                }
-                Some(destination_chain) => destination_chain.gateway.address,
-                // messages with unknown destination chains are routed to
-                // the nexus gateway if the sender is not the nexus gateway
-                // itself
-                None if sender != config.nexus_gateway => config.nexus_gateway.clone(),
-                _ => return Err(report!(Error::ChainNotFound)),
-            };
+            let gateway =
+                match state::load_chain_by_chain_name(storage, &destination_chain.amplifier()?)? {
```
cjcobb23, line 212: Is this line right? Isn't the `?` going to cause this whole closure to return an error if the chain is a legacy chain, instead of routing to nexus?

packages/router-api/src/primitives.rs
```diff
@@ -245,6 +223,219 @@ impl AsRef<str> for ChainName {
...
+    pub fn amplifier(self) -> Result<ChainName, Error> {
+        match self {
+            GeneralizedChainName::Amplifier(chain) => Ok(chain),
+            GeneralizedChainName::Legacy(_) => Err(Error::InvalidChainName),
+        }
+    }
+
+    pub fn legacy(self) -> Result<LegacyChainName, Error> {
+        match self {
+            GeneralizedChainName::Amplifier(_) => Err(Error::InvalidChainName),
```
cjcobb23, line 325: I feel like this could probably use it's own error, such as `InvalidChainType` or somesuch

packages/router-api/src/primitives.rs
```diff
@@ -245,6 +223,219 @@ impl AsRef<str> for ChainName {
...
+}
+
+impl PartialEq<LegacyChainName> for GeneralizedChainName {
+    fn eq(&self, other: &LegacyChainName) -> bool {
+        match self {
+            GeneralizedChainName::Amplifier(_) => false,
+            GeneralizedChainName::Legacy(chain) => chain.eq(other),
+        }
+    }
+}
```
cjcobb23, line 393: Just curious, why do we need to define this? It looks like a default implementation to me.

packages/router-api/src/primitives.rs
```diff
@@ -245,6 +223,219 @@ impl AsRef<str> for ChainName {
...
+
+impl FromStr for GeneralizedChainName {
+    type Err = Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        ChainName::from_str(s).map_or_else(
+            |_| Ok(GeneralizedChainName::Legacy(LegacyChainName::from_str(s)?)),
+            |chain| Ok(GeneralizedChainName::Amplifier(chain)),
+        )
```
cjcobb23, line 347: So this means that a chain registered in core will actually parse to a `GeneralizedChainName::Amplifier` variant if the name is all lowercase?

cgorenflo:
we'll solve this differently

PR #505, opened 2024-07-11
feat(minor-voting-verifier): query poll by id

haiyizxx:
## Description
AXE-4400
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

### DIFF
--- a/contracts/voting-verifier/src/msg.rs
+++ b/contracts/voting-verifier/src/msg.rs
@@ -3,7 +3,7 @@ use cosmwasm_schema::{cw_serde, QueryResponses};
 use axelar_wasm_std::{
     msg_id::MessageIdFormat,
     nonempty,
-    voting::{PollId, Vote},
+    voting::{PollId, Vote, WeightedPoll},
     MajorityThreshold, VerificationStatus,
 };
 use multisig::verifier_set::VerifierSet;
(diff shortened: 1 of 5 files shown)


cjcobb23 (commented):
I feel like we should return the messages as well

cjcobb23 (commented):
also, the msg.rs file says this query is supposed to return a different type

contracts/voting-verifier/src/query.rs
```diff
@@ -48,6 +48,26 @@ pub fn message_status(
     ))
 }
 
+pub fn poll_response(deps: Deps, poll_id: PollId) -> Result<PollResponse, ContractError> {
+    let poll = POLLS.load(deps.storage, poll_id)?;
+    let data = match poll {
+        Poll::Messages(_) => {
+            PollData::Messages(poll_messages().idx.load_messages(deps.storage, poll_id)?)
```
cjcobb23, line 55: Can we add a check here that the number of messages matches the number of tallies in the `poll` object? An assert is fine.

contracts/voting-verifier/src/msg.rs
```diff
@@ -68,15 +68,20 @@ pub enum ExecuteMsg {
...
-    poll_id: PollId,
-    messages: Vec<Message>,
+pub enum PollData {
+    Messages(Vec<Message>),
+    VerifierSet(VerifierSet),
+}
+#[cw_serde]
+pub struct PollResponse {
+    pub poll: WeightedPoll,
+    pub data: PollData,
```
cjcobb23, line 78: Can we add the poll status? It's computed dynamically now and not stored in WeightedPoll

cjcobb23 (approved):
LGTM, left a few comments

cjcobb23 (approved):

PR #510, opened 2024-07-12
feat(minor-ampd): make handle sleep and max_attempts parameters configurable

maancham:
## Description
AXE-4396

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


cgorenflo (changes requested):
when you make it configurable, it should actually be configurable, i.e. bubble it up all the way so it can be set in the configuration file. Instead of passing in the parameters down the line, I think it's better to build the RetryPolicy at a higher level and pass that down

ampd/src/event_processor.rs
```diff
@@ -39,12 +39,15 @@ pub enum Error {
 /// Let the `handler` consume events from the `event_stream`. The token is checked for cancellation
 /// at the end of each consumed block or when the `event_stream` times out. If the token is cancelled or the
 /// `event_stream` is closed, the function returns
+#[allow(clippy::too_many_arguments)]
```
cgorenflo, line 42: pass in the config instead?

ampd/src/config.rs
```diff
@@ -9,15 +9,34 @@ use crate::handlers::{self, config::deserialize_handler_configs};
...
+    fn default() -> Self {
+        Self {
+            retry_timeout: Duration::from_secs(1),
+            retry_max_attempts: 3,
+            stream_timeout: Duration::from_secs(15),
+            buffer_cap: 100000,
+        }
+    }
+}
+
```
cgorenflo, line 32: move to event_processor.rs

ampd/src/config.rs
```diff
@@ -9,15 +9,34 @@ use crate::handlers::{self, config::deserialize_handler_configs};
...
 use crate::url::Url;
 
+#[derive(Debug, Deserialize, Serialize, Clone, PartialEq)]
+pub struct EventProcessorConfig {
+    #[serde(with = "humantime_serde")]
+    pub retry_timeout: Duration,
+    pub retry_max_attempts: u64,
+    #[serde(with = "humantime_serde")]
+    pub stream_timeout: Duration,
+    pub buffer_cap: usize,
```
cgorenflo, line 19:
```suggestion
    pub stream_buffer_size: usize,
```

ampd/src/config.rs
```diff
@@ -9,15 +9,34 @@ use crate::handlers::{self, config::deserialize_handler_configs};
 use crate::tofnd::Config as TofndConfig;
 use crate::url::Url;
 
+#[derive(Debug, Deserialize, Serialize, Clone, PartialEq)]
+pub struct EventProcessorConfig {
+    #[serde(with = "humantime_serde")]
+    pub retry_timeout: Duration,
```
cgorenflo, line 15:
```suggestion
    pub retry_delay: Duration,
```
compare to `stream_timeout`, where it fails if the action takes longer than the set duration. Here we wait for x amount of time before retrying

ampd/src/config.rs
```diff
@@ -9,15 +9,34 @@ use crate::handlers::{self, config::deserialize_handler_configs};
...
 #[derive(Deserialize, Serialize, Debug, PartialEq)]
 #[serde(default)]
 pub struct Config {
     pub health_check_bind_addr: SocketAddrV4,
     pub tm_jsonrpc: Url,
     pub tm_grpc: Url,
-    pub event_buffer_cap: usize,
-    #[serde(with = "humantime_serde")]
-    pub event_stream_timeout: Duration,
+    pub event_processor_config: EventProcessorConfig,
```
cgorenflo, line 39:
```suggestion
    pub event_processor: EventProcessorConfig,
```

ampd/src/config.rs
maancham: Makes sense!

ampd/src/lib.rs
```diff
@@ -353,8 +373,15 @@ where
         let broadcaster = self.broadcaster.client();
         let sub = self.event_subscriber.subscribe();
 
+        let event_config = EventProcessorConfig {
+            retry_delay,
+            retry_max_attempts,
+            stream_timeout,
+            stream_buffer_size: 10000,
+        };
```
cgorenflo, line 381: why is the config defined here?

maancham: Because we have a `consume_events` a few lines down that needs a event config as input. But `create_handler_task` receives the individual fields.

cgorenflo (approved):

PR #513, opened 2024-07-12
fix(multisig-prover): drop checksum check for destination chain

haiyizxx:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


cgorenflo (dismissed):

contracts/multisig-prover/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "multisig-prover"
-version = "0.6.0"
```
talalashraf, line 3: shouldn't have to manually change these :/

cgorenflo: we needed to for the manual release, because the release action can only work from the HEAD of main, but here this should be indeed reverted

cgorenflo (approved):

milapsheth (approved):

PR #523, opened 2024-07-19
feat(minor): add legacy (uppercase) chain name support

cgorenflo:


packages/router-api/src/primitives.rs
```diff
@@ -191,15 +183,47 @@ impl TryFrom<String> for ChainName {
...
+    fn from(value: &ChainNameRaw) -> Self {
+        ChainName(value.0.to_lowercase())
+    }
+}
+
+impl From<ChainNameRaw> for ChainName {
+    fn from(value: ChainNameRaw) -> Self {
+        ChainName(value.0.to_lowercase())
+    }
+}
```
cjcobb23, line 204: Do we need these? I feel like these could be dangerous.

contracts/gateway/src/state.rs
```diff
@@ -53,7 +53,7 @@ pub(crate) enum Error {
 const CONFIG_NAME: &str = "config";
 const CONFIG: Item<Config> = Item::new(CONFIG_NAME);
 const OUTGOING_MESSAGES_NAME: &str = "outgoing_messages";
-const OUTGOING_MESSAGES: Map<CrossChainId, Message> = Map::new(OUTGOING_MESSAGES_NAME);
+const OUTGOING_MESSAGES: Map<&CrossChainId, Message> = Map::new(OUTGOING_MESSAGES_NAME);
```
cjcobb23, line 56: will this need a migration?

cjcobb23 (approved):
We should probably add tests somewhere that ensure routing works as we expect:

1. Mixed case chain names from core are preserved.
2. Cannot route two different versions of the same message by just changing the case of the chain name.

Not sure where these should go though.

contracts/gateway/src/state.rs
cgorenflo: no, the data is identical

PR #525, opened 2024-07-19
feat(multisig-prover): bcs payload encoding

fish-sammy:
## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


ampd/src/sui/verifier.rs
```diff
@@ -133,20 +73,21 @@ impl PartialEq<&VerifierSetConfirmation> for &SuiEvent {
...
             Ok(SignersRotated {
                 signers:
                     WeightedSigners {
                         mut signers,
                         threshold,
                         nonce,
                     },
+                ..
             }) => {
                 signers.sort();
```
milapsheth, line 86: they're expected to be in sorted order already in the event. It would be concerning if they're not. Perhaps avoid this?

ampd/src/sui/verifier.rs
```diff
@@ -133,20 +73,21 @@ impl PartialEq<&VerifierSetConfirmation> for &SuiEvent {
             .chain(expected.created_at.to_be_bytes())
```
milapsheth, line 73: Use sort_by for expected_signers to be explicit?

ampd/src/sui/verifier.rs
```diff
@@ -492,26 +431,28 @@ mod tests {
...
+                    .collect(),
+                threshold: threshold.u128(),
+                nonce: <[u8; 32]>::try_from(
+                    [0u8; 24]
+                        .into_iter()
+                        .chain(created_at.to_be_bytes())
+                        .collect::<Vec<_>>(),
+                )
+                .unwrap()
+                .into(),
```
milapsheth, line 453:
```suggestion
                nonce: U256::from(created_at).to_be_bytes(),
```

contracts/multisig-prover/src/encoding/mod.rs
```diff
@@ -1,10 +1,92 @@
...
+use crate::payload::Payload;
 
 #[cw_serde]
 #[derive(Copy)]
 pub enum Encoder {
     Abi,
     Bcs,
 }
+
+impl Encoder {
```
milapsheth, line 22: I think we should create an `Encoder` trait instead that AbiEncoder and BcsEncoder types implement.
`EncoderType` enum can be used to denote which encoders are supported, and returns the appropriate encoders. Avoids having to match cases for each method below then

contracts/multisig-prover/src/encoding/mod.rs
```diff
@@ -1,10 +1,92 @@
...
+            signer
+        })
+        .collect()
+}
+
+fn add_27(recovery_byte: k256::ecdsa::RecoveryId) -> u8 {
+    recovery_byte
+        .to_byte()
+        .checked_add(27)
+        .expect("overflow when adding 27 to recovery byte")
```
milapsheth, line 87: return errors instead?

contracts/multisig-prover/src/encoding/bcs.rs
```diff
@@ -0,0 +1,284 @@
...
+fn encode_payload(payload: &Payload) -> Result<Vec<u8>, ContractError> {
+    let encoded: Vec<u8> = match payload {
+        Payload::Messages(messages) => bcs::to_bytes(
+            &messages
+                .iter()
+                .map(Message::try_from)
+                .collect::<Result<Vec<_>, _>>()
+                .change_context(ContractError::InvalidMessage)?,
+        )
+        .expect("failed to serialize messages"),
```
milapsheth, line 27: Convert expects to errors

contracts/multisig-prover/src/encoding/bcs.rs
```diff
@@ -0,0 +1,284 @@
...
+                    },
+                ),
+            ]
+            .into_iter()
+            .collect(),
+            threshold: 2u128.into(),
+            created_at: 2025,
+        });
+
+        assert_eq!(
```
milapsheth, line 185: use goldie::assert_json! for testing against golden files (you can refer the existing usage)

contracts/multisig-prover/src/encoding/bcs.rs
```diff
@@ -0,0 +1,284 @@
...
+
+pub fn payload_digest(
+    domain_separator: &Hash,
+    verifier_set: &VerifierSet,
+    payload: &Payload,
+) -> Result<Hash, ContractError> {
+    let command_type = match payload {
+        Payload::Messages(_) => COMMAND_TYPE_APPROVE_MESSAGES,
+        Payload::VerifierSet(_) => COMMAND_TYPE_ROTATE_SIGNERS,
+    };
```
milapsheth, line 45:
```suggestion
    let command_type = payload.variant_to_u8();
```
perhaps? You can define a Sui `CommandType` and implement a `From<Payload> for CommandType` that uses `variant_to_u8` internally.

contracts/multisig-prover/src/encoding/bcs.rs
```diff
@@ -0,0 +1,284 @@
...
+    let msg = MessageToSign {
+        domain_separator: (*domain_separator).into(),
+        signers_hash: WeightedSigners::from(verifier_set.clone()).hash().into(),
+        data_hash: <[u8; 32]>::from(Keccak256::digest(data)).into(),
+    };
+
+    Ok(msg.hash())
+}
+
+pub fn encode_execute_data(
```
milapsheth, line 58: Document that BCS encoded execute data returns payload and proof in a certain format so it's clearer for anyone who queries on how to decode

milapsheth (changes requested):

ampd/src/sui/verifier.rs
fish-sammy: Don't think that's relevant to the context of PartialEq. That's external knowledge that this function should not have known. Testing for that ordering is of the job of the contracts.

ampd/src/sui/verifier.rs
fish-sammy: Why though? For the context of this method, it doesn't care about the "correct" order nor does it know it, it only wants to compare equality of two vectors.

contracts/multisig-prover/src/encoding/bcs.rs
fish-sammy: Sure

fish-sammy: I don't know about using that `variant_to_u8` though. That's like coupling the u8 we use in amlifier contracts with the command type thing we use in the Sui contracts. I can make it an enum and do the From trait but I wouldn't wanna use that `variant_to_u8`.

contracts/multisig-prover/src/encoding/bcs.rs
fish-sammy: Where do I document that?

packages/sui-gateway/src/gateway/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+use serde::{Deserialize, Serialize};
+use sha3::{Digest, Keccak256};
+
+use crate::base_types::SuiAddress;
+use crate::error::Error;
+
+pub mod events;
+
+pub const COMMAND_TYPE_APPROVE_MESSAGES: u8 = 0;
+pub const COMMAND_TYPE_ROTATE_SIGNERS: u8 = 1;
```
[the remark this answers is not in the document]
fish-sammy: sure

packages/sui-gateway/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "sui-gateway"
+version = "0.1.0"
+edition = "2021"
+rust-version = { workspace = true }
+
```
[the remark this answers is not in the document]
fish-sammy: I don't see the difference?

packages/sui-gateway/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+cosmwasm-std = { workspace = true }
+error-stack = { workspace = true }
+hex = "0.4.3"
+multisig = { workspace = true, features = ["library"] }
+router-api = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
+sha3 = { workspace = true }
+thiserror = { workspace = true }
+
```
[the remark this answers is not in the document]
fish-sammy: I think due to merging from main this is off. Which do you mean here?

packages/sui-gateway/src/gateway/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+                pub_key: HexBinary::from(signer.pub_key.clone()).to_vec(),
+                weight: signer.weight.into(),
+            })
+            .collect();
+        signers.sort_by_key(|signer| signer.pub_key.clone());
+
+        let nonce = [0u8; 24]
+            .into_iter()
+            .chain(verifier_set.created_at.to_be_bytes())
+            .collect::<Vec<_>>();
```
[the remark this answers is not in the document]
fish-sammy: oh nice didn't think of this way of doing it. clean

packages/sui-gateway/src/gateway/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+    pub nonce: Bytes32,
+}
+
+impl From<VerifierSet> for WeightedSigners {
+    fn from(verifier_set: VerifierSet) -> Self {
+        let mut signers: Vec<_> = verifier_set
+            .signers
+            .values()
+            .map(|signer| WeightedSigner {
+                pub_key: HexBinary::from(signer.pub_key.clone()).to_vec(),
```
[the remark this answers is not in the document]
fish-sammy: This one makes sense.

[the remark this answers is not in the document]
fish-sammy: I don't think the Ethereum side is doing this though.

contracts/multisig-prover/src/encoding/mod.rs
fish-sammy: Not sure if that is better and I also want to avoid making this PR more complex and bigger for some refactor that we can do later.

contracts/multisig-prover/src/encoding/mod.rs
fish-sammy: According to the `RecoveryId` doc, this can only be 0, 1, 2 and 3. I'd prefer a panic here cuz that should never really happen.

contracts/multisig-prover/src/encoding/bcs.rs
fish-sammy: I am also not sure about this. Why?

packages/sui-gateway/src/gateway/mod.rs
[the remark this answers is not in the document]
fish-sammy: done

packages/sui-gateway/src/gateway/mod.rs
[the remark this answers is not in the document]
fish-sammy: done

packages/sui-gateway/src/gateway/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+            .into_iter()
+            .chain(verifier_set.created_at.to_be_bytes())
+            .collect::<Vec<_>>();
+
+        Self {
+            signers,
+            threshold: verifier_set.threshold.into(),
+            nonce: <[u8; 32]>::try_from(nonce)
+                .expect("nonce must be 32 bytes")
+                .into(),
```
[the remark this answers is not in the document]
fish-sammy: done

packages/sui-gateway/src/gateway/mod.rs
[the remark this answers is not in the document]
fish-sammy: done

packages/sui-gateway/src/gateway/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+    fn from(verifier_set: VerifierSet) -> Self {
+        let mut signers: Vec<_> = verifier_set
+            .signers
+            .values()
+            .map(|signer| WeightedSigner {
+                pub_key: HexBinary::from(signer.pub_key.clone()).to_vec(),
+                weight: signer.weight.into(),
+            })
+            .collect();
+        signers.sort_by_key(|signer| signer.pub_key.clone());
```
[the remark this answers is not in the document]
fish-sammy: done

packages/sui-gateway/src/gateway/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+    }
+}
+
+#[derive(Serialize)]
+pub struct Signature {
+    bytes: Vec<u8>,
+}
+
+impl From<multisig::key::Signature> for Signature {
+    fn from(signature: multisig::key::Signature) -> Self {
```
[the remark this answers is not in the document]
fish-sammy: done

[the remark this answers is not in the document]
fish-sammy: Again, I do not think Ethereum is doing that.

packages/sui-gateway/src/gateway/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+
+#[derive(Serialize)]
+pub struct Proof {
+    signers: WeightedSigners,
+    signatures: Vec<Signature>,
+}
+
+impl From<(VerifierSet, Vec<SignerWithSig>)> for Proof {
+    fn from((verifier_set, mut signatures): (VerifierSet, Vec<SignerWithSig>)) -> Self {
+        signatures.sort_by_key(|signer| signer.signer.pub_key.clone());
```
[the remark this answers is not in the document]
fish-sammy: done

packages/sui-gateway/src/base_types.rs
```diff
@@ -0,0 +1,40 @@
...
+            .map(Self)
+            .change_context(Error::InvalidAddressBytes(bytes.to_vec()))
+    }
+}
+
+impl FromStr for SuiAddress {
+    type Err = Report<Error>;
+
+    fn from_str(s: &str) -> Result<Self, Error> {
+        hex::decode(s.trim_start_matches(ADDRESS_PREFIX))
```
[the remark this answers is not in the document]
fish-sammy: done

ampd/src/sui/verifier.rs
fish-sammy: done

packages/sui-gateway/src/gateway/events.rs
```diff
@@ -0,0 +1,41 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::{ContractCall, SignersRotated};
+
+    #[test]
+    fn signers_rotated_should_serialize_and_deserialize_correct_bcs() {
+        let bcs_bytes = bs58::decode("7VmsFWvv3atGxVB1oUDZAx6nRWoE1xiWYxqLP6mcWD3Rf4wFjhfMwjKmxjXHyVc9ZTdCRdt25rUCWxCBRGEuFogKfgXkUXSkHPxirMeS4R8kJn1pgA7bkMwQywgFnrmd1ceeWNCCBQuUGr83Gx8mbyZeN7W2EAct14rKQ5zukNZH6mKv7n4dmBasFm8vT").into_vec().unwrap();
+
+        let event = bcs::from_bytes::<SignersRotated>(&bcs_bytes).unwrap();
```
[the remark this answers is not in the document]
fish-sammy: done

packages/sui-gateway/src/base_types.rs
```diff
@@ -0,0 +1,40 @@
...
+            .try_into()
+            .map(Self)
+            .change_context(Error::InvalidAddressBytes(bytes.to_vec()))
+    }
+}
+
+impl FromStr for SuiAddress {
+    type Err = Report<Error>;
+
+    fn from_str(s: &str) -> Result<Self, Error> {
```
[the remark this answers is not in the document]
fish-sammy: done

contracts/multisig-prover/src/encoding/bcs.rs
fish-sammy: done

packages/sui-gateway/Cargo.toml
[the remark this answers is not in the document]
milapsheth: yeah, not sure what happened here

packages/sui-gateway/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "sui-gateway"
+version = "1.0.0"
+edition = "2021"
```
milapsheth, line 4:
```suggestion
edition = { workspace = true }
```

milapsheth: switched this on main recently

packages/sui-gateway/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+rust-version = { workspace = true }
+
+[dependencies]
+bcs = "0.1.5"
```
milapsheth, line 8: add this to the workspace Cargo.toml since it's used by ampd?

packages/sui-gateway/Cargo.toml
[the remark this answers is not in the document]
milapsheth: yeah, reposted in the correct place

packages/sui-gateway/src/base_types.rs
```diff
@@ -0,0 +1,40 @@
...
+use error_stack::{Report, Result, ResultExt};
+use serde::{Deserialize, Serialize};
+
+use crate::error::Error;
+
+const ADDRESS_PREFIX: &str = "0x";
+const SUI_ADDRESS_LENGTH: usize = 32;
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct SuiAddress([u8; SUI_ADDRESS_LENGTH]);
```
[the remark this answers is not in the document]
milapsheth: Sure, it'll be needed for voting to work

packages/sui-gateway/src/base_types.rs
```diff
@@ -0,0 +1,94 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use rand::RngCore;
+
+    use super::*;
+
+    #[test]
+    fn test_sui_address_try_from() {
```
milapsheth, line 52: test_ prefix isn't needed i think
```suggestion
    fn sui_address_try_from() {
```

contracts/multisig-prover/src/encoding/mod.rs
milapsheth: Ok, worth doing it sooner rather than later still since we'll be adding Stellar as well

contracts/multisig-prover/src/encoding/mod.rs
milapsheth: sure, i was thinking that we were trying to use errors everywhere, but panic is fine here

contracts/multisig-prover/src/encoding/mod.rs
```diff
@@ -1,10 +1,92 @@
...
+    M: AsRef<[u8]>,
+{
+    let recovery_transform = match encoder {
+        Encoder::Abi => add_27,
+        Encoder::Bcs => no_op,
+    };
+    signers
+        .into_iter()
+        .map(|mut signer| {
+            if let Signature::Ecdsa(nonrecoverable) = signer.signature {
```
milapsheth, line 71: Panic if signature is ed25519? From what I understand since pub keys are registered as ecdsa type, this should never occur

packages/sui-gateway/src/gateway/mod.rs
```diff
@@ -0,0 +1,178 @@
...
+pub struct Signature {
+    bytes: Vec<u8>,
+}
+
+impl TryFrom<multisig::key::Signature> for Signature {
+    type Error = Report<Error>;
+
+    fn try_from(signature: multisig::key::Signature) -> Result<Self, Self::Error> {
+        match signature {
+            multisig::key::Signature::EcdsaRecoverable(signature) => Ok(Self {
```
milapsheth, line 135: Add a comment that this is only being called during encoding after signatures have been normalized to EcdsaRecoverable

milapsheth: It's not clear when just looking at this implementation this invariant holds. Maybe it should support converting from both

milapsheth (approved):

packages/sui-gateway/src/gateway/mod.rs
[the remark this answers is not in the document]
milapsheth: looks like it's being done here for EVM
https://github.com/axelarnetwork/axelar-amplifier/blob/main/packages/evm-gateway/src/lib.rs#L135

ampd/src/sui/verifier.rs
milapsheth: Hmm, while it wouldn't matter for a generic PartialEq implementation, in this case we're trying to validate the event. It would be concerning if the event order isn't correct since the verifier set hash computation changes between internal/external gateway.  Maybe the hash should be checked explicitly. This is more of a sanity check I suppose.

contracts/multisig-prover/src/encoding/bcs.rs
milapsheth: Sure, could still wrap inside a From<Payload> for CommandType implementation

contracts/multisig-prover/src/encoding/bcs.rs
milapsheth: On this method, and/or a README in this folder?

packages/sui-gateway/src/base_types.rs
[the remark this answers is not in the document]
fish-sammy: https://axelarnetwork.atlassian.net/jira/software/c/projects/AXE/boards/18?selectedIssue=AXE-4629

contracts/multisig-prover/src/encoding/mod.rs
fish-sammy: dont't think this should be a panic. Panic should only happen with local knowledge.

packages/sui-gateway/src/gateway/mod.rs
fish-sammy: Let me see how hard it is to make it work for both. Kinda like that idea.

contracts/multisig-prover/src/encoding/mod.rs
fish-sammy: hmmm I didn't touch this piece of code but just moved it though. Does `ed25519` have this property of being recoverable?

packages/sui-gateway/src/gateway/mod.rs
fish-sammy: Ok don't think handling both is realistic here. It requires the pub key and msg to convert from non-recoverable to recoverable. I'll add the comment.

contracts/multisig-prover/src/encoding/mod.rs
milapsheth: No it doesn't support recovery. Let's revisit the check after the Stellar integration, we should clean it up

PR #526, opened 2024-07-21
feat(multisig-prover): add permission control to multisig-prover

cgorenflo:
AXE-4487


contracts/multisig-prover/src/contract.rs
```diff
@@ -133,17 +116,21 @@ pub fn migrate(
 ) -> Result<Response, axelar_wasm_std::ContractError> {
     cw2::set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
```
haiyizxx, line 117: version should be set after migration

haiyizxx (approved):

contracts/multisig-prover/src/contract.rs
cgorenflo: thanks

PR #527, opened 2024-07-22
feat: return structured errors from contracts

cgorenflo:
instead of flattened strings, the errors now preserve the error stack


packages/report/src/loggable.rs
```diff
@@ -81,7 +81,10 @@ impl<T> From<&Report<T>> for LoggableError {
             // because of the stack order of attachments we need to reverse them to get the historical order
             attachments.reverse();
             error.attachments = attachments;
-            errors.push(error)
+
+            if !error.msg.is_empty() || !error.attachments.is_empty() {
+                errors.push(error)
```
haiyizxx, line 86: it will skip backtrace?

haiyizxx (approved):

packages/report/src/loggable.rs
cgorenflo: this is a minor hack. The only way I could make `ContractError` work with typed `Report`s was to wrap the error with an "invisible" context, i.e. an error without message or attachments. I don't want that to be printed, because then every error would start with a `:` without message. So I added this exception, which I think is reasonable. If an error neither has a message nor attachments it'll be skipped

PR #528, opened 2024-07-22
feat: add only tx hash message id

raress96:
Added only tx hash message id, considering event index as 0 in events.


packages/axelar-wasm-std/src/msg_id/tx_hash.rs
```diff
@@ -0,0 +1,128 @@
...
+    where
+        Self: Sized,
+    {
+        // the PATTERN has exactly two capture groups, so the groups can be extracted safely
+        REGEX
+            .captures(message_id)
+            .ok_or(Error::InvalidMessageID {
+                id: message_id.to_string(),
+                expected_format: PATTERN.to_string(),
+            })?;
```
cgorenflo, line 50:
```suggestion
        ensure!(
            REGEX.is_match(message_id),
            Error::InvalidMessageID {
                id: message_id.to_string(),
                expected_format: PATTERN.to_string(),
            }
        );
```

cgorenflo: in your case you don't care about the captures, just that the whole string matches the pattern

cgorenflo: `ensure!` is in the `error_stack` crate

cgorenflo (approved):
just some minor suggestions for cleaner code

packages/axelar-wasm-std/src/msg_id/tx_hash.rs
```diff
@@ -0,0 +1,128 @@
...
+use crate::hash::Hash;
+use crate::nonempty;
+
+pub struct HexTxHash {
+    pub tx_hash: Hash,
+}
+
+impl HexTxHash {
+    pub fn tx_hash_as_hex(&self) -> nonempty::String {
+        format!("{}", HexBinary::from(self.tx_hash).to_hex())
```
k4m4, line 20:
```suggestion
        HexBinary::from(self.tx_hash).to_hex()
```
`format` should be redundant here.

packages/axelar-wasm-std/src/msg_id/tx_hash.rs
```diff
@@ -0,0 +1,128 @@
...
+        assert!(res.is_err());
+    }
+
+    #[test]
+    fn should_not_parse_msg_id_with_non_hex_tx_hash() {
+        let msg_id = "82GKYvWv5EKm7jnYksHoh3u5M2RxHN2boPreM8Df4ej9";
+        let res = HexTxHash::from_str(msg_id);
+        assert!(res.is_err());
+    }
+}
```
k4m4, line 128:
```suggestion
    #[test]
    fn should_not_parse_msg_id_with_0x_prefix() {
        let msg_id = "0x7cedbb3799cd99636045c84c5c55aef8a138f107ac8ba53a08cad1070ba4385b-1";
        let res = HexTxHash::from_str(msg_id);
        assert!(res.is_err());
    }

    #[test]
    fn should_not_parse_msg_id_with_event_index() {
        let tx_hash = random_hash();
        let res = HexTxHash::from_str(&format!("{}-1", tx_hash));
        assert!(res.is_err());
    }
}
```

Just copy-pasted from our duplicate PR: https://github.com/axelarnetwork/axelar-amplifier/pull/532/files#diff-21e519f3544fb0dc94d4d87daef8ee69a1bd012bc43723dd93417cf95284680bR128-R141.

packages/axelar-wasm-std/src/msg_id/tx_hash.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    use super::*;
+
+    fn random_hash() -> String {
+        let mut bytes = vec![];
+        for _ in 0..32 {
+            let byte: u8 = rand::random();
+            bytes.push(byte)
+        }
+        format!("{}", HexBinary::from(bytes).to_hex())
```
k4m4, line 82:
```suggestion
        HexBinary::from(bytes).to_hex()
```
Redundant `format`.

contracts/voting-verifier/src/events.rs
```diff
@@ -143,6 +143,12 @@ fn parse_message_id(
 
             Ok((id.signature_as_base58(), id.event_index))
         }
+        MessageIdFormat::HexTxHash => {
+            let id = HexTxHash::from_str(&message_id)
+                .map_err(|_| ContractError::InvalidMessageID(message_id.into()))?;
+
+            Ok((id.tx_hash_as_hex(), 0))
```
cjcobb23, line 150: this seems a bit odd to put 0 as the event index here. We should probably refactor the event a bit to not need to include an event index here. Maybe the `TxEventConfirmation` event should have a `tx_info` field that is an enum, which can be different types.

cjcobb23 (changes requested):

contracts/voting-verifier/src/events.rs
raress96: That would mean also refactoring ampd and other parts, I don't think that is in the scope of this PR.

packages/axelar-wasm-std/src/msg_id/tx_hash.rs
raress96: Done

packages/axelar-wasm-std/src/msg_id/tx_hash.rs
raress96: Done

packages/axelar-wasm-std/src/msg_id/tx_hash.rs
raress96: Done

packages/axelar-wasm-std/src/msg_id/tx_hash.rs
raress96: Done

contracts/voting-verifier/src/events.rs
cjcobb23: I think that change is worth doing, but would you prefer to do it in a follow up PR?

cjcobb23 (approved):
Looks good to me, but please follow up with another PR about the event refactor when you can

packages/axelar-wasm-std/src/msg_id/tx_hash.rs
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    pub fn new(tx_id: impl Into<[u8; 32]>) -> Self {
+        Self {
+            tx_hash: tx_id.into(),
+        }
+    }
+}
+
+const PATTERN: &str = "^[0-9a-f]{64}$";
```
cjcobb23, line 33:
```suggestion
const PATTERN: &str = "^0x[0-9a-f]{64}$";
```
this should start with 0x, sorry I missed this the first time.

The existing tx hash + event index message id starts with 0x, so we should be consistent

cjcobb23 (changes requested):

packages/axelar-wasm-std/src/msg_id/tx_hash.rs
raress96: Done

cjcobb23 (approved):

PR #530, opened 2024-07-22
feat(nexus-gateway): add permission control to nexus-gateway

cgorenflo:
AXE-4498


contracts/nexus-gateway/src/contract.rs
```diff
@@ -102,4 +95,228 @@ mod tests {
...
+            deps,
+            mock_env(),
+            mock_info("admin", &[]),
+            InstantiateMsg {
+                nexus: NEXUS.to_string(),
+                router: ROUTER.to_string(),
+            },
+        )
+        .unwrap();
+    }
```
cgorenflo, line 321: moved from the execute file

contracts/nexus-gateway/src/contract/execute.rs
cgorenflo, line 1: removed Contract and Store because they interfered with `ensure_permissions`

packages/axelar-wasm-std/src/error.rs
```diff
@@ -76,3 +76,13 @@ pub fn extend_err<T, E: Context>(
...
+
+#[macro_export]
+macro_rules! err_contains {
+    ($expression:expr, $error_type:ty, $pattern:pat $(if $guard:expr)? $(,)?) => {
+        match $expression.downcast_ref::<$error_type>() {
+            Some($pattern) $(if $guard)? => true,
+            _ => false,
+        }
+    };
+}
```
cgorenflo, line 88: small helper to make it easier to test for specific error patterns. Example: 
```
assert!(res.is_err_and(|err| err_contains!(
            err.report,
            permission_control::Error,
            permission_control::Error::AddressNotWhitelisted { .. }
        )));
```

packages/msgs-derive/src/lib.rs
cgorenflo, line 1: the nexus gateway msg has only specific checks, and no general permission checks. Because of that, generated code after line 266 is (correctly) not reachable and clippy complains. These changes get around that by using `?` instead of `return Err(..)`

haiyizxx (approved):

PR #532, opened 2024-07-23
feat: add support for XRPL message ID

k4m4:
## Description

This PR adds support for a custom XRPL message ID, which comprises of just a 32-byte hash.
To initiate bridging from XRPL, a [`Payment`](https://xrpl.org/docs/references/protocol/transactions/types/payment/) transaction is made into an XRPL multisig (which can be though of as a replacement for the `AxelarGateway`).
Since XRPL does not support batched transactions, there can only be one bridging transaction per XRPL transaction, so an event index is redundant.

## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Other Notes

We do not have full context of all the moving pieces that this PR affects. Please review and let us know if you see any issues or recommend any changes.


k4m4:
Closing since this seems to be an almost exact duplicate of #528.

PR #538, opened 2024-07-24
refactor(minor): rename id field in CrosschainId to message_id

cgorenflo:


cgorenflo:
todo: migrate data that has a crosschain id in it

packages/router-api/src/primitives.rs
```diff
@@ -103,7 +103,7 @@ impl TryFrom<String> for Address {
 #[derive(Eq, Hash)]
 pub struct CrossChainId {
     pub chain: ChainNameRaw,
```
cjcobb23, line 105: do you think we should rename this to `source_chain` while we are here?

contracts/gateway/src/contract/migrations/v0_2_3.rs
```diff
@@ -0,0 +1,221 @@
...
+
+use crate::contract::{CONTRACT_NAME, CONTRACT_VERSION};
+use crate::state;
+
+const BASE_VERSION: &str = "0.2.3";
+
+pub fn migrate(storage: &mut dyn Storage) -> Result<(), ContractError> {
+    cw2::assert_contract_version(storage, CONTRACT_NAME, BASE_VERSION)?;
+
+    migrate_outgoing_messages(storage)?;
```
cjcobb23, line 21: you could also just drop the messages as well, they can always be reverifier and rerouted

cjcobb23 (approved):

PR #544, opened 2024-07-25
fix(multisig-prover): better error message on chain name mismatch

cjcobb23:


contracts/multisig-prover/src/contract/execute.rs
```diff
@@ -98,7 +98,9 @@ fn messages(
         .iter()
         .any(|msg| msg.destination_chain != chain_name)
     {
-        panic!("violated invariant: messages from different chain found");
+        panic!(
+            "violated invariant: destination_chain in message does not match configured chain name"
```
haiyizxx, line 102: include the configured chain name?

haiyizxx (approved):

contracts/multisig-prover/src/error.rs
```diff
@@ -61,4 +63,10 @@ pub enum ContractError {
 
     #[error("not enough verifiers")]
     NotEnoughVerifiers,
+
+    #[error("invalid destination chain '{actual}', expected '{expected}")]
```
haiyizxx, line 67:
```suggestion
    #[error("invalid destination chain '{actual}', expected '{expected}'")]
```

PR #553, opened 2024-07-31
fix(minor-multisig-prover): generate payload id uniquely

milapsheth:
## Description

[AXE-4571]

## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Other Notes

[AXE-4571]: https://axelarnetwork.atlassian.net/browse/AXE-4571?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/multisig-prover/src/payload.rs
```diff
@@ -19,16 +19,30 @@ pub enum Payload {
...
-                message_ids.as_slice().into()
+                Keccak256::digest(message_ids.join(&FIELD_DELIMITER.to_string()).as_bytes())
+                    .to_vec()
             }
-            Payload::VerifierSet(verifier_set) => verifier_set.hash().as_slice().into(),
+            Payload::VerifierSet(verifier_set) => verifier_set.hash().to_vec(),
+        };
+
+        let mut id = vec![self.variant_to_u8()];
+        id.extend(hash);
```
cjcobb23, line 37: shouldn't this be added to the value before it's hashed? Now the PayloadId is 33 bytes right?

contracts/multisig-prover/src/contract/execute.rs
```diff
@@ -34,18 +34,22 @@ pub fn construct_proof(
...
+    let payload_id = payload.id();
+
+    match PAYLOAD
         .may_load(deps.storage, &payload_id)
         .map_err(ContractError::from)?
     {
-        Some(payload) => payload,
+        Some(stored_payload) => {
+            if stored_payload != payload {
+                return Err(report!(ContractError::PayloadMismatch));
```
cjcobb23, line 46: might be useful to return the original payload here

cjcobb23: the stored payload I mean

milapsheth: attached as a string, since implementing Display for Payload is trickier

contracts/multisig-prover/src/payload.rs
milapsheth: yeah, switched to double hashing instead since we decided to preserve 32 byte length

cjcobb23 (approved):

PR #555, opened 2024-07-31
fix(service-registry): require verifier addresses to be 20 bytes

milapsheth:
## Description

[AXE-4586]

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Other Notes

[AXE-4586]: https://axelarnetwork.atlassian.net/browse/AXE-4586?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/service-registry/src/contract.rs
+++ b/contracts/service-registry/src/contract.rs
@@ -70,7 +70,7 @@ pub fn execute(
         } => {
             let verifiers = verifiers
                 .into_iter()
-                .map(|verifier| deps.api.addr_validate(&verifier))
+                .map(|verifier| validate_verifier_address(&deps, verifier))
                 .collect::<Result<Vec<_>, _>>()?;
             execute::update_verifier_authorization_status(
                 deps,
@@ -85,7 +85,7 @@ pub fn execute(
         } => {
             let verifiers = verifiers
                 .into_iter()
-                .map(|verifier| validate_verifier_address(&deps, verifier))
+                .map(|verifier| deps.api.addr_validate(&verifier))
                 .collect::<Result<Vec<_>, _>>()?;
             execute::update_verifier_authorization_status(
                 deps,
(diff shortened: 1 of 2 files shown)


cgorenflo (commented):
I really don't like fixing the length to something that is only an implicit convention not under our control

contracts/service-registry/src/contract.rs
```diff
@@ -67,7 +70,7 @@ pub fn execute(
         } => {
             let verifiers = verifiers
                 .into_iter()
-                .map(|veriier| deps.api.addr_validate(&veriier))
+                .map(|verifier| deps.api.addr_validate(&verifier))
```
cjcobb23, line 73: shouldn't this be 
```suggestion
                .map(|verifier| validate_verifier_address(&deps, verifier))
```

cjcobb23 (commented):
Should we check the addresses passed to other messages as well? Also, should we block bonding if the address is not the right length?

milapsheth:
> I really don't like fixing the length to something that is only an implicit convention not under our control

Verifiers are expected to generate their addresses via ampd/tofnd which uses 20 bytes as well. I think it's fine to require it for verifiers specifically (I'd definitely be concerned about doing this for any cosmos address)

cgorenflo (changes requested):
If we absolutely MUST do this, we should introduce a Verifier type and back the checks into that. Spreading disconnected checks throughout the codebase that leave to guarantee on the object that is actually correct makes the code brittle and susceptible to bugs

milapsheth:
> If we absolutely MUST do this, we should introduce a Verifier type and bake the checks into that. Spreading disconnected checks throughout the codebase that leave to guarantee on the object that is actually correct makes the code brittle and susceptible to bugs

Yes, we should definitely introduce a dedicated Verifier type. Although, that's a big change to make everywhere atm. I can introduce the type in this PR but only use it for AuthorizeVerifiers for now, which introduces a verifier into the Amplifier system. We should migrate to that everywhere else in a separate task.

PR #566, opened 2024-08-01
refactor: add derive feature to axelar-wasm-std

cgorenflo:


contracts/coordinator/Cargo.toml
```diff
@@ -33,8 +33,7 @@ optimize = """docker run --rm -v "$(pwd)":/code \
 """
 
 [dependencies]
-axelar-wasm-std = { workspace = true }
-axelar-wasm-std-derive = { workspace = true }
+axelar-wasm-std = { workspace = true, features = ["derive"] }
```
milapsheth, line 36: perhaps add this feature to the entry in the top level `Cargo.toml` since it's used pretty much everywhere

cgorenflo: features are purely additive, so if it’s enabled in the workspace it’s not possible to disable it anywhere else. Therefore, I’ll keep it as is

milapsheth (approved):

PR #570, opened 2024-08-06
feat(axelarnet-gateway): add axelarnet gateway

milapsheth:
## Description

[AXE-4184]

- [x] Add more tests

## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Other Notes

[AXE-4184]: https://axelarnetwork.atlassian.net/browse/AXE-4184?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/axelarnet-gateway/src/contract/execute.rs
```diff
@@ -0,0 +1,135 @@
...
+    state::update_message_status(store, cc_id.clone(), message.clone())
+        .change_context(Error::MessageStatusUpdateFailed(cc_id))?;
+
+    let destination_contract = api
+        .addr_validate(&message.destination_address)
+        .change_context(Error::InvalidAddress(
+            message.destination_address.to_string(),
+        ))?;
+
+    // Apps are required to expose ExecuteMsg::Execute interface
```
milapsheth, line 120: App contracts need to expose the same Execute method which is allowed to be called by the gateway. I'll create an AxelarExecutable client to wrap this, although enforcing that this exists on the app would likely require using a macro.

contracts/axelarnet-gateway/src/msg.rs
```diff
@@ -0,0 +1,48 @@
...
+    /// Messages initiated via `CallContract` can be forwarded again to the router.
+    /// If the messages are coming from the router, then they are marked ready for execution.
+    #[permission(Any)]
+    RouteMessages(Vec<Message>),
+
+    /// Execute the message at the destination contract with the corresponding payload.
+    #[permission(Any)]
+    Execute {
+        message: Message,
+        payload: HexBinary,
```
milapsheth, line 38: Thoughts on flattening the message struct here? Avoids the need to validate that payload hash matches what's in the message struct. Another option is to only pass in cc_id and payload and query the message from the state. Deviates a bit from the Axelar executable API, but it's cleaner

contracts/axelarnet-gateway/src/contract/execute.rs
```diff
@@ -0,0 +1,134 @@
...
+use sha3::{Digest, Keccak256};
+
+use crate::contract::Error;
+use crate::events::AxelarnetGatewayEvent;
+use crate::msg::AxelarExecutableExecuteMsg;
+use crate::state::{self};
+
+// TODO: Retrieve the actual tx hash from core, since cosmwasm doesn't provide it. Use a placeholder in the meantime.
+const PLACEHOLDER_TX_HASH: &str =
+    "0x0000000000000000000000000000000000000000000000000000000000000000";
```
cjcobb23, line 15: Idk how I feel about this. What's the purpose of adding placeholder tx hash? Why not just use a counter by itself, or blockhash and counter?

contracts/axelarnet-gateway/src/contract/execute.rs
```diff
@@ -0,0 +1,134 @@
...
+    destination_chain: ChainName,
+    destination_address: Address,
+    payload: HexBinary,
+) -> Result<Response, Error> {
+    let counter =
+        state::increment_message_counter(store).change_context(Error::InvalidStoreAccess)?;
+
+    let cc_id = CrossChainId {
+        source_chain: chain_name.into(),
+        message_id: nonempty::String::try_from(format!("{0}-{1}", PLACEHOLDER_TX_HASH, counter))
```
cjcobb23, line 31: Use the actual message id type, instead of formatting manually like this

contracts/axelarnet-gateway/src/contract/execute.rs
```diff
@@ -0,0 +1,134 @@
...
+            .change_context(Error::MessageIdConstructionFailed)?,
+    };
+
+    let payload_hash = Keccak256::digest(payload.as_slice()).into();
+
+    let msg = Message {
+        cc_id: cc_id.clone(),
+        source_address: Address::try_from(sender.clone().into_string())
+            .change_context(Error::InvalidSender(sender))?,
```
cjcobb23, line 40: in what situation would this ever error? Seems like it should be an asset or expect

contracts/axelarnet-gateway/src/contract/execute.rs
```diff
@@ -0,0 +1,134 @@
...
+        destination_chain,
+        destination_address,
+        payload_hash,
+    };
+
+    state::save_incoming_msg(store, cc_id, &msg).change_context(Error::InvalidStoreAccess)?;
+
+    let wasm_msg = router
+        .route(vec![msg.clone()])
+        .ok_or(Error::RoutingFailed)?;
```
cjcobb23, line 50: I would just construct this inline where it's used, instead of making a variable for it

contracts/axelarnet-gateway/src/contract/execute.rs
```diff
@@ -0,0 +1,134 @@
...
+) -> Result<Response, Error> {
+    for msg in msgs.iter() {
+        let stored_msg = state::may_load_incoming_msg(store, &msg.cc_id)
+            .change_context(Error::MessageNotFound(msg.cc_id.clone()))?;
+
+        match stored_msg {
+            Some(message) if msg != &message => {
+                Err(report!(Error::MessageMismatch(msg.cc_id.clone())))
+            }
+            _ => Ok(()),
```
cjcobb23, line 86: I think we should error if the message is not found

contracts/axelarnet-gateway/src/contract/execute.rs
```diff
@@ -0,0 +1,134 @@
...
+        }?
+    }
+
+    let wasm_msg = router.route(msgs.clone()).ok_or(Error::RoutingFailed)?;
```
cjcobb23, line 90: prefer to construct this inline where it's used

contracts/axelarnet-gateway/src/contract/execute.rs
```diff
@@ -0,0 +1,134 @@
...
+    let events = msgs
+        .into_iter()
+        .map(|msg| AxelarnetGatewayEvent::Routing { msg }.into());
```
cjcobb23, line 93: this could also be constructed inline

contracts/axelarnet-gateway/src/msg.rs
```diff
@@ -0,0 +1,60 @@
...
+        destination_chain: ChainName,
+        destination_address: Address,
+        payload: HexBinary,
+    },
+
+    /// Forward the given messages to the next step of the routing layer.
+    /// Messages initiated via `CallContract` can be forwarded again to the router.
+    /// If the messages are coming from the router, then they are marked ready for execution.
+    #[permission(Any)]
+    RouteMessages(Vec<Message>),
```
cjcobb23, line 32: It's a bit cumbersome to have to pass the whole message in the rerouting case. I don't know that it makes sense to change the API from the standard gateway API. Who is the primary caller of this? Axelarons?

contracts/axelarnet-gateway/src/msg.rs
```diff
@@ -0,0 +1,48 @@
...
+        cc_id: CrossChainId,
+        payload: HexBinary,
+    },
+}
+
+#[cw_serde]
+#[derive(QueryResponses)]
+pub enum QueryMsg {
+    /// Returns the list of messages with their status destined for Axelar.
+    #[returns(Vec<MessageWithStatus>)]
```
milapsheth, line 46: Doesn't quite match the generic `gateway` query, which we might want for consistency with other gateways (e.g. for relayers). Thoughts on splitting this into two queries (one that returns Vec<Message>, and another that returns the statuses)?

contracts/axelarnet-gateway/src/msg.rs
```diff
@@ -0,0 +1,60 @@
...
+    OutgoingMessages { message_ids: Vec<CrossChainId> },
```
cjcobb23, line 47: Probably would be useful to have a query for the messages initiated via CallContract as well, though we could add in a separate PR, since this is already quite large.

Also, I find the nomenclature confusing here. For the other gateways, we call the messages returned by this query "outgoing messages", and it makes sense because those messages are going to be relayed from the gateway on axelar to a gateway on another chain (those messages are leaving axelar, in a sense). Versus the messages relayed from the source to axelar are incoming. But in this case, I feel like it's more natural to call the messages initiated via CallContract as outgoing (they will leave axelar), and messages received from the router as incoming (they will be executed on axelar). 

I don't want to switch the standard nomenclature we've been using, because I feel like that would also be confusing. I think you could just call this query `GetMessages` and add a status for messages initiated via CallContract ('Sent'?).

contracts/axelarnet-gateway/src/msg.rs
```diff
@@ -0,0 +1,60 @@
...
+pub enum AxelarExecutableExecuteMsg {
+    /// Execute the message at the destination contract with the corresponding payload, via the gateway.
+    Execute {
+        cc_id: CrossChainId,
+        source_address: Address,
+        destination_chain: ChainName,
+        destination_address: Address,
+        payload: HexBinary,
+    },
+}
```
cjcobb23: this should probably be in a package

contracts/axelarnet-gateway/src/contract/execute.rs
```diff
@@ -0,0 +1,134 @@
...
+            .change_context(Error::SaveOutgoingMessage)?;
+    }
+
+    Ok(Response::new().add_events(
+        msgs.into_iter()
+            .map(|msg| AxelarnetGatewayEvent::Routing { msg }.into()),
+    ))
+}
+
+pub(crate) fn route_incoming_messages(
```
cjcobb23, line 73: same comment about nomenclature here, I find these names confusing

contracts/axelarnet-gateway/src/msg.rs
```diff
@@ -0,0 +1,60 @@
...
+#[cw_serde]
+#[derive(QueryResponses)]
+pub enum QueryMsg {
+    /// Returns the list of messages with their status destined for Axelar.
+    #[returns(Vec<MessageWithStatus>)]
+    OutgoingMessages { message_ids: Vec<CrossChainId> },
+}
+
+#[cw_serde]
+pub enum AxelarExecutableExecuteMsg {
```
cjcobb23: idk that this needs to be an enum. Executable contracts will probably have their own `ExecuteMsg` enum, and just one case will be this struct here. I think it might be cleaner to just make this a struct. Otherwise there will be an extra layer of unnecessary nesting.

contracts/axelarnet-gateway/src/state.rs
```diff
@@ -0,0 +1,231 @@
...
+}
+
+const CONFIG_NAME: &str = "config";
+const CONFIG: Item<Config> = Item::new(CONFIG_NAME);
+const COUNTER_NAME: &str = "counter";
+const COUNTER: Counter<u64> = Counter::new(COUNTER_NAME);
+const INCOMING_MESSAGES_NAME: &str = "incoming_messages";
+const INCOMING_MESSAGES: Map<CrossChainId, Message> = Map::new(INCOMING_MESSAGES_NAME);
+const OUTGOING_MESSAGES_NAME: &str = "outgoing_messages";
+const OUTGOING_MESSAGES: Map<CrossChainId, MessageWithStatus> = Map::new(OUTGOING_MESSAGES_NAME);
```
cjcobb23, line 33: add some spaces here between the groups

contracts/axelarnet-gateway/src/contract/execute.rs
milapsheth: yeah, just switched it

contracts/axelarnet-gateway/src/contract/execute.rs
milapsheth: it'll be switched to the actual tx hash, so just a placeholder for now

contracts/axelarnet-gateway/src/msg.rs
milapsheth: Yeah, I want to conform to the gateway API, since the router also calls this method. Re-routing from source will likely be done manually.

contracts/axelarnet-gateway/src/msg.rs
milapsheth: yeah, will need to think about how this should be enforced for apps though
https://github.com/axelarnetwork/axelar-amplifier/pull/570#discussion_r1706148964

contracts/axelarnet-gateway/src/msg.rs
milapsheth: Yeah, I'd prefer that although we need it for the executable client to serialize it correctly I believe

contracts/axelarnet-gateway/src/msg.rs
milapsheth: Hmm, I wanted to preserve the API that other gateways expose (another name will also break it). Agreed that it's confusing, although not sure about calling it just `GetMessages`

contracts/axelarnet-gateway/src/executable.rs
```diff
@@ -0,0 +1,81 @@
...
+    },
+}
+
+impl<'a> From<client::Client<'a, AxelarExecutableMsg, ()>> for AxelarExecutableClient<'a> {
+    fn from(client: client::Client<'a, AxelarExecutableMsg, ()>) -> Self {
+        AxelarExecutableClient { client }
+    }
+}
+
+pub struct AxelarExecutableClient<'a> {
```
milapsheth, line 21: the client is only relevant to the gateway itself (since it executes it). Only the ExecuteMsg needs to be public for apps to implement it

contracts/axelarnet-gateway/src/contract/execute.rs
milapsheth: sure, i thought we were always using errors instead of unwrapping

contracts/axelarnet-gateway/src/contract/execute.rs
milapsheth: hmm, feels less readable to have a huge expression. switched it for comparison

contracts/axelarnet-gateway/src/contract/execute.rs
milapsheth: we should change in the gateway as well. Since the frame of reference is confusing, we can try to be more explicit. Perhaps something like `route_from_router`, `router_to_gateway`, `route_to_chain`

contracts/axelarnet-gateway/src/msg.rs
cjcobb23: Well the return type is already different, so the API is already not preserved. For normal gateways, this is also mainly used by the prover, but there is no prover in this case. Furthermore, this gateway is different than the others, since it's not an intermediate gateway. So I would rather this gateway have an API that is clear and understandable based on what it does, rather than try to conform to what the other gateways do and be confusing.

contracts/axelarnet-gateway/src/contract/execute.rs
cjcobb23: I'd rather use the blockhash as a placeholder for now, instead of this dummy string. The blockhash provides more info than this.

contracts/axelarnet-gateway/src/msg.rs
cjcobb23: im ok with flattening here, cleaner is better

contracts/axelarnet-gateway/src/contract/execute.rs
milapsheth: Sure, blockhash isn't available, but added block height

contracts/axelarnet-gateway/src/msg.rs
haiyizxx: I also found incoming / outgoing confusing. In the context of gateway itself,
- messages from router are incoming
- call contract is outgoing

I am open to name it more explicitly

contracts/axelarnet-gateway/src/client.rs
```diff
@@ -0,0 +1,131 @@
...
+    pub fn execute(&self, cc_id: CrossChainId, payload: HexBinary) -> WasmMsg {
+        self.client.execute(&ExecuteMsg::Execute { cc_id, payload })
+    }
+
+    pub fn route_messages(&self, msgs: Vec<Message>) -> Option<WasmMsg> {
+        ignore_empty(msgs).map(|messages| self.client.execute(&ExecuteMsg::RouteMessages(messages)))
+    }
+}
+
+fn ignore_empty(msgs: Vec<Message>) -> Option<Vec<Message>> {
```
haiyizxx, line 39: it can be extracted to package

milapsheth: the whole client, or just the helper?

haiyizxx: just the helper is good for now, we redefined the same fn multiple times already

contracts/axelarnet-gateway/src/msg.rs
milapsheth: I renamed it to `send_messages` and `receive_messages`

contracts/axelarnet-gateway/src/msg.rs
milapsheth: So, the current plan is to not conform with the gateway for queries. But I'll keep `RouteMessages` the same since it's a core API for the gateway and the router.

contracts/axelarnet-gateway/src/msg.rs
milapsheth: A macro to enforce this should be straightforward. But I'll open a separate task for that.

contracts/axelarnet-gateway/src/msg.rs
haiyizxx: could you also rename in state file

contracts/axelarnet-gateway/src/contract/execute.rs
```diff
@@ -0,0 +1,147 @@
...
+    let message_id = HexTxHashAndEventIndex {
+        tx_hash: Uint256::from(block_height).to_be_bytes(),
+        event_index: counter,
+    }
+    .to_string();
+
+    let cc_id = CrossChainId {
+        source_chain: chain_name.into(),
+        message_id: nonempty::String::try_from(message_id)
+            .change_context(Error::InvalidMessageId)?,
```
cjcobb23, line 39: this should probably be an assert too right?

contracts/axelarnet-gateway/src/state.rs
```diff
@@ -0,0 +1,235 @@
...
+pub(crate) fn may_load_outgoing_msg(
+    storage: &dyn Storage,
+    cc_id: &CrossChainId,
+) -> Result<Option<MessageWithStatus>, Error> {
+    OUTGOING_MESSAGES
+        .may_load(storage, cc_id.clone())
+        .map_err(Error::from)
+}
+
+pub(crate) fn save_outgoing_msg(
```
cjcobb23, line 99: rename these functions? save_received and save_sent

contracts/axelarnet-gateway/src/state.rs
```diff
@@ -0,0 +1,235 @@
...
+const CONFIG: Item<Config> = Item::new(CONFIG_NAME);
+
+const COUNTER_NAME: &str = "counter";
+const COUNTER: Counter<u32> = Counter::new(COUNTER_NAME);
+
+const INCOMING_MESSAGES_NAME: &str = "incoming_messages";
+const INCOMING_MESSAGES: Map<CrossChainId, Message> = Map::new(INCOMING_MESSAGES_NAME);
+
+const OUTGOING_MESSAGES_NAME: &str = "outgoing_messages";
+const OUTGOING_MESSAGES: Map<CrossChainId, MessageWithStatus> = Map::new(OUTGOING_MESSAGES_NAME);
```
cjcobb23, line 36: rename these?

contracts/axelarnet-gateway/src/state.rs
```diff
@@ -0,0 +1,235 @@
...
+                    status: MessageStatus::Approved,
+                },
+            )
+            .map_err(Error::from)?
+            .then(Ok),
+    }
+}
+
+/// Update the status of a message to executed if it is in approved status, error otherwise.
+pub(crate) fn update_msg_status(
```
cjcobb23, line 128:
```suggestion
pub(crate) fn set_msg_as_executed(
```

contracts/axelarnet-gateway/src/state.rs
```diff
@@ -0,0 +1,235 @@
...
+
+const OUTGOING_MESSAGES_NAME: &str = "outgoing_messages";
+const OUTGOING_MESSAGES: Map<CrossChainId, MessageWithStatus> = Map::new(OUTGOING_MESSAGES_NAME);
+
+#[derive(thiserror::Error, Debug, IntoContractError)]
+pub enum Error {
+    #[error(transparent)]
+    Std(#[from] StdError),
+    #[error("gateway got into an invalid state, its config is missing")]
+    MissingConfig,
```
cjcobb23, line 43: I personally think this sort of thing should be an assert, not an error

contracts/axelarnet-gateway/src/state.rs
```diff
@@ -0,0 +1,235 @@
...
+        .ok_or(Error::MissingConfig)
+}
+
+pub(crate) fn save_incoming_msg(
+    storage: &mut dyn Storage,
+    key: CrossChainId,
+    msg: &Message,
+) -> Result<(), Error> {
+    match INCOMING_MESSAGES.may_load(storage, key.clone())? {
+        Some(_) => Err(Error::MessageAlreadyExists(key)),
```
cjcobb23, line 73: should this be an assert? When would this happen?

contracts/axelarnet-gateway/src/contract/execute.rs
```diff
@@ -0,0 +1,147 @@
...
+use router_api::client::Router;
+use router_api::{Address, ChainName, CrossChainId, Message};
+use sha3::{Digest, Keccak256};
+
+use crate::contract::Error;
+use crate::events::AxelarnetGatewayEvent;
+use crate::executable::AxelarExecutableClient;
+use crate::state::{self};
+
+#[allow(clippy::too_many_arguments)]
```
cjcobb23, line 16: should probably find a way to avoid having to use this. Maybe just pass in the config?

contracts/axelarnet-gateway/src/contract/execute.rs
```diff
@@ -0,0 +1,147 @@
...
+    sender: Addr,
+    destination_chain: ChainName,
+    destination_address: Address,
+    payload: HexBinary,
+) -> Result<Response, Error> {
+    let counter = state::increment_msg_counter(store).change_context(Error::InvalidStoreAccess)?;
+
+    // TODO: Retrieve the actual tx hash from core, since cosmwasm doesn't provide it. Use the block height as the placeholder in the meantime.
+    let message_id = HexTxHashAndEventIndex {
+        tx_hash: Uint256::from(block_height).to_be_bytes(),
```
cjcobb23, line 31: oh well this is not as clean as using the blockhash, but oh well

contracts/axelarnet-gateway/src/contract/execute.rs
```diff
@@ -0,0 +1,147 @@
...
+        .add_events(events))
+}
+
+// Because the messages came from the router, we can assume they are already verified
+pub(crate) fn receive_messages(
+    store: &mut dyn Storage,
+    msgs: Vec<Message>,
+) -> Result<Response, Error> {
+    for msg in msgs.iter() {
+        state::save_outgoing_msg(store, msg.cc_id.clone(), msg.clone())
```
cjcobb23, line 69:
```suggestion
        state::save_received_msg(store, msg.cc_id.clone(), msg.clone())
```

contracts/axelarnet-gateway/src/contract/execute.rs
```diff
@@ -0,0 +1,147 @@
...
+    let msg = Message {
+        cc_id: cc_id.clone(),
+        source_address: Address::try_from(sender.clone().into_string())
+            .expect("failed to convert sender address"),
+        destination_chain,
+        destination_address,
+        payload_hash,
+    };
+
+    state::save_incoming_msg(store, cc_id, &msg).change_context(Error::InvalidStoreAccess)?;
```
cjcobb23, line 53:
```suggestion
    state::save_sent_msg(store, cc_id, &msg).change_context(Error::InvalidStoreAccess)?;
```

contracts/axelarnet-gateway/src/contract/execute.rs
```diff
@@ -0,0 +1,147 @@
...
+    let payload_hash: [u8; 32] = Keccak256::digest(payload.as_slice()).into();
+    if payload_hash != msg.payload_hash {
+        return Err(report!(Error::PayloadHashMismatch));
+    }
+
+    if chain_name != msg.destination_chain {
+        return Err(report!(Error::InvalidDestinationChain(
+            msg.destination_chain
+        )));
+    }
```
cjcobb23, line 133: Should this be an assert? How could this happen?

cjcobb23: Or maybe this check should be moved to `receive_messages`?

contracts/axelarnet-gateway/src/msg.rs
```diff
@@ -0,0 +1,42 @@
...
+        payload: HexBinary,
+    },
+
+    /// Forward the given messages to the next step of the routing layer.
+    /// Messages initiated via `CallContract` can be forwarded again to the router.
+    /// If the messages are coming from the router, then they are marked ready for execution.
+    #[permission(Any)]
+    RouteMessages(Vec<Message>),
+
+    /// Execute the message at the destination contract with the corresponding payload.
```
cjcobb23, line 32: add a comment about the message being marked as executed and thus cannot be executed again

contracts/axelarnet-gateway/src/contract/execute.rs
```diff
@@ -0,0 +1,147 @@
...
+
+    let executable: AxelarExecutableClient =
+        client::Client::new(querier, destination_contract).into();
+
+    // Call the destination contract
+    // Apps are required to expose AxelarExecutableMsg::Execute interface
+    Ok(Response::new()
+        .add_message(executable.execute(msg.cc_id.clone(), msg.source_address.clone(), payload))
+        .add_event(AxelarnetGatewayEvent::MessageExecuted { msg }.into()))
+}
```
cjcobb23, line 147: shouldn't this file have some tests?

contracts/axelarnet-gateway/Cargo.toml
```diff
@@ -0,0 +1,54 @@
+[package]
+name = "axelarnet-gateway"
+version = "0.1.0"
+rust-version = { workspace = true }
+edition = "2021"
```
cjcobb23, line 5: is this the right year?

contracts/axelarnet-gateway/src/state.rs
milapsheth: ah, forgot to update state.rs

contracts/axelarnet-gateway/src/state.rs
milapsheth: I usually favour using errors in Rust instead of having to think about whether certain cases can be considered panic-worthy or not (and decide again whenever the related code changes). It's nice to have extra info in case it does get triggered due to some issue. Not a strong opinion for contract code, but we should be consistent everywhere if we decide to switch.

contracts/axelarnet-gateway/src/contract/execute.rs
milapsheth: If we add more fields to the config, then we're potentially moving more data around than needed. And deviates from the pattern of only passing data that's needed by the method, which is a good one to follow.

contracts/axelarnet-gateway/src/contract/execute.rs
milapsheth: Yeah, weird that they haven't exposed blockhash / txhash in cosmwasm yet. Regardless, this will be changed soon

contracts/axelarnet-gateway/src/contract/execute.rs
milapsheth: yep, planning to add test coverage today

contracts/axelarnet-gateway/Cargo.toml
milapsheth: lol, afaik Rust 2024 edition hasn't been released yet

contracts/axelarnet-gateway/src/contract/execute.rs
milapsheth: implemented From conversion to make this cleaner

contracts/axelarnet-gateway/src/state.rs
milapsheth: Only due to coding issues atm (incorrect message id usage, or migration issue). But I prefer `state::*` methods to not assume what the caller exactly does, and just return an error.

contracts/axelarnet-gateway/src/contract/execute.rs
milapsheth: Config issue potentially. Gateway is registered under a different chain name in the router. Since the gateway set the chain_name for send_messages, I think it's a fine sanity check to apply for receiving messages (still prefer returning a structured error vs an assert, but not a strong opinion). I prefer checking it here instead of the state file, since it's being checked against `chain_name` which `receive_messages` doesn't have internally (or it has to load the config again).

contracts/axelarnet-gateway/src/state.rs
cjcobb23: you can always use `expect` to output an error message and have the extra info. I prefer to use asserts/expect for things that should never happen. I think this is general best practice, and the purpose of assert/expect/panic. But we are not consistent about this, and there are many places in the code where we return errors for things that should never happen.

contracts/axelarnet-gateway/src/contract/execute.rs
cjcobb23: I don't think you need to add more fields to the config. The config already has chain name and router, so you'd be compressing two arguments into one. I think that's enough to avoid having to use the allow macro

contracts/axelarnet-gateway/src/contract/execute.rs
cjcobb23: I meant we could check this when receiving messages from the router, instead of when trying to execute. I prefer failing earlier rather than later.

cjcobb23 (approved):

packages/router-api/src/client.rs
```diff
@@ -17,15 +18,7 @@ impl Router {
...
-    }
-}
-
-// TODO: unify across contract clients
-fn ignore_empty(msgs: Vec<Message>) -> Option<Vec<Message>> {
-    if msgs.is_empty() {
-        None
-    } else {
-        Some(msgs)
+        msgs.to_none_if_empty()
```
haiyizxx, line 21: 👍

haiyizxx (approved):

contracts/axelarnet-gateway/src/contract/execute.rs
haiyizxx: I also prefer to shrink args, hard to read.

milapsheth: I meant in the scenario where we added more fields to the config in the future. Also requires each execute method to set up it's own router client then causing some duplication. I'm ok with switching to an approach where we always pass in the Config in each contract instead of individual fields, but don't want to create inconsistency with just changing it here.

contracts/axelarnet-gateway/src/contract/execute.rs
milapsheth: Sure, since it's a sanity check, fine to fail there

PR #572, opened 2024-08-07
feat(interchain-token-service): add its hub contract

milapsheth:
## Description

[AXE-3820]

- Allow setting/removing trusted ITS addresses by chain
- Route ITS call to the true destination

## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Other Notes

[AXE-3820]: https://axelarnetwork.atlassian.net/browse/AXE-3820?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


interchain-token-service/src/msg.rs
```diff
@@ -0,0 +1,43 @@
...
+use axelarnet_gateway::AxelarExecutableMsg;
+use cosmwasm_schema::{cw_serde, QueryResponses};
+use msgs_derive::EnsurePermissions;
+use router_api::{Address, ChainName, ChainNameRaw};
+
+#[cw_serde]
+pub struct InstantiateMsg {
+    pub chain_name: ChainNameRaw,
+    pub gateway_address: String,
+    pub trusted_addresses: Option<HashMap<ChainName, Address>>,
```
haiyizxx, line 12: what is trusted addresses, ITS edge contracts for different chains?

interchain-token-service/src/contract.rs
```diff
@@ -0,0 +1,125 @@
...
+    msg: InstantiateMsg,
+) -> Result<Response, axelar_wasm_std::error::ContractError> {
+    cw2::set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
+
+    let gateway = deps
+        .api
+        .addr_validate(&msg.gateway_address)
+        .change_context(Error::InvalidAddress)
+        .attach_printable(msg.gateway_address.clone())?;
+
```
haiyizxx, line 67: need set gov and admin

interchain-token-service/src/contract/execute.rs
```diff
@@ -0,0 +1,287 @@
...
+
+    let its_hub_message =
+        ItsHubMessage::abi_decode(&payload).change_context(Error::InvalidPayload)?;
+
+    match its_hub_message {
+        ItsHubMessage::SendToHub {
+            destination_chain,
+            message: its_message,
+        } => {
+            let receive_from_hub = ItsHubMessage::ReceiveFromHub {
```
haiyizxx, line 39: don't need the additional var declare

interchain-token-service/src/msg.rs
milapsheth: Yep, maybe I'll rename to `trusted_its_addresses` or just `its_addresses`?

interchain-token-service/src/msg.rs
```diff
@@ -0,0 +1,45 @@
...
+use cosmwasm_schema::{cw_serde, QueryResponses};
+use msgs_derive::EnsurePermissions;
+use router_api::{Address, ChainName, ChainNameRaw};
+
+#[cw_serde]
+pub struct InstantiateMsg {
+    pub governance_address: String,
+    pub admin_address: String,
+    pub chain_name: ChainNameRaw,
+    pub gateway_address: String,
```
cjcobb23, line 13: this is the axelarnet gateway right? Should add a comment explaining so

interchain-token-service/src/msg.rs
```diff
@@ -0,0 +1,45 @@
...
+    pub its_addresses: HashMap<ChainName, Address>,
```
cjcobb23, line 14: add a comment about what these are for?

interchain-token-service/src/msg.rs
```diff
@@ -0,0 +1,45 @@
...
+
+#[cw_serde]
+#[derive(EnsurePermissions)]
+pub enum ExecuteMsg {
+    #[permission(Specific(gateway))]
+    Execute(AxelarExecutableMsg),
+    #[permission(Governance)]
+    SetItsAddress { chain: ChainName, address: Address },
+    #[permission(Elevated)]
+    RemoveItsAddress { chain: ChainName },
```
cjcobb23, line 25: Add comments describing what these are meant to do?

interchain-token-service/src/msg.rs
```diff
@@ -0,0 +1,45 @@
...
+}
+
+#[cw_serde]
+#[derive(QueryResponses)]
+pub enum QueryMsg {
+    #[returns(ItsAddressResponse)]
+    SetItsAddress { chain: ChainName },
```
cjcobb23, line 32:
```suggestion
   ItsAddress { chain: ChainName },
```

interchain-token-service/src/msg.rs
```diff
@@ -0,0 +1,45 @@
...
+    #[permission(Governance)]
+    SetItsAddress { chain: ChainName, address: Address },
+    #[permission(Elevated)]
+    RemoveItsAddress { chain: ChainName },
+}
+
+#[cw_serde]
+#[derive(QueryResponses)]
+pub enum QueryMsg {
+    #[returns(ItsAddressResponse)]
```
cjcobb23, line 31:
```suggestion
    #[returns(Option<Address>)]
```

interchain-token-service/src/msg.rs
```diff
@@ -0,0 +1,45 @@
...
+    SetItsAddress { chain: ChainName },
+    #[returns(AllItsAddressesResponse)]
+    AllItsAddresses {},
+}
+
+#[cw_serde]
+pub struct ItsAddressResponse {
+    pub address: Option<Address>,
+}
```
cjcobb23, line 40: Do we need this struct?

interchain-token-service/src/msg.rs
```diff
@@ -0,0 +1,45 @@
...
+
+#[cw_serde]
+pub struct AllItsAddressesResponse {
+    pub addresses: HashMap<ChainName, Address>,
+}
```
cjcobb23, line 45: Also not sure that we need this struct

interchain-token-service/src/msg.rs
```diff
@@ -0,0 +1,45 @@
...
+    #[permission(Elevated)]
+    RemoveItsAddress { chain: ChainName },
+}
+
+#[cw_serde]
+#[derive(QueryResponses)]
+pub enum QueryMsg {
+    #[returns(ItsAddressResponse)]
+    SetItsAddress { chain: ChainName },
+    #[returns(AllItsAddressesResponse)]
```
cjcobb23, line 33:
```suggestion
    #[returns(HashMap<ChainName, Address>)]
```

interchain-token-service/src/contract.rs
```diff
@@ -0,0 +1,184 @@
...
+    BalanceUpdateFailed(ChainName, TokenId),
+}
+
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn migrate(
+    _deps: DepsMut,
+    _env: Env,
+    _msg: Empty,
+) -> Result<Response, axelar_wasm_std::error::ContractError> {
+    // Implement migration logic if needed
```
cjcobb23, line 49:
```suggestion
    // Implement migration logic if needed
    cw2::set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
```

interchain-token-service/src/contract/execute.rs
```diff
@@ -0,0 +1,291 @@
...
+pub fn execute_message(
+    deps: DepsMut,
+    cc_id: CrossChainId,
+    source_address: Address,
+    payload: HexBinary,
+) -> Result<Response, Error> {
+    let config = load_config(deps.storage).change_context(Error::InvalidStoreAccess)?;
+
+    let source_chain = ChainName::try_from(cc_id.source_chain.clone().to_string())
+        .change_context(Error::InvalidPayload)?;
```
cjcobb23, line 24: Isn't this going to error if source chain is a legacy chain with uppercase letters?

interchain-token-service/src/msg.rs
```diff
@@ -0,0 +1,45 @@
...
+}
+
+#[cw_serde]
+#[derive(EnsurePermissions)]
+pub enum ExecuteMsg {
+    #[permission(Specific(gateway))]
+    Execute(AxelarExecutableMsg),
+    #[permission(Governance)]
+    SetItsAddress { chain: ChainName, address: Address },
+    #[permission(Elevated)]
```
cjcobb23, line 24: What's the reasoning behind the different permission levels here?

interchain-token-service/src/state.rs
```diff
@@ -0,0 +1,133 @@
...
+    Std(#[from] StdError),
+    #[error("ITS contract got into an invalid state, its config is missing")]
+    MissingConfig,
+    #[error("its address for chain {0} not found")]
+    ItsAddressNotFound(ChainName),
+}
+
+#[cw_serde]
+pub struct Config {
+    pub chain_name: ChainNameRaw,
```
cjcobb23, line 21:
```suggestion
    pub chain_name: ChainName,
```

interchain-token-service/src/msg.rs
```diff
@@ -0,0 +1,45 @@
...
+    RemoveItsAddress { chain: ChainName },
+}
+
+#[cw_serde]
+#[derive(QueryResponses)]
+pub enum QueryMsg {
+    #[returns(ItsAddressResponse)]
+    SetItsAddress { chain: ChainName },
+    #[returns(AllItsAddressesResponse)]
+    AllItsAddresses {},
```
cjcobb23, line 34: Do you know if there is a limit to how large of a hashmap we can return here? I'm wondering at what point we need to paginate the query

interchain-token-service/src/contract/execute.rs
```diff
@@ -0,0 +1,291 @@
...
+            message: its_message,
+        } => {
+            let destination_payload = ItsHubMessage::ReceiveFromHub {
+                source_chain: source_chain.clone(),
+                message: its_message.clone(),
+            }
+            .abi_encode();
+
+            let destination_address = load_its_address(deps.storage, &destination_chain)
+                .change_context(Error::InvalidStoreAccess)?;
```
cjcobb23, line 46: Is this safe? It seems a little weird that when someone creates the original message, they don't necessarily know the final destination address for the message, but rather it's supplied by the ITS hub. And it looks like this is modifiable? So between when I send the message and when it actually arrives at the ITS hub, the destination ITS contract could change. In what situation would the ITS contract for a particular chain change?

interchain-token-service/src/events.rs
```diff
@@ -0,0 +1,103 @@
...
+fn make_its_message_event(
+    event_name: &str,
+    source_chain: ChainName,
+    destination_chain: ChainName,
+    msg: ItsMessage,
+) -> Event {
+    let mut attrs = vec![
+        Attribute::new("source_chain", source_chain.to_string()),
+        Attribute::new("destination_chain", destination_chain.to_string()),
+        Attribute::new("message_type", format!("{:?}", msg)),
```
cjcobb23, line 55: Does this not display the whole msg?

interchain-token-service/src/msg.rs
```diff
@@ -0,0 +1,45 @@
...
+use axelarnet_gateway::AxelarExecutableMsg;
+use cosmwasm_schema::{cw_serde, QueryResponses};
+use msgs_derive::EnsurePermissions;
+use router_api::{Address, ChainName, ChainNameRaw};
+
+#[cw_serde]
+pub struct InstantiateMsg {
+    pub governance_address: String,
+    pub admin_address: String,
+    pub chain_name: ChainNameRaw,
```
cjcobb23, line 12:
```suggestion
    pub chain_name: ChainName,
```

interchain-token-service/src/state.rs
cjcobb23: Also, where is this used?

interchain-token-service/src/state.rs
```diff
@@ -0,0 +1,133 @@
...
+    pub gateway: Addr,
```
cjcobb23, line 22:
```suggestion
    pub axelarnet_gateway: Addr,
```

interchain-token-service/src/contract.rs
```diff
@@ -0,0 +1,184 @@
...
+) -> Result<Binary, axelar_wasm_std::error::ContractError> {
+    match msg {
+        QueryMsg::SetItsAddress { chain } => query::its_address(deps, chain)?,
+        QueryMsg::AllItsAddresses {} => query::all_its_addresses(deps)?,
+    }
+    .then(Ok)
+}
+
+#[cfg(test)]
+mod tests {
```
cjcobb23, line 132: would be good to have a test that covers the execute methods, at least at a basic level, if only just to test the permissions

interchain-token-service/src/contract/execute.rs
milapsheth: ITS hub is a router for ITS calls. Similar to how external gateway contracts don't know about each other's address, since the Amplifier protocol does the translation, there's no need for each ITS edge contract to also be aware. In fact, this is one of the powerful properties of our hub design that we want to extend to ITS as well as we move away from the p2p configuration.

milapsheth: Yeah, I'm also considering preventing the overwrite for it and making it require an explicit migration like for gateway address. It's unlikely we redeploy ITS contracts, and would require a migration externally as well

interchain-token-service/src/msg.rs
milapsheth: Yeah, no point of wrapping. I'm used to axelar-core wrapper protobufs for query responses lol

interchain-token-service/src/msg.rs
milapsheth: I'll test out gas usage with 200+ chains

interchain-token-service/src/contract/execute.rs
milapsheth: Thought about allowing overwrite more. Unlike external gateway addresses that are maintained per voting verifier, ITS hub has to track all ITS addresses. So if an integrator whitelists an ITS address incorrectly, or needs to redeploy (quite probable during testing), it'll be convenient to allow overwriting, so we don't have to migrate ITS hub

interchain-token-service/src/events.rs
milapsheth: good catch, fixed. Thoughts about using a string vs a number for the event. String representation is not great if we rename something, but it's more readable

interchain-token-service/src/contract.rs
milapsheth: Yeah, planning to add tests in the newer style of axelarnet-gateway today

interchain-token-service/src/primitives.rs
```diff
@@ -56,13 +64,13 @@ pub enum ItsHubMessage {
     /// ITS edge source contract -> ITS Hub
     SendToHub {
         /// True destination chain of the ITS message
-        destination_chain: ChainName,
+        destination_chain: ChainNameRaw,
```
milapsheth, line 67: I'll change this back to `ChainName` since it'll auto normalize the underlying string so I don't have to normalize explicitly. A bit annoying that source has to use the raw name, but we can't get rid of legacy naming in ITS without breaking consistency.

talalashraf (approved):

interchain-token-service/src/contract/execute.rs
```diff
@@ -0,0 +1,330 @@
...
+    match ItsHubMessage::abi_decode(&payload).change_context(Error::InvalidPayload)? {
+        ItsHubMessage::SendToHub {
+            destination_chain,
+            message,
+        } => {
+            let destination_chain = normalize(&destination_chain);
+            let destination_address = load_its_address(deps.storage, &destination_chain)
+                .change_context_lazy(|| Error::UnknownChain(destination_chain.clone()))?;
+
+            let destination_payload = ItsHubMessage::ReceiveFromHub {
```
fish-sammy, line 44: Why is the source address not encoded in the payload?

interchain-token-service/src/contract/execute.rs
```diff
@@ -0,0 +1,330 @@
...
+
+            Ok(send_to_destination(
+                deps.storage,
+                deps.querier,
+                destination_chain.clone(),
+                destination_address,
+                destination_payload,
+            )?
+            .add_event(
+                ItsContractEvent::ItsMessageReceived {
```
fish-sammy, line 58: Maybe just me, but I kept reading it as the word "its" instead of "ITS". Does all-cap it violate rust linter? If not, I'd prefer ITS over Its/its

interchain-token-service/src/msg.rs
fish-sammy:
```suggestion
    AllItsAddresses,
```

interchain-token-service/src/state.rs
fish-sammy: Same question. What is this?

.github/workflows/basic.yaml
```diff
@@ -93,7 +93,7 @@ jobs:
 
       - name: Build ITS release
         working-directory: ./interchain-token-service
```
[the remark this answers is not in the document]
milapsheth: Chris suggested to keep it in a separate folder since it's not part of the core amplifier protocol contracts

interchain-token-service/src/contract/execute.rs
```diff
@@ -0,0 +1,330 @@
...
+                    message,
+                }
+                .into(),
+            ))
+        }
+        _ => bail!(Error::InvalidPayload),
+    }
+}
+
+fn normalize(chain: &ChainNameRaw) -> ChainName {
```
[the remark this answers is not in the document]
milapsheth: I think we want to avoid implementing `TryFrom<ChainNameRaw> for ChainName` for safety in other contracts so users don't lose the casing by mistake. I suppose the router contract didn't run into this issue since it queries registered ChainName by the gateway address and then equates it to `cc_id.source_chain` which is `ChainNameRaw`

Maybe we can name it `to_chain_name`? Another option is to use the raw chain name for ITS addresses, but I prefer to avoid introducing case sensitivity in storage.

cc @cjcobb23

interchain-token-service/src/contract/execute.rs
milapsheth: The `source_address` is of the ITS edge contract on source chain which is not relevant for the destination. The real `source_address` corresponding to the caller of ITS is inside the [its message](https://github.com/axelarnetwork/axelar-amplifier/blob/e38843b9bb62c79c0038ef96d24df559f5960157/interchain-token-service/src/primitives.rs#L41). Since ITS is an app on top of GMP, we can't get around this wrapping issue.

interchain-token-service/src/contract/execute.rs
milapsheth: Lol, yeah, Rust [recommends](https://rust-lang.github.io/api-guidelines/naming.html) this convention for acronyms (e.g. `Uuid`).

interchain-token-service/src/state.rs
milapsheth: The chain name of Axelar. It'll require special handling if ITS calls are received to Axelar itself. I'll remove it for now and add it back when needed.

.github/workflows/basic.yaml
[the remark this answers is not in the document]
cjcobb23: It is in a different class than the rest of the contracts, but I actually don't feel strongly that it's in a separate folder. I see the rationale for keeping it separate, but keeping everything in the same folder is also a bit more readable. Either way is fine with me.

interchain-token-service/src/primitives.rs
cjcobb23, line 67: Are you planning to change this? I thought core was able to accept a normalized name, and do the conversion if necessary. I was under the impression that ChainNameRaw was only needed for the source chain for messages to amplifier

packages/axelar-wasm-std/src/msg.rs
```diff
@@ -0,0 +1,32 @@
...
+    Std(#[from] StdError),
+    #[error("no wasm msg found")]
+    NotFound,
+    #[error("multiple msgs found in the response")]
+    MultipleMsgsFound,
+}
+
+/// Get a msg wrapped inside a `WasmMsg::Execute` from a `Response`.
+/// If there are no wasm messages or more than one message in the response, this returns an error.
+pub fn from_response<T>(response: Response) -> Result<T, Error>
```
cjcobb23: where is this being used?

interchain-token-service/src/primitives.rs
```diff
@@ -1,6 +1,8 @@
+use std::fmt::Display;
+
```
cjcobb23, line 2: I dislike using `primitives.rs` files, it's sort of a catch all, which gets very messy. You could just call this `its_message.rs` or somesuch.

interchain-token-service/src/contract/execute.rs
cjcobb23: I agree, I keep reading it as the word "its", would prefer "ITS"

interchain-token-service/src/contract/execute.rs
```diff
@@ -0,0 +1,329 @@
...
+                }
+                .into(),
+            ))
+        }
+        _ => bail!(Error::InvalidPayload),
+    }
+}
+
+fn to_chain_name(chain: &ChainNameRaw) -> ChainName {
+    ChainName::try_from(chain.as_ref()).expect("invalid chain name")
```
cjcobb23, line 71: this one probably shouldn't be a panic right? This is being used on data supplied externally, so I think we should return an error

interchain-token-service/src/events.rs
```diff
@@ -0,0 +1,97 @@
+use cosmwasm_std::{Attribute, Event};
+use router_api::{Address, ChainName, CrossChainId};
+
+use crate::primitives::ItsMessage;
+
+pub enum ItsContractEvent {
```
cjcobb23, line 6: you could just name this `Event`, it's sort of redundant to call it `ItsContractEvent` since the containing package gives you the scoping

interchain-token-service/src/contract/execute.rs
cgorenflo: It does look a bit weird, but rust is very clear about this convention. Linting severity could change in the future (the lint exists, it's just not on by default https://doc.rust-lang.org/nightly/clippy/lint_configuration.html#upper-case-acronyms-aggressive). Capitalizing can also lead to issues when generated code is derived from such a type this (e.g. call json fields `i_t_s_foo`. So I would prefer to either come up with an alternative, e.g. call it fully specified and remove the prefix like `interchain-token-service::ContractEvent` or spell it out, rather than breaking with the convention

cgorenflo: Furthermore, `save_its_address` for example looks just as weird, and we definitely shouldn't convert that into `save_ITS_address`, so you can't get rid of the weirdness anyway, so embrace the convention 🙂

interchain-token-service/src/contract/execute.rs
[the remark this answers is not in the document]
cjcobb23: I'm ok with a `TryFrom<ChainNameRaw> for ChainName` as long as it fails if the casing is wrong. I'm confused about why we need this though? Why can't we just use `ChainName` to begin with? If the name has uppercase letters, this is going to fail anyways.

interchain-token-service/src/contract/execute.rs
```diff
@@ -0,0 +1,329 @@
...
+
+    let gateway: axelarnet_gateway::Client =
+        client::Client::new(querier, config.axelarnet_gateway).into();
+
+    let call_contract_msg = gateway.call_contract(destination_chain, destination_address, payload);
+
+    Ok(Response::new().add_message(call_contract_msg))
+}
+
+pub fn set_its_address(
```
cgorenflo, line 108: I think it makes more sense to call all these references to `its_address` `its_contract_address` instead

interchain-token-service/src/contract/execute.rs
```diff
@@ -0,0 +1,329 @@
...
+        let expected_event = ItsContractEvent::ItsMessageReceived {
+            cc_id,
+            destination_chain,
+            message: its_message,
+        };
+        assert_eq!(result.events, vec![Event::from(expected_event)]);
+    }
+
+    #[test]
+    fn execute_message_units_address() {
```
cgorenflo, line 227: what does this mean?

cgorenflo (commented):
can you use more descriptive test names?

interchain-token-service/src/contract/execute.rs
cgorenflo: yes, no panic

interchain-token-service/src/events.rs
```diff
@@ -0,0 +1,97 @@
...
+        destination_chain: ChainName,
+        message: ItsMessage,
+    },
+    ItsAddressSet {
+        chain: ChainName,
+        address: Address,
+    },
+    ItsAddressRemoved {
+        chain: ChainName,
+    },
```
cgorenflo, line 18: can we use different verbs than set and remove (everywhere, not just for the event). It's not just a store entry, it also implies access control, right?

cgorenflo: e.g. allow, register, authorize

interchain-token-service/src/lib.rs
```diff
@@ -1,5 +1,8 @@
 mod primitives;
-
-pub mod error;
 pub use primitives::*;
+
 pub mod abi;
```
cgorenflo, line 4: does this need to be public?

interchain-token-service/src/msg.rs
```diff
@@ -0,0 +1,45 @@
...
+
+#[cw_serde]
+#[derive(QueryResponses)]
+pub enum QueryMsg {
+    /// Query the ITS contract address of a chain
+    #[returns(Option<Address>)]
+    ItsAddress { chain: ChainName },
+    /// Query all configured ITS contract addresses
+    #[returns(HashMap<ChainName, Address>)]
+    AllItsAddresses,
```
cgorenflo, line 44: do we need to paginate this?

interchain-token-service/src/primitives.rs
```diff
@@ -26,7 +34,7 @@ pub enum TokenManagerType {
 /// ITS message type that can be sent between ITS contracts for transfers/token deployments
 /// `ItsMessage` that are routed via the ITS hub get wrapped inside `ItsHubMessage`
 #[cw_serde]
-#[derive(Eq)]
+#[derive(Eq, strum::IntoStaticStr)]
```
cgorenflo, line 37: Why do we need that?

interchain-token-service/src/primitives.rs
cgorenflo: what do you mean by that? Why can't we use `ChainName` at the source?

interchain-token-service/src/state.rs
```diff
@@ -0,0 +1,127 @@
...
+pub struct Config {
+    pub axelarnet_gateway: Addr,
+}
+
+const CONFIG: Item<Config> = Item::new("config");
+const ITS_ADDRESSES: Map<&ChainName, Address> = Map::new("its_addresses");
+
+pub fn load_config(storage: &dyn Storage) -> Result<Config, Error> {
+    CONFIG.may_load(storage)?.ok_or(Error::MissingConfig)
+}
```
cgorenflo, line 29: something to discuss: in the axelarnet-gateway I ended up making this panic if the config isn't there and just return `Config`. Missing the config means that either instantiation or migration went wrong, i.e. there is a bug in the code.

interchain-token-service/tests/execute.rs
cgorenflo, line 1: please be more descriptive in the test names

packages/axelar-wasm-std/src/msg.rs
cgorenflo, line 1: response.rs?

interchain-token-service/src/primitives.rs
milapsheth: The `source_chain` comes is the case sensitive identifier that comes from the `message.cc_id.source_chain`. ITS allows optionally calling a contract along with the transfer (i.e GMP call with token). Similar to GMP, the app will receive the true `source_chain` and might have a whitelist based on the string. Hence, the raw chain name needs to be passed to avoid inconsistencies with GMP

.github/workflows/basic.yaml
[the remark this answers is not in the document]
milapsheth: I slightly prefer keeping it in contracts folder for readability and to avoid custom handling in our gh workflows

interchain-token-service/src/contract/execute.rs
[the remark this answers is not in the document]
milapsheth: No, this is converting any upper case chars to lower case when casting to `ChainName`. That is the behaviour we want, since we're trying to look up the remote ITS address for this chain. Say the message came from eth ITS, which has a case sensitive name `Ethereum`. The ITS address map will store `ethereum -> ITS_address`, and then we need to convert `ChainNameRaw` to `ChainName`.

I'd like to avoid having ITS address map use `ChainNameRaw` as the key to prevent casing issues. And also if I used that, it causes the opposite issue for `destination_chain` which now has `ChainName` type that needs to be converted to the raw name.

interchain-token-service/src/contract/execute.rs
milapsheth: The JSON conversion issue is a good example. I'm also ok with being verbose `InterchainTokenServiceMessage` or external crates qualifying it with the name and keeping the types simpler `interchain_token_service::Message`, with preference for the latter.

interchain-token-service/src/primitives.rs
milapsheth: Yes, I changed it back to `ChainName` to mimic GMP

packages/axelar-wasm-std/src/msg.rs
milapsheth: in the integration tests (for ITS and axelarnet-gateway)

interchain-token-service/src/primitives.rs
milapsheth: Open to other naming suggestions. I don't like `its_message.rs` since it's too specific to the message type, but this file does contain core ITS types (like `TokenId`, `TokenManagerType`) that are relevant independent of the message itself.

interchain-token-service/src/contract/execute.rs
milapsheth: user can't cause an error since `ChainName` just lower cases the already validated `ChainNameRaw` string. It would be a bug if validation between the two diverges.

interchain-token-service/src/contract.rs
```diff
@@ -0,0 +1,160 @@
...
+        }) => execute::execute_message(deps, cc_id, source_address, payload)
+            .change_context(Error::Execute),
+        ExecuteMsg::SetItsAddress { chain, address } => {
+            execute::set_its_address(deps, chain, address).change_context(Error::SetItsAddress)
+        }
+        ExecuteMsg::RemoveItsAddress { chain } => {
+            execute::remove_its_address(deps, chain).change_context(Error::RemoveItsAddress)
+        }
+    }?
+    .then(Ok)
```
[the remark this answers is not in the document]
milapsheth: What Chris said

interchain-token-service/src/contract.rs
```diff
@@ -0,0 +1,160 @@
...
+    Ok(state::load_config(storage)?.axelarnet_gateway)
+}
+
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn query(deps: Deps, _: Env, msg: QueryMsg) -> Result<Binary, ContractError> {
+    match msg {
+        QueryMsg::ItsAddress { chain } => query::its_address(deps, chain)?,
+        QueryMsg::AllItsAddresses => query::all_its_addresses(deps)?,
+    }
+    .then(Ok)
```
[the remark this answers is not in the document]
milapsheth: https://github.com/axelarnetwork/axelar-amplifier/pull/572#discussion_r1737249141

interchain-token-service/src/abi.rs
```diff
@@ -562,7 +568,7 @@ mod tests {
     #[test]
     fn encode_decode_unicode_strings() {
         let original = ItsHubMessage::SendToHub {
-            destination_chain: ChainName::from_str("chain").unwrap(),
+            destination_chain: ChainNameRaw::from_str("chain").unwrap(),
```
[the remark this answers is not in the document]
milapsheth: switched it back. Already kept it for `source_chain` to mimic what we do for `router::Message`

interchain-token-service/src/contract.rs
```diff
@@ -0,0 +1,160 @@
...
+pub fn instantiate(
+    deps: DepsMut,
+    _: Env,
+    _: MessageInfo,
+    msg: InstantiateMsg,
+) -> Result<Response, ContractError> {
+    cw2::set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
+
+    let admin = deps.api.addr_validate(&msg.admin_address)?;
+    let governance = deps.api.addr_validate(&msg.governance_address)?;
```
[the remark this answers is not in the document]
milapsheth: How about doing the validation in `permission_control::set_*` instead? Requires passing in `deps.api` though but simplifies the usage everywhere. I can refactor this in another PR

interchain-token-service/src/lib.rs
milapsheth: Yeah, non smart contract chains like XRPL will implement ITS edge contract functionality in their gateways, so they need the `abi_encode`/`abi_decode` methods.

interchain-token-service/src/state.rs
milapsheth: sounds good to me

interchain-token-service/src/primitives.rs
milapsheth: In the event to make it more readable
https://github.com/axelarnetwork/axelar-amplifier/pull/572/files#diff-3fafd412ea6177bfd7b62305e64a5e58a51dd84ed8db07fe15d118c30c95cf9dR49

interchain-token-service/Cargo.toml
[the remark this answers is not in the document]
milapsheth: Since the app is called ITS and because it's a mouthful already. This contract will implement both ITS hub and edge functionality (i.e ITS transfers to/from Axelar/Cosmos).

interchain-token-service/src/events.rs
milapsheth: Yeah, I'll use register/deregister for consistency

interchain-token-service/src/contract/execute.rs
milapsheth: `its_address` being a contract feels implied, other potential names: `external_its_address`, `its_edge_address`, `remote_its_address`.

interchain-token-service/src/msg.rs
milapsheth: planning to test what gas consumption looks like for 100+ chains

interchain-token-service/src/contract/execute.rs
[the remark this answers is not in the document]
cjcobb23: We should definitely not implement a `TryFrom` that converts to lowercase, I think that is dangerous. It should fail if there are uppercase letters. Otherwise we could lose the casing by mistake.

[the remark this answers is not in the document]
cjcobb23: Just call this function `normalize`, like you did. And keep it local here, I think this is dangerous to expose more broadly.

I actually would prefer to just use ChainNameRaw for the keys. ItsHub is basically a router, and some of the chains registered with it need casing, so it should use the case sensitive type for chain name. All of our other contracts use the case insensitive type, because they only work with amplifier chains. But ItsHub works with legacy chains, and so I think should use a different type. I think the hidden conversions are confusing tbh.

interchain-token-service/src/contract/execute.rs
fish-sammy: I like `interchain_token_service::Message` the best.

interchain-token-service/src/contract/execute.rs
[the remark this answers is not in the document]
milapsheth: >Just call this function normalize, like you did. And keep it local here, I think this is dangerous to expose more broadly.
Yeah, I don't want to expose it.

If we use `ChainNameRaw` for ITS address mapping, we create an issue with `destination_chain` instead as I mentioned above. The `destination_chain` would now need to be `ChainNameRaw`, and needs to be normalized to `ChainName` to initiate `axelarnet_gateway::call_contract`. I found this to be worse than normalizing the `source_chain` since `destination_chain` casing doesn't matter for the protocol.

interchain-token-service/src/primitives.rs
cgorenflo: I don't like catch-all files either, but if it's just TokenId and ItsMessage I think it's fine. It just mustn't grow further than this

interchain-token-service/src/msg.rs
```diff
@@ -0,0 +1,45 @@
...
+pub enum ExecuteMsg {
+    /// Execute a cross-chain message received by the axelarnet-gateway from another chain
+    #[permission(Specific(gateway))]
+    Execute(AxelarExecutableMsg),
+    /// Register the ITS contract address of another chain. Each chain's ITS contract has to be whitelisted before
+    /// ITS Hub can send cross-chain messages to it, or receive messages from it.
+    /// If an ITS address is already set for the chain, it will be overwritten.
+    /// This allows easier management of ITS contracts without the need for migration.
+    #[permission(Governance)]
+    RegisterItsAddress { chain: ChainName, address: Address },
```
milapsheth, line 29: remove overwrite ability

interchain-token-service/src/contract/execute.rs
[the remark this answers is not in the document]
milapsheth: don't normalize for ITS address map, but normalize destination for call_contract

interchain-token-service/src/primitives.rs
milapsheth: preserve primitives.rs for now

packages/axelar-wasm-std/src/response.rs
```diff
@@ -0,0 +1,32 @@
...
+    Std(#[from] StdError),
+    #[error("no wasm msg found")]
+    NotFound,
+    #[error("multiple msgs found in the response")]
+    MultipleMsgsFound,
+}
+
+/// Get a msg wrapped inside a `WasmMsg::Execute` from a `Response`.
+/// If there are no wasm messages or more than one message in the response, this returns an error.
+pub fn inspect_response_msg<T>(response: Response) -> Result<T, Error>
```
fish-sammy, line 16: This is only used in tests? Then why is it placed here?

interchain-token-service/src/contract/execute.rs
```diff
@@ -0,0 +1,125 @@
...
+    state::save_its_contract(deps.storage, &chain, &address)
+        .change_context_lazy(|| Error::FailedItsAddressRegistration(chain.clone()))?;
+
+    Ok(Response::new().add_event(Event::ItsContractRegistered { chain, address }.into()))
+}
+
+pub fn deregister_its_contract(deps: DepsMut, chain: ChainNameRaw) -> Result<Response, Error> {
+    state::remove_its_contract(deps.storage, &chain);
+
+    Ok(Response::new().add_event(Event::ItsContractDeregistered { chain }.into()))
```
fish-sammy, line 124: This doesn't seem to fail if the chain isn't registered in the first place. Wouldn't the event be confusing then?

interchain-token-service/src/abi.rs
```diff
@@ -62,10 +61,22 @@ sol! {
...
+    MessageDecodeFailed,
+    #[error("invalid message type")]
+    InvalidMessageType,
+    #[error("invalid chain name")]
+    InvalidChainName,
+    #[error("invalid token manager type")]
+    InvalidTokenManagerType,
+}
+
+impl Message {
```
fish-sammy, line 76: I get the difference between `Message` and `HubMessage` but it's still quite confusing that they are put in the same enum in the solidity code.

fish-sammy (approved):

interchain-token-service/README.md
```diff
@@ -0,0 +1,15 @@
+# Interchain Token Service Hub
```
cjcobb23, line 1: would be nice to have a simple diagram in here, but can do at a later time

cjcobb23 (approved):

interchain-token-service/README.md
milapsheth: yeah, planning to add more docs (especially with the balance tracking feature), until then the notion doc is the reference

interchain-token-service/src/abi.rs
milapsheth: Yeah, it's confusing but we wanted to keep backwards compatibility with existing ITS connections and thus the message format

packages/axelar-wasm-std/src/response.rs
milapsheth: It's used in tests across multiple crates so needs to be in wasm-std. We should consider adding a `testutils` feature for `axelar-wasm-std` for test only code (such as assert macros). But this PR is already quite large so I didn't want to introduce more refactoring here.

interchain-token-service/src/contract/execute.rs
milapsheth: Good catch, added a check since I'd like to keep the event

cgorenflo (approved):

PR #573, opened 2024-08-07
feat(interchain-token-service): apply balance invariants for ITS transfers

milapsheth:
## Description

[AXE-4185]

## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Other Notes

[AXE-4185]: https://axelarnetwork.atlassian.net/browse/AXE-4185?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


interchain-token-service/src/state.rs
```diff
@@ -20,8 +30,62 @@ pub struct Config {
...
 }
 
+/// Token balance for a given token id and chain
+#[cw_serde]
+pub enum TokenBalance {
+    /// Token balance is tracked on the chain
+    Tracked(Uint256),
+    /// Token balance is not tracked
+    Untracked,
+}
```
milapsheth, line 40: Open to naming suggestions here. We need to distinguish between whether the token was already deployed once or not, and whether the balance invariant should be updated for this token

For custom tokens, we don't want to track balances since ITS doesn't have sole control over the mint/burn rights there. I used an enum instead of `Option` to make it more explicit.

milapsheth: There is also more token related info that we will likely store, such as rate limits. I'm thinking about wrapping this into a TokenInfo struct vs having separate maps for each. For transfers we'd be reading both everytime, but for deployments we only need to read the token balance though (but when stored under the same struct, we end up read both)

PR #576, opened 2024-08-08
feat(minor-rewards)!: set params per pool

cjcobb23:
https://axelarnetwork.atlassian.net/browse/AXE-4625


contracts/rewards/src/contract.rs
```diff
@@ -25,7 +25,7 @@ pub fn migrate(
     _env: Env,
     _msg: Empty,
 ) -> Result<Response, axelar_wasm_std::error::ContractError> {
-    v0_4_0::migrate(deps.storage)?;
+    v1_0_0::migrate(deps.storage)?;
```
cgorenflo, line 28: should we move the `set_contract_version` call into the migration after all?

contracts/rewards/src/contract/execute.rs
```diff
@@ -59,7 +61,10 @@ pub(crate) fn distribute_rewards(
     epoch_process_limit: Option<u64>,
 ) -> Result<HashMap<Addr, Uint128>, ContractError> {
     let epoch_process_limit = epoch_process_limit.unwrap_or(DEFAULT_EPOCHS_TO_PROCESS);
-    let cur_epoch = Epoch::current(&state::load_params(storage), cur_block_height)?;
+    let cur_epoch = Epoch::current(
+        &state::load_rewards_pool_params(storage, pool_id.clone())?,
+        cur_block_height,
+    )?;
```
cgorenflo, line 67: create a `state::current_epoch(storage, pool_id, block_height)` function to wrap this?

contracts/rewards/src/contract/execute.rs
```diff
@@ -114,12 +119,45 @@ fn iterate_epoch_tallies<'a>(
...
-pub(crate) fn update_params(
+pub(crate) fn create_pool(
+    storage: &mut dyn Storage,
+    params: Params,
+    block_height: u64,
+    pool_id: &PoolId,
+) -> Result<(), ContractError> {
+    if state::load_rewards_pool(storage, pool_id.clone()).is_ok() {
+        return Err(ContractError::RewardsPoolAlreadyExists.into());
+    }
```
cgorenflo, line 130: create a `state::hash_rewards_pool` wrapper and do
```
ensure!(state::has_rewards_pool(storage, pool_id), ContractError::RewardsPoolAlreadyExists)
```

contracts/rewards/src/contract/execute.rs
```diff
@@ -114,12 +119,45 @@ fn iterate_epoch_tallies<'a>(
...
+    };
+
+    state::save_rewards_pool(storage, &pool)
+}
+
+pub(crate) fn update_pool_params(
     storage: &mut dyn Storage,
     new_params: Params,
     block_height: u64,
+    pool_id: &PoolId,
```
cgorenflo, line 155: put it after storage

contracts/rewards/src/contract/execute.rs
```diff
@@ -114,12 +119,45 @@ fn iterate_epoch_tallies<'a>(
...
 ) -> Result<(), ContractError> {
-    let cur_epoch = Epoch::current(&state::load_params(storage), block_height)?;
+    let cur_epoch = Epoch::current(
+        &state::load_rewards_pool_params(storage, pool_id.clone())?,
+        block_height,
+    )?;
```
cgorenflo, line 160: use the wrapper mentioned above

contracts/rewards/src/contract/migrations/v1_0_0.rs
```diff
@@ -0,0 +1,174 @@
...
+        )?;
+    }
+    PARAMS.remove(storage);
+
+    Ok(())
+}
+
+const POOLS: Map<PoolId, RewardsPool> = Map::new("pools");
+
+pub(crate) fn get_all_pools(storage: &mut dyn Storage) -> Result<Vec<RewardsPool>, ContractError> {
```
cgorenflo, line 41: why is this public?

contracts/rewards/src/contract/migrations/v1_0_0.rs
```diff
@@ -0,0 +1,174 @@
...
+        .range(storage, None, None, Order::Ascending)
+        .map(|res| res.map(|(_, pool)| pool).map_err(|err| err.into()))
+        .collect::<Result<Vec<_>, _>>()
+}
+#[deprecated(since = "1.0.0", note = "only used during migration")]
+pub const PARAMS: Item<ParamsSnapshot> = Item::new("params");
+
+#[cw_serde]
+#[deprecated(since = "1.0.0", note = "only used during migration")]
+pub struct RewardsPool {
```
cgorenflo, line 52: we can make these private here too

contracts/rewards/src/contract/migrations/v1_0_0.rs
```diff
@@ -0,0 +1,174 @@
...
+    use cosmwasm_std::{Addr, DepsMut, Env, MessageInfo, Response, Uint128};
+
+    use super::{RewardsPool, PARAMS, POOLS};
+    use crate::contract::migrations::v1_0_0;
+    use crate::contract::CONTRACT_NAME;
+    use crate::msg::{InstantiateMsg, Params};
+    use crate::state::{self, Config, Epoch, ParamsSnapshot, PoolId, CONFIG};
+
+    #[deprecated(since = "0.4.0", note = "only used during migration tests")]
+    fn instantiate(
```
cgorenflo, line 70: move to the end

contracts/rewards/src/error.rs
```diff
@@ -1,9 +1,12 @@
 use axelar_wasm_std::IntoContractError;
-use cosmwasm_std::OverflowError;
+use cosmwasm_std::{OverflowError, StdError};
 use thiserror::Error;
 
 #[derive(Error, Debug, PartialEq, IntoContractError)]
 pub enum ContractError {
+    #[error(transparent)]
+    Std(#[from] StdError),
```
cgorenflo, line 8: where is this needed? As far as I can tell it's wrapped everywhere already

contracts/rewards/src/msg.rs
```diff
@@ -59,12 +60,18 @@ pub enum ExecuteMsg {
 
     /// Start a new reward pool for the given contract if none exists. Otherwise, add tokens to an existing reward pool.
```
cgorenflo, line 61: is this still correct?

contracts/rewards/src/state.rs
cgorenflo, line 1: can you move all the storage entries to the top so it's easier to see what we actually store?

contracts/rewards/src/contract/migrations/v1_0_0.rs
cjcobb23: mistake, shouldn't be public

contracts/rewards/src/contract.rs
cjcobb23: I'd rather leave it here, since some migrations are just code updates and have no migration logic, so the only place to set the version would be here. I'd rather it just always be set here, to keep things simple.

contracts/rewards/src/error.rs
cjcobb23: Map::update requires this for some reason: https://docs.rs/cw-storage-plus/latest/cw_storage_plus/struct.Map.html#method.update

contracts/rewards/src/msg.rs
cjcobb23: no, fixed it

cgorenflo (approved):

PR #582, opened 2024-08-12
fix(contracts): restrict crate visibility

milapsheth:
## Description

[AXE-4646]

## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Other Notes

[AXE-4646]: https://axelarnetwork.atlassian.net/browse/AXE-4646?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/axelarnet-gateway/src/lib.rs
```diff
@@ -1,7 +1,7 @@
 pub mod contract;
 pub mod events;
 pub mod msg;
-pub mod state;
+mod state;
```
milapsheth, line 4: Do we want to re-export state::Error publicly?

contracts/axelarnet-gateway/src/contract/execute.rs
```diff
@@ -13,7 +13,7 @@ use crate::executable::AxelarExecutableClient;
 use crate::state::{self};
 
 #[allow(clippy::too_many_arguments)]
-pub(crate) fn call_contract(
+pub fn call_contract(
```
milapsheth, line 16: It's more common to manage visibility at a level higher since the module doesn't need to worry about whether the method is used only within the crate or externally.

contracts/service-registry/src/lib.rs
```diff
@@ -2,6 +2,10 @@ pub mod contract;
 pub mod error;
 pub mod helpers;
 pub mod msg;
-pub mod state;
+mod state;
+
+pub use state::{
+    AuthorizationState, BondingState, Service, Verifier, WeightedVerifier, VERIFIER_WEIGHT,
```
milapsheth, line 8: Should we consider defining public types in `types.rs` module instead of re-exporting definitions from state?

contracts/multisig/src/types.rs
```diff
@@ -44,7 +45,7 @@ impl TryFrom<HexBinary> for MsgToSign {
             });
         }
 
-        Ok(MsgToSign::unchecked(other))
+        Ok(MsgToSign(other))
```
cgorenflo, line 48: why change this?

contracts/service-registry/src/lib.rs
milapsheth: Some of these types are being returned as part of queries as well

cgorenflo: I prefer re-exporting

contracts/gateway/tests/contract.rs
```diff
@@ -314,8 +313,8 @@ fn reject_reroute_outgoing_message_with_different_contents() {
     );
     assert!(response.is_err_and(|err| err_contains!(
         err.report,
-        state::Error,
-        state::Error::MessageMismatch { .. }
+        Error,
+        Error::RouteOutgoingMessages
```
cgorenflo, line 317: why did this change?

contracts/multisig/src/lib.rs
```diff
@@ -15,7 +15,7 @@ mod secp256k1;
 #[cfg(feature = "ed25519")]
 mod ed25519;
 
-#[cfg(feature = "test")]
+#[cfg(any(test, feature = "test"))]
```
cgorenflo, line 18: what does this do?

contracts/multisig/src/types.rs
milapsheth: `MsgToSign::unchecked` is now only defined for tests. So, this needs to be constructed directly.

contracts/gateway/tests/contract.rs
milapsheth: `state::Error` isn't exposed anymore, see below. Thoughts?
https://github.com/axelarnetwork/axelar-amplifier/pull/582#discussion_r1714140803

contracts/axelarnet-gateway/src/lib.rs
cgorenflo: I don't think it's necessary

cgorenflo: Maybe just for tests? I don't think other crates should use them, but in tests it is very helpful to ensure the correct error in a type-safe manner

milapsheth: Ok, it's not needed for unit tests within the same crate. And due to the limitation of Rust, exporting it to others will require a `test` feature flag. We can add that when needed (mainly for integration tests)

milapsheth: Ok, it's a bit ugly because of the same limitation of Rust which requires explicit feature gating for exposing test only methods.

contracts/gateway/tests/contract.rs
```diff
@@ -13,6 +13,8 @@ use cosmwasm_std::{
 };
 use gateway::contract::*;
```
milapsheth, line 14: Maybe these should just be unit tests instead of integration tests, similar to what we do for other contracts

cgorenflo (approved):

PR #583, opened 2024-08-13
refactor(minor-axelarnet-gateway): simplify the initial implementation

cgorenflo:
AXE-4654


contracts/axelarnet-gateway/src/state.rs
```diff
@@ -53,286 +30,297 @@ pub enum Error {
...
 
+#[cw_serde]
+pub(crate) struct Config {
+    pub chain_name: ChainName,
+    pub router: Addr,
+}
+
+#[cw_serde]
+pub enum ExecutableMessage {
+    NotExecuted(Message),
```
cjcobb23, line 41:
```suggestion
    Approved(Message),
```

contracts/axelarnet-gateway/src/state.rs
```diff
@@ -53,286 +30,297 @@ pub enum Error {
...
     storage: &mut dyn Storage,
-    key: CrossChainId,
+    cc_id: &CrossChainId,
     msg: &Message,
 ) -> Result<(), Error> {
-    match SENT_MESSAGES.may_load(storage, key.clone())? {
-        Some(_) => Err(Error::MessageAlreadyExists(key)),
-        None => SENT_MESSAGES.save(storage, key, msg).map_err(Error::from),
+    // these cc IDs are generated by the gateway and should be unique.
+    // If there is a collision, there must be a bug in the logic of the caller.
```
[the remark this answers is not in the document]
cgorenflo: I think at this location it should be an error, because this function cannot control how the cc_id is generated. We could panic further up the stack instead.

contracts/axelarnet-gateway/src/state.rs
cgorenflo: I don't really like `Approved`, because it's not immediately clear what it means, but I'll follow the team quorum on this

contracts/axelarnet-gateway/src/state.rs
```diff
@@ -1,39 +1,16 @@
...
-const SENT_MESSAGES_NAME: &str = "sent_messages";
-const SENT_MESSAGES: Map<CrossChainId, Message> = Map::new(SENT_MESSAGES_NAME);
-
-const RECEIVED_MESSAGES_NAME: &str = "received_messages";
-const RECEIVED_MESSAGES: Map<CrossChainId, MessageWithStatus> = Map::new(RECEIVED_MESSAGES_NAME);
+const CONFIG: Item<Config> = Item::new("config");
+const MESSAGES_FROM_CONTRACT_CALL: Map<&CrossChainId, Message> = Map::new("sent_messages");
+pub(crate) const MESSAGES_FROM_CONTRACT_CALL_COUNTER: Counter<u32> =
+    Counter::new("sent_message_counter");
+const MESSAGES_TO_EXECUTE: Map<&CrossChainId, ExecutableMessage> = Map::new("received_messages");
```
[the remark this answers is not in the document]
cgorenflo: I shortened the names a bit

contracts/axelarnet-gateway/src/clients/external.rs
```diff
@@ -11,36 +12,35 @@ pub struct AxelarExecutableMsg {
...
 
-impl<'a> AxelarExecutableClient<'a> {
     pub fn execute(
         &self,
         cc_id: CrossChainId,
         source_address: Address,
         payload: HexBinary,
     ) -> WasmMsg {
-        self.client
-            .execute(&AxelarExecutableExecuteMsg::Execute(AxelarExecutableMsg {
```
milapsheth, line 39: we need this wrapping since the app does this wrapping: `ExecuteMsg::Execute(AxelarExecutableMsg)`

contracts/axelarnet-gateway/src/clients/external.rs
```diff
@@ -11,36 +12,35 @@ pub struct AxelarExecutableMsg {
     pub payload: HexBinary,
 }
 
-/// Crate-specific `ExecuteMsg` type wraps the `AxelarExecutableMsg` for the AxelarExecutable client.
-#[cw_serde]
-pub(crate) enum AxelarExecutableExecuteMsg {
-    /// Execute the message at the destination contract with the corresponding payload.
-    Execute(AxelarExecutableMsg),
+pub struct PayloadExecutor<'a> {
```
milapsheth, line 15:
```suggestion
pub struct AxelarExecutableClient<'a> {
```
I'd prefer to still call this AxelarExecutable client to match the terminology used by external gateways. E.g. https://github.com/axelarnetwork/axelar-gmp-sdk-solidity/blob/main/contracts/interfaces/IAxelarGMPExecutable.sol

Makes it easier for devrel/docs since people are used to the EVM side.

cgorenflo: since we're not calling it `IAxelarGMPExecutable` I think it's better to have a name that actually means something 😉

cgorenflo: I renamed it though, let me know what you think

milapsheth: Linked to the another one, here's the [AxelarExecutable](https://github.com/axelarnetwork/axelar-gmp-sdk-solidity/blob/main/contracts/executable/AxelarExecutable.sol#L8). `Ixyz` is just the solidity convention for an interface.

Similar to `CallContract`, I prefer reusing existing terminology for GMP, instead of using new ones, and `CrossChainExecutor` isn't really much better in my mind (is it executing on a remote chain or locally, is it some app built on top of GMP etc.)

contracts/axelarnet-gateway/src/contract/execute.rs
```diff
@@ -1,76 +1,177 @@
...
-    router: &Router,
-    chain_name: ChainName,
     sender: Addr,
-    destination_chain: ChainName,
-    destination_address: Address,
-    payload: HexBinary,
+    call_contract: CallContractData,
 ) -> Result<Response, Error> {
-    let counter = state::increment_msg_counter(store).change_context(Error::InvalidStoreAccess)?;
+    let payload = call_contract.payload.clone();
```
milapsheth, line 76: feels unnecessary to have to clone the whole payload, just because of how `into_message` works even though it only uses it as a slice. Maybe it takes a reference instead, or change it to something like `into_message_and_payload` that returns `(msg, payload)`.

contracts/axelarnet-gateway/src/clients/external.rs
```diff
@@ -11,39 +12,47 @@ pub struct AxelarExecutableMsg {
...
-            .execute(&AxelarExecutableExecuteMsg::Execute(AxelarExecutableMsg {
+            .execute(&ExecuteMsg::Execute(AxelarExecutableMsg {
                 cc_id,
                 source_address,
                 payload,
             }))
     }
 }
 
+impl<'a> From<client::Client<'a, ExecuteMsg, ()>> for CrossChainExecutor<'a> {
```
milapsheth, line 50: I like including `Client` suffix in the names to make it obvious it's a client

contracts/axelarnet-gateway/src/clients/external.rs
```diff
@@ -11,39 +12,47 @@ pub struct AxelarExecutableMsg {
...
+    pub fn new(deps: Deps<'a>, destination: &str) -> error_stack::Result<Self, address::Error> {
+        let destination = address::validate_cosmwasm_address(deps.api, destination)?;
+        Ok(CrossChainExecutor {
+            client: client::Client::new(deps.querier, destination),
+        })
+    }
 
-impl<'a> AxelarExecutableClient<'a> {
-    pub fn execute(
+    pub fn prepare_execute_msg(
```
milapsheth, line 35:
```suggestion
    pub fn execute(
```
We name the client methods by the msg that they create. Should be consistent. Alternatively, don't create a new client struct, and define a single method `prepare_execute_msg` that constructs the client `client::Client<'a, AxelarExecutableExecuteMsg, ()>` and returns the msg in one-shot

contracts/axelarnet-gateway/src/contract/execute.rs
```diff
@@ -79,297 +180,37 @@ pub fn receive_messages(
...
-    querier: QuerierWrapper,
-    cc_id: CrossChainId,
-    payload: HexBinary,
-) -> Result<Response, Error> {
-    let msg = state::set_msg_as_executed(store, cc_id.clone())
-        .change_context(Error::SetMessageStatusExecutedFailed(cc_id))?;
+    let msgs: Vec<_> = msgs
+        .into_iter()
+        .unique()
+        .filter_map(|msg| verify_message(store, msg).transpose())
```
milapsheth, line 192: need to see if return value from verify can be avoided. otherwise may be rename it to something that implies a return value is being used

contracts/axelarnet-gateway/src/state.rs
```diff
@@ -51,288 +27,103 @@ pub enum Error {
...
-        // Test setting a non-existent message
-        let non_existent_id = CrossChainId::new("non-existent", "id").unwrap();
-        let result = super::set_msg_as_executed(deps.as_mut().storage, non_existent_id.clone());
-        assert_eq!(result, Err(Error::MessageNotApproved(non_existent_id)));
-    }
+/// Update the status of a message to executed if it is in approved status, error otherwise.
+pub fn update_as_executed(
+    storage: &mut dyn Storage,
+    cc_id: &CrossChainId,
+    action: impl FnOnce(Message) -> Result<Message, Error>,
```
milapsheth, line 118: This is not very readable. Maybe just pass in the msg and payload as a reference, and check the payload_hash in here?

contracts/axelarnet-gateway/src/lib.rs
```diff
@@ -3,8 +3,6 @@ pub mod events;
 pub mod msg;
 mod state;
 
-mod client;
-pub use client::Client;
-
-mod executable;
-pub use executable::AxelarExecutableMsg;
+mod clients;
+pub use clients::{AxelarExecutableMsg, ExecuteMsg as ExternalExecuteMsg, GatewayClient};
```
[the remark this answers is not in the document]
cgorenflo: In order to deserialize it I have to expose it. Alternatively I could redefine `ExecuteMsg` in the tests

contracts/axelarnet-gateway/src/clients/external.rs
cgorenflo: that's the point though, it's not really a client.the only thing it does is prepare a message, there are no queries, and the message only works with other contracts by convention

contracts/axelarnet-gateway/src/contract/execute.rs
cgorenflo: I don't get it

contracts/axelarnet-gateway/src/state.rs
cgorenflo: I don't want to move logic concerns into the data layer if possible. This follows the same pattern as the update function in cw_storage_plus. I'll think about it some more

contracts/axelarnet-gateway/src/clients/external.rs
milapsheth: I suggested an alternate approach here that makes it more clear it's not a client:
https://github.com/axelarnetwork/axelar-amplifier/pull/583#discussion_r1729519076

contracts/axelarnet-gateway/src/lib.rs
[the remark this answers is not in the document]
milapsheth: I'm ok with redefining in tests, since apps we'll defining there own ExecuteMsg anyways

contracts/axelarnet-gateway/src/contract/execute.rs
milapsheth: Sorry, referring to the `verify_message` returning a Message discussion

contracts/axelarnet-gateway/src/state.rs
```diff
@@ -51,288 +27,103 @@ pub enum Error {
...
+) -> Result<Message, Error> {
+    let msg = match may_load_executable_msg(storage, cc_id)? {
+        None => Err(Error::MessageNotApproved(cc_id.clone())),
+        Some(ExecutableMessage::Executed(_)) => Err(Error::MessageAlreadyExecuted(cc_id.clone())),
+        Some(ExecutableMessage::Approved(msg)) => Ok(action(msg)?),
```
milapsheth, line 123:
```suggestion
        Some(ExecutableMessage::Approved(msg)) if validate(msg) => Ok(msg),
```
instead of an arbitrary action that could transform the message, only provide a validation function?

contracts/axelarnet-gateway/src/state.rs
milapsheth: Yeah, it's not ideal. How about this suggestion?
https://github.com/axelarnetwork/axelar-amplifier/pull/583/files#r1729954821

contracts/axelarnet-gateway/src/contract/execute.rs
milapsheth: How about creating a `try_filter` similar to the `try_map` that we have that handles this internally. Then the usage here is more clear, and `verify_message` just verifies instead of transforming the value.
`.try_filter(|msg| verify_message(store, &msg))?;`

contracts/axelarnet-gateway/src/clients/external.rs
cgorenflo: this client should not be exposed anyway. I ended up calling it `external::Client`. I believe that's clear enough

milapsheth (approved):

PR #585, opened 2024-08-15
feat: integrate address format with SuiAddress

fish-sammy:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


packages/sui-types/src/lib.rs
```diff
@@ -32,13 +39,17 @@ impl FromStr for SuiAddress {
...
+        match s.chars().any(|c| c.is_uppercase()) {
+            true => Err(report!(Error::InvalidAddressHex(s.to_string()))),
+            false => hex::decode(
+                s.strip_prefix(ADDRESS_PREFIX)
+                    .ok_or(report!(Error::InvalidAddressHex(s.to_string())))?,
+            )
+            .change_context(Error::InvalidAddressHex(s.to_string()))?
+            .as_slice()
+            .try_into(),
+        }
```
cgorenflo, line 52:
```suggestion
        ensure!(s.chars().all(|c| c.is_lowercase()), Error::InvalidAddressHex(s.to_string()));
        
        hex::decode(
                s.strip_prefix(ADDRESS_PREFIX)
                    .ok_or(report!(Error::InvalidAddressHex(s.to_string())))?,
            )
            .change_context(Error::InvalidAddressHex(s.to_string()))?
            .as_slice()
            .try_into(),

```

cgorenflo (approved):

packages/sui-types/src/lib.rs
```diff
@@ -32,6 +40,12 @@ impl FromStr for SuiAddress {
     type Err = Report<Error>;
 
     fn from_str(s: &str) -> Result<Self, Error> {
+        // disallow uppercase characters for the sui addresses
+        ensure!(
```
milapsheth, line 46: Add unit test for failing on correctly hex encoded but upper and mixed case strings

packages/sui-types/src/lib.rs
```diff
@@ -1,13 +1,20 @@
...
-use crate::error::Error;
+use thiserror::Error;
 
 const ADDRESS_PREFIX: &str = "0x";
 const SUI_ADDRESS_LENGTH: usize = 32;
 
+#[derive(Error, Debug)]
+pub enum Error {
+    #[error("invalid address bytes: {:?}", .0)]
+    InvalidAddressBytes(Vec<u8>),
```
milapsheth, line 13:
```suggestion
    InvalidSuiAddressLength(Vec<u8>),
```
perhaps?

packages/sui-types/src/lib.rs
```diff
@@ -1,13 +1,20 @@
...
+    #[error("invalid address hex: {0}")]
+    InvalidAddressHex(String),
```
milapsheth, line 15:
```suggestion
    InvalidSuiAddress(String),
```
since this error is used for a few checks (lowercase + hex encoding + prefix stripping)

packages/sui-gateway/src/lib.rs
```diff
@@ -1,3 +1,184 @@
...
+use sha3::{Digest, Keccak256};
+use sui_types::SuiAddress;
+
 pub mod error;
-pub mod gateway;
+pub mod events;
+
+use error::Error;
+
+#[repr(u8)]
```
milapsheth, line 17: Move this to a different file, say `types.rs`? I don't like making `lib.rs` files complex, instead of keeping their focus on module inclusions, visibility restriction/re-exports.

packages/sui-types/src/lib.rs
```diff
@@ -32,13 +39,19 @@ impl FromStr for SuiAddress {
     type Err = Report<Error>;
 
     fn from_str(s: &str) -> Result<Self, Error> {
```
milapsheth, line 41: This means that we'll enforce lower case for both source address and destination address. While it's not strictly needed for destination, I prefer the consistency and it's straightforward to adhere to it

milapsheth (approved):

packages/sui-types/src/lib.rs
cgorenflo: it's already in the sui crate, no need to add sui to the variant name

packages/sui-types/src/lib.rs
cgorenflo: it's already in the sui crate, no need to add sui to the variant name

packages/sui-types/src/lib.rs
cgorenflo: I'm not sure what your suggestion is

milapsheth: Sorry, no suggestion, just a callout that we're restricting Sui destination addresses more than before. I think that's fine for Sui. We don't do the same for EVM though (restrict EVM destination address to be EIP 155 checksum encoded since it's hard to encode to that on-chain)

packages/sui-gateway/src/lib.rs
milapsheth: It's not a random collection. The crate is `sui-gateway`, and these are the Sui gateway type definitions, so `sui-gateway::types` makes sense to me. We can make it more verbose like `gateway-types.rs` if you like (could also break it out into message.rs and proof.rs type definitions). But this crate already has a few modules, so don't see the point of bloating lib.rs

packages/sui-types/src/lib.rs
milapsheth: Sure, I'd still change it to `InvalidAddress` and not just `InvalidAddressHex` since some checks aren't about hex encoding.

packages/sui-types/src/lib.rs
milapsheth: Sure, can make it InvalidAddressLength

packages/sui-types/src/lib.rs
fish-sammy: Ya it's better to have the consistency througout amplifier.

packages/sui-types/src/lib.rs
fish-sammy: It's in the `address.rs` tests

packages/sui-gateway/src/lib.rs
fish-sammy: I agree with Chris. I don't like having mods in a crate before it needs the separation.

PR #588, opened 2024-08-15
fix(minor-service-registry): prevent zero bond

cjcobb23:
https://axelarnetwork.atlassian.net/browse/AXE-4656


milapsheth (approved):

contracts/service-registry/src/contract/execute.rs
```diff
@@ -84,25 +85,30 @@ pub fn bond_verifier(
...
-            .iter()
-            .find(|coin| coin.denom == service.bond_denom)
-            .ok_or(ContractError::WrongDenom)?
-            .amount
+    let bond: Option<nonempty::Uint128> = if !info.funds.is_empty() {
+        Some(
+            info.funds
+                .iter()
+                .find(|coin| coin.denom == service.bond_denom)
+                .ok_or(ContractError::NoFundsToBond)?
```
cgorenflo, line 93: this is misleading. This will also trigger if someone sends the wrong denomination

packages/axelar-wasm-std/src/nonempty/uint.rs
```diff
@@ -111,6 +111,12 @@ impl From<Uint128> for cosmwasm_std::Uint128 {
     }
 }
 
+impl AsRef<cosmwasm_std::Uint128> for Uint128 {
+    fn as_ref(&self) -> &cosmwasm_std::Uint128 {
+        &self.0
+    }
+}
+
```
cgorenflo, line 119: use Deref instead of `AsRef`, you're not returning a reference here

cgorenflo (changes requested):

packages/into-inner-derive/src/lib.rs
```diff
@@ -0,0 +1,31 @@
...
+
+#[proc_macro_derive(IntoInner)]
+pub fn into_inner_derive(input: TokenStream) -> TokenStream {
+    let ast: DeriveInput = syn::parse(input).unwrap();
+
+    let name = &ast.ident;
+    let ty = match ast.data {
+        syn::Data::Struct(val) => match val.fields {
+            Fields::Unnamed(fields) if fields.unnamed.len() == 1 => {
+                fields.unnamed.first().map(|field| field.ty.clone())
```
cgorenflo, line 13: is the clone necessary?

cgorenflo (approved):

packages/into-inner-derive/src/lib.rs
cjcobb23: yes, otherwise it's a reference that doesn't live long enough to be returned. I changed it to `to_owned()` to make it clearer though

PR #589, opened 2024-08-19
feat(minor-multisig-prover): stellar xdr encoding message to sign

haiyizxx:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


contracts/multisig-prover/src/error.rs
```diff
@@ -74,4 +74,7 @@ pub enum ContractError {
 
     #[error("payload does not match the stored value")]
     PayloadMismatch,
+
+    #[error(transparent)]
+    StellarXdrError(#[from] stellar_xdr::curr::Error),
```
cgorenflo, line 79: I don't see that used anywhere and don't think there should be a Stellar-specific error at this level

cgorenflo (commented):
can we put all those gateway crates into their own subfolder? I find them confusing to encounter in the main packages folder

PR #592, opened 2024-08-19
chore(ampd): include libssl3 in docker image

jcs47:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

### DIFF
--- a/ampd/Dockerfile
+++ b/ampd/Dockerfile
@@ -1,4 +1,4 @@
-FROM rust:1.78-bookworm as builder
+FROM rust:1.78-trixie as builder
 RUN apt-get update && apt-get install -y clang protobuf-compiler cmake
 
 WORKDIR /ampd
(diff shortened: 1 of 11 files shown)


ampd/Dockerfile
```diff
@@ -22,6 +22,7 @@ COPY ./ampd/build.rs ./ampd/build.rs
 RUN cargo install --locked --path ./ampd
 
 FROM debian:bookworm-slim AS runner
+RUN apt update && apt install libssl3
```
milapsheth, line 25: why is this needed now? There's already an apt update and install at the top

milapsheth: Looks like the previous apt install is for a different image

jcs47: yes, currently we are unable to execute ampd in the runner image because libssl3 is missing.

milapsheth (approved):

ampd/Dockerfile
milapsheth: maybe because we don't pin the bookworm-slim version

PR #598, opened 2024-08-26
feat(minor-multisig-prover): stellar encode execute data

haiyizxx:
## Description
AXE-4667

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


contracts/multisig-prover/src/encoding/mod.rs
```diff
@@ -70,7 +70,7 @@ where
     let recovery_transform = match encoder {
         Encoder::Abi => add_27,
         Encoder::Bcs => no_op,
-        Encoder::StellarXdr => no_op,
+        _ => panic!("unsupported encoder"),
```
cgorenflo, line 73: is it not being called by stellar?

haiyizxx: yes, stellar use ed225519, this function is for ecdsa recover. probably the function itself needs some factor.

cgorenflo (approved):

PR #599, opened 2024-08-27
feat(nexus-gateway)!: call contract with token

fish-sammy:
## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

### DIFF
--- a/contracts/nexus-gateway/src/contract/execute.rs
+++ b/contracts/nexus-gateway/src/contract/execute.rs
@@ -1,9 +1,69 @@
-use cosmwasm_std::{to_json_binary, Response, Storage, WasmMsg};
+use axelar_wasm_std::msg_id::HexTxHashAndEventIndex;
+use axelar_wasm_std::nonempty;
+use cosmwasm_std::{BankMsg, HexBinary, MessageInfo, QuerierWrapper, Response, Storage, Uint256};
+use error_stack::{bail, ResultExt};
+use router_api::{Address, ChainName};
+use sha3::{Digest, Keccak256};
 
 use crate::error::ContractError;
+use crate::state::load_config;
 use crate::{nexus, state};
 
 type Result<T> = error_stack::Result<T, ContractError>;
+
+pub fn call_contract_with_token(
+    storage: &mut dyn Storage,
+    querier: QuerierWrapper,
+    block_height: u64,
+    info: MessageInfo,
+    destination_chain: ChainName,
+    destination_address: Address,
+    payload: HexBinary,
+) -> Result<Response<nexus::Message>> {
+    let config = load_config(storage)?;
+    let axelarnet_gateway: axelarnet_gateway::Client =
+        client::Client::new(querier, config.axelar_gateway).into();
+    let source_address: Address = info
+        .sender
+        .into_string()
+        .parse()
+        .expect("invalid sender address");
+    let token = match info.funds.as_slice() {
+        [token] => token.clone(),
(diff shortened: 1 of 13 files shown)


contracts/nexus-gateway/src/contract/execute.rs
```diff
@@ -1,10 +1,70 @@
...
+        .parse()
+        .expect("invalid sender address");
+    let token = match info.funds.as_slice() {
+        [token] => token.clone(),
+        _ => bail!(ContractError::InvalidToken(info.funds)),
+    };
+    // TODO: Retrieve the actual tx hash and event index from core, since cosmwasm doesn't provide it. Use the block height as the placeholder in the meantime.
+    let tx_hash = Uint256::from(block_height).to_be_bytes();
+    let event_index = 0u32;
+    let id = HexTxHashAndEventIndex::new(tx_hash, event_index);
```
cjcobb23, line 38: We can't use only block height, txs in the same block will have the same message id

contracts/nexus-gateway/src/state.rs
```diff
@@ -40,4 +40,5 @@ pub fn is_message_routed(storage: &dyn Storage, id: &CrossChainId) -> Result<boo
 pub struct Config {
     pub nexus: Addr,
     pub router: Addr,
+    pub axelar_gateway: Addr,
```
cjcobb23, line 43: this requires a migration, right?

contracts/nexus-gateway/src/msg.rs
```diff
@@ -1,17 +1,26 @@
...
 
 #[cw_serde]
 #[derive(EnsurePermissions)]
 pub enum ExecuteMsg {
+    #[permission(Any)]
+    CallContractWithToken {
+        destination_chain: ChainName,
+        destination_address: Address,
+        payload: HexBinary,
+    },
```
cjcobb23, line 23: we should add a comment saying this only works for legacy chains. Also, what is the behavior if the destination chain is not known to core? Will the caller receive an error?

cjcobb23 (changes requested):

contracts/nexus-gateway/src/contract/execute.rs
fish-sammy: This is known, and this is the case in axelarnet-gateway for the moment as well. Another JIRA ticket is out there for implementing this in core so that contracts can retrieve.

contracts/nexus-gateway/src/state.rs
fish-sammy: Yes, I'll do it separately.

contracts/nexus-gateway/src/msg.rs
fish-sammy: The tx would fail. Well, everything in the nexus gateway works only for the legacy chains, not just this msg type.

cjcobb23: yeah would be good to add some comments so that this is clear

contracts/nexus-gateway/src/contract/execute.rs
fish-sammy: which is equally as bad cuz the counter isn't global and even worse hidden. Core should handle be able to return both the tx id and index so that the index can be unique in the context of the same tx. I do not want to extend the mistake.

cjcobb23 (approved):

PR #600, opened 2024-08-28
feat(minor-service-registry): add client for queries

cjcobb23:


contracts/service-registry/src/client.rs
```diff
@@ -0,0 +1,169 @@
...
+use router_api::ChainName;
+
+use crate::msg::{ExecuteMsg, QueryMsg};
+use crate::{Service, Verifier, WeightedVerifier};
+
+type Result<T> = error_stack::Result<T, Error>;
+
+#[derive(thiserror::Error, Debug, PartialEq)]
+pub enum Error {
+    #[error("failed query the service registry for active verifiers for service {service_name} and chain {chain_name}")]
```
cgorenflo, line 11:
```suggestion
    #[error("failed to query the service registry for active verifiers for service {service_name} and chain {chain_name}")]
```

contracts/service-registry/src/client.rs
```diff
@@ -0,0 +1,169 @@
...
+    ) -> Result<Vec<WeightedVerifier>> {
+        self.client
+            .query(&QueryMsg::ActiveVerifiers {
+                service_name: service_name.clone(),
+                chain_name: chain_name.clone(),
+            })
+            .change_context_lazy(|| Error::QueryActiveVerifiers {
+                service_name,
+                chain_name,
+            })
```
cgorenflo, line 53: not sure lazy is really helpful in these cases. You still need to clone the parameters and instead of instantiating an error you instantiate a closure

contracts/service-registry/src/client.rs
```diff
@@ -0,0 +1,169 @@
...
+    fn from(client: client::Client<'a, ExecuteMsg, QueryMsg>) -> Self {
+        Client { client }
+    }
+}
+
+pub struct Client<'a> {
+    client: client::Client<'a, ExecuteMsg, QueryMsg>,
+}
+
+impl<'a> Client<'a> {
```
cgorenflo, line 37: suggestion for the error handling:
implement `From` for the QueryMsg for conversion, then you can do something like:
```
let msg = ...;
self.client.query(&msg).change_context_lazy(|| msg.into())
```

contracts/service-registry/src/client.rs
```diff
@@ -0,0 +1,169 @@
...
+        let client: Client = client::Client::new(QuerierWrapper::new(&querier), addr).into();
+        let service_name = "verifiers".to_string();
+        let res = client.service(service_name.clone());
+        assert!(res.is_err());
+
+        assert_eq!(
+            res.unwrap_err().to_string(),
+            Error::QueryService(service_name).to_string()
+        );
+    }
```
cgorenflo, line 138: consider comparing with a golden files. That might make adding the non-error cases easier too

cgorenflo (approved):

PR #601, opened 2024-08-28
feat(minor-voting-verifier): use service registry client

cjcobb23:
https://axelarnetwork.atlassian.net/browse/AXE-4716


contracts/voting-verifier/src/contract.rs
```diff
@@ -80,19 +79,28 @@ pub fn execute(
 }
 
 #[cfg_attr(not(feature = "library"), entry_point)]
-pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> StdResult<Binary> {
+pub fn query(
```
cgorenflo, line 82: consider moving the `map_err` calls to the end:
```
match msg {
...
}.map_error(|err| err.into())
```
or 
```

match msg {
...
}?.then(Ok)
```

contracts/voting-verifier/src/contract/execute.rs
```diff
@@ -256,19 +275,25 @@ pub fn vote(
 }
 
 pub fn end_poll(deps: DepsMut, env: Env, poll_id: PollId) -> Result<Response, ContractError> {
-    let config = CONFIG.load(deps.storage)?;
+    let config = CONFIG
+        .load(deps.storage)
+        .change_context(ContractError::StorageError)?;
```
cgorenflo, line 280: should we change loading the config to a panic instead of result? Unless there is something disastrously wrong with the contract, the config should always be loadable

contracts/voting-verifier/src/contract/execute.rs
```diff
@@ -130,12 +139,15 @@ pub fn verify_messages(
...
-            .map_err(ContractError::from)?;
+            .change_context(ContractError::StorageError)?;
     }
 
     let messages = msgs_to_verify
         .into_iter()
-        .map(|msg| (msg, &config.msg_id_format).try_into())
+        .map(|msg| {
+            TxEventConfirmation::try_from((msg.clone(), &config.msg_id_format))
+                .map_err(|err| report!(err))
```
cgorenflo, line 149: just curious, have you ever tried if `map_err(report!)` works?

contracts/voting-verifier/src/contract/execute.rs
cjcobb23: I am all for that, I dislike treating this as an error

contracts/voting-verifier/src/contract/execute.rs
cjcobb23: Doesn't seem to:
```
150 |                 .map_err(report!)
    |                                 ^ expected one of `(`, `[`, or `{`
```

PR #602, opened 2024-08-29
feat(nexus-gateway): migrate the config

fish-sammy:
## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


contracts/nexus-gateway/src/contract.rs
```diff
@@ -1,29 +1,31 @@
...
 pub fn migrate(
-    deps: DepsMut,
+    mut deps: DepsMut,
     _env: Env,
-    _msg: Empty,
+    msg: MigrateMsg,
 ) -> Result<Response, axelar_wasm_std::error::ContractError> {
-    // any version checks should be done before here
-
-    cw2::set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
```
cjcobb23, line 26: This isn't supposed to be removed

contracts/nexus-gateway/src/contract/migrations/v1_0_0.rs
```diff
@@ -0,0 +1,126 @@
...
+const BASE_VERSION: &str = "1.0.0";
+const CONFIG: Item<Config> = Item::new("config");
+
+pub fn migrate(deps: &mut DepsMut, msg: MigrateMsg) -> Result<(), ContractError> {
+    cw2::assert_contract_version(deps.storage, CONTRACT_NAME, BASE_VERSION)?;
+
+    let axelarnet_gateway = address::validate_cosmwasm_address(deps.api, &msg.axelarnet_gateway)?;
+    migrate_config(deps, axelarnet_gateway)?;
+
+    cw2::set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
```
cjcobb23, line 20: can we set this in the migrate entry point, instead of here? So that way we are consistent with contracts that don't actually have any migration logic, but just need to migrate the code

contracts/nexus-gateway/src/contract/migrations/v1_0_0.rs
```diff
@@ -0,0 +1,126 @@
...
+        state::Config {
+            nexus: config.nexus,
+            router: config.router,
+            axelarnet_gateway,
+        },
+    )
+    .map_err(Into::into)
+}
+
+#[cw_serde]
```
cjcobb23, line 40: Can we mark this as deprecated?
```suggestion
#[cw_serde]
#[deprecated(since = "1.0.0", note = "only used during migration")]
```

contracts/nexus-gateway/src/contract/migrations/v1_0_0.rs
```diff
@@ -0,0 +1,126 @@
...
+    migrate_config(deps, axelarnet_gateway)?;
+
+    cw2::set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
+
+    Ok(())
+}
+
+fn migrate_config(deps: &mut DepsMut, axelarnet_gateway: Addr) -> Result<(), ContractError> {
+    let config = CONFIG.load(deps.storage)?;
+    CONFIG.remove(deps.storage);
```
cjcobb23, line 27: do you actually need to remove?

contracts/nexus-gateway/src/nexus.rs
```diff
@@ -22,6 +22,7 @@ pub struct Message {
     pub source_tx_id: nonempty::Vec<u8>,
     pub source_tx_index: u64,
     pub id: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
```
cjcobb23, line 25: So what does this mean for deserializing older versions of the struct, does it just deserialize to None? If so, can we test this?

contracts/nexus-gateway/src/contract/migrations/v1_0_0.rs
fish-sammy: I don't see how is it consistent https://github.com/axelarnetwork/axelar-amplifier/blob/2acf237d4c8cbe20e844436c2b94259408020597/contracts/voting-verifier/src/contract/migrations/v0_5_0.rs#L26.

contracts/nexus-gateway/src/contract/migrations/v1_0_0.rs
fish-sammy: hmmm I am guessing not.

contracts/nexus-gateway/src/nexus.rs
fish-sammy: There's no "older" version cuz the field was just added in my previous PR. Yes it deserializes `null` or `undefined` or not exist to None. I am not sure if I should test it cuz I don't wanna test every single serde feature.

contracts/nexus-gateway/src/contract.rs
```diff
@@ -1,29 +1,31 @@
...
+    v1_0_0::migrate(deps.borrow_mut(), msg)?;
```
cgorenflo, line 28: why is this necessary?

contracts/nexus-gateway/src/contract/migrations/v1_0_0.rs
```diff
@@ -0,0 +1,127 @@
...
+use cw_storage_plus::Item;
+
+use crate::contract::{CONTRACT_NAME, CONTRACT_VERSION};
+use crate::msg::MigrateMsg;
+use crate::state;
+
+const BASE_VERSION: &str = "1.0.0";
+const CONFIG: Item<Config> = Item::new("config");
+
+pub fn migrate(deps: &mut DepsMut, msg: MigrateMsg) -> Result<(), ContractError> {
```
cgorenflo, line 16: why use a &mut instead of the owned type?

contracts/nexus-gateway/src/contract/migrations/v1_0_0.rs
cgorenflo: we need to come up with a clear migration strategy and then set up consistent rules from there. Right now it's way to easy to release versions with incompatible upgrade paths

contracts/nexus-gateway/src/nexus.rs
cgorenflo: why do we need this in the first place?

fish-sammy: It's not entirely needed. It's just silly to send extra data cross the VM and core.

contracts/nexus-gateway/src/contract/migrations/v1_0_0.rs
fish-sammy: oh cuz I initially was setting the contract version in the contract, but decided later to move it inside for the sake of easier testing. Will make it take over the ownership.

cgorenflo (approved):

PR #606, opened 2024-09-04
fix(ampd): resolve deadlock between broadcaster and confirmer on shutdown

cgorenflo:
AXE-4740
AXE-4741

This resolves the deadlock and adds additional logging on task shut down, so it's easier to understand which process is shutting down

### DIFF
--- a/ampd/src/lib.rs
+++ b/ampd/src/lib.rs
@@ -113,6 +113,21 @@ async fn prepare_app(cfg: Config) -> Result<App<impl Broadcaster>, Error> {
         .await
         .change_context(Error::Broadcaster)?;
 
+    let broadcaster = QueuedBroadcaster::new(
+        broadcaster,
+        broadcast.batch_gas_limit,
+        broadcast.queue_cap,
+        interval(broadcast.broadcast_interval),
+    );
+
+    let tx_confirmer = TxConfirmer::new(
+        service_client,
+        RetryPolicy::RepeatConstant {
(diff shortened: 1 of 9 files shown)


ampd/src/lib.rs
```diff
@@ -113,6 +113,21 @@ async fn prepare_app(cfg: Config) -> Result<App<impl Broadcaster>, Error> {
...
+        broadcast.batch_gas_limit,
+        broadcast.queue_cap,
+        interval(broadcast.broadcast_interval),
+    );
+
+    let tx_confirmer = TxConfirmer::new(
+        service_client,
+        RetryPolicy::RepeatConstant {
+            sleep: broadcast.tx_fetch_interval,
+            max_attempts: broadcast.tx_fetch_max_retries.saturating_add(1).into(),
```
haiyizxx, line 127: don't think we need add 1, attempt_count starts from 0

ampd/src/queue/queued_broadcaster.rs
```diff
@@ -190,42 +160,81 @@ where
...
+                log.events
+                    .iter()
+                    .enumerate()
+                    .for_each(|(event_index, event)| {
+                        debug!(tx_hash, msg_index, event_index, "tx event {:?}", event);
+                    });
+            });
+        }
+        TxStatus::Failure => {
+            warn!(tx_hash, "tx failed");
```
haiyizxx, line 233: would be good log error code and description

haiyizxx (approved):

PR #610, opened 2024-09-09
feat(service-registry-api): add service-registry-api package

maancham:
## Description

AXE-4789

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


cjcobb23:
why do we need this?

cgorenflo:
> why do we need this?

to break the cyclic dependency between service-registry and coordinator

cjcobb23:
> > why do we need this?
> 
> to break the cyclic dependency between service-registry and coordinator

For what exactly? Is there a planned change that will create a cyclic dependency?

packages/service-registry-api/src/primitives.rs
```diff
@@ -0,0 +1,148 @@
...
+                Self {
+                    bonding_state: BondingState::Unbonded,
+                    ..self
+                },
+                amount,
+            )),
+            _ => Err(ContractError::InvalidBondingState(self.bonding_state)),
+        }
+    }
+}
```
cgorenflo, line 102: we'll have to look into how the verifier is used in other contracts. It doesn't make sense to provide this logic outside of the service-registry

packages/service-registry-api/src/primitives.rs
```diff
@@ -0,0 +1,148 @@
...
+
+#[cw_serde]
+pub struct WeightedVerifier {
+    pub verifier_info: Verifier,
+    pub weight: nonempty::Uint128,
+}
+
+/// For now, all verifiers have equal weight, regardless of amount bonded
+pub const VERIFIER_WEIGHT: nonempty::Uint128 = nonempty::Uint128::one();
```
cgorenflo, line 111: state definitely shouldn't go into this crate

packages/service-registry-api/src/primitives.rs
```diff
@@ -0,0 +1,148 @@
...
+    NotAuthorized,
+    Authorized,
+    Jailed,
+}
+
+type ServiceName = String;
+type VerifierAddress = Addr;
+
+pub const SERVICES: Map<&ServiceName, Service> = Map::new("services");
+pub const VERIFIERS: Map<(&ServiceName, &VerifierAddress), Verifier> = Map::new("verifiers");
```
cgorenflo, line 148: no state

cgorenflo (changes requested):

cgorenflo (approved):

PR #611, opened 2024-09-09
feat(minor-voting-verifier): Stellar source address validation

haiyizxx:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


packages/axelar-wasm-std/src/address.rs
```diff
@@ -29,6 +31,13 @@ pub fn validate_address(address: &str, format: &AddressFormat) -> Result<(), Err
             SuiAddress::from_str(address)
                 .change_context(Error::InvalidAddress(address.to_string()))?;
         }
+        AddressFormat::Stellar => {
+            if address != address.to_uppercase() {
+                bail!(Error::InvalidAddress(address.to_string()))
+            }
+            ScAddress::from_str(address)
```
milapsheth, line 38: did you look into this implementation for any malleability concerns for sanity?

packages/axelar-wasm-std/src/address.rs
```diff
@@ -29,6 +31,13 @@ pub fn validate_address(address: &str, format: &AddressFormat) -> Result<(), Err
...
+                .change_context(Error::InvalidAddress(address.to_string()))?;
```
milapsheth, line 39: can you measure the impact on bytecode increase and gas usage due to using stellar libraries? we should add some automation around this

packages/axelar-wasm-std/src/address.rs
```diff
@@ -104,4 +113,25 @@ mod tests {
...
+        // account id
+        let addr = "GA7QYNF7SOWQ3GLR2BGMZEHXAVIRZA4KVWLTJJFC7MGXUA74P7UJVSGZ";
+        assert!(address::validate_address(addr, &address::AddressFormat::Stellar).is_ok());
+
+        let lower_case = addr.to_lowercase();
+        assert!(address::validate_address(&lower_case, &address::AddressFormat::Stellar).is_err());
+
+        // contract
+        let addr = "CA7QYNF7SOWQ3GLR2BGMZEHXAVIRZA4KVWLTJJFC7MGXUA74P7UJUWDA";
+        assert!(address::validate_address(addr, &address::AddressFormat::Stellar).is_ok());
```
milapsheth, line 128: use assert_ok and assert_err_contains

milapsheth (approved):

packages/axelar-wasm-std/src/address.rs
haiyizxx: yes, only thing I come up is non-canonical string representation, it's case insensitive. didn't find anything else.

packages/axelar-wasm-std/src/address.rs
haiyizxx: not much, around 20k bytes. don't think it will increase gas, we validate all source address. the conversion is not expensive.

PR #617, opened 2024-09-10
feat(minor-ampd): stellar verifier set verifier

haiyizxx:
## Description
AXE-4640
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


ampd/src/handlers/config.rs
```diff
@@ -202,6 +206,22 @@ where
...
+        .iter()
+        .filter(|config| matches!(config, Config::StellarMsgVerifier { .. }))
+        .count()
+    {
+        count if count > 1 => Err(de::Error::custom(
+            "only one Stellar verifier set verifier config is allowed",
+        )),
+        _ => Ok(()),
+    }
+}
```
cgorenflo, line 223: can we create a macro for this, so it's `ensure_unique_config!("Stellar verifier set", Config::StellarMsgVerifier)` and can be reused for all other checks as well, or do these validations differ from config to config?

cgorenflo (approved):

ampd/src/handlers/config.rs
haiyizxx: sg, will work on this

PR #618, opened 2024-09-10
feat(minor-coordinator): add client for queries

cjcobb23:


contracts/coordinator/src/lib.rs
```diff
@@ -1,3 +1,4 @@
+pub mod client;
```
cgorenflo, line 1: I suggest to reexport the client instead to prevent `coordinator::client::Client` calls

cgorenflo (approved):

PR #621, opened 2024-09-11
fix(minor-coordinator)!: use String instead of Addr in messages

cjcobb23:
Addr should not be used in the public facing API, because then the addresses bypass any sort of validation. Switched to string.

https://axelarnetwork.atlassian.net/browse/AXE-4842


contracts/coordinator/src/msg.rs
```diff
@@ -16,15 +15,15 @@ pub enum ExecuteMsg {
...
-    SetActiveVerifiers { verifiers: HashSet<Addr> },
+    SetActiveVerifiers { verifiers: HashSet<String> },
 }
 
 #[cw_serde]
 #[derive(QueryResponses)]
 pub enum QueryMsg {
     #[returns(bool)]
-    ReadyToUnbond { worker_address: Addr },
+    ReadyToUnbond { worker_address: String },
```
milapsheth, line 28: looks like this still uses `worker_address` naming

contracts/coordinator/src/msg.rs
```diff
@@ -16,15 +15,15 @@ pub enum ExecuteMsg {
     #[permission(Governance)]
     RegisterProverContract {
         chain_name: ChainName,
-        new_prover_addr: Addr,
+        new_prover_addr: String,
     },
     #[permission(Specific(prover))]
-    SetActiveVerifiers { verifiers: HashSet<Addr> },
+    SetActiveVerifiers { verifiers: HashSet<String> },
```
milapsheth, line 21: We should introduce a `VerifierAddress` type to length restrict verifier addresses and expose a validation function
https://axelarnetwork.atlassian.net/browse/AXE-4586

milapsheth (approved):

contracts/coordinator/src/msg.rs
cjcobb23: yeah, i will do this in a follow up

contracts/coordinator/src/client.rs
```diff
@@ -1,21 +1,23 @@
...
+        "failed to execute ReadyToUnbond query at coordinator contract. verifier_address: {0}"
+    )]
+    ReadyToUnbond(String),
 }
 
 impl From<QueryMsg> for Error {
     fn from(value: QueryMsg) -> Self {
         match value {
-            QueryMsg::ReadyToUnbond { worker_address } => Error::ReadyToUnbond(worker_address),
+            QueryMsg::ReadyToUnbond { verifier_address } => Error::ReadyToUnbond(verifier_address),
```
cgorenflo, line 20: how did this slip through for so long? 😆

cgorenflo (approved):

PR #624, opened 2024-09-13
feat(axelarnet-gateway): move CallContractWithToken from the nexus gateway to axelarnet gateway

fish-sammy:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


contracts/nexus-gateway/src/contract/execute.rs
```diff
@@ -86,23 +26,66 @@ pub fn route_to_router(
...
 }
 
-pub fn route_to_nexus(
+pub fn route_message_with_token_to_nexus(
     storage: &mut dyn Storage,
+    querier: QuerierWrapper,
+    info: MessageInfo,
+    msg: router_api::Message,
+) -> Result<Response<nexus::execute::Message>> {
+    let token = match info.funds.as_slice() {
```
milapsheth, line 35: Use `info.funds.pop()` + a length check instead to avoid cloning the token? Could make it into a helper for better readability.

contracts/nexus-gateway/src/contract/execute.rs
```diff
@@ -86,23 +26,66 @@ pub fn route_to_router(
...
+
+    let mut msg: nexus::execute::Message = msg.into();
+    msg.token.clone_from(&token);
 
-    Ok(Response::new().add_messages(msgs))
+    vec![client.route_message(msg)]
+        .then(move |mut msgs| {
+            if let Some(token) = token {
+                msgs.extend(iter::once(
+                    BankMsg::Send {
```
milapsheth, line 80: The send will happen after the execute, but it should be done before so nexus has the token to process.

contracts/nexus-gateway/src/contract.rs
```diff
@@ -115,80 +111,68 @@ mod tests {
...
-                destination_address: "0xD419".parse().unwrap(),
-                payload: decode("bb9b5566c2f4876863333e481f4698350154259ffe6226e283b16ce18a64bcf1")
-                    .unwrap()
-                    .into(),
-            },
+            mock_info("unauthorized", &[Coin::new(100, "test")]),
+            ExecuteMsg::RouteMessageWithToken(router_messages()[0].clone()),
         );
-        assert!(res.is_err_and(|err| err_contains!(err.report, Error, Error::InvalidToken { .. })));
+        assert!(res.is_err_and(|err| err_contains!(
```
milapsheth, line 124: Switch to using `assert_err_contains` and `assert_ok`

contracts/nexus-gateway/src/contract.rs
```diff
@@ -115,80 +111,68 @@ mod tests {
...
+                AXELARNET_GATEWAY,
+                &[Coin::new(100, "test"), Coin::new(100, "test")],
+            ),
+            ExecuteMsg::RouteMessageWithToken(router_messages()[0].clone()),
+        );
+        assert!(res.is_err_and(|err| { err.to_string().contains("invalid token") }));
+    }
+
+    #[test]
+    fn route_message_with_token() {
```
milapsheth, line 157: Can we start moving new integration tests to the tests folder to follow the new pattern used in axelarnet-gateway?

contracts/nexus-gateway/src/contract.rs
```diff
@@ -115,80 +111,68 @@ mod tests {
...
-            }
-            _ => panic!("unexpected query: {:?}", msg),
-        });
         instantiate_contract(deps.as_mut());
 
-        let token = Coin {
-            denom: "denom".to_string(),
-            amount: 100u128.into(),
-        };
         let res = execute(
```
milapsheth, line 161: Use `assert_ok!` to check ok and unwrap the result

contracts/nexus-gateway/src/contract.rs
```diff
@@ -115,80 +111,68 @@ mod tests {
...
-        }));
+        goldie::assert_json!(res.unwrap());
+
+        let res = execute(
+            deps.as_mut(),
+            mock_env(),
+            mock_info(AXELARNET_GATEWAY, &[Coin::new(100, "test")]),
+            ExecuteMsg::RouteMessageWithToken(router_messages()[0].clone()),
+        );
+        assert!(res.is_ok_and(|res| res.messages.is_empty()));
```
milapsheth, line 175: Same

contracts/axelarnet-gateway/tests/execute.rs
```diff
@@ -309,6 +313,40 @@ fn contract_call_returns_correct_message() {
     goldie::assert_json!(msg)
 }
 
+#[test]
+fn contract_call_with_token_returns_correct_message() {
+    let tx_hash: [u8; 32] =
+        hex::decode("b795e27bcb71c3dfd108c18e031ec966e37c7c95927c2a9fd88ec573ee690c2c")
+            .unwrap()
+            .try_into()
+            .unwrap();
```
milapsheth, line 322:
```suggestion
    let tx_hash: [u8; 32] = [1; 32];
```
keep it simple since value doesn't matter?

contracts/axelarnet-gateway/src/contract.rs
```diff
@@ -76,7 +77,7 @@ pub fn execute(
         } => execute::call_contract(
             deps.storage,
             deps.querier,
-            info.sender,
+            info,
```
milapsheth, line 80: Create a separate `ContractCalledWithToken` msg type. The routing is quite different between the two call types and call contract with token has more restrictions (token can only be a gateway token, and destination chain has to be a consensus connection), so I prefer to not combine them. This also makes it consistent with the legacy EVM gateway API of having two entrypoints.

contracts/axelarnet-gateway/src/events.rs
```diff
@@ -15,13 +17,26 @@ pub enum AxelarnetGatewayEvent {
...
-            AxelarnetGatewayEvent::ContractCalled { msg, payload } => {
-                make_message_event("contract_called", msg)
-                    .add_attributes(vec![("payload", payload.to_string())])
-            }
+            AxelarnetGatewayEvent::ContractCalled {
+                msg,
+                payload,
+                token,
+            } => make_message_event("contract_called", msg)
+                .then(add_token_attribute_if_some(token))
```
milapsheth, line 38: add optional attribute at the end?

contracts/axelarnet-gateway/src/contract/execute.rs
```diff
@@ -88,31 +97,56 @@ pub fn call_contract(
...
 
     state::save_unique_routable_msg(storage, &msg.cc_id, &msg)
         .inspect_err(|err| panic_if_already_exists(err, &msg.cc_id))
         .change_context(Error::SaveRoutableMessage)?;
 
-    Ok(
-        route_to_router(storage, &Router::new(router), vec![msg.clone()])?.add_event(
-            AxelarnetGatewayEvent::ContractCalled {
-                msg,
+    let res = match info.funds.as_slice() {
```
milapsheth, line 108: Use `info.funds.pop()` to avoid cloning the token

contracts/nexus-gateway/src/contract.rs
fish-sammy: I am not sure if I like that pattern

contracts/axelarnet-gateway/src/contract/execute.rs
fish-sammy: That would require `info` to be a mutable reference which I strongly do not like.

contracts/nexus-gateway/src/contract/execute.rs
fish-sammy: Same here, I don't like a mutable reference to the MessageInfo.

contracts/nexus-gateway/src/contract/execute.rs
fish-sammy: good catch

contracts/nexus-gateway/src/msg.rs
```diff
@@ -21,12 +19,8 @@ pub struct InstantiateMsg {
 pub enum ExecuteMsg {
     /// Initiate a cross-chain contract call with token from Axelarnet to another chain.
```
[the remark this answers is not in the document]
fish-sammy: What is it that has to be updated here?

contracts/axelarnet-gateway/src/contract.rs
fish-sammy: It feels awkward to create a separate `CallContractWithToken` where its msg struct is identical to `CallContract`, i.e. `token` is no where seen in its msg struct.

contracts/nexus-gateway/src/msg.rs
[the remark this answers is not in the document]
milapsheth: It's not initiating a contract call anymore but routing a Message with a single nexus registered token attached to legacy/nexus registered chains only

contracts/nexus-gateway/src/contract.rs
milapsheth: we should discuss how we want to do this in the sync for consistency

contracts/axelarnet-gateway/src/contract.rs
milapsheth: Yeah, I understand, but I still find it worse to allow `CallContract` to support it given all the asterisks around legacy token/legacy chains, and since it'll be eventually removed. Would like to hear thoughts from others as well.

fish-sammy: I don't see why is it worse but we can see what others think.

contracts/axelarnet-gateway/src/contract/execute.rs
milapsheth: The helper can take `info.funds` by value, and mark it as a mutable arg since it owns it, so it's an implementation detail, and doesn't require `info` to marked as mutable.

fish-sammy: It doesn't make sense for it to own it though. Other code may need the `info` for a completely different reason.

contracts/nexus-gateway/src/contract.rs
```diff
@@ -115,80 +112,68 @@ mod tests {
...
+        let mut deps = mock_dependencies();
+        instantiate_contract(deps.as_mut());
 
+        let res = execute(
+            deps.as_mut(),
+            mock_env(),
+            mock_info(AXELARNET_GATEWAY, &[]),
+            ExecuteMsg::RouteMessageWithToken(router_messages()[0].clone()),
+        );
+        assert!(res.is_err_and(|err| { err.to_string().contains("invalid token") }));
```
milapsheth, line 143: use `assert_err_contains`

contracts/nexus-gateway/src/contract.rs
```diff
@@ -115,80 +112,68 @@ mod tests {
...
+        let res = execute(
+            deps.as_mut(),
+            mock_env(),
+            mock_info(
+                AXELARNET_GATEWAY,
+                &[Coin::new(100, "test"), Coin::new(100, "test")],
+            ),
+            ExecuteMsg::RouteMessageWithToken(router_messages()[0].clone()),
+        );
+        assert!(res.is_err_and(|err| { err.to_string().contains("invalid token") }));
```
milapsheth, line 154: same

contracts/nexus-gateway/src/contract.rs
```diff
@@ -115,80 +112,68 @@ mod tests {
...
-        }));
+        goldie::assert_json!(assert_ok!(res));
+
+        let res = execute(
+            deps.as_mut(),
+            mock_env(),
+            mock_info(AXELARNET_GATEWAY, &[Coin::new(100, "test")]),
+            ExecuteMsg::RouteMessageWithToken(router_messages()[0].clone()),
+        );
+        assert!(assert_ok!(res).messages.is_empty());
```
milapsheth, line 176:
```suggestion
        let response = assert_ok!(execute(
            deps.as_mut(),
            mock_env(),
            mock_info(AXELARNET_GATEWAY, &[Coin::new(100, "test")]),
            ExecuteMsg::RouteMessageWithToken(router_messages()[0].clone()),
        ));
        assert!(response.messages.is_empty());
```

contracts/nexus-gateway/src/contract.rs
```diff
@@ -115,80 +112,68 @@ mod tests {
...
-                msg:
-                    CosmosMsg::Custom(nexus::execute::Message {
-                        token: Some(actual_token),
-                        ..
-                    }),
-                ..
-            }] => *actual_token == token && to_address == NEXUS && *amount == vec![token],
-            _ => false,
-        }));
+        goldie::assert_json!(assert_ok!(res));
```
milapsheth, line 168:
```suggestion
        goldie::assert_json!(response);
```
assert_ok earlier, same as below

contracts/nexus-gateway/src/contract/execute.rs
```diff
@@ -86,23 +26,61 @@ pub fn route_to_router(
...
+    querier: QuerierWrapper,
+    info: MessageInfo,
+    msg: router_api::Message,
+) -> Result<Response<nexus::execute::Message>> {
+    let token = match info.funds.as_slice() {
+        [token] => token.clone(),
+        _ => bail!(Error::InvalidToken(info.funds)),
+    };
+
+    route_message_to_nexus(storage, querier, msg, Some(token))?
```
milapsheth, line 40: Create the BankMsg here and then call `route_message_to_nexus` and chain the response to simplify `route_message_to_nexus` and `route_messages_to_nexus`

contracts/axelarnet-gateway/src/contract/execute.rs
fish-sammy: Did it but it isn't anything cleaner.

contracts/nexus-gateway/src/contract/execute.rs
fish-sammy: Not a bad idea.

fish-sammy: `route_message_to_nexus` has to take the token in cuz the nexus message itself needs it. In that case, I don't think the proposal makes it cleaner.

contracts/nexus-gateway/src/contract.rs
fish-sammy: What's the problem with inlining it?

contracts/nexus-gateway/src/contract.rs
fish-sammy: Same here. What's the problem with inlining it?

fish-sammy: I honestly don't even understand the purpose of this `assert_ok` thing. What's wrong with just unwrap?

contracts/nexus-gateway/src/contract/execute.rs
```diff
@@ -86,23 +27,61 @@ pub fn route_to_router(
...
+
+    state::set_message_routed(storage, &msg.cc_id)?;
 
-    let msgs: Vec<nexus::execute::Message> = msgs.into_iter().map(Into::into).collect();
+    let client: nexus::Client = client::CosmosClient::new(querier).into();
+    let config = load_config(storage)?;
 
-    Ok(Response::new().add_messages(msgs))
+    let mut msg: nexus::execute::Message = msg.into();
+    msg.token.clone_from(&token);
```
milapsheth, line 75: Ah, yes, forgot about this. I'd suggest implementing a `impl From<(router_api::Message, Option<Coin>) for nexus::execute::Message` instead to make this a bit safer. `route_message_to_nexus` could then take in just `nexus::execute::Message` (and caller handles the BankMsg) to make it cleaner perhaps?

contracts/nexus-gateway/src/contract/execute.rs
```diff
@@ -86,23 +27,61 @@ pub fn route_to_router(
...
+    storage: &mut dyn Storage,
+    querier: QuerierWrapper,
+    msg: router_api::Message,
+    token: Option<Coin>,
+) -> Result<Vec<CosmosMsg<nexus::execute::Message>>> {
+    if state::is_message_routed(storage, &msg.cc_id)? {
+        return Ok(vec![]);
+    }
+
+    state::set_message_routed(storage, &msg.cc_id)?;
```
milapsheth, line 69: Not relevant to this PR but core handles this already, shall we remove the nexus-gateway state instead of duplicating it between core and amplifier?

contracts/nexus-gateway/src/contract.rs
milapsheth: Want to make it explicit that the test is asserting that `execute` succeeds. See this [thread](https://interoplabs.slack.com/archives/C03HHV83DAS/p1725475605505229?thread_ts=1725388225.441939&cid=C03HHV83DAS)

Calling `assert_ok` on `execute` makes this more clear instead of the next line. Also, a nicer pattern if we want to do more with the unwrapped response later

contracts/nexus-gateway/src/contract.rs
milapsheth: replied above

contracts/nexus-gateway/src/contract/execute.rs
milapsheth: replied [here](https://github.com/axelarnetwork/axelar-amplifier/pull/624#discussion_r1761970028)

contracts/nexus-gateway/src/contract/execute.rs
fish-sammy: I don't like the idea of "caller handles the BankMsg". It can easily become a future bug to me cuz one may not realize they are a package of one thing, especially now even the order matters. I don't hate the From trait though.

contracts/nexus-gateway/src/contract/execute.rs
fish-sammy: Maybe, but definitely not gonna do in this PR.

contracts/axelarnet-gateway/src/contract/execute.rs
milapsheth: you can move just `info.funds` instead of all of `info`. But not a strong opinion

contracts/nexus-gateway/src/contract.rs
```diff
@@ -115,80 +112,70 @@ mod tests {
...
+        goldie::assert_json!(res);
+
+        let res = execute(
+            deps.as_mut(),
+            mock_env(),
+            mock_info(AXELARNET_GATEWAY, &[Coin::new(100, "test")]),
+            ExecuteMsg::RouteMessageWithToken(router_messages()[0].clone()),
+        );
+        let res = assert_ok!(res);
+        assert!(res.messages.is_empty());
```
milapsheth, line 178:
```suggestion
        let response = assert_ok!(execute(
            deps.as_mut(),
            mock_env(),
            mock_info(AXELARNET_GATEWAY, &[Coin::new(100, "test")]),
            ExecuteMsg::RouteMessageWithToken(router_messages()[0].clone()),
        ));
        assert!(response.messages.is_empty());
```
Sorry, I meant this, similar to what we do [here](https://github.com/axelarnetwork/axelar-amplifier/pull/624/files#diff-5f32f8b97bf50fa78ba3747e50ecf568d72d30f3e191a3fc20fd98633c76c243R333)

milapsheth (approved):

contracts/nexus-gateway/src/contract/execute.rs
milapsheth: yeah, just raising it here

contracts/axelarnet-gateway/src/contract/execute.rs
fish-sammy: I think I like what I did better, i.e. implemented a method on `MessageInfo`.

contracts/nexus-gateway/src/contract/execute.rs
milapsheth: Yeah, I agree with the trade off of the caller handling it. Since the caller is in this module itself, I was ok with it to reduce the complexity. And yeah, From trait would be good to have still

contracts/axelarnet-gateway/src/contract/execute.rs
```diff
@@ -88,31 +98,55 @@ pub fn call_contract(
...
+
+            Response::new()
+                .add_message(WasmMsg::Execute {
+                    contract_addr: nexus_gateway.to_string(),
+                    msg: to_json_binary(&nexus_gateway::msg::ExecuteMsg::RouteMessageWithToken(
+                        msg,
+                    ))
+                    .expect("failed to serialize route message with token"),
+                    funds: vec![token],
+                })
```
cgorenflo, line 134: don't we have a client yet?

contracts/axelarnet-gateway/src/contract/execute.rs
```diff
@@ -88,31 +98,55 @@ pub fn call_contract(
...
 
     state::save_unique_routable_msg(storage, &msg.cc_id, &msg)
         .inspect_err(|err| panic_if_already_exists(err, &msg.cc_id))
         .change_context(Error::SaveRoutableMessage)?;
 
-    Ok(
-        route_to_router(storage, &Router::new(router), vec![msg.clone()])?.add_event(
-            AxelarnetGatewayEvent::ContractCalled {
-                msg,
+    let res = match info.token().change_context(Error::InvalidToken)? {
```
cgorenflo, line 109: consider
```
let token = info.token().change_context(Error::InvalidToken)?;
let event = AxelarnetGatewayEvent::ContractCalled {
                msg: msg.clone(),
                payload: call_contract.payload,
                token: token.clone(),
            };

let res = match token {
...
}
```

contracts/axelarnet-gateway/src/events.rs
```diff
@@ -15,13 +17,26 @@ pub enum AxelarnetGatewayEvent {
...
 
+fn add_token_attribute_if_some(token: Option<Coin>) -> impl Fn(Event) -> Event {
+    move |event| {
+        if let Some(token) = &token {
+            event.add_attribute("token", token.to_string())
+        } else {
+            event
+        }
+    }
+}
```
cgorenflo, line 28: generalize this as a trait on the Event? I think that can be quite useful in more cases, i.e.
```
fn add_attribute_if_some<T>(label: &str, value: Option<T>){...}
```
(or however the label is defined)

contracts/nexus-gateway/src/contract/execute.rs
```diff
@@ -86,23 +27,60 @@ pub fn route_to_router(
...
-        .into_iter()
-        .filter_map(|msg| match state::is_message_routed(storage, &msg.cc_id) {
-            Ok(true) => None,
-            Ok(false) => Some(Ok(msg)),
-            Err(err) => Some(Err(err)),
-        })
-        .collect::<Result<Vec<_>>>()?;
+    msgs.into_iter()
+        .map(|msg| route_message_to_nexus(storage, querier, msg, None))
+        .collect::<Result<Vec<_>>>()?
```
cgorenflo, line 53:
```suggestion
        .try_collect::<Vec<_>>()?
```

contracts/nexus-gateway/src/contract/execute.rs
cgorenflo: how does core handle it? Here we ignore already routed messages. If core fails when you retry, then we need to keep it.

contracts/nexus-gateway/src/contract/execute.rs
```diff
@@ -86,23 +27,60 @@ pub fn route_to_router(
...
 
-    let msgs: Vec<nexus::execute::Message> = msgs.into_iter().map(Into::into).collect();
+    let client: nexus::Client = client::CosmosClient::new(querier).into();
+    let config = load_config(storage)?;
 
-    Ok(Response::new().add_messages(msgs))
+    let msg: nexus::execute::Message = (msg, token.clone()).into();
+
+    token
+        .into_iter()
```
cgorenflo, line 77:
```suggestion
```
Option already has a `map`

packages/axelar-wasm-std/src/token.rs
```diff
@@ -0,0 +1,23 @@
...
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum Error {
+    #[error("invalid token: one and only one token is required for this operation, got {0:?}")]
+    MultipleTokens(Vec<Coin>),
+}
+
+pub trait GetToken {
+    fn token(&self) -> Result<Option<Coin>, Error>;
```
cgorenflo, line 12: maybe `single_token()`?

packages/axelar-wasm-std/src/token.rs
```diff
@@ -0,0 +1,23 @@
...
+use error_stack::{bail, Result};
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum Error {
+    #[error("invalid token: one and only one token is required for this operation, got {0:?}")]
+    MultipleTokens(Vec<Coin>),
+}
+
+pub trait GetToken {
```
cgorenflo, line 11: consider
```
mod sealed {
   pub trait Sealed {}
}

pub trait GetToken: sealed::Sealed{
...
}

impl sealed::Sealed for MessageInfo {}

...
```
that way, no other struct can implement the `GetToken` trait

packages/axelar-wasm-std/src/token.rs
```diff
@@ -0,0 +1,23 @@
...
+}
+
+impl GetToken for MessageInfo {
+    fn token(&self) -> Result<Option<Coin>, Error> {
+        match self.funds.as_slice() {
+            [] => Ok(None),
+            [token] => Ok(Some(token.clone())),
+            _ => bail!(Error::MultipleTokens(self.funds.clone())),
+        }
+    }
```
cgorenflo, line 22: alternative
```suggestion
        ensure!(self.len <=1, Error::MultipleTokens(self.funds.clone()));
        
        Ok(self.funds.first().cloned())
    }
```

contracts/nexus-gateway/src/contract/execute.rs
```diff
@@ -86,23 +27,60 @@ pub fn route_to_router(
...
+        .map(|token| BankMsg::Send {
+            to_address: config.nexus.to_string(),
+            amount: vec![token],
+        })
+        .map(Into::into)
+        .chain(iter::once(client.route_message(msg)))
```
cgorenflo, line 83: I'm slightly worried about this split into multiple messages, but it's probably fine?
Is there any chance the tokens can be abused through reentry?

contracts/axelarnet-gateway/src/contract.rs
fish-sammy: Oh ya, that as well. We didn't have this option for the EVM gateways since ERC20 tokens are not first-class citizens. In Cosmwasm, however, things are different.

contracts/nexus-gateway/src/contract/execute.rs
fish-sammy: That is true. Core does fail it rather than ignore it.

contracts/axelarnet-gateway/src/contract/execute.rs
fish-sammy: Nope. I don't mind doing it in this same PR, but it's already pretty big.

contracts/axelarnet-gateway/src/events.rs
fish-sammy: That's an interesting thought. Let me try.

packages/axelar-wasm-std/src/token.rs
fish-sammy: luv it

contracts/nexus-gateway/src/contract/execute.rs
fish-sammy: I am confused. Do you mean use the `Option::map` first and then `.into_iter`?

contracts/nexus-gateway/src/contract/execute.rs
fish-sammy:
```
use of unstable library feature 'iterator_try_collect'
see issue #94047 <https://github.com/rust-lang/rust/issues/94047> for more informationrustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20%5B0%5D?0#file%3A%2F%2F%2FUsers%2Fsammyliu%2FDocuments%2Faxelar%2Famplifier%2Fcontracts%2Fnexus-gateway%2Fsrc%2Fcontract%2Fexecute.rs)
```

packages/axelar-wasm-std/src/token.rs
fish-sammy: Not sure if I wanna prevent it this hard.

contracts/nexus-gateway/src/contract/execute.rs
cgorenflo: I had exactly the same train of thought for an optimization, but `mut MessageInfo` is a big no for me too

contracts/axelarnet-gateway/src/contract/execute.rs
cgorenflo: we can do a follow-up

cgorenflo (approved):

PR #625, opened 2024-09-13
feat(minor-service-registry): add UpdateService execute message

cjcobb23:
https://axelarnetwork.atlassian.net/browse/AXE-4860


contracts/service-registry/src/contract.rs
```diff
@@ -272,6 +276,145 @@ mod test {
         ));
     }
 
+    #[test]
```
cgorenflo, line 279: can you split this test into multiple or find another way to compact it? It's so long that it's hard to understand what's going on

contracts/service-registry/src/contract/execute.rs
```diff
@@ -78,6 +79,33 @@ pub fn update_verifier_authorization_status(
...
 }
 
+pub fn update_service(
+    deps: DepsMut,
+    service_name: String,
+    updated_service_params: UpdatedServiceParams,
+) -> Result<Response, ContractError> {
+    SERVICES.update(deps.storage, &service_name, |service| match service {
+        None => Err(ContractError::ServiceNotFound),
+        // TODO: write a macro to automate this?
```
cgorenflo, line 89: Unless we go all out and create a derive macro like `UpdatableParams` that automatically generates the Params struct generically and the update function with it the effort of creating a macro is not worth it. So I suggest to check how often we would need that and then either commit to doing that right now (follow-up PR), or remove this TODO because it's not worth it

cgorenflo (approved):

PR #626, opened 2024-09-16
refactor(multisig-prover): non-zero nonce for verifier set test data

haiyizxx:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


contracts/multisig-prover/src/encoding/mod.rs
```diff
@@ -45,7 +45,7 @@ impl Encoder {
         payload: &Payload,
     ) -> Result<HexBinary, ContractError> {
         match self {
-            Encoder::Abi => abi::execute_data::encode(
+            Encoder::Abi => abi::encode_execute_data(
```
milapsheth, line 48: Make this into a direct dispatch (call `payload_digest` internally instead)

contracts/multisig-prover/src/encoding/mod.rs
```diff
@@ -45,7 +45,7 @@ impl Encoder {
...
                 verifier_set,
                 sigs,
                 &self.digest(domain_separator, verifier_set, payload)?,
```
milapsheth, line 51: Similarly, `to_recoverable` should be an internal transform specific to ABI and BCS module (BCS doesn't need to transform it as well so it can be more efficient)

contracts/multisig-prover/src/encoding/abi.rs
```diff
@@ -113,11 +217,45 @@ mod tests {
...
-        let execute_data =
-            encode(&verifier_set, signers_with_sigs, &payload_hash, &payload).unwrap();
-        let data_hash = Keccak256::digest(execute_data.as_slice());
+        goldie::assert!(encode_execute_data(
+            &verifier_set,
+            signers_with_sigs,
+            &payload_hash,
+            &payload
+        )
+        .unwrap()
```
milapsheth, line 226: switch to assert_ok and assert_err_contains for the tests

cgorenflo (approved):

contracts/multisig-prover/src/encoding/abi.rs
```diff
@@ -1,17 +1,78 @@
...
+
+pub fn encode_payload(payload: &Payload) -> Result<Vec<u8>, ContractError> {
+    let command_type = CommandType::from(payload).into();
+
+    match payload {
+        Payload::Messages(messages) => {
+            let messages = messages
+                .iter()
+                .map(Message::try_from)
+                .map_ok(|m| Token::Tuple(m.into_tokens()))
```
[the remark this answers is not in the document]
haiyizxx: sure we can evaluate

PR #628, opened 2024-09-17
fix(minor-multisig-prover): set active verifiers fix

cjcobb23:
* set the active verifiers with the coordinator when first updating the verifier set

https://axelarnetwork.atlassian.net/browse/AXE-4871


contracts/multisig-prover/src/contract/execute.rs
```diff
@@ -230,16 +230,28 @@ pub fn update_verifier_set(
...
                 .map_err(ContractError::from)?;
 
-            Ok(Response::new().add_message(
-                wasm_execute(
-                    config.multisig,
-                    &multisig::msg::ExecuteMsg::RegisterVerifierSet {
-                        verifier_set: new_verifier_set,
-                    },
-                    vec![],
+            let verifier_union_set = all_active_verifiers(&deps)?;
```
haiyizxx, line 233: what's the difference between new_verifier_set?

cjcobb23: no difference, this function will just end up transforming the type in this case. But maybe not the best because it will read from state again. I'll change it

contracts/multisig-prover/src/contract/execute.rs
```diff
@@ -230,16 +230,31 @@ pub fn update_verifier_set(
...
+            Ok(Response::new()
+                .add_message(
+                    wasm_execute(
+                        config.multisig,
+                        &multisig::msg::ExecuteMsg::RegisterVerifierSet {
+                            verifier_set: new_verifier_set.clone(),
+                        },
+                        vec![],
+                    )
+                    .map_err(ContractError::from)?,
```
cgorenflo, line 242: do we have a client?

contracts/multisig-prover/src/contract/execute.rs
```diff
@@ -230,16 +230,31 @@ pub fn update_verifier_set(
...
+                        &coordinator::msg::ExecuteMsg::SetActiveVerifiers {
+                            verifiers: new_verifier_set
+                                .signers
+                                .values()
+                                .map(|signer| signer.address.to_string())
+                                .collect::<HashSet<String>>(),
+                        },
+                        vec![],
+                    )
+                    .map_err(ContractError::from)?,
```
cgorenflo, line 256: do we have a client?

cgorenflo (approved):

PR #633, opened 2024-09-19
fix(minor-ampd): reject missing chain in register chain support command

cjcobb23:
https://axelarnetwork.atlassian.net/browse/AXE-5103

### DIFF
--- a/ampd/src/commands/register_chain_support.rs
+++ b/ampd/src/commands/register_chain_support.rs
@@ -14,18 +14,15 @@ use crate::{Error, PREFIX};
 #[derive(clap::Args, Debug, Valuable)]
 pub struct Args {
     pub service_name: nonempty::String,
-    pub chains: Vec<ChainName>,
+    pub chain: ChainName,
 }
 
 pub async fn run(config: Config, args: Args) -> Result<Option<String>, Error> {
-    if args.chains.len() == 0 {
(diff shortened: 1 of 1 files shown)


ampd/src/commands/register_chain_support.rs
```diff
@@ -18,6 +18,9 @@ pub struct Args {
 }
 
 pub async fn run(config: Config, args: Args) -> Result<Option<String>, Error> {
+    if args.chains.len() == 0 {
+        return Err(Error::InvalidInput.into());
+    }
```
milapsheth, line 23:
```suggestion
    ensure!(args.chains.is_empty(), Error::InvalidInput.into());
```
perhaps?

ampd/src/commands/register_chain_support.rs
```diff
@@ -18,6 +18,9 @@ pub struct Args {
 }
```
milapsheth, line 18: how about using a `nonempty::Vec` here instead?

milapsheth (approved):

ampd/src/commands/register_chain_support.rs
cjcobb23: this was a bit difficult, we need to derive Valuable on nonempty::Vec, which cascades to a lot of changes across the codebase

PR #636, opened 2024-09-20
chore(ampd): make default gas adjustment 1.2 to increase chance of broadcast sucess

cgorenflo:


ampd/src/broadcaster/mod.rs
```diff
@@ -87,7 +87,7 @@ impl Default for Config {
             chain_id: "axelar-dojo-1".parse().unwrap(),
             tx_fetch_interval: Duration::from_millis(500),
             tx_fetch_max_retries: 10,
-            gas_adjustment: 1.0,
+            gas_adjustment: 1.1,
```
milapsheth, line 90:
```suggestion
            gas_adjustment: 1.2,
```
i'd suggest 1.2, 1.1 has issues as well sometimes

milapsheth (approved):

PR #637, opened 2024-09-24
feat(minor-coordinator): verifier information query

maancham:
## Description

AXE-4756

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


contracts/coordinator/src/client.rs
```diff
@@ -12,12 +12,20 @@ pub enum Error {
         "failed to execute ReadyToUnbond query at coordinator contract. verifier_address: {0}"
     )]
     ReadyToUnbond(String),
+    #[error(
+        "failed to execute VerifierDetailsWithProvers query at coordinator contract. service_name: {0}, verifier_address: {1}"
+    )]
+    VerifierDetailsWithProvers(String, String),
```
cgorenflo, line 18: prefer `VerifierDetailsWithProvers{service_name:String, verifier_address:String}`, it's not immediately obvious what the parameters mean and it's easy to get the order wrong since they have the same type

contracts/coordinator/src/contract/migrations/v0_2_0.rs
cgorenflo, line 1: this needs a new migration to set the service registry if the contract is already instantiated

contracts/coordinator/src/contract/query.rs
```diff
@@ -14,3 +21,47 @@ fn is_verifier_in_any_verifier_set(deps: Deps, verifier_address: &Addr) -> bool
...
+        .by_verifier
+        .prefix(verifier_address)
+        .range(deps.storage, None, None, Order::Ascending)
+        .map(|result| result.map(|(_, record)| record.prover))
+        .try_collect()?;
+
+    Ok(provers)
+}
+
+pub fn verifier_details_with_provers(
```
cgorenflo, line 40: prefer to move public functions to the top of the file and the private functions they use below them

contracts/coordinator/src/contract/query.rs
```diff
@@ -14,3 +21,47 @@ fn is_verifier_in_any_verifier_set(deps: Deps, verifier_address: &Addr) -> bool
...
+    deps: Deps,
+    service_name: String,
+    verifier_address: Addr,
+) -> Result<VerifierInfo, ContractError> {
+    let config = CONFIG.load(deps.storage).map_err(ContractError::from)?;
```
cgorenflo, line 45: create a panicking function for this (see other contracts)

contracts/coordinator/src/contract/query.rs
```diff
@@ -14,3 +21,47 @@ fn is_verifier_in_any_verifier_set(deps: Deps, verifier_address: &Addr) -> bool
...
+        .querier
+        .query(&QueryRequest::Wasm(WasmQuery::Smart {
+            contract_addr: config.service_registry.to_string(),
+            msg: to_json_binary(&Verifier {
+                service_name,
+                verifier: verifier_address.to_string(),
+            })
+            .map_err(ContractError::from)?,
+        }))
+        .map_err(ContractError::from)?;
```
cgorenflo, line 57: use service registry client instead

contracts/coordinator/src/contract/query.rs
```diff
@@ -1,14 +1,42 @@
...
+pub fn check_verifier_ready_to_unbond(deps: Deps, verifier_address: Addr) -> StdResult<bool> {
+    Ok(!is_verifier_in_any_verifier_set(deps, &verifier_address))
+}
+
+pub fn verifier_details_with_provers(
+    deps: Deps,
+    service_name: String,
+    verifier_address: Addr,
+) -> Result<VerifierInfo, ContractError> {
+    let config = CONFIG.load(deps.storage).expect("couldn't load config");
```
cgorenflo, line 21: what I meant is create a `state::load_config` function that hides the panic, compare https://github.com/axelarnetwork/axelar-amplifier/blob/58a7ea7f1bf998bee60ec90f66befaebd4fa5c4a/contracts/axelarnet-gateway/src/state.rs#L59

integration-tests/tests/test_utils/mod.rs
```diff
@@ -332,7 +332,7 @@ pub fn verifier_info_from_coordinator(
 pub fn assert_verifier_details_are_equal(
     verifier_info: VerifierInfo,
     verifier: &Verifier,
-    chains: &Vec<Chain>,
+    chains: &[Chain],
```
cgorenflo, line 335: nice change

contracts/coordinator/src/contract/migrations/v0_2_0.rs
cgorenflo, line 1: coordinator has v1.0.0 right now, so you need to move it to v1_0_0 and change the `BASE_VERSION` to v1.0.0

contracts/coordinator/src/contract/migrations/v0_2_0.rs
```diff
@@ -0,0 +1,118 @@
...
+    #[error(transparent)]
+    Version(#[from] VersionError),
+    #[error(transparent)]
+    NonEmpty(#[from] nonempty::Error),
+}
+
+pub fn migrate(storage: &mut dyn Storage, service_registry: Addr) -> Result<(), Error> {
+    cw2::assert_contract_version(storage, CONTRACT_NAME, BASE_VERSION)?;
+
+    killswitch::init(storage, State::Disengaged)?;
```
cgorenflo, line 25: I don't think the coordinator needs a killswitch, confirm with CJ. Since the coordinator only aggregates data and doesn't modify it itself I can't see a scenario where we would have to quickly deactivate it, but I can also see an argument for adding a killswitch to any contract

contracts/coordinator/src/contract/migrations/v0_2_0.rs
```diff
@@ -0,0 +1,118 @@
...
+        cw2::set_contract_version(deps.storage, CONTRACT_NAME, BASE_VERSION)?;
+
+        let governance = address::validate_cosmwasm_address(deps.api, &msg.governance_address)?;
+        permission_control::set_governance(deps.storage, &governance)?;
+
+        Ok(Response::default())
+    }
+
+    #[cw_serde]
+    #[deprecated(since = "0.2.0", note = "only used to test the migration")]
```
cgorenflo, line 114: the deprecated messages also need to be adjusted to 1.0.0

contracts/coordinator/src/contract/migrations/v0_2_0.rs
maancham: He said he does not think we need it here. Removing...

cgorenflo (approved):

PR #640, opened 2024-09-26
fix(integration-test): app type for custom querier

cgorenflo:


integration-tests/src/coordinator_contract.rs
```diff
@@ -30,7 +31,6 @@ impl CoordinatorContract {
         CoordinatorContract { contract_addr }
     }
 }
-
```
cgorenflo, line 33: add the empty line back in

haiyizxx (approved):

PR #656, opened 2024-10-11
fix(minor-interchain-token-service): ignore empty event attributes

milapsheth:
## Description

[AXE-5913]

## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Other Notes

[AXE-5913]: https://axelarnetwork.atlassian.net/browse/AXE-5913?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -158,7 +158,8 @@ fn gateway_token_transfer(
     match (gateway_denom, message) {
         (Some(denom), Message::InterchainTransfer { amount, .. }) => Ok(Some(Coin {
             denom: denom.to_string(),
-            amount: Uint128::try_from(*amount).change_context(Error::TransferAmountOverflow)?,
+            amount: Uint128::try_from(Uint256::from(*amount))
```
haiyizxx, line 161: what's the purpose of this change?

packages/axelar-wasm-std/src/nonempty/hexbinary.rs
```diff
@@ -0,0 +1,68 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::nonempty::{Error, HexBinary};
+
+    #[test]
+    fn test_non_empty_hexbinary() {
+        assert!(HexBinary::try_from(cosmwasm_std::HexBinary::from(&[1, 2, 3])).is_ok())
```
haiyizxx, line 58: assert_ok

packages/axelar-wasm-std/src/nonempty/hexbinary.rs
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    #[test]
+    fn test_empty_non_empty_hexbinary() {
+        assert_eq!(
+            HexBinary::try_from(cosmwasm_std::HexBinary::from(&[])).unwrap_err(),
+            Error::InvalidValue("empty".to_string())
```
haiyizxx, line 65: does it work with err_contains?

contracts/interchain-token-service/src/abi.rs
```diff
@@ -122,49 +126,58 @@ impl Message {
     }
 
     pub fn abi_decode(payload: &[u8]) -> Result<Self, Report<Error>> {
-        ensure!(payload.len() >= 32, Error::MessageDecodeFailed);
+        ensure!(payload.len() >= 32, Error::InsufficientMessageLength);
 
         let message_type = MessageType::abi_decode(&payload[0..32], true)
+            .map_err(Error::AbiDecodeFailed)
```
haiyizxx, line 132: why add additional map?

haiyizxx (approved):

contracts/interchain-token-service/src/contract/execute.rs
milapsheth: nonempty::Uint256 -> Uint256 -> Uint128. I'll try to add a direct conversion method instead

packages/axelar-wasm-std/src/nonempty/hexbinary.rs
milapsheth: `Error` doesn't implement `IntoContractError` trait, and it cannot implement it since the macro is defined within the same crate (since it can't refer itself `axelar_wasm_std` when expanded). I don't know of a solution where Rust handles local vs foreign namespacing in the macro

contracts/interchain-token-service/src/abi.rs
milapsheth: hmm, fair enough, i was wrapping the decoding error for consistency, but I suppose there's no point when changing context to the current Error. Removed

contracts/interchain-token-service/src/contract/execute.rs
milapsheth: removing gateway token logic so this won't be needed anyways

PR #657, opened 2024-10-11
feat(minor-router): route to axelarnet gateway

haiyizxx:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


fish-sammy (approved):

contracts/router/src/msg.rs
```diff
@@ -6,8 +6,13 @@ pub struct InstantiateMsg {
     pub admin_address: String,
     // governance votes on chains being added or upgraded
     pub governance_address: String,
-    // the address of the nexus gateway
-    pub nexus_gateway: String,
+    // the address of the axelarnet gateway
+    pub axelarnet_gateway: String,
```
milapsheth, line 10: update deployment scripts as well

contracts/router/src/contract/migrations/v1_0_0.rs
```diff
@@ -0,0 +1,121 @@
+use axelar_wasm_std::error::ContractError;
+use cosmwasm_std::{Addr, Storage};
+
+use crate::contract::CONTRACT_NAME;
+use crate::state;
+
+const BASE_VERSION: &str = "1.0.0";
+
+pub fn migrate(storage: &mut dyn Storage, axelarnet_gateway: Addr) -> Result<(), ContractError> {
+    cw2::assert_contract_version(storage, CONTRACT_NAME, BASE_VERSION)?;
```
milapsheth, line 10: we had issued a v1.0.1 patch to the router. Will that cause an issue here since the base version isn't v1.0.0 on some networks I think? cc @eguajardo

haiyizxx: I don't know the release action was not on main branch.

@eguajardo can you confirm if all contracts deployed across networks are on 1.0.1?

eguajardo: Router devnet-amplifier is on 1.0.1, all others (stagenet, testnet and mainnet) Router is 1.0.0. What version are we migrating to? If the migration is for 1.0.1 to 1.0.2, then BASE_VERSION should be set to 1.0.1 and all contracts must be migrated sequentially 1.0.0 -> 1.0.1 -> 1.0.2

PR #662, opened 2024-10-18
feat(minor-ampd): update stellar gateway event validation

ahramy:
## Description
[AXE-6089](https://axelarnetwork.atlassian.net/browse/AXE-6089)

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Other Notes

[AXE-6089]: https://axelarnetwork.atlassian.net/browse/AXE-6089?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ampd/src/stellar/verifier.rs
```diff
@@ -15,27 +14,18 @@ impl PartialEq<ContractEventBody> for Message {
     fn eq(&self, event: &ContractEventBody) -> bool {
         let ContractEventBody::V0(body) = event;
 
-        if body.topics.len() != 3 {
+        if body.topics.len() != 5 {
             return false;
         }
 
-        let [symbol, source_address, payload_hash] = &body.topics[..] else {
+        let [symbol, source_address, dest_chain, dest_address, payload_hash] = &body.topics[..]
```
milapsheth, line 21:
```suggestion
        let [symbol, source_address, destination_chain, destination_address, payload_hash] = &body.topics[..]
```

ampd/src/stellar/verifier.rs
```diff
@@ -48,34 +38,21 @@ impl PartialEq<ContractEventBody> for VerifierSetConfirmation {
...
-        };
-
-        WeightedSigners::try_from(&self.verifier_set)
-            .ok()
-            .and_then(|signers| ScVal::try_from(signers).ok())
-            .map_or(false, |signers: ScVal| {
-                symbol == &expected_topic && signers == rotated_signers
-            })
+        expected_topic == *symbol
+            && ScVal::Bytes(ScBytes(BytesM::try_from(self.verifier_set.hash()).unwrap()))
```
milapsheth, line 54: Avoid using `unwrap`. Return `false` similar to other failures above on error

milapsheth, line 54: `self.verifier_set.hash()` is different from `WeightedSigners::hash`. We need the latter which is Stellar specific hashing for it to match.

ampd/src/stellar/verifier.rs
```diff
@@ -48,34 +38,21 @@ impl PartialEq<ContractEventBody> for VerifierSetConfirmation {
...
+                == *rotate_signers_hash
```
milapsheth, line 55: instead of dereferencing this, take the reference of the LHS

ampd/src/stellar/verifier.rs
milapsheth: move this hash computation into a separate statement for readability

ampd/src/stellar/verifier.rs
```diff
@@ -48,34 +38,21 @@ impl PartialEq<ContractEventBody> for VerifierSetConfirmation {
     fn eq(&self, event: &ContractEventBody) -> bool {
         let ContractEventBody::V0(body) = event;
 
-        if body.topics.len() != 1 {
+        if body.topics.len() != 3 {
             return false;
         }
 
-        let [symbol] = &body.topics[..] else {
+        let [symbol, _, rotate_signers_hash] = &body.topics[..] else {
```
milapsheth, line 45:
```suggestion
        let [symbol, _, signers_hash] = &body.topics[..] else {
```

ampd/src/stellar/verifier.rs
milapsheth, line 55: similar to this test, add a golden test for WeightedSigners `hash()` as well to detect any changes. You can add a similar test in soroban repo with the same weighted signers values so they match
```
    #[test]
    fn signers_rotation_hash() {
```

ampd/src/stellar/verifier.rs
ahramy: good catch, updated.

ahramy: agreed. updated.

ampd/src/stellar/verifier.rs
ahramy: updated

ahramy: that's a good call. updated.

ampd/src/stellar/verifier.rs
```diff
@@ -392,4 +418,37 @@ mod test {
                 .unwrap(),
         }
     }
+
+    #[test]
+    fn test_weighted_signers_hash() {
```
milapsheth, line 423: test should be defined in the file where `hash` is defined and follow a similar pattern of the this [test](https://github.com/axelarnetwork/axelar-amplifier/blob/102d41b76da6c94f9805c51a4779735e4bc7e84f/external-gateways/stellar/src/lib.rs#L447)

ampd/src/stellar/verifier.rs
```diff
@@ -15,67 +15,61 @@ impl PartialEq<ContractEventBody> for Message {
...
-        let rotated_signers = match &body.data {
-            ScVal::Vec(Some(data)) if data.len() == 1 => {
-                let [rotated_signers] = &data[..] else {
-                    return false;
-                };
-                rotated_signers.clone()
-            }
-            _ => return false,
+        let weighted_signers = match WeightedSigners::try_from(&self.verifier_set) {
+            Ok(ws) => ws,
```
milapsheth, line 56: Do this is in a more idiomatic manner. If it looks ugly, then there's usually a better way to do this in Rust (you can see how the previous check was being chained)

ampd/src/stellar/verifier.rs
```diff
@@ -15,67 +15,61 @@ impl PartialEq<ContractEventBody> for Message {
...
+            Err(_) => return false,
         };
 
-        WeightedSigners::try_from(&self.verifier_set)
-            .ok()
-            .and_then(|signers| ScVal::try_from(signers).ok())
-            .map_or(false, |signers: ScVal| {
-                symbol == &expected_topic && signers == rotated_signers
-            })
+        let weighted_signers_hash = ScVal::Bytes(ScBytes(bytes));
```
milapsheth, line 70:
```suggestion
        let Some(weighted_signers_hash) = WeightedSigners::try_from(&self.verifier_set)
            .ok()
            .and_then(|weighted_signers| weighted_signers.hash().ok())
            .and_then(|signers_hash| BytesM::try_from(signers_hash).ok())
            .map(ScBytes)
            .map(ScVal::Bytes) else {
                return false
            };
```
a better refactor for later would be to create a PartialEqResult trait type whose `eq` returns a Result type, so we can use `?`, and then `PartialEq` can be implemented by calling this and returning false on error

ampd/src/stellar/verifier.rs
```diff
@@ -319,18 +314,16 @@ mod test {
             topics: vec![
                 ScVal::Symbol(ScSymbol(StringM::from_str(TOPIC_CALLED).unwrap())),
                 ScVal::Address(msg.source_address.clone()),
+                ScVal::String(msg.destination_chain.clone()),
+                ScVal::String(msg.destination_address.clone()),
                 ScVal::Bytes(msg.payload_hash.clone()),
             ]
             .try_into()
             .unwrap(),
             data: ScVal::Vec(Some(
```
milapsheth, line 323: Event data is not a tuple containing payload anymore, but the payload itself `ScVal::Bytes(..)`

ampd/src/stellar/verifier.rs
ahramy: make sense. updated.

ampd/src/stellar/verifier.rs
ahramy: just change this to non vec
```

File: ampd/src/stellar/verifier.rs (hunk 6 above)
Comment (ahramy) [reply ampd/src/stellar/verifier.rs]: updated.

File: ampd/src/stellar/verifier.rs (hunk 7 above)
Comment (ahramy) [reply ampd/src/stellar/verifier.rs]: Thank you for the comment. it looks much better, updated.

File: ampd/src/stellar/verifier.rs (hunk 8 above)
Comment (ahramy) [reply ampd/src/stellar/verifier.rs]: updated.

File: external-gateways/stellar/src/lib.rs [hunk 9]
```diff
@@ -443,6 +443,50 @@ mod test {
...
+                    weight: 2,
+                },
+            ],
+            threshold: 8,
+            nonce: "8784bf7be5a9baaeea47e12d9e8ad0dec29afcbc3617d97f771e3c24fa945dce"
+                .parse()
+                .unwrap(),
+        };
+
+        goldie::assert!(HexBinary::from(weighted_signers.hash().unwrap()).to_hex());
```
milapsheth, line 487:
```suggestion
        goldie::assert_json!([
            hex::encode(assert_ok!(weighted_signers.hash())),
            hex::encode(assert_ok!(weighted_signers.signer_rotation_hash())),
        ]);
```
instead of duplicating the dummy data, I recommend combining the two tests into one and using `assert_json!`, since they're very related anyways. Alternative would be to create a helper `dummy_signers` method that returns the data.

Also suggested using `hex::encode` and `assert_ok!` instead

ahramy: I've updated the code, similar to what `proof_signature_encode` test does. Adding assert_ok introduces new dependencies, so I need to add `assert_ok = "1.0" to Cargo.toml`, and I'd prefer to avoid adding new dependencies unless necessary. Let me know if you think we should still use `assert_ok!`

ampd/src/stellar/verifier.rs
```diff
@@ -372,19 +353,23 @@ mod test {
...
+        .unwrap();
 
         let event_body = ContractEventBody::V0(ContractEventV0 {
-            topics: vec![ScVal::Symbol(ScSymbol(
-                StringM::from_str(TOPIC_ROTATED).unwrap(),
-            ))]
+            topics: vec![
+                ScVal::Symbol(ScSymbol(StringM::from_str(TOPIC_ROTATED).unwrap())),
+                ScVal::U64(1),
+                ScVal::Bytes(ScBytes(weighted_signers_hash)),
```
milapsheth, line 368:
```suggestion
                ScVal::Bytes(ScBytes(BytesM::try_from(weighted_signers_hash).unwrap())),
```
convert it here perhaps for consistency with other topic values above

milapsheth (approved):

ampd/src/stellar/verifier.rs
ahramy: Agreed. updated.

ahramy: after updating the code, it gives lint error:
```
error: useless conversion to the same type: `stellar_xdr::curr::BytesM<u32::MAX>`
   --> ampd/src/stellar/verifier.rs:368:38
    |
368 |                 ScVal::Bytes(ScBytes(BytesM::try_from(weighted_signers_hash).unwrap())),
    |                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: consider removing `BytesM::try_from()`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#useless_conversion
    = note: `-D clippy::useless-conversion` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::useless_conversion)]`

error: could not compile `ampd` (lib test) due to 2 previous errors
```

Thus, reverted

milapsheth: well, you need to remove the BytesM conversion where the variable is being created first since it's being done here

PR #664, opened 2024-10-21
fix(multisig-prover): update stellar gateway message definition

ahramy:
## Description
[AXE-6113](https://axelarnetwork.atlassian.net/browse/AXE-6113)

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Other Notes

[AXE-6113]: https://axelarnetwork.atlassian.net/browse/AXE-6113?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


external-gateways/stellar/src/lib.rs
```diff
@@ -75,19 +75,19 @@ impl TryFrom<Message> for ScVal {
...
     fn try_from(value: Message) -> Result<Self, XdrError> {
         let keys: [&'static str; 5] = [
-            "contract_address",
+            "source_chain",
             "message_id",
-            "payload_hash",
             "source_address",
-            "source_chain",
+            "contract_address",
+            "payload_hash",
```
milapsheth, line 82: note that this doesn't start with `source_chain`. I believe that's because XDR sorts the keys first, so the order within the type doesn't matter. Add a golden test in Soroban for the message XDR encoding to compare. Pretty sure this shouldn't be changed in this case (add a comment on the behaviour if it's sorting to clarify this in the future)

ahramy: Added comment, updated.

milapsheth (approved):

PR #668, opened 2024-10-25
feat(minor-its-hub): add support for global freeze

cjcobb23:
* governance or admin can disable/enable execution of its messages

https://axelarnetwork.atlassian.net/browse/AXE-6233

### DIFF
--- a/contracts/interchain-token-service/src/contract.rs
+++ b/contracts/interchain-token-service/src/contract.rs
@@ -1,7 +1,7 @@
 use std::fmt::Debug;
 
 use axelar_wasm_std::error::ContractError;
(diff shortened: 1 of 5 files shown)


contracts/interchain-token-service/src/contract.rs
cgorenflo, line 1: is it supposed to only disable message passing and not (de)registering contracts?

cjcobb23: that is my intention, I think it would be useful to still be able to execute those admin/governance commands while message passing is frozen, but @milapsheth should confirm

cgorenflo (approved):

PR #669, opened 2024-10-28
feat(interchain-token-service): add message type SetChainConfig

fish-sammy:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


contracts/interchain-token-service/src/msg.rs
```diff
@@ -33,6 +34,13 @@ pub enum ExecuteMsg {
     /// The admin is allowed to remove the ITS address of a chain for emergencies.
     #[permission(Elevated)]
     DeregisterItsContract { chain: ChainNameRaw },
+
+    #[permission(Governance)]
+    SetChainConfig {
+        chain: ChainNameRaw,
+        max_uint: Uint256,
+        max_target_decimals: u8,
+    },
```
cjcobb23, line 43: can't this be part of `RegisterItsContract`? Why make it a separate execute msg?

contracts/interchain-token-service/src/state.rs
```diff
@@ -25,8 +26,15 @@ pub struct Config {
...
 
+#[cw_serde]
+pub struct ChainConfig {
+    max_uint: Uint256,
+    max_target_decimals: u8,
+}
+
 const CONFIG: Item<Config> = Item::new("config");
 const ITS_CONTRACTS: Map<&ChainNameRaw, Address> = Map::new("its_contracts");
+const CHAIN_CONFIGS: Map<&ChainNameRaw, ChainConfig> = Map::new("chain_configs");
```
cjcobb23, line 37: Would it make sense to combine these two maps?

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -132,3 +136,25 @@ pub fn deregister_its_contract(deps: DepsMut, chain: ChainNameRaw) -> Result<Res
...
 }
+
+pub fn set_chain_config(
+    deps: DepsMut,
+    chain: ChainNameRaw,
+    max_uint: Uint256,
+    max_target_decimals: u8,
+) -> Result<Response, Error> {
+    match state::load_chain_config(deps.storage, &chain).change_context(Error::StorageError)? {
+        Some(_) => bail!(Error::ChainConfigAlreadySet(chain)),
```
cjcobb23, line 147: So no way to update once set?

fish-sammy: correct. It's permanent once set cuz if it does, it means previous transfers are all of a sudden worth more or less on the original chain or elsewhere. This is also why it resides in a separate message and state. It CANNOT be deregistered, unlike the ITS contract address.

contracts/interchain-token-service/src/msg.rs
fish-sammy: see below

contracts/interchain-token-service/src/state.rs
fish-sammy: see below

contracts/interchain-token-service/src/msg.rs
```diff
@@ -33,6 +34,13 @@ pub enum ExecuteMsg {
     /// The admin is allowed to remove the ITS address of a chain for emergencies.
     #[permission(Elevated)]
     DeregisterItsContract { chain: ChainNameRaw },
+
+    #[permission(Governance)]
+    SetChainConfig {
```
milapsheth, line 39: Hmm, how about we define `RegisterChainConfig`, and `UpdateChainItsContract` msg types instead? Register config will allow setting the config, whereas update will only allow updating specific fields. It's similar to what we do in the [router](https://github.com/axelarnetwork/axelar-amplifier/blob/main/packages/router-api/src/msg.rs#L21) where the msg id format isn't allowed to be changed, but the gateway address is

contracts/interchain-token-service/src/msg.rs
```diff
@@ -33,6 +34,13 @@ pub enum ExecuteMsg {
...
+        chain: ChainNameRaw,
+        max_uint: Uint256,
+        max_target_decimals: u8,
+    },
 }
 
 #[cw_serde]
```
milapsheth, line 46: add a query to see the registered config per chain. If we combine the two configs as per above, we can simplify the query as well

contracts/interchain-token-service/src/msg.rs
```diff
@@ -33,6 +34,13 @@ pub enum ExecuteMsg {
     /// The admin is allowed to remove the ITS address of a chain for emergencies.
     #[permission(Elevated)]
     DeregisterItsContract { chain: ChainNameRaw },
+
+    #[permission(Governance)]
+    SetChainConfig {
+        chain: ChainNameRaw,
+        max_uint: Uint256,
```
milapsheth, line 41:
```suggestion
        max_uint: u8,
```
Hmm, do we want to support non power-of-two aligned max values? I suppose it's more future proof even if we don't need it atm, and saves some computation

contracts/interchain-token-service/src/msg.rs
milapsheth, line 39: add a comment to explain what these values mean

contracts/interchain-token-service/src/state.rs
```diff
@@ -38,6 +46,29 @@ pub fn save_config(storage: &mut dyn Storage, config: &Config) -> Result<(), Err
...
+    Ok(CHAIN_CONFIGS.may_load(storage, chain)?)
+}
+
+pub fn save_chain_config(
+    storage: &mut dyn Storage,
+    chain: &ChainNameRaw,
+    max_uint: Uint256,
+    max_target_decimals: u8,
+) -> Result<(), Error> {
+    Ok(CHAIN_CONFIGS.save(
```
milapsheth, line 62: Prevent overwriting in this method so it can't be misused, similar to `register_its_contract`

contracts/interchain-token-service/src/msg.rs
fish-sammy: I prefer the current approach instead of saving the powers. It also doesn't save any computation.

contracts/interchain-token-service/src/state.rs
fish-sammy: I don't like the idea of embedding business logic in I/O methods.

cjcobb23: I actually also prefer preventing overwriting here. Doesn't feel like business logic to me, but rather a state invariant, which I think is appropriate for this module to handle.

fish-sammy: The state being not able to overwrite is business logic though.

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -132,3 +136,25 @@ pub fn deregister_its_contract(deps: DepsMut, chain: ChainNameRaw) -> Result<Res
...
+        None => state::save_chain_config(
+            deps.storage,
+            &chain,
+            max_uint
+                .try_into()
+                .map_err(Report::new)
+                .change_context(Error::InvalidChainMaxUint)?,
```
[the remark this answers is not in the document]
fish-sammy: The `StorageError` to me means anything that comes from the `state` module. It is indeed unknow to here.

contracts/interchain-token-service/src/msg.rs
fish-sammy: Are you suggesting a removal of `DeregisterItsContract`?

contracts/interchain-token-service/src/contract/execute.rs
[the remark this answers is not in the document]
cjcobb23: that's not how we use that error across the contracts though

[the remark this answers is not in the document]
cjcobb23: https://github.com/axelarnetwork/axelar-amplifier/blob/4adfcefe088e5707a67c5798b2fa5b55f605b7d5/contracts/voting-verifier/src/error.rs#L50
https://github.com/axelarnetwork/axelar-amplifier/blob/4adfcefe088e5707a67c5798b2fa5b55f605b7d5/contracts/multisig-prover/src/error.rs#L96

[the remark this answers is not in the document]
fish-sammy: `StateError` reflects anything that comes from the state module, whatever it is. Functions in execute can attach specific info, such as InvalidMaxUint, but it cannot attach anything to `StateError` since it doesn't know anything about it. Now `contract.rs` should be the place to say `FailedToSetChainConfig` since it knows that execute function fails and that's all it knows. 

"We use StorageError across many contracts to just wrap cw_storage_plus errors" if that's the case, then `StorageError` not never be one of the contract errors but the state modules's Error. The better way of using error stack is to attach info at each layer if available, instead of attching anything. Also, the worst possible way of using it is for a function such as `set_chain_config` to have to attach `FailedToSetChainConfig` at every single error it possibly returns.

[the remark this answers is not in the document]
cjcobb23: `StateError` is fine with me. And I also agree that `StorageError` should be an error of the state module. Unfortunately I think a lot of our contracts don't define errors for each module, but just errors used across the whole contract. But in ITS hub, I would prefer `StorageError` to live in the state module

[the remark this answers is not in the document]
fish-sammy: Anyway, I'll likely re-write the whole thing with the interface change we just talked about. I'll get rid of all the unused errors. There are a couple others I noticed as well.

contracts/interchain-token-service/tests/execute.rs
```diff
@@ -314,3 +314,58 @@ fn execute_message_when_invalid_message_type_fails() {
...
+        chain.clone(),
+        max_uint,
+        decimals
+    ));
+    assert_err_contains!(
+        utils::set_chain_config(deps.as_mut(), chain, max_uint, decimals),
+        ExecuteError,
+        ExecuteError::ChainConfigAlreadySet(_)
+    )
+}
```
cjcobb23, line 371: I think these tests should be in the execute module file, I don't see why they need to be here instead

contracts/interchain-token-service/src/msg.rs
cjcobb23, line 41: why not:
```suggestion
        max_uint: nonempty::Uint256,
```

cjcobb23 (approved):
I still have some outstanding comments, but not going to block on those, address them at your discretion. Looks good to me.

contracts/interchain-token-service/tests/execute.rs
fish-sammy: I 100% agree with you, but all execute mod tests were here before my change. The move can be done later.

contracts/interchain-token-service/src/msg.rs
fish-sammy: later

PR #670, opened 2024-10-29
feat(minor-interchain-token-service): freeze/unfreeze

cjcobb23:
* Add support for freezing and unfreezing specific chains

https://axelarnetwork.atlassian.net/browse/AXE-6265

### DIFF
--- a/contracts/interchain-token-service/src/contract/execute.rs
+++ b/contracts/interchain-token-service/src/contract/execute.rs
@@ -64,8 +64,7 @@ pub fn execute_message(
             message,
         } => {
             ensure!(
-                !is_chain_frozen(deps.storage, &destination_chain)
-                    .change_context(Error::State)?,
+                !is_chain_frozen(deps.storage, &destination_chain).change_context(Error::State)?,
                 Error::ChainFrozen(destination_chain)
             );
             let destination_address = load_its_contract(deps.storage, &destination_chain)
@@ -174,7 +173,6 @@ pub fn enable_execution(deps: DepsMut) -> Result<Response, Error> {
     killswitch::disengage(deps.storage, Event::ExecutionEnabled).change_context(Error::State)
 }
 
-
 pub fn set_chain_config(
     deps: DepsMut,
     chain: ChainNameRaw,
@@ -201,7 +199,8 @@ mod tests {
     use router_api::{ChainNameRaw, CrossChainId};
 
     use crate::contract::execute::{
-        disable_execution, enable_execution, execute_message, freeze_chain, register_its_contract, unfreeze_chain, Error
+        disable_execution, enable_execution, execute_message, freeze_chain, register_its_contract,
+        unfreeze_chain, Error,
     };
     use crate::state::{self, Config};
     use crate::{HubMessage, Message};
--- a/contracts/interchain-token-service/src/contract/execute.rs
+++ b/contracts/interchain-token-service/src/contract/execute.rs
@@ -53,20 +53,14 @@ pub fn execute_message(
         Error::ExecutionDisabled
     );
     ensure_its_source_address(deps.storage, &cc_id.source_chain, &source_address)?;
-    ensure!(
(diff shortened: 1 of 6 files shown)


contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -21,12 +21,12 @@ pub enum Error {
...
     #[error("failed to deregister its contract for chain {0}")]
     FailedItsContractDeregistration(ChainNameRaw),
-    #[error("failed to execute message")]
-    FailedExecuteMessage,
     #[error("failed to query nexus")]
     NexusQueryError,
-    #[error("storage error")]
-    StorageError,
+    #[error("state error")]
+    StateError,
```
fish-sammy, line 27:
```suggestion
    #[error(transparent)]
    State(#[from] state::Error)
```
Does this work?

cgorenflo: not sure that works with `change_context`, but either way, it should be `State` and not `StateError`

cjcobb23: yeah `transparent` doesn't work very cleanly with `change_context`

contracts/interchain-token-service/src/state.rs
```diff
@@ -76,9 +86,27 @@ pub fn remove_its_contract(storage: &mut dyn Storage, chain: &ChainNameRaw) -> R
...
+pub fn is_chain_frozen(storage: &dyn Storage, chain: &ChainNameRaw) -> Result<bool, Error> {
+    FROZEN_CHAINS
+        .may_load(storage, chain)
+        .change_context(Error::Storage)
+        .map(|res| res.is_some())
+}
+
+pub fn freeze_chain(storage: &mut dyn Storage, chain: &ChainNameRaw) -> Result<(), Error> {
+    FROZEN_CHAINS
+        .save(storage, chain, &())
```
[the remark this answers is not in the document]
cjcobb23: There is a test for this that passes, and I tested it on a devnet as well just now, no issues. So appears to work correctly

contracts/interchain-token-service/src/contract/execute.rs
milapsheth: Instead of `.change_context(Error::State)?` shall we do `.map_err(Error::State)?` if we want to wrap it. Not a strong opinion either way

contracts/interchain-token-service/src/state.rs
```diff
@@ -29,6 +34,7 @@ pub struct ChainConfig {
 
 const CONFIG: Item<Config> = Item::new("config");
 const ITS_CONTRACTS: Map<&ChainNameRaw, Address> = Map::new("its_contracts");
+const FROZEN_CHAINS: Map<&ChainNameRaw, ()> = Map::new("frozen_chains");
 const CHAIN_CONFIGS: Map<&ChainNameRaw, ChainConfig> = Map::new("chain_configs");
```
milapsheth, line 38: shall we add freeze status in the chain config instead? It also ensures that we only freeze registered chains, and not any chain name

contracts/interchain-token-service/src/state.rs
```diff
@@ -102,9 +118,27 @@ pub fn remove_its_contract(storage: &mut dyn Storage, chain: &ChainNameRaw) -> R
...
+pub fn is_chain_frozen(storage: &dyn Storage, chain: &ChainNameRaw) -> Result<bool, Error> {
+    FROZEN_CHAINS
+        .may_load(storage, chain)
+        .change_context(Error::Storage)
+        .map(|res| res.is_some())
+}
+
+pub fn freeze_chain(storage: &mut dyn Storage, chain: &ChainNameRaw) -> Result<(), Error> {
+    FROZEN_CHAINS
+        .save(storage, chain, &())
```
milapsheth, line 136: check if the chain is actually registered?

PR #671, opened 2024-10-30
feat(major-interchain-token-service): apply token balance invariants

milapsheth:
## Description

[AXE-4185]

- [ ] Add more tests

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Other Notes

[AXE-4185]: https://axelarnetwork.atlassian.net/browse/AXE-4185?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -167,6 +174,80 @@ pub fn set_chain_config(
...
+///    - Increases the token balance on the destination chain, unless it's the origin chain in which case it's decreased.
+///    - If the balance underflows for either case, an error is returned.
+///
+/// 2. DeployInterchainToken:
+///    - If a custom minter is not set, then the token balance is tracked for the source/destination chain.
+///    - If the custom minter is set, then the balance is not tracked, but the deployment is recorded.
+///
+/// 3. DeployTokenManager:
+///    - Same as the custom minter being set above. ITS Hub can't know if the existing token on the destination chain has a custom minter set.
+fn apply_balance_invariant(
```
milapsheth, line 198: Base on suggestion from Sammy, I'll separate this into individual invariants, e.g.
- deploy invariant: initialize token balance
- redeploy invariant: check if token isn't already deployed
- balance invariant: update balances and fail on underflow

contracts/interchain-token-service/README.md
```diff
@@ -10,6 +10,10 @@ The Interchain Token Service (ITS) Hub contract is a crucial component of a cros
...
 
-### Cross-chain messaging
+## Cross-chain messaging
 
 The ITS Hub makes use of the Axelarnet gateway [contract](../contracts/axelarnet-gateway/) to facilitate sending or receiving cross-chain messages. Messages are sent via `CallContract`, and received when the Axelarnet gateway is executed (by a relayer / user) through `Execute`, which in turn executes ITS Hub's `Execute` method.
+
+## Balance Invariants
+
+ITS Hub maintains balance invariants for native interchain tokens for every chain they're deployed to. This helps isolate the security risk between chains. A compromise on one chain can only affect the token balance that was moved to that chain in the worst case. For e.g. say if USDC was deployed from Ethereum to Solana via the ITS Hub, and 10M USDC was moved to Solana (in total). If there's a compromise on Solana (potentially of the chain itself, or the Solana ITS contract or another related contract), an attacker can only withdraw at most 10M USDC back to Ethereum or another chain (and not all the bridged USDC locked on the Ethereum ITS contract). ITS Hub will prevent all USDC transfers from Solana once 10M USDC has been moved back out from it.
```
cgorenflo, line 19:
```suggestion
ITS Hub maintains balance invariants for native interchain tokens for every chain they're deployed to. This helps isolate the security risk between chains. A compromised chain can only affect the token balance that was moved to that chain in the worst case. For e.g. if USDC was deployed from Ethereum to Solana via the ITS Hub, and 10M USDC was moved to Solana (in total). If there's a compromise on Solana (potentially the chain itself, or the Solana ITS contract or another related contract), an attacker can only withdraw at most 10M USDC back to Ethereum or another chain (and not all the bridged USDC locked on the Ethereum ITS contract). ITS Hub will prevent all USDC transfers from Solana once 10M USDC has been moved back out from it.
```

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -167,6 +174,80 @@ pub fn set_chain_config(
...
+    storage: &mut dyn Storage,
+    message: &Message,
+    chain: ChainNameRaw,
+    is_source_chain: bool,
+) -> Result<(), Error> {
```
cgorenflo, line 203: that's a pretty messy signature, and we can do better in Rust. Consider something like:
```
enum Chain{
   Source(ChainNameRaw),
   Destination(ChainNameRaw)
```

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -167,6 +174,80 @@ pub fn set_chain_config(
...
+    match message {
+        Message::InterchainTransfer {
+            token_id, amount, ..
+        } => {
+            state::update_token_info(
```
cgorenflo, line 208: why is this `update` when all the other cases are `save`?

contracts/interchain-token-service/src/contract/query.rs
```diff
@@ -12,3 +12,12 @@ pub fn all_its_contracts(deps: Deps) -> Result<Binary, state::Error> {
...
     Ok(to_json_binary(&contract_addresses)?)
 }
+
+pub fn token_info(
+    deps: Deps,
+    chain: ChainNameRaw,
+    token_id: TokenId,
+) -> Result<Binary, state::Error> {
+    let config = state::may_load_token_info(deps.storage, &chain, &token_id)?;
+    Ok(to_json_binary(&config)?)
```
cgorenflo, line 22: rename config

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -31,6 +32,8 @@ pub enum Error {
     LoadChainConfig(ChainNameRaw),
     #[error("failed to save chain config for chain {0}")]
     SaveChainConfig(ChainNameRaw),
+    #[error("failed to apply invariants for token {0}")]
+    InvariantViolated(TokenId),
```
cgorenflo, line 36: why do we have a invariant violated error on multiple levels?

contracts/interchain-token-service/src/state.rs
```diff
@@ -27,9 +45,24 @@ pub struct ChainConfig {
...
+    /// If the token is native to this chain, this will be the total supply bridged to other chains.
+    balance: Option<Uint256>,
+    /// Whether the token was deployed from this chain.
+    is_origin_chain: bool,
+}
+
+#[cw_serde]
+pub struct TokenInfo {
+    balance: TokenBalance,
+}
```
cgorenflo, line 60: what's the point of this wrapping?

contracts/interchain-token-service/src/state.rs
```diff
@@ -27,9 +45,24 @@ pub struct ChainConfig {
     max_target_decimals: u8,
 }
 
+#[cw_serde]
+pub struct TokenBalance {
+    /// The total balance of the token bridged from/to the chain.
+    /// If the token is native to this chain, this will be the total supply bridged to other chains.
```
cgorenflo, line 51: this comment is super confusing. How can it be both to and from?

cgorenflo: and why is it an option?

contracts/interchain-token-service/src/state.rs
```diff
@@ -107,6 +140,89 @@ pub fn load_all_its_contracts(
         .collect::<Result<HashMap<_, _>, _>>()?)
 }
 
+pub fn save_token_info(
+    storage: &mut dyn Storage,
+    chain: ChainNameRaw,
+    token_id: TokenId,
+    is_origin_chain: bool,
+    track_balance: bool,
```
cgorenflo, line 148: again, this is so hard to follow, try to encode it into the actual types

contracts/interchain-token-service/src/state.rs
```diff
@@ -107,6 +140,89 @@ pub fn load_all_its_contracts(
...
+
+    Ok(())
+}
+
+pub fn update_token_info(
+    storage: &mut dyn Storage,
+    chain: ChainNameRaw,
+    token_id: TokenId,
+    amount: nonempty::Uint256,
+    is_deposit: bool,
```
cgorenflo, line 170: what does that mean?

cgorenflo (changes requested):

contracts/interchain-token-service/src/state.rs
```diff
@@ -107,6 +140,89 @@ pub fn load_all_its_contracts(
...
+                b.checked_sub(amount.into())
+            }
+        })?;
+
+        Ok(TokenBalance {
+            balance,
+            is_origin_chain: *is_origin_chain,
+        })
+    }
+}
```
cjcobb23, line 224: move this near the struct definition

contracts/interchain-token-service/src/state.rs
```diff
@@ -107,6 +140,89 @@ pub fn load_all_its_contracts(
...
+                is_origin_chain,
+            };
+            TOKEN_INFO.save(storage, &(chain, token_id), &TokenInfo { balance })?;
+        }
+    }
+
+    Ok(())
+}
+
+pub fn update_token_info(
```
cjcobb23, line 165:
```suggestion
pub fn update_token_balance(
```

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -60,6 +63,10 @@ pub fn execute_message(
             let destination_address = load_its_contract(deps.storage, &destination_chain)
                 .change_context_lazy(|| Error::UnknownChain(destination_chain.clone()))?;
 
+            apply_balance_invariant(deps.storage, &message, cc_id.source_chain.clone(), true)?;
+
+            apply_balance_invariant(deps.storage, &message, destination_chain.clone(), false)?;
```
cjcobb23, line 68: this looks a bit odd. Maybe combine this into one function call, and just pass in the source chain and destination chain, to avoid the `true` and `false`? The boolean parameters are quite cryptic

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -167,6 +174,80 @@ pub fn set_chain_config(
...
+///    - If a custom minter is not set, then the token balance is tracked for the source/destination chain.
+///    - If the custom minter is set, then the balance is not tracked, but the deployment is recorded.
+///
+/// 3. DeployTokenManager:
+///    - Same as the custom minter being set above. ITS Hub can't know if the existing token on the destination chain has a custom minter set.
+fn apply_balance_invariant(
+    storage: &mut dyn Storage,
+    message: &Message,
+    chain: ChainNameRaw,
+    is_source_chain: bool,
```
cjcobb23, line 202: I would just pass in the chain or chains, instead of this bool

contracts/interchain-token-service/src/state.rs
```diff
@@ -107,6 +140,89 @@ pub fn load_all_its_contracts(
...
+    /// Deposit an amount to the token chain balance.
+    /// If the token originated on this chain, the amount is subtracted from the balance.
+    fn update(&self, amount: nonempty::Uint256, is_deposit: bool) -> Result<Self, OverflowError> {
+        let TokenBalance {
+            balance,
+            is_origin_chain,
+        } = self;
+
+        // If the token originated on this chain, deposits/withdrawals are reversed
+        // since the balance is tracking the total supply that was bridged to other chains.
```
cjcobb23, line 209: this is confusing to me. What exactly are we even tracking for the origin chain? It looks like the total balance that was bridged out of the chain? Which is the reverse of what we track on the other chains (the total balance bridged to the chain?). If that is the case, why do we even need to track anything for the origin chain?

contracts/interchain-token-service/src/state.rs
cjcobb23, line 170: this flag is quite confusing, the way it is being passed around, and inverted, then x'or'd.

contracts/interchain-token-service/src/state.rs
```diff
@@ -27,9 +45,24 @@ pub struct ChainConfig {
     max_target_decimals: u8,
 }
 
+#[cw_serde]
+pub struct TokenBalance {
+    /// The total balance of the token bridged from/to the chain.
+    /// If the token is native to this chain, this will be the total supply bridged to other chains.
+    balance: Option<Uint256>,
```
cjcobb23, line 52: this optional is to indicate whether the balance should be tracked at all right? If so, I'd prefer to just not create this struct at all in that case.

contracts/interchain-token-service/src/state.rs
```diff
@@ -107,6 +140,89 @@ pub fn load_all_its_contracts(
...
+    chain: ChainNameRaw,
+    token_id: TokenId,
+    amount: nonempty::Uint256,
+    is_deposit: bool,
+) -> Result<(), Error> {
+    let Some(mut token_config) =
+        TOKEN_INFO.may_load(storage, &(chain.clone(), token_id.clone()))?
+    else {
+        return Err(Error::TokenNotDeployed { token_id, chain });
+    };
```
cjcobb23, line 176: I feel like this logic should be separated out into it's own function: `is_token_deployed`

contracts/interchain-token-service/src/contract/execute.rs
cjcobb23: yes I like that, this function is way too broad at the moment

cjcobb23 (changes requested):

contracts/interchain-token-service/src/state.rs
```diff
@@ -27,9 +44,65 @@ pub struct ChainConfig {
...
+    Untracked,
+}
+
+#[cw_serde]
+pub enum TokenType {
+    /// The token originated on this chain.
+    Origin,
+    /// The token was deployed to this chain.
+    Remote,
+}
```
cjcobb23, line 62:
```suggestion
#[cw_serde]
pub enum TokenType {
    /// The token originated on this chain.
    Native,
    /// The token was deployed to this chain.
    Bridged,
}
```

contracts/interchain-token-service/src/state.rs
```diff
@@ -107,6 +180,104 @@ pub fn load_all_its_contracts(
...
+    token_config.balance = match directional_chain {
+        DirectionalChain::Source(_) => token_config.balance.checked_sub(amount),
+        DirectionalChain::Destination(_) => token_config.balance.checked_add(amount),
+    }
+    .map_err(|_| Error::TokenBalanceInvariantViolated {
+        token_id: token_id.clone(),
+        chain: chain.clone(),
+    })?;
+
+    TOKEN_INFO.save(storage, &(chain, token_id), &token_config)?;
```
cjcobb23, line 234: I think you could use `update` for this: https://docs.rs/cw-storage-plus/latest/cw_storage_plus/struct.Map.html#method.update

contracts/interchain-token-service/src/state.rs
```diff
@@ -107,6 +180,104 @@ pub fn load_all_its_contracts(
         .collect::<Result<HashMap<_, _>, _>>()?)
 }
 
+pub fn save_token_info(
```
cjcobb23, line 183:
```suggestion
pub fn init_token_info(
```

contracts/interchain-token-service/src/state.rs
```diff
@@ -107,6 +180,104 @@ pub fn load_all_its_contracts(
...
+    storage: &mut dyn Storage,
+    directional_chain: DirectionalChain,
```
cjcobb23, line 185: For this function, I would just pass in both chains as `source` and `dest`, and create both of the necessary structs

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -167,6 +184,71 @@ pub fn set_chain_config(
...
+/// 2. DeployInterchainToken:
+///    - If a custom minter is not set, then the token balance is tracked for the destination chain.
+///    - If a custom minter is set, then the balance is not tracked, but the deployment is recorded as `TokenInfo`.
+///
+/// 3. DeployTokenManager:
+///    - Same as the custom minter being set above. ITS Hub can't know if the existing token on the destination chain has a custom minter set.
+fn apply_balance_invariant(
+    storage: &mut dyn Storage,
+    message: &Message,
+    directional_chain: DirectionalChain,
```
cjcobb23, line 211: I would pass in source and destination chain here, so you can just call this function once in execute. The double call in execute looks messy.

contracts/interchain-token-service/src/state.rs
cjcobb23: actually nvm, this is fine here

contracts/interchain-token-service/src/state.rs
```diff
@@ -107,6 +180,104 @@ pub fn load_all_its_contracts(
...
+        match (token_type, token_deployment_type) {
+            // Token balances are tracked for remote trustless tokens
+            (TokenType::Remote, TokenDeploymentType::Trustless) => {
+                TokenBalance::Tracked(Uint256::zero())
+            }
+            _ => TokenBalance::Untracked,
+        }
+    }
+}
+
```
cjcobb23, line 280: The added logic in this file deserves some unit tests

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -167,6 +184,71 @@ pub fn set_chain_config(
...
+                directional_chain,
+                token_id.clone(),
+                TokenDeploymentType::CustomMinter,
+            )
+            .change_context_lazy(|| Error::SaveTokenInfo(token_id.clone()))?;
+        }
+    };
+
+    Ok(())
+}
```
cjcobb23, line 250: this function deserves unit tests

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -186,16 +209,179 @@ pub fn set_chain_config(
...
+                chain: message_direction.clone().into(),
+            }
+        );
+    }
+
+    Ok(())
+}
+
+/// Ensures that the token is not being redeployed to the same destination chain.
+fn token_redeployment_check(
```
fish-sammy, line 314:
```suggestion
fn check_token_deployment(
```

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -186,16 +209,179 @@ pub fn set_chain_config(
...
+        message_direction.into(),
+        message.token_id(),
+        &token_chain_info,
+    )
+    .change_context(Error::State)
+}
+
+/// Ensures that the token is being redeployed from the same origin chain.
+/// If it's being deployed for the first time, the from chain is saved as the origin chain in the token config.
+fn token_deployment_origin_chain_check(
```
fish-sammy, line 255:
```suggestion
fn check_token_deployment_origin_chain(
```
Actually, `check` isn't even a good name here as this function writes to the storage but `check` suggests that it does not.

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -186,16 +209,179 @@ pub fn set_chain_config(
...
+    ) && matches!(message_direction, MessageDirection::To(_))
+        && token_chain_info.is_some()
+    {
+        bail!(Error::TokenAlreadyDeployed {
+            token_id: message.token_id(),
+            chain: message_direction.clone().into(),
+        });
+    }
+
+    Ok(())
```
fish-sammy, line 331:
```suggestion
    match (message, message_direction, token_chain_info) {
        (
            Message::DeployInterchainToken { .. } | Message::DeployTokenManager { .. },
            MessageDirection::To(_),
            Some(_),
        ) => {
            bail!(Error::TokenAlreadyDeployed {
                token_id: message.token_id(),
                chain: message_direction.clone().into(),
            });
        }
        _ => Ok(()),
    }
```
Chain of `&&` is very hard to read.

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -186,16 +209,179 @@ pub fn set_chain_config(
...
+        ensure!(
+            token_chain_info.is_none(),
+            Error::TokenAlreadyDeployed {
+                token_id: message.token_id(),
+                chain: message_direction.clone().into(),
+            }
+        );
+    }
+
+    Ok(())
```
fish-sammy, line 310:
```suggestion
    match (message, message_direction) {
        (
            Message::DeployInterchainToken { .. } | Message::DeployTokenManager { .. },
            MessageDirection::From(_),
        ) => {
            ...
        }
        _ => return Ok(()),
    }
```

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -186,16 +209,179 @@ pub fn set_chain_config(
...
+        });
+    }
+
+    Ok(())
+}
+
+/// Ensures that the token info is recorded on deployment.
+/// The token supply tracking is also enabled on deployment for trustless token deployments (i.e no minter set).
+/// Tokens that haven't been deployed yet cannot be transferred.
+fn token_deployment_check(
```
fish-sammy, line 337:
```suggestion
fn check_token_deployment(
```

contracts/interchain-token-service/README.md
```diff
@@ -2,14 +2,18 @@
...
 
-### Cross-chain messaging
+## Cross-chain messaging
 
 The ITS Hub makes use of the Axelarnet gateway [contract](../contracts/axelarnet-gateway/) to facilitate sending or receiving cross-chain messages. Messages are sent via `CallContract`, and received when the Axelarnet gateway is executed (by a relayer / user) through `Execute`, which in turn executes ITS Hub's `Execute` method.
+
+## Token Supply Invariant
+
+ITS Hub maintains the token supply for native interchain tokens for every chain they're deployed to. This helps isolate the security risk between chains. A compromised chain can only affect the total token supply that was sent to that chain in the worst case. For e.g. if USDC was deployed from Ethereum to Solana via the ITS Hub, and 10M USDC was moved to Solana (in total). If there's a compromise on Solana (potentially the chain itself, or the Solana ITS contract, or another related contract), an attacker can only withdraw at most 10M USDC back to Ethereum or another chain (and not all the bridged USDC locked on the Ethereum ITS contract). ITS Hub will prevent all USDC transfers from Solana once 10M USDC has been moved back out from it.
```
cgorenflo, line 19: Either "For example" or "e.g.", not bot.

```suggestion
ITS Hub maintains the token supply for native interchain tokens for every chain they're deployed to. This helps isolate the security risk between chains. A compromised chain can only affect the total token supply that was sent to that chain in the worst case. For example, if USDC was deployed from Ethereum to Solana via the ITS Hub, and 10M USDC was moved to Solana (in total). If there's a compromise on Solana (potentially the chain itself, or the Solana ITS contract, or another related contract), an attacker can only withdraw at most 10M USDC back to Ethereum or another chain (and not all the bridged USDC locked on the Ethereum ITS contract). ITS Hub will prevent all USDC transfers from Solana once 10M USDC has been moved back out from it.
```

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -186,16 +209,179 @@ pub fn set_chain_config(
...
+        storage,
+        message_direction.clone().into(),
+        message.token_id(),
+    )
+    .change_context(Error::State)?;
+
+    // Validation checks
+    chain_frozen_check(storage, &message_direction.clone().into())?;
+    token_redeployment_check(message, &message_direction, &token_chain_info)?;
+    token_deployment_origin_chain_check(storage, message, &message_direction, &token_chain_info)?;
```
cgorenflo, line 236: rename to `check_*`, verb first

cgorenflo: I would even prefer naming like `ensure_chain_not_frozen`, etc.

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -186,16 +209,179 @@ pub fn set_chain_config(
...
+    .change_context(Error::State)
+}
+
+/// Ensures that the token is being redeployed from the same origin chain.
+/// If it's being deployed for the first time, the from chain is saved as the origin chain in the token config.
+fn token_deployment_origin_chain_check(
+    storage: &mut dyn Storage,
+    message: &Message,
+    message_direction: &MessageDirection,
+    token_chain_info: &Option<TokenChainInfo>,
```
cgorenflo, line 259: always use `Option<&>` instead of `&Option<>`

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -186,16 +209,179 @@ pub fn set_chain_config(
...
+        }
+        Message::InterchainTransfer { .. } => bail!(Error::TokenNotDeployed {
+            token_id: message.token_id(),
+            chain: message_direction.clone().into(),
+        }),
+    };
+
+    Ok(TokenChainInfo::new(
+        (message_direction, token_deployment_type).into(),
+    ))
```
fish-sammy, line 359:
```suggestion
    match (message, token_chain_info) {
        (_, Some(token_chain_info)) => Ok(token_chain_info),
        (Message::InterchainTransfer { .. }, None) => bail!(Error::TokenNotDeployed {
            token_id: message.token_id(),
            chain: message_direction.clone().into(),
        }),
        (Message::DeployInterchainToken { minter: None, .. }, None) => Ok(TokenChainInfo::new(
            (message_direction, TokenDeploymentType::Trustless).into(),
        )),
        (Message::DeployInterchainToken { .. } | Message::DeployTokenManager { .. }, None) => Ok(
            TokenChainInfo::new((message_direction, TokenDeploymentType::CustomMinter).into()),
        ),
    }
```

fish-sammy (changes requested):

contracts/interchain-token-service/src/state.rs
```diff
@@ -159,6 +199,116 @@ pub fn unfreeze_chain(
...
+    fn checked_sub(&self, amount: nonempty::Uint256) -> Result<Self, OverflowError> {
+        match self {
+            TokenSupply::Tracked(supply) => {
+                TokenSupply::Tracked(supply.checked_sub(amount.into())?)
+            }
+            TokenSupply::Untracked => TokenSupply::Untracked,
+        }
+        .then(Ok)
+    }
+}
```
cgorenflo, line 262: I don't think this should be implemented on the supply, this can lead to unexpected coding errors when `Untracked` can also add and sub

cgorenflo (approved):

contracts/interchain-token-service/src/state.rs
```diff
@@ -159,6 +199,116 @@ pub fn unfreeze_chain(
...
+
+impl From<MessageDirection> for ChainNameRaw {
+    fn from(message_direction: MessageDirection) -> Self {
+        match message_direction {
+            MessageDirection::From(chain) => chain,
+            MessageDirection::To(chain) => chain,
+        }
+    }
+}
```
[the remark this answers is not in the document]
milapsheth: Yeah, I'm thinking that'll be cleaner as well

contracts/interchain-token-service/src/contract/execute.rs
milapsheth: Yeah, they are a mixture of checks and transformations. I'm thinking whether we need any consistency to identify all such functions (in which case we can name each in a more apt way), or we use some other word like `handler` instead of `check`

contracts/interchain-token-service/src/contract/execute.rs
milapsheth: I didn't find either to be too clean, but i can go with whatever folks prefer

contracts/interchain-token-service/src/contract/execute.rs
milapsheth: I didn't find that to be too readable so I separated the check

contracts/interchain-token-service/src/state.rs
milapsheth: In my mind, the caller shouldn't have to worry whether the supply underthehood is being tracked or not. But not a strong opinion, we could handle it more like an `Option`

fish-sammy:
https://interoplabs.slack.com/archives/C03HHV83DAS/p1730307028137419?thread_ts=1730305631.136499&cid=C03HHV83DAS

contracts/interchain-token-service/src/state.rs
```diff
@@ -48,10 +48,28 @@ pub enum TokenSupply {
 
 /// Information about a token on a specific chain.
 #[cw_serde]
-pub struct TokenChainInfo {
+pub struct TokenInstantiation {
+    pub token_id: TokenId,
```
milapsheth, line 52: why is token_id being added? it's already part of the key

contracts/interchain-token-service/src/state.rs
```diff
@@ -48,10 +48,28 @@ pub enum TokenSupply {
 
 /// Information about a token on a specific chain.
 #[cw_serde]
-pub struct TokenChainInfo {
+pub struct TokenInstantiation {
```
milapsheth, line 51: This name is worse imo, since supply is being updated and not just instantiated

contracts/interchain-token-service/src/state.rs
```diff
@@ -60,24 +78,25 @@ pub enum MessageDirection {
...
+    /// The token is trustless, i.e. only owned by ITS.
     Trustless,
     /// The token has a custom minter.
     CustomMinter,
 }
 
 /// Global config for a token.
 #[cw_serde]
-pub struct TokenConfig {
+pub struct GlobalTokenConfig {
```
milapsheth, line 89: Weird to explicitly label it as Global, I'd prefer renaming the other struct to make it more obvious that it's per chain. Similar to all verifiers in service registry contract vs verifiers registered by chain.

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -74,50 +75,81 @@ pub fn execute_message(
...
+    source_chain: &ChainNameRaw,
+    destination_chain: &ChainNameRaw,
+    message: &Message,
+) -> Result<(), Error> {
+    ensure_chain_not_frozen(storage, source_chain)?;
+    ensure_chain_not_frozen(storage, destination_chain)?;
+
+    match message {
+        Message::InterchainTransfer { amount, .. } => {
+            apply_transfer(amount)?;
```
milapsheth, line 139: the supply change for destination chain will need to be done after the scaling has been applied so `apply_*` should consume the message and return the transformed message for further processing

contracts/interchain-token-service/tests/execute.rs
```diff
@@ -573,3 +552,411 @@ fn set_chain_config_should_fail_if_chain_config_is_already_set() {
...
+            source_its_contract.clone(),
+            msg,
+        ),
+        ExecuteError,
+        ExecuteError::TokenAlreadyDeployed { .. }
+    );
+}
+
+#[test]
+fn deploy_interchain_token_from_non_origin_chain_fails() {
```
cgorenflo, line 889: I didn't fix this test on purpose. The deployment fails as expected, but the reason is that there already is a token deployed on that chain. I believe a test that checks that you can't deploy from a non-origin chain should actually try to deploy to a chain without token.

cgorenflo: So this is up for discussion

milapsheth: Yeah, the test was relying on the ordering of the checks so it would hit the origin chain check first and fail. But I agree it's better to use a new destination chain as well where the token is not deployed to make it more clear. It should fail in both cases, but the latter failure is more obviously a result of the check in question

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -217,169 +255,158 @@ pub fn set_chain_config(
...
-    }
+    // Token is being deployed for the first time
+    let token_config = GlobalTokenConfig {
+        token_id: token_id.clone(),
+        origin_chain: source_chain.clone(),
+    };
+    let instance = TokenInstance::new_on_origin(token_id.clone());
 
+    state::save_global_token_config(storage, &token_id, &token_config)
+        .and_then(|_| state::save_token_instantiation(storage, source_chain, token_id, &instance))
```
milapsheth, line 366: I'd rather see these as separate statements for readability and since the ordering is independent

contracts/interchain-token-service/src/state.rs
```diff
@@ -60,24 +79,25 @@ pub enum MessageDirection {
...
     pub origin_chain: ChainNameRaw,
 }
 
 const CONFIG: Item<Config> = Item::new("config");
 const ITS_CONTRACTS: Map<&ChainNameRaw, Address> = Map::new("its_contracts");
 const CHAIN_CONFIGS: Map<&ChainNameRaw, ChainConfig> = Map::new("chain_configs");
-const TOKEN_CHAIN_INFO: Map<&(ChainNameRaw, TokenId), TokenChainInfo> =
-    Map::new("token_chain_info");
-const TOKEN_CONFIGS: Map<&TokenId, TokenConfig> = Map::new("token_configs");
+const TOKEN_INSTANTIATIONS: Map<&(ChainNameRaw, TokenId), TokenInstance> =
```
milapsheth, line 98:
```suggestion
const TOKEN_INSTANTIATIONS: Map<(&ChainNameRaw, &TokenId), TokenInstance> =
```
Can we switch to this to avoid some cloning?

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -217,169 +255,158 @@ pub fn set_chain_config(
...
-        token_id, amount, ..
-    } = message
-    {
-        token_chain_info
-            .update_supply(*amount, message_direction.clone())
-            .change_context_lazy(|| Error::TokenSupplyInvariantViolated {
-                token_id: token_id.clone(),
-                chain: message_direction.clone().into(),
-            })?;
+trait DeploymentType {
```
milapsheth, line 392: :+1:

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -186,16 +247,168 @@ pub fn set_chain_config(
...
+    storage: &mut dyn Storage,
+    source_chain: ChainNameRaw,
+    token_id: TokenId,
+) -> Result<(), Error> {
+    // Token is being deployed for the first time
+    let token_config = GlobalTokenConfig {
+        token_id: token_id.clone(),
+        origin_chain: source_chain.clone(),
+    };
+    let instance = TokenInstance::new_on_origin(token_id.clone());
```
fish-sammy, line 363: Why make the `GlobalTokenConfig` contain `token_id`? If there's a reason, why doesn't `TokenInstance` contain the `origin_chain`?

fish-sammy: Also, why is the `TokenInstance` on the original chain always untracked?

contracts/interchain-token-service/src/state.rs
```diff
@@ -192,104 +210,77 @@ pub fn unfreeze_chain(
...
-impl TokenSupply {
-    fn checked_add(&self, amount: nonempty::Uint256) -> Result<Self, OverflowError> {
-        match self {
-            TokenSupply::Tracked(supply) => {
-                TokenSupply::Tracked(supply.checked_add(amount.into())?)
-            }
+impl TokenInstance {
+    pub fn update_supply(
+        source: &mut Self,
+        destination: &mut Self,
```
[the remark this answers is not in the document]
fish-sammy: If I understand it correctly, the amount won't be the same for the source and the destination later on so this probably wouldn't work. The source will use the amount before applying the scaling factor and destination will use the amount after.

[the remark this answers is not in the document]
fish-sammy: I also would prefer making this method return the new token insteaces after updating the supply, instead of mutating them.

contracts/interchain-token-service/src/contract/query.rs
```diff
@@ -23,3 +27,19 @@ pub fn all_its_contracts(deps: Deps) -> Result<Binary, Error> {
     let contract_addresses = load_all_its_contracts(deps.storage).change_context(Error::State)?;
     to_json_binary(&contract_addresses).change_context(Error::JsonSerialization)
 }
+
+pub fn token_instantiation(
```
fish-sammy, line 31:
```suggestion
pub fn token_instance(
```

contracts/interchain-token-service/src/contract/query.rs
```diff
@@ -23,3 +27,19 @@ pub fn all_its_contracts(deps: Deps) -> Result<Binary, Error> {
...
+    deps: Deps,
+    chain: ChainNameRaw,
+    token_id: TokenId,
+) -> Result<Binary, Error> {
+    let token_instantiation =
```
fish-sammy, line 36:
```suggestion
    let token_instance =
```

contracts/interchain-token-service/src/contract.rs
```diff
@@ -36,14 +36,18 @@ pub enum Error {
...
-    QueryAllItsContracts,
     #[error("failed to disable execution")]
     DisableExecution,
     #[error("failed to enable execution")]
     EnableExecution,
+    #[error("failed to query its address")]
+    QueryItsContract,
+    #[error("failed to query all its addresses")]
+    QueryAllItsContracts,
+    #[error("failed to query a specific token instantiation")]
```
fish-sammy, line 47: Why `instantiation`?

contracts/interchain-token-service/src/state.rs
```diff
@@ -35,9 +37,67 @@ pub struct ChainConfig {
...
+            TokenDeploymentType::Trustless => TokenSupply::Tracked(Uint256::zero()),
+            _ => TokenSupply::Untracked,
+        };
+
+        Self { token_id, supply }
+    }
+}
+
+#[derive(Clone)]
+pub enum MessageDirection {
```
fish-sammy, line 75: Is this still needed?

fish-sammy (changes requested):
https://interoplabs.slack.com/archives/C03HHV83DAS/p1730520768280459?thread_ts=1730506331.710519&cid=C03HHV83DAS

contracts/interchain-token-service/src/state.rs
```diff
@@ -48,8 +48,27 @@ pub enum TokenSupply {
 
 /// Information about a token on a specific chain.
 #[cw_serde]
-pub struct TokenChainInfo {
-    supply: TokenSupply,
+pub struct TokenInstance {
+    pub token_id: TokenId,
+    pub supply: TokenSupply,
```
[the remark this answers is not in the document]
cgorenflo: any code outside of `state` can't instantiate a `TokenSupply`, so I don't think it's an issue

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -217,169 +255,158 @@ pub fn set_chain_config(
...
+            &destination_instance,
+        )
+    })
     .change_context(Error::State)
 }
 
-/// Ensures that the token is being redeployed from the same origin chain.
-/// If it's being deployed for the first time, the from chain is saved as the origin chain in the token config.
-fn token_deployment_origin_chain_check(
+fn try_load_token_instance(
```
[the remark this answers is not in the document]
cgorenflo: i think it is pretty generic. You try to load the token instance and if it's not there you get a "not deployed" error

contracts/interchain-token-service/src/contract/execute.rs
cgorenflo: because I need to change context on both statement otherwise I didn't find it more readable the other way around, but I don't feel strongly about it

contracts/interchain-token-service/src/state.rs
cgorenflo: I'll have to check if the key needs to get deserialized when you read a value. If so your suggestion wouldn't work because of lifetimes. Otherwise I'll change it

contracts/interchain-token-service/src/contract/execute.rs
cgorenflo: I'll remove the token_ids again. I'm not really happy with what these structs contain, but for now I guess it's good enough to get it across the finish line with all the necessary data available and then we can adjust later.

cgorenflo: I guess the idea is that the original chain can always mint tokens and ITS doesn't have control over the token management

contracts/interchain-token-service/src/state.rs
cgorenflo: changed it to `Instance` (which was the intended name, I got the word wrong 😅 ). I'm open to suggestions though

contracts/interchain-token-service/src/state.rs
cgorenflo: removed "global" again

contracts/interchain-token-service/src/contract/execute.rs
cgorenflo: can we not just pass in the scaling factors or something similar here?

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -74,50 +73,93 @@ pub fn execute_message(
         killswitch::is_contract_active(deps.storage),
         Error::ExecutionDisabled
     );
-    ensure_its_source_address(deps.storage, &cc_id.source_chain, &source_address)?;
+    ensure_is_its_source_address(deps.storage, &cc_id.source_chain, &source_address)?;
 
     match HubMessage::abi_decode(&payload).change_context(Error::InvalidPayload)? {
```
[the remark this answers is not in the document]
cgorenflo: `if let` looks pretty terrible here

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -74,50 +73,93 @@ pub fn execute_message(
...
-            )?
-            .add_event(
-                Event::MessageReceived {
-                    cc_id,
-                    destination_chain,
-                    message,
-                }
-                .into(),
-            ))
+fn execute_hub_message(
```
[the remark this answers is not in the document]
cgorenflo: changed it to `execute_message_on_hub`

packages/router-api/src/primitives.rs
```diff
@@ -314,6 +314,11 @@ impl ChainNameRaw {
     /// Maximum length of a chain name (in bytes).
     /// This MUST NOT be changed without a corresponding change to the ChainName validation in axelar-core.
     const MAX_LEN: usize = 20;
+
+    pub fn normalize(&self) -> ChainName {
```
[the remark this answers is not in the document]
cgorenflo: I'll add a comment. That's also why I didn't change it to a `From` implementation. The reason I moved it was that it relied on the internal details of `ChainNameRaw` and that it only differs in casing from `ChainName`

contracts/interchain-token-service/src/abi.rs
```diff
@@ -297,43 +300,47 @@ mod tests {
         let cases = vec![
             HubMessage::SendToHub {
                 destination_chain: remote_chain.clone(),
-                message: Message::InterchainTransfer {
+                message: primitives::InterchainTransfer {
```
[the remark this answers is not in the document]
cgorenflo: it's used because of naming collisions

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -217,169 +255,158 @@ pub fn set_chain_config(
     }
 }
 
-/// Applies various checks on the message.
-fn apply_checks(
+fn apply_transfer(
```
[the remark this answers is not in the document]
cgorenflo: I generally don't like comments that reiterate what the code is doing, because it's too easy to drift apart. If it's not clear what's going on the code should be improved. Happy to here other opinions

contracts/interchain-token-service/src/state.rs
cgorenflo: doesn't work

contracts/interchain-token-service/src/state.rs
cgorenflo: removed

contracts/interchain-token-service/src/contract.rs
cgorenflo: My brain just went with the wrong word. I meant instance

fish-sammy (approved):

contracts/interchain-token-service/src/abi.rs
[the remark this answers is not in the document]
milapsheth: Ah, yes, the solidity structs

cjcobb23 (approved):

PR #675, opened 2024-11-01
feat(interchain-token-service): apply scaling factor on token transfer based on token decimals

fish-sammy:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

### DIFF
--- a/contracts/interchain-token-service/tests/execute.rs
+++ b/contracts/interchain-token-service/tests/execute.rs
@@ -15,7 +15,7 @@ use interchain_token_service::{
 };
 use router_api::{Address, ChainName, ChainNameRaw, CrossChainId};
 use serde_json::json;
-use utils::{instantiate_contract, make_deps, params, TestMessage};
+use utils::{params, TestMessage};
 
 mod utils;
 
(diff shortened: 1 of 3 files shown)


contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -255,6 +263,78 @@ pub fn set_chain_config(
...
 
+/// Calculates the destination on token transfer amount.
+///
+/// The amount is calculated based on the token decimals on the source and destination chains.
+/// The calculation is done as following:
+/// 1) `destination_amount` = `source_amount` * 10 ^ (`destination_chain_decimals` - `source_chain_decimals`)
+/// 3) If new_amount is greater than the destination chain's `max_uint`, the translation
+/// fails.
+/// 4) If new_amount is zero, the translation fails.
+fn destination_amount(
```
cjcobb23, line 274: needs some unit tests. can we add unit tests for a variety of inputs to this function? want to make sure this is right.

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -255,6 +263,78 @@ pub fn set_chain_config(
...
+    let source_token = try_load_token_instance(storage, source_chain.clone(), token_id)?;
+    let destination_token = try_load_token_instance(storage, destination_chain.clone(), token_id)?;
+    let (source_decimals, destination_decimals) =
+        match (source_token.decimals, destination_token.decimals) {
+            (Some(source_decimals), Some(destination_decimals)) => {
+                (source_decimals, destination_decimals)
+            }
+            (None, None) => return Ok(source_amount),
+            _ => unreachable!(
+                "decimals are either set in both the source and destination, or set in neither"
```
cjcobb23, line 290:
```suggestion
                "decimals should be set in both the source and destination, or set in neither"
```

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -255,6 +263,78 @@ pub fn set_chain_config(
...
+    source_amount: nonempty::Uint256,
+) -> Result<nonempty::Uint256, Error> {
+    let source_token = try_load_token_instance(storage, source_chain.clone(), token_id)?;
+    let destination_token = try_load_token_instance(storage, destination_chain.clone(), token_id)?;
+    let (source_decimals, destination_decimals) =
+        match (source_token.decimals, destination_token.decimals) {
+            (Some(source_decimals), Some(destination_decimals)) => {
+                (source_decimals, destination_decimals)
+            }
+            (None, None) => return Ok(source_amount),
```
cjcobb23, line 288: this early return here is easy to miss. I think it would be easier to read if you move this return out of the match. You could return an `Option<(u8, u8)>` from the match, and return `None` in this case. Then just check if it's `None` after the match. I think you should also check if the decimals are the same after the match as well and return early. We don't need to do the computation in that case, and I think it's easer to read.

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -255,6 +263,78 @@ pub fn set_chain_config(
...
+                amount: source_amount,
+            })?
+    };
+
+    if destination_amount.gt(&destination_max_uint) {
+        bail!(Error::InvalidTransferAmount {
+            source_chain: source_chain.to_owned(),
+            destination_chain: destination_chain.to_owned(),
+            amount: source_amount,
+        })
```
cjcobb23, line 326: Feel like this should be something like `OverflowingTransfer` and also include the destination_max_uint in the error

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -291,54 +371,69 @@ fn apply_transfer(
...
+        ..transfer.clone()
+    })
 }
 
 fn subtract_amount_from_source(
     storage: &mut dyn Storage,
     source_chain: ChainNameRaw,
-    transfer: &InterchainTransfer,
+    token_id: TokenId,
+    amount: nonempty::Uint256,
```
cjcobb23, line 402: I think it was cleaner before, why do we need to make this change?

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -291,54 +371,69 @@ fn apply_transfer(
...
+    state::save_token_instance(storage, source_chain, token_id, &source_instance)
         .change_context(Error::State)
 }
 
 fn add_amount_to_destination(
     storage: &mut dyn Storage,
     destination_chain: ChainNameRaw,
-    transfer: &InterchainTransfer,
+    token_id: TokenId,
+    amount: nonempty::Uint256,
```
cjcobb23, line 422: Think it was cleaner before. Why not just transform the `InterchainTransfer` object before passing into this function?

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -291,54 +371,69 @@ fn apply_transfer(
...
+        destination_chain,
+        transfer.token_id,
+        destination_amount,
+    )?;
+
+    Ok(InterchainTransfer {
+        amount: destination_amount,
+
+        ..transfer.clone()
+    })
```
cjcobb23, line 395: I would continue passing in the `InterchainTransfer` object to `subtract_amount_from_source` and `add_amount_to_destination`, and just transform the amount in the object in between those two calls.

contracts/interchain-token-service/src/contract/execute.rs
fish-sammy: doctest would be the best for this kind of functions, sadly it doesn't work with private fn.

contracts/interchain-token-service/src/contract/execute.rs
fish-sammy: I don't necessarilly think the returning `Option<(u8, u8)>` makes it more readable but I do like the check to see if they are the same.

contracts/interchain-token-service/src/contract/execute.rs
fish-sammy: I'd prefer keeping this simple and generic since all the other cases also suggest some very specific number issue.

contracts/interchain-token-service/src/contract/execute.rs
fish-sammy: Don't like having function taking unnecessary info.

cjcobb23 (approved):

contracts/interchain-token-service/src/contract/execute.rs
```diff
@@ -257,6 +265,83 @@ pub fn set_chain_config(
...
+        })
+    }
+
+    nonempty::Uint256::try_from(destination_amount).change_context_lazy(|| {
+        Error::InvalidTransferAmount {
+            source_chain: source_chain.to_owned(),
+            destination_chain: destination_chain.to_owned(),
+            amount: source_amount,
+        }
+    })
```
cgorenflo, line 342: this calculation could get extracted into it's own function

cgorenflo (approved):

PR #676, opened 2024-11-04
feat(ampd): add Stacks handlers with custom ITS logic & verify logic for contract deployments

raress96:
## Description

Besides the normal `stacks_verify_msg` and `stacks_verify_verifier_set` handlers, we needed to add custom logic for Stacks in order to support integration with the ITS Hub:

1. On Stacks, a smart contract can NOT deploy another smart contract
    - because the ITS contract needs to deploy NativeInterchainToken & Token Manager contracts, we need an untrusted 3rd party (eg: the Relayer or any other user account) that deploys the contracts first
    - then we need a way to verify that the deployed contracts are actually correct
    - we use the Axelar Network in order to verify that the deployed contracts have the correct code (on Stacks the code is interpreted and stored on chain), by first deploying reference contract implementations that Axelar Validators support Stacks need to add in their config
    - this flow is also triggered only in the case of a cross chain call from Stacks -> Stacks and from the Stacks ITS contract to itself
    - Axelar Network validators ONLY verify that the deployed contract code is correct and hasn't been modified at all. The Stacks ITS contract will further call the contract(s) and check if they have the required data stored (correct name/symbol/decimals, correct operator/minter etc)

2. The Stacks ITS contract can NOT support ABI encoding
    - due to the fact that the Clarity smart contract language used for Stacks is non-turing complete, there was no way to implement ABI encoding/decoding on Stacks
    - because of that, for outbound contract calls from Stacks -> other blockchains, we will use the Axelar Validator code to translate the Clarity payload into ABI encoded payload
    - if there is a cross chain call from the Stacks ITS contract known to Axelar Validators supporting Stacks, then the `payloadHash` in the Message should be the hash of the ABI encoded payload, and NOT of the actual Clarity payload
    - the whole flow works like this:
        - the Relayer will convert the Clarity payload into ABI encoded payload and send it & the new payloadHash to the Axelar GMP API
        - when receiving this message, the Axelar Validators for Stacks will also convert the Clarity payload into ABI and check that the payloadHash is correct
        - this way the flow is supported out of the box, without any modifications needed on GMP API or other CosmWasm contracts


ampd/src/handlers/stacks_verify_msg.rs
```diff
@@ -0,0 +1,409 @@
...
+use crate::stacks::http_client::{Client, Transaction};
+use crate::stacks::verifier::verify_message;
+use crate::types::{Hash, TMAddress};
+
+type Result<T> = error_stack::Result<T, Error>;
+
+#[derive(Deserialize, Debug)]
+pub struct Message {
+    pub tx_id: Hash,
+    pub event_index: u32,
```
milapsheth, line 31: We updated voting verifier recently to emit the message_id string in the event instead and deprecated the old fields. Now you can directly decode into the typed MsgIdFormat, see example [here](https://github.com/axelarnetwork/axelar-amplifier/blob/main/ampd/src/handlers/stellar_verify_msg.rs#L33). We'll be announcing this in the upcoming release's changelog. But you can update multiversX/stacks handlers to use this already. The new voting verifier contract will need to be deployed first so the new field is emitted for ampd to parse

raress96: Done

PR #678, opened 2024-11-05
doc: review comments

milapsheth:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


cgorenflo (changes requested):
as is, the changed/added comments boil down to mostly change requests. This is not how documentation should be used

PR #682, opened 2024-11-06
refactor(interchain-token-service): create message execution interceptors

fish-sammy:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


contracts/interchain-token-service/src/state.rs
cgorenflo, line 1: this change doesn't make sense to me

contracts/interchain-token-service/src/contract/execute/mod.rs
```diff
@@ -150,17 +150,61 @@ fn apply_to_hub(
...
+        .and_then(|_| {
+            interceptors::apply_scaling_factor_to_amount(
+                storage,
+                &source_chain,
+                &destination_chain,
+                &mut transfer,
+            )
+        })
+        .and_then(|_| interceptors::add_supply_amount(storage, &destination_chain, &transfer))
+        .map(|_| transfer)
```
cgorenflo, line 180: not that I don't like more functional patterns, but is this necessary here?
Wouldn't it just read
```
interceptors::subtract_supply_amount(...)?;
interceptors::apply_scaling_factor_to_amount(...)?;
interceptors::add_supply_amount(...)?;

Ok(transfer);
```

contracts/interchain-token-service/src/contract/execute/mod.rs
```diff
@@ -150,17 +150,61 @@ fn apply_to_hub(
...
+            )
+        })
+        .and_then(|_| {
+            interceptors::save_token_instance_for_destination_chain(
+                storage,
+                destination_chain,
+                &deploy_token,
+            )
+        })
+        .map(|_| deploy_token)
```
cgorenflo, line 205: same question

contracts/interchain-token-service/src/state.rs
fish-sammy: TokenId implements Copy so it feels extra to me to pass it by reference, but not a very strong opinion.

contracts/interchain-token-service/src/contract/execute/mod.rs
fish-sammy: IMO, leaving it functional would encourage people to do one thing at a step in the future when they need to touch this function again. What I want to avoid for this function to happen is people start adding procedural things to it again.

cgorenflo: would it be possible for these functions to return the appropriate message, i.e. take in `transfer` and pass it back out? That way, the `and_then` wouldn't be wasted and you wouldn't need to map in the end

fish-sammy: Thought of doing it, but decided to land like this so that it's clearer which interceptors would mutate the messages and which ones would not when someone looks at the apply function.

cgorenflo: It's a good argument, but is that a concern?

contracts/interchain-token-service/src/state.rs
cgorenflo: ah gotcha

contracts/interchain-token-service/src/contract/execute/mod.rs
fish-sammy: No, it's just better readability.

cjcobb23: I actually disagree, I think the mutability here is a bit hard to follow. I think it would be better if `apply_scaling_factor_to_amount` actually returned the new transfer, instead of mutating. Same with the other functions. We can then save the returned values in this function (if necessary).

contracts/interchain-token-service/src/contract/execute/interceptors.rs
```diff
@@ -0,0 +1,648 @@
...
+
+    state::save_token_instance(storage, chain.clone(), transfer.token_id, &token)
+        .change_context(Error::State)
+}
+
+pub fn apply_scaling_factor_to_amount(
+    storage: &dyn Storage,
+    source_chain: &ChainNameRaw,
+    destination_chain: &ChainNameRaw,
+    transfer: &mut InterchainTransfer,
```
cjcobb23, line 52: just return this, instead of mutating. The function doesn't return anything as is

contracts/interchain-token-service/src/contract/execute/interceptors.rs
```diff
@@ -0,0 +1,648 @@
...
+
+use super::Error;
+use crate::state::{self, TokenDeploymentType};
+use crate::{DeployInterchainToken, InterchainTransfer, TokenConfig, TokenId, TokenInstance};
+
+pub fn subtract_supply_amount(
+    storage: &mut dyn Storage,
+    chain: &ChainNameRaw,
+    transfer: &InterchainTransfer,
+) -> Result<(), Error> {
```
cjcobb23, line 14: I think this would read better if you passed in the `TokenInstance`, subtracted the amount, and returned the new instance. And then the caller can save. I don't think the name and signature of this function gives a good idea of what it's doing

contracts/interchain-token-service/src/contract/execute/interceptors.rs
```diff
@@ -0,0 +1,648 @@
...
+
+    state::save_token_instance(storage, chain.clone(), transfer.token_id, &token)
+        .change_context(Error::State)
+}
+
+pub fn add_supply_amount(
+    storage: &mut dyn Storage,
+    chain: &ChainNameRaw,
+    transfer: &InterchainTransfer,
+) -> Result<(), Error> {
```
cjcobb23, line 33: Same comment as above

contracts/interchain-token-service/src/contract/execute/interceptors.rs
```diff
@@ -0,0 +1,648 @@
...
+        source_chain,
+        destination_chain,
+        transfer.token_id,
+        transfer.amount,
+    )?;
+
+    Ok(())
+}
+
+pub fn save_token_instance_for_source_chain(
```
cjcobb23, line 65: I don't think this is a good name for this function, I generally think functions that are called `save_something`, just save something. But actually this is doing quite a bit of initialization and checks.

contracts/interchain-token-service/src/contract/execute/interceptors.rs
```diff
@@ -0,0 +1,648 @@
...
+    )
+    .change_context(Error::State)
+    .map(|_| ())
+}
+
+pub fn calculate_scaling_factor(
+    storage: &dyn Storage,
+    source_chain: &ChainNameRaw,
+    destination_chain: &ChainNameRaw,
+    deploy_token: &mut DeployInterchainToken,
```
cjcobb23, line 132: why not just return this? This mutability is hard to read.

cjcobb23 (changes requested):
I dislike the mutation of function arguments that's in this change, it feels like a bit of an anti pattern. I'm happy to be wrong, but would like to discuss with the team before merging. I don't think we do this anywhere else in the code.

contracts/interchain-token-service/src/contract/execute/mod.rs
fish-sammy: Returning it hides the fact that it may have changed during the process. Someone looking at the caller's flow can not tell which interceptors transform the message and which do not. It's not an anti-pattern. Rust forces people to use the `mut` keyword for a reason.

contracts/interchain-token-service/src/contract/execute/interceptors.rs
fish-sammy: What do u suggest?

contracts/interchain-token-service/src/contract/execute/mod.rs
```diff
@@ -151,20 +150,56 @@ fn apply_to_hub(
...
+    interceptors::subtract_supply_amount(storage, &source_chain, &transfer)?;
+    let transfer = interceptors::apply_scaling_factor_to_amount(
+        storage,
+        &source_chain,
+        &destination_chain,
+        transfer,
+    )?;
+    interceptors::add_supply_amount(storage, &destination_chain, &transfer)?;
+
+    Ok(transfer)
```
cgorenflo, line 179: here `and_then` might look better now 😆

contracts/interchain-token-service/src/contract/execute/mod.rs
```diff
@@ -151,20 +150,56 @@ fn apply_to_hub(
...
+    interceptors::deploy_token_to_source_chain(storage, source_chain, &deploy_token)?;
+    let deploy_token = interceptors::calculate_scaling_factor(
+        storage,
+        source_chain,
+        destination_chain,
+        deploy_token,
+    )?;
+    interceptors::deploy_token_to_destination_chain(storage, destination_chain, &deploy_token)?;
+
+    Ok(deploy_token)
```
cgorenflo, line 197: same

contracts/interchain-token-service/src/contract/execute/mod.rs
cgorenflo: my biggest issue before was that it was just used to unpack the result, but the passed value was ignored. With the new fluid interface that is no longer the case

cgorenflo:
```suggestion
    interceptors::apply_scaling_factor_to_amount(
        storage,
        &source_chain,
        &destination_chain,
        transfer,
    )
    .inspect(interceptors::add_supply_amount(storage, &destination_chain, transfer))
```

cjcobb23 (approved):

PR #683, opened 2024-11-06
fix(minor-interchain-token-service): fix message type encoding

cjcobb23:


contracts/interchain-token-service/src/abi.rs
```diff
@@ -16,6 +16,7 @@ sol! {
     enum MessageType {
         InterchainTransfer,
         DeployInterchainToken,
+        Unused,
```
fish-sammy, line 19: Can't we just keep `DeployTokenManager` here? It's not used but this file can still reflect what's in the solidity code.

cjcobb23: It can, but I feel like a reader might then wonder why it's not handled by the rest of the code. Maybe a comment will suffice then

cjcobb23: i changed it

fish-sammy: a comment is good, but the reader should also understand that this file is a reflection of something external, not internal.

fish-sammy (approved):

contracts/interchain-token-service/src/abi.rs
fish-sammy: For example, if I compiles some Rust mod from protobuf cuz I need to use some of those types in there, the rest is gonna be unused but that's totally fine.

PR #689, opened 2024-11-12
feat(router-minor): support Message Id format for Aleo

anstylian:
## Description
Support `MessageIdFormat` for Aleo


cjcobb23 (changes requested):
this should be implemented like the other message ids, particularly with a regex. This is not very safe to use the aleo_types crate to do the parsing. For instance, what if that crate allows different representations of the same transaction id? like upper vs lowercase, 0x vs not, etc. What if it does not allow that today, but does in the future? This part of the code is very sensitive and needs to be extremely secure.

anstylian:
@cjcobb23 thanks for the review. 
A regex to verify message ids has been added.

Please let me know if further changes are needed.

packages/axelar-wasm-std/src/msg_id/aleo_message_id.rs
```diff
@@ -0,0 +1,64 @@
...
+    pub fn new(transaction_id: Transaction, transition_id: Transition) -> Self {
+        Self {
+            transaction_id,
+            transition_id,
+        }
+    }
+}
+
+const PATTERN: &str =
+    "^(at1[qpzry9x8gf2tvdw0s3jn54khce6mua7l]{58})-(au1[qpzry9x8gf2tvdw0s3jn54khce6mua7l]{58})$";
```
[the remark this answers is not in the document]
anstylian: We can put the in alphanumeric order if you prefer it.
The order follows the order of "Bech32"
https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#bech32

I have copied the table here for reference:

|     | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 |
|-----|---|---|---|---|---|---|---|---|
| +0  | p | q | z | r | y | 9 | x | 8 |
| +8  | g | f | 2 | t | v | d | w | 0 |
| +16 | s | 3 | j | n | 5 | 4 | k | h |
| +24 | c | e | 6 | m | u | a | 7 | l |

Just let me know how you prefer it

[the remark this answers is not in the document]
cgorenflo: I would prefer to make the pattern as readable as possible, otherwise regex is easy to get wrong. You could also consider using ranges to make it more obvious which letters are skipped. I think `[02-9a-hj-np-z]` works, but please double-check

packages/aleo-types/src/address.rs
```diff
@@ -0,0 +1,70 @@
...
+
+const ADDRESS_PREFIX: &str = "aleo";
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Address {
+    address: String,
+}
+
+impl Address {
+    pub fn address(&self) -> &str {
```
cgorenflo, line 18: I would call this `as_str()` as well if you need it. Otherwise, you'll have weird calls like 'address.address()'

packages/aleo-types/src/lib.rs
cgorenflo, line 1: is it necessary to export all modules as they are _and_ export everything inside of those modules in a flattened form with the wildcard?

packages/aleo-types/src/transaction.rs
```diff
@@ -0,0 +1,77 @@
...
+
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Hash, Clone)]
+pub struct Transaction {
+    transition_id: String,
+}
+
+impl FromStr for Transaction {
+    type Err = Report<Error>;
+
+    fn from_str(message_id: &str) -> Result<Self, Self::Err>
```
cjcobb23, line 19: This is pretty much the same implementation as Transaction::from_str. Can we unify the logic and make something generic? And also call it something not specific to Aleo? This is just bech32 with a prefix and checksum right?

packages/aleo-types/src/program.rs
```diff
@@ -0,0 +1,117 @@
...
+
+use error_stack::{ensure, Report, Result};
+use serde::{Deserialize, Serialize};
+
+use crate::Error;
+
+const PROGRAM_SUFFIX: &str = ".aleo";
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Program {
```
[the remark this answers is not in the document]
anstylian: removed

packages/aleo-types/src/address.rs
```diff
@@ -0,0 +1,70 @@
...
+use serde::{Deserialize, Serialize};
+
+use crate::Error;
+
+const ADDRESS_PREFIX: &str = "aleo";
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Address {
+    address: String,
+}
```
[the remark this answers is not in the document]
anstylian: is used at packages/axelar-wasm-std/src/address.rs
```
use aleo_types::address::Address as AleoAddress;
```

packages/axelar-wasm-std/src/msg_id/mod.rs
```diff
@@ -25,6 +27,12 @@ pub enum Error {
     InvalidTxHash(String),
     #[error("invalid tx digest in message id '{0}'")]
     InvalidTxDigest(String),
+    #[error("Invalid Aleo message Id: '{0}'")]
+    InvalidAleoMessageId(String),
+    #[error(
+        "Expected format: 'AleoTransitionId-AleoTransactionId-Index', index is a u32 number, '{0}'"
```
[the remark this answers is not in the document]
anstylian: Thanks for noticing, fixed

packages/aleo-types/src/transaction.rs
anstylian: done

packages/axelar-wasm-std/src/address.rs
```diff
@@ -19,6 +20,7 @@ pub enum AddressFormat {
     Eip55,
     Sui,
     Stellar,
+    Aleo,
```
cjcobb23, line 23: I don't think we need this change in the main repo. The address format is only used by the voting verifier. The aleo voting verifier should be maintained on your own fork. The message id is the only change we need in the universal repo.

packages/axelar-wasm-std/src/msg_id/mod.rs
```diff
@@ -45,6 +51,7 @@ pub enum MessageIdFormat {
     Base58TxDigestAndEventIndex,
     Base58SolanaTxSignatureAndEventIndex,
     HexTxHash,
+    AleoTransaction,
```
cjcobb23, line 54: I would prefer a more generic name that doesn't include Aleo if possible. Something like `Bech32TransactionIdAndTransitionId`

packages/aleo-types/src/transition.rs
```diff
@@ -0,0 +1,67 @@
...
+
+impl FromStr for Transition {
+    type Err = Report<Error>;
+
+    fn from_str(transition_id: &str) -> Result<Self, Self::Err>
+    where
+        Self: Sized,
+    {
+        crate::verify_bech32m(transition_id, TRANSITION_PREFIX)
+            .change_context(Error::InvalidAleoTransition(transition_id.to_string()))?;
```
cjcobb23, line 24: Now that you've simplified this function, I don't know that you even need the `Transition` and `Transaction` type. Why not just call this `verify_bech32m` function directly when parsing the message id? I'm sorry to want to minimize so much, it's just I want to avoid adding code for every new chain. I would like to keep chain specific code very minimal, else this repo will explode.

packages/axelar-wasm-std/src/msg_id/mod.rs
```diff
@@ -45,6 +49,7 @@ pub enum MessageIdFormat {
     Base58TxDigestAndEventIndex,
     Base58SolanaTxSignatureAndEventIndex,
     HexTxHash,
+    Bech32m(String, u8), // (prefix, length)
```
cjcobb23, line 52: probably cleaner to use named fields here

packages/axelar-wasm-std/src/msg_id/bech32m.rs
```diff
@@ -0,0 +1,70 @@
...
+    pub fn new(encoded: String) -> Self {
+        Self { encoded }
+    }
+
+    pub fn from_str(prefix: &str, length: u8, message_id: &str) -> Result<Self, Report<Error>> {
+        let pattern = format!(
+            "^({}1[02-9ac-hj-np-z]{{{}}})$",
+            prefix,
+            length
+                .saturating_sub(prefix.len().try_into().unwrap())
```
cjcobb23, line 24: let's not unwrap in production code, just return an error if this errors

packages/axelar-wasm-std/src/msg_id/bech32m.rs
```diff
@@ -0,0 +1,70 @@
...
+                .saturating_sub(1)
+        );
+
+        let regex = Regex::new(pattern.as_str()).expect("invalid regex");
+
+        let (_, [string]) = regex
+            .captures(message_id)
+            .ok_or(Error::InvalidMessageID {
+                id: message_id.to_string(),
+                expected_format: "Foo Bech32m".to_string(),
```
cjcobb23, line 34: this expected format does not seem correct

packages/axelar-wasm-std/src/msg_id/bech32m.rs
```diff
@@ -0,0 +1,70 @@
...
+}
+
+impl Bech32mFormat {
+    pub fn new(encoded: String) -> Self {
+        Self { encoded }
+    }
+
+    pub fn from_str(prefix: &str, length: u8, message_id: &str) -> Result<Self, Report<Error>> {
+        let pattern = format!(
+            "^({}1[02-9ac-hj-np-z]{{{}}})$",
```
cjcobb23, line 21: Why do we need `{{{}}}`? Can't it just be `{{}}`?

packages/axelar-wasm-std/src/msg_id/mod.rs
```diff
@@ -25,6 +27,10 @@ pub enum Error {
     InvalidTxHash(String),
     #[error("invalid tx digest in message id '{0}'")]
     InvalidTxDigest(String),
+    #[error("Invalid bech32m: '{0}'")]
+    InvalidBech32mLocalChecks(String),
```
cjcobb23, line 31:
```suggestion
    InvalidBech32mFormat(String),
```

packages/axelar-wasm-std/src/msg_id/mod.rs
```diff
@@ -25,6 +27,10 @@ pub enum Error {
...
+    #[error("Invalid bech32m: '{0}'")]
+    InvalidBech32mExternalChecks(String),
```
cjcobb23, line 33:
```suggestion
    InvalidBech32m(String),
```

cjcobb23 (approved):

PR #690, opened 2024-11-13
docs: add ARC-1 proposal

milapsheth:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

### DIFF
--- a/ARCs/ARC-1.md
+++ b/ARCs/ARC-1.md
@@ -1,0 +1,116 @@
+# ARC-1: ITS Hub Multi-Chain Token Linking
+
+## Metadata
+
+- **ARC ID**: 1
+- **Author(s)**: Milap Sheth
+- **Status**: Draft
+- **Created**: 2024-11-14
+- **Last Updated**: 2024-11-14
+- **Target Implementation**: Q4 2024
+
+## Summary
+
+Extend the Interchain Token Service (ITS) Hub to support linking existing tokens across different chains, potentially using different decimals. This improvement enables token transfers for custom token links across both EVM and non-EVM chains by supporting appropriate decimal scaling.
+
+## Motivation
+
+### Background
+
+The current ITS `DeployTokenManager` message type only supports linking existing tokens with identical decimals. This limitation is particularly problematic when dealing with non-EVM chains, which often cannot support the high decimal values (e.g., 18) common in EVM chains due to token amount type constraints.
+
+### Goals
+
+- Enable linking of existing tokens across connected chains
+- Support linking tokens with different decimal precisions
+
+## Detailed Design
+
+### Overview
+
+Two potential solutions are proposed to implement this enhancement:
+
+1. Modify the existing `DeployTokenManager` message type
+2. Introduce new message types for token registration and linking
+
+### Technical Specification
+
+#### Solution 1: Enhanced DeployTokenManager
+
+```solidity
+struct DeployTokenManager {
+    bytes32 tokenId;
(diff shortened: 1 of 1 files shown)


ARCs/ARC-1.md
```diff
@@ -0,0 +1,116 @@
...
+    uint256 tokenManagerType;
+    bytes params;
+    uint8 sourceDecimals;
+    uint8 destinationDecimals;
+}
+```
+
+ITS Hub can use the source and destination token decimals to determine the appropriate scaling factor to apply when intercepting this message. ITS edge contract can populate the source token decimals by reading the token metadata, however the destination token decimals will need to be provided by the deployer.
+
+Setting this incorrectly however can be lead to the ITS hub recording the deployment incorrectly, while the deployment tx on the destination ITS will fail due to mismatching decimals, which isn’t recoverable.
```
cjcobb23, line 52: why will it fail on the destination? Because the decimals would be greater than the chain supports? Can't we check this in the ITS hub?

ARCs/ARC-1.md
```diff
@@ -0,0 +1,116 @@
...
+
+One potential mitigation is to allow the deployer to resubmit the msg but with a different destination decimals, and have ITS hub overwrite the recorded deployment. ITS Hub can allow this overwrite until the first Transfer msg is encountered to allow the user to fix their deployment. This adds some more complexity in ITS hub setup.
```
cjcobb23, line 54: this overwrite until transfer logic sounds a bit complicated, would prefer to avoid. Can the successful deployment send back some sort of confirmation to the hub?

ARCs/ARC-1.md
```diff
@@ -0,0 +1,116 @@
...
+
+Flow:
+
+1. Deployer submits `DeployTokenManager`
+2. ITS Hub calculates scaling factor and records deployment
+3. Destination chain deploys token manager with the corresponding token address (extracted from the params)
+
+#### Solution 2: Token Registration and Linking
```
cjcobb23, line 62: this solution seems much better. Is there a downside to this approach?

ARCs/ARC-1.md
milapsheth: ITS hub doesn't know about the destination token's decimals. Solution 2 addresses this

ARCs/ARC-1.md
milapsheth: Extra complexity and worse UX for a developer setting up their token link. Although, I think it's acceptable for supporting custom tokens. The common flow of deploying a token is still very simple

ARCs/ARC-1.md
milapsheth: Yeah, a modified version of Solution 2 is to send a GMP call back on deploy token manager going through with the decimals of the destination token but I find it more complex to have a 2-stage setup in ITS Hub

ARCs/ARC-1.md
```diff
@@ -0,0 +1,116 @@
...
+Two new ITS message types are introduced: `RegisterToken` and `LinkToken`.
+
+```solidity
+struct RegisterToken {
+    bytes tokenAddress;
+    uint8 decimals;
+}
+
+struct LinkToken {
+    bytes32 tokenId;
```
cjcobb23, line 73: how/when is this tokenId generated?

haiyizxx:
moved to [ARCs repo](https://github.com/axelarnetwork/ARCs/pull/1)

PR #691, opened 2024-11-13
chore: update changelog with migration notes

cjcobb23:


cgorenflo (approved):
do we want to include specific version in this? Are these general notes or something we update with every release?

cjcobb23:
> do we want to include specific version in this? Are these general notes or something we update with every release?

good question. I think it would be useful to keep a running log of migration notes for each version, so I will add the version number.

PR #692, opened 2024-11-13
fix(minor-rewards): remove old migration handler

cjcobb23:


contracts/rewards/src/contract.rs
```diff
@@ -25,10 +25,7 @@ pub fn migrate(
     _env: Env,
     _msg: Empty,
 ) -> Result<Response, axelar_wasm_std::error::ContractError> {
-    migrations::v1_0_0::migrate(deps.storage)?;
-
     // any version checks should be done before here
-
     cw2::set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
```
cgorenflo, line 29: I think the current version should still be checked, so it's not possible to accidentally (or maliciously) roll back to this from whichever version we have

haiyizxx (approved):

PR #693, opened 2024-11-13
fix(minor-voting-verifier): remove old migration handler

cjcobb23:


cgorenflo (dismissed):

cgorenflo:
I think the current version should still be checked, so it's not possible to accidentally (or maliciously) roll back to this from whichever version we have

haiyizxx (approved):

PR #694, opened 2024-11-14
refactor(ampd): switch from horizon to rpc client for Stellar

haiyizxx:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


ampd/src/stellar/rpc_client.rs
```diff
@@ -0,0 +1,120 @@
...
+    ) -> error_stack::Result<HashMap<String, TxResponse>, Error> {
+        let tx_hashes: Vec<_> = tx_hashes
+            .into_iter()
+            .map(|tx_hash| Hash::from_str(tx_hash.as_str()).change_context(Error::TxHash))
+            .collect::<std::result::Result<Vec<_>, _>>()?;
+
+        let responses = join_all(
+            tx_hashes
+                .iter()
+                .map(|tx_hash| self.0.get_transaction(tx_hash)),
```
milapsheth, line 86: Looks like we don't support doing a batched JSON RPC query in ampd? Seems to be the same as EVM. Is there a task to add this later?

milapsheth, line 86: I imagine that `0x` prefix shouldn't be included for this rpc query as well?

milapsheth (approved):

ampd/src/stellar/rpc_client.rs
haiyizxx: The crate doesn’t support batch yet; we can update once they add that support.

haiyizxx: yes, same as before. the handler trim the prefix

PR #697, opened 2024-11-15
test(interchain-token-service): add test for 3 chain cycle

cjcobb23:


cgorenflo (approved):

contracts/interchain-token-service/tests/execute.rs
```diff
@@ -314,6 +314,159 @@ fn execute_message_interchain_transfer_should_scale_the_amount_when_source_decim
...
+        deps.as_mut(),
+        CrossChainId {
+            source_chain: configs[2].0.clone(),
+            message_id: router_message.cc_id.message_id.clone()
+        },
+        source_its_contract,
+        hub_message,
+    ));
+
+    goldie::assert_json!(json!({"response_to_destination": response_to_destination}));
```
fish-sammy, line 467:
```suggestion
    goldie::assert_json!(response_to_destination);
```
Don't think you need `json!` when you have only one object to assert.

contracts/interchain-token-service/tests/execute.rs
```diff
@@ -314,6 +314,312 @@ fn execute_message_interchain_transfer_should_scale_the_amount_when_source_decim
...
+            "stellar".parse().unwrap(),
+            source_its_contract.clone(),
+            Uint256::from(u128::MAX).try_into().unwrap(),
+            12,
+        ),
+        (
+            "sui".parse().unwrap(),
+            source_its_contract.clone(),
+            Uint256::from(u64::MAX).try_into().unwrap(),
+            6,
```
milapsheth, line 340: I think we can increase the coverage for this like below:

For each chain in the config, start a deployment of the token from it to other chains (token decimals can be set to chain's decimals maybe), and then do a transfer cycle that ends on the same chain.

This will test various scaling combinations:
18 -> 12 -> 6 -> 18
12 -> 6 -> 18 -> 12
6 -> 18 -> 12 -> 6

Generalizing this also makes it easier to add any chain config in this list in the future

contracts/interchain-token-service/tests/execute.rs
```diff
@@ -314,6 +314,312 @@ fn execute_message_interchain_transfer_should_scale_the_amount_when_source_decim
...
+    let TestMessage {
+        source_its_contract,
+        ..
+    } = TestMessage::dummy();
+    let configs = vec![
+        (
+            "ethereum".parse().unwrap(),
+            source_its_contract.clone(),
+            Uint256::MAX.try_into().unwrap(),
+            18,
```
milapsheth, line 328: use `u8::MAX` to mimic actual config

contracts/interchain-token-service/tests/execute.rs
```diff
@@ -314,6 +314,312 @@ fn execute_message_interchain_transfer_should_scale_the_amount_when_source_decim
...
+        },
+        source_its_contract,
+        hub_message,
+    ));
+
+    goldie::assert_json!(response_to_destination);
+}
+
+#[test]
+fn execute_message_interchain_transfer_should_scale_correctly_in_3_chain_cycle_with_dust() {
```
milapsheth, line 471: add multiple transfer amount cases in the previous test perhaps to reduce duplication. Or move shared logic into helpers, such as deploying tokens from source to a list of other chains

contracts/interchain-token-service/tests/execute.rs
cjcobb23: I will implement this in a follow up, want to get this test in for the release

contracts/interchain-token-service/tests/execute.rs
cjcobb23: will implement in a follow up PR

PR #705, opened 2024-11-19
fix(multisig): remove old migration

cjcobb23:


fish-sammy (approved):

contracts/multisig/src/contract.rs
```diff
@@ -25,28 +25,15 @@ mod query;
 
 const CONTRACT_NAME: &str = env!("CARGO_PKG_NAME");
 const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
+const BASE_VERSION: &str = "1.0.0";
```
milapsheth, line 28: this test will break post release, so needs to be fixed again

cgorenflo: that is kind of the point of these tests, we should make sure we have them everywhere. That way we don't forget to remove the migration

cgorenflo (approved):

PR #707, opened 2024-11-20
chore(ampd-minor): update signers hash for multiversx after contracts audit

raress96:
## Description

As per the Gateway contract audit, the signers hash computation has changed to use MultiversX nested encoding. Since I didn't want to add a new Rust dependency just for this and further complicate the code, a simple manual solution was used instead which yields the same result.


ampd/src/mvx/mod.rs
```diff
@@ -76,6 +82,10 @@ fn uint256_to_compact_vec(value: Uint256) -> Vec<u8> {
     bytes[slice_from..].to_vec()
 }
 
+fn usize_to_u32(value: usize) -> Vec<u8> {
+    (value as u32).to_be_bytes().to_vec()
+}
```
cgorenflo, line 87: I'd prefer to move the `to_vec()` call out of this function and return a fixed length byte array. This being fixed length is security relevant, so I wouldn't hide it behind a vec conversion

ampd/src/mvx/mod.rs
```diff
@@ -24,15 +25,20 @@ pub struct WeightedSigners {
 
 impl WeightedSigners {
     pub fn hash(&self) -> Hash {
-        let mut encoded = Vec::new();
+        let mut encoded: Vec<Vec<u8>> = Vec::new();
```
cgorenflo, line 28: Do you need `Vec<Vec<u8>>`? `digest` takes a `Ref<[u8]>`. If possible I'd prefer to use `Vec<&[u8]>` as before. Defining encoded as mutable with nested vectors allows you to modify the vector elements as well I believe, which seems dangerous

raress96: The reason I didn't use `Vec<&[u8]>` is because of the needed call to `usize_to_u32` which is done in the `for` loop. That creates a new owned variable that doesn't live long enough, hence I needed to change this to to a `Vec<Vec<u8>>`.

The solution would have been to save the length in the `WeightedSigner` & `WeightedSigners` structs so I can reference it, but I don't think that is a good idea.

ampd/src/mvx/mod.rs
raress96: Fixed.

ampd/src/mvx/mod.rs
cgorenflo: but now that issue doesn't exist anymore, does it? You can just call `encoded.push(&usize_to_u32(self.signers.len()));

raress96: Still not possible since the value will live only long enough for the `for` loop. This code does not work because the `len` from the for doesn't live long enough. I tried multiple ways of refactoring to storing refs, but I didn't find any solution:
```rust
pub fn hash(&self) -> Hash {
        let mut encoded: Vec<&[u8]> = Vec::new();

        let len = usize_to_u32(self.signers.len());
        encoded.push(&len);

        for signer in self.signers.iter() {
            encoded.push(&signer.signer);

            let len = usize_to_u32(signer.weight.len());
            encoded.push(&len);
            encoded.push(&signer.weight);
        }

        let len = usize_to_u32(self.threshold.len());
        encoded.push(&len);
        encoded.push(&self.threshold);
        encoded.push(&self.nonce);

        Keccak256::digest(encoded.concat()).into()
    }
```

```
let len = usize_to_u32(signer.weight.len());
   |                 --- binding `len` declared here
37 |             encoded.push(&len);
   |                          ^^^^ borrowed value does not live long enough
```

PR #708, opened 2024-11-21
Update vote event

Jeongseup:
## Description

I thought verifier's vote should fire vote status about voting for poll_id. I'm not sure this implement is good or not. but I want to suggest add a event vote status.

### DIFF
--- a/contracts/voting-verifier/src/events.rs
+++ b/contracts/voting-verifier/src/events.rs
@@ -249,6 +249,15 @@ impl TryFrom<(Message, &MessageIdFormat)> for TxEventConfirmation {
 pub struct Voted {
     pub poll_id: PollId,
     pub voter: Addr,
+    pub votes: Vec<Vote>,
+}
+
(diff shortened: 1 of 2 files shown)


contracts/voting-verifier/src/events.rs
```diff
@@ -259,6 +268,7 @@ impl From<Voted> for Event {
                 serde_json::to_string(&other.poll_id).expect("failed to serialize poll_id"),
             )
             .add_attribute("voter", other.voter)
+            .add_attribute("votes", votes_to_string(other.votes))
```
fish-sammy, line 271:
```suggestion
            .add_attribute("votes", serde_json::to_string(&votes).expect("failed to serialize votes"))
```

fish-sammy (changes requested):

fish-sammy:
https://github.com/axelarnetwork/axelar-amplifier/pull/732

PR #722, opened 2024-12-12
Router contract msg_id_format upgrade

diogofriggo:
The router contract allows for the update of the gateway address associated with a chain, this PR adds that same functionality also for the MessageIdFormat. If the incorrect MessageIdFormat gets specified during contract instantiation then it can never be corrected.

### DIFF
--- a/contracts/router/src/contract.rs
+++ b/contracts/router/src/contract.rs
@@ -98,6 +98,10 @@ pub fn execute(
                 msg_id_format,
             )?)
         }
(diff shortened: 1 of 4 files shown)


cjcobb23:
This is not a safe feature to allow, since this could theoretically open the door to double spends (a message routed with the old format and then routed again under the new format). There's no need for this except if a mistake was made operationally. If so, just register a new chain.

PR #723, opened 2024-12-12
feat: implementation of Starknet GMP

ctoyan:
## Description
This PR consists of:
- small `voting-verifier` CosmWASM contract changes, mostly related to `MessageIdFormat` and `AddressFormat`
- small `multisig-prover` CosmWASM contract changes, mostly related to and ABI encoding
- starknet specific types and events de/serialisation implementations
- verifier set verification
- call contract verification

## Todos
- [x] Unit tests
- [x] Manual tests
- [x] Documentation
- [ ] Connect epics/issues

## Steps to Test
You would need the Starknet relayer, in order to manually test, which is not open sourced yet.

## Notes
- We haven't tested with the custom MessageIdFormat yet. This will happen once this PR is merged and the changes related to the format are deployed.


packages/axelar-wasm-std/src/msg_id/starknet_field_element_event_index.rs
ctoyan, line 1: The main idea of this MessageIdFormat is that it's checking if the FieldElement (Felt) overflows.
`Felt::MAX` is `2^251+17*2^192+1`, which fits in 252 bits. This means that a 64 hex value might overflow it. We enforce 64 hex string everywhere with padded zeros.

ctoyan:
Closing this issue, in order to split the PR into two smaller chunks - one for ampd and one for the cosm wasm contracts.

PR #725, opened 2024-12-13
feat(minor-interchain-token-service): support linking custom tokens

cjcobb23:
* Adds RegisterToken and LinkToken message types

https://axelarnetwork.atlassian.net/browse/AXE-6826


contracts/interchain-token-service/src/primitives.rs
```diff
@@ -110,7 +145,9 @@ impl Message {
     pub fn token_id(&self) -> TokenId {
         match self {
             Message::InterchainTransfer(InterchainTransfer { token_id, .. })
-            | Message::DeployInterchainToken(DeployInterchainToken { token_id, .. }) => *token_id,
+            | Message::DeployInterchainToken(DeployInterchainToken { token_id, .. })
+            | Message::LinkToken(LinkToken { token_id, .. }) => *token_id,
+            Message::RegisterToken(_) => panic!("no token id associated with this message type"), // TODO is this right?
```
cgorenflo, line 150: I don't like circumventing the type system here. Either we need to treat RegisterToken separately or token_id() should return an option

contracts/interchain-token-service/src/contract/execute/mod.rs
```diff
@@ -100,12 +111,54 @@ pub fn execute_message(
...
+    )?;
+    Ok(Response::new().add_event(
+        Event::MessageReceived {
+            cc_id,
+            destination_chain,
+            message,
+        }
+        .into(),
+    ))
+}
```
cgorenflo, line 141: idea to unify the behaviour and get rid of this special case:
- move loading the destination address and destination payload into the `send_to_destination` function
- add the special case to send_to_destination, that if the destination is axelarnet then be a noop
as far as I can tell, then execute_message_on_hub does the same as execute_axelar_message

cgorenflo (approved):

contracts/interchain-token-service/src/contract/execute/mod.rs
cjcobb23: we would also need to somehow skip the frozen check, since axelar is not a registered chain. Or somehow register axelar as a chain, even though there is no ITS contract yet on Axelar

contracts/interchain-token-service/src/contract/execute/mod.rs
```diff
@@ -157,10 +173,76 @@ fn apply_to_hub(
...
+            destination_chain.clone(),
+            link_token.destination_token_address.clone(),
+        )
+        .change_context(Error::State)?
+        .ok_or(Error::TokenNotRegistered(
+            link_token.destination_token_address.clone(),
+        ))?
+        .decimals
+    } else {
+        source_token.decimals
```
fish-sammy, line 232: Correct me if I am wrong, say chain A registers token T with X decimals and chain B registers token T with Y decimals, and then we deploy link token T from chain A to chain B with autoscaling disabled. Now token T has both X decimals for chain A & B. If now chain C also registers token T, it can never be linked from chain B but has to be from chain A because `deploy_token_to_source_chain` is gonna see a decimal mismatch of Token T on chain B. Is this an issue?

contracts/interchain-token-service/src/contract/execute/mod.rs
```diff
@@ -227,9 +321,22 @@ fn send_to_destination(
     storage: &dyn Storage,
     querier: QuerierWrapper,
     destination_chain: &ChainNameRaw,
-    destination_address: Address,
     payload: HexBinary,
 ) -> Result<Response, Error> {
+    if *destination_chain == axelar_chain_name(storage, querier)? {
+        // messages sent to axelar are handled in a special way
```
fish-sammy, line 327: What's the special way right now?

fish-sammy (changes requested):

contracts/interchain-token-service/src/events.rs
```diff
@@ -80,6 +80,26 @@ fn make_message_event(
             .add_attribute("symbol", symbol)
             .add_attribute("decimals", decimals.to_string())
             .add_attribute_if_some("minter", minter.map(|minter| minter.to_string())),
+        Message::RegisterToken(RegisterToken { address, decimals }) => event
+            .add_attribute("decimals", decimals.to_string())
```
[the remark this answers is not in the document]
milapsheth: We should just add a derive macro so the event struct can be converted into a cosmwasm event by requiring field types to implement ToString

contracts/interchain-token-service/src/primitives.rs
```diff
@@ -74,6 +79,36 @@ impl From<DeployInterchainToken> for Message {
     }
 }
 
+#[cw_serde]
+#[derive(Eq)]
+pub struct RegisterToken {
```
milapsheth, line 84:
```suggestion
pub struct RegisterTokenMetadata {
```

contracts/interchain-token-service/src/state.rs
```diff
@@ -126,10 +126,19 @@ pub struct TokenConfig {
     pub origin_chain: ChainNameRaw,
 }
 
+type TokenAddress = nonempty::HexBinary;
+#[cw_serde]
+pub struct CustomToken {
```
milapsheth, line 131:
```suggestion
pub struct TokenMetadata {
```
Custom feels weird, this can be metadata for any token

contracts/interchain-token-service/src/state.rs
```diff
@@ -126,10 +126,19 @@ pub struct TokenConfig {
...
+    pub chain: ChainNameRaw,
```
milapsheth, line 132: Why is the chain included here?

contracts/interchain-token-service/src/state.rs
```diff
@@ -281,6 +290,34 @@ pub fn save_token_config(
         .change_context(Error::Storage)
 }
 
+pub fn save_custom_token(
+    storage: &mut dyn Storage,
+    source_chain: ChainNameRaw,
+    register_token: RegisterToken,
```
milapsheth, line 296:
```suggestion
    RegisterToken { address, decimals, chain }: RegisterToken,
```
Destructure the arg maybe so the following code is cleaner

contracts/interchain-token-service/src/abi.rs
```diff
@@ -144,6 +187,47 @@ impl Message {
                 }
                 .into()
             }
+            MessageType::RegisterToken => {
+                let decoded = RegisterToken::abi_decode_params(payload, true)
```
milapsheth, line 191: Maybe destructure in the let assignment?

contracts/interchain-token-service/src/contract/execute/interceptors.rs
```diff
@@ -290,18 +287,22 @@ fn destination_amount(
...
+        storage,
+        source_chain.clone(),
+        register_token.address.clone(),
+    )
+    .change_context(Error::State)?;
+    ensure!(
+        existing_token.is_none(),
+        Error::TokenAlreadyRegistered(register_token.address)
+    );
+    state::save_custom_token(storage, source_chain, register_token).change_context(Error::State)
```
milapsheth, line 305:
```suggestion

    state::save_custom_token(storage, source_chain, register_token).change_context(Error::State)
```

contracts/interchain-token-service/src/contract/execute/mod.rs
milapsheth: `deploy_token_to_source_chain` will prevent linking from a different chain other than A since a global token config is stored for that tokenId after the first deployment and the method ensures the source chain for LinkToken doesn't change. And ITS edge won't allow linking from a different origin chain since tokenId derivation will be different

contracts/interchain-token-service/src/primitives.rs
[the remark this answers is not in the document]
cjcobb23: yes we should, but not as part of this PR

contracts/interchain-token-service/src/state.rs
cjcobb23: wow i didn't know you could do this

contracts/interchain-token-service/src/state.rs
cjcobb23: it's not necessary, but we do this in some maps and not others (include the key in the object itself). Not sure what our pattern should be

contracts/interchain-token-service/src/primitives.rs
```diff
@@ -74,6 +79,36 @@ impl From<DeployInterchainToken> for Message {
...
+    fn from(value: RegisterToken) -> Self {
+        Message::RegisterToken(value)
+    }
+}
+
+#[cw_serde]
+#[derive(Eq)]
+pub struct LinkToken {
+    pub token_id: TokenId,
+    pub token_manager_type: Uint256,
```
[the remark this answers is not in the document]
cgorenflo: might be difficult to do because Uint256 is not a rust primitive. Maybe treat it as black box data for now and handle it in more detail if we actually have to differentiate types?

contracts/interchain-token-service/src/events.rs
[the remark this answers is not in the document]
cgorenflo: I've been meaning to actually convert all the values to json to be consistent for a while, but never had the time to do so. We definitely shouldn't go further into the direction of more customized outputs, axelarons basically has to build a custom parser for each event they process

contracts/interchain-token-service/src/contract/execute/mod.rs
cgorenflo: they don't send anything to another destination chain after the hub is done with the message

contracts/interchain-token-service/src/state.rs
cgorenflo: NonItsToken? I want it to be clear that this is a special case

contracts/interchain-token-service/src/state.rs
cgorenflo: I'm still of the opinion that we shouldn't need to rely on the storage key to get relevant data. Keys should be pretty much one way, i.e. you can compose them from intelligible data and you can use them to filter directly in the store, but they shouldn't be treated as part of the value. Otherwise, we tie the storage model to the contract logic and it's a pain to make changes. I'm fine with normalizing data in the store, i.e. we could fill a state object with multiple calls to different keys in the storage.

contracts/interchain-token-service/src/state.rs
cjcobb23: yeah this is a special case, it can't be metadata for any token

contracts/interchain-token-service/src/contract/execute/mod.rs
cjcobb23: yeah it's what I describe below. Maybe I can just remove this specific line though

contracts/interchain-token-service/src/state.rs
milapsheth: I don't follow, this could be any token on that chain. Even ITS tokens adhere to the chain's token standard and could be linked as a custom token.

This functionality allows registering metadata for any existing token (whether it's a standard ERC20, ITS ERC20 token, custom token or some other standard doesn't matter to ITS Hub)

contracts/interchain-token-service/src/state.rs
milapsheth: This pattern feels pretty weird when we have structured keys that are part of the storage themselves. I understand that this was problematic in core, and can be an issue in other VMs like EVM where the key isn't stored as well, but cosmwasm guarantees both key and value can be retrieved from storage so this feels unnecessary. It introduces a new concern where the chain name / token address in the key don't match what's stored in the value which I don't really like. I'm fine with sticking to this if you strongly favour it, but I don't quite see the benefit of doing this everywhere

contracts/interchain-token-service/src/state.rs
milapsheth: Haha, not commonly used, but it's neat sometimes

contracts/interchain-token-service/src/primitives.rs
```diff
@@ -29,6 +30,10 @@ pub enum Message {
     InterchainTransfer(InterchainTransfer),
     /// Deploy a new interchain token on the destination chain
     DeployInterchainToken(DeployInterchainToken),
+
+    RegisterToken(RegisterTokenMetadata),
```
milapsheth, line 34:
```suggestion
    RegisterTokenMetadata(RegisterTokenMetadata),
```
Rename message type as well

contracts/interchain-token-service/src/primitives.rs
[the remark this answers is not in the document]
milapsheth: We had defined an enum before for it. There's a tradeoff here: the benefit is extra validation but the value itself isn't useful to the hub. The downside of the validation is also that when new token manager types are added to the edge, the Hub has to be updated as well which is additional overhead. For e.g. EVM ITS might add support for some more types, and the other ITS contracts don't really care and will continue accepting only the types they support. So I'm leaning towards a passthrough approach as well

contracts/interchain-token-service/src/contract/execute/mod.rs
```diff
@@ -154,10 +170,72 @@ fn apply_to_hub(
...
+    ))?;
+
+    deploy_token_to_source_chain(
+        storage,
+        &source_chain,
+        link_token.token_id,
+        source_token.decimals,
+    )?;
+
+    let destination_decimals = state::may_load_custom_token(
```
milapsheth, line 217: What's the plan with autoscaling? Adding it in a separate PR or release, or not adding it, or want to do it differently?

contracts/interchain-token-service/src/state.rs
cjcobb23: hmm yes i guess that's true (though I think this sort of behavior is very hard to reason about and we should probably find a way to improve this aspect of ITS). But the issue is that not all tokens will have metadata stored in this struct. So we need some way to reflect that special case

fish-sammy (approved):

contracts/interchain-token-service/src/state.rs
cjcobb23: I couldn't think of a better name (don't like `TokenMetadata`), so I just left it as is for now, we can change it later if we want

contracts/interchain-token-service/src/state.rs
cjcobb23: I do prefer just storing all the relevant data together in one object

PR #726, opened 2024-12-14
feat: starknet message_id and address format

ctoyan:
## Description
The PR consists of:
- `voting-verifier` CosmWASM contract changes, mostly related to MessageIdFormat and AddressFormat
- `MessageIdFormat::FieldElementAndEventIndex` and `AddressFormat:Starknet` 

## Todos

- [x] Unit tests
- [x] Manual tests
- [x] Documentation
- [ ] Connect epics/issues

## Steps to Test
You would need the Starknet relayer, in order to manually test, which is not open sourced yet.

## Expected Behaviour
To restrict a value, which overflows the FieldElement max value.

## Other Notes
Once this PR is merged and the changes related to the format are deployed on your devnet contracts, we'll re-test manually.


packages/axelar-wasm-std/src/msg_id/starknet_field_element_event_index.rs
ctoyan, line 1: This MessageIdFormat is very similar to the `HexTxHashAndEventIndex`.
The main difference is that it's checking if the FieldElement (Felt) overflows.
`Felt::MAX` is 2^251+17*2^192+1, which fits in 252 bits. This means that a 64 hex string, representing 256 bits, might overflow it. We enforce 64 hex string everywhere with padded zeros.

If you find this check useless, I believe we can get rid of the custom message id format and re-use the `HexTxHashAndEventIndex`.

packages/axelar-wasm-std/src/utils.rs
```diff
@@ -19,6 +22,22 @@ impl<T> TryMapExt<T> for Vec<T> {
     }
 }
 
+/// since the `Felt` type doesn't error on overflow, we have to implement that check
```
ctoyan, line 25: I was also contemplating creating something like a `FeltChecked` type, that implements this check within it, but ultimately decided that adding an extra type won't result in a cleaner result.

packages/axelar-wasm-std/src/utils.rs
```diff
@@ -19,6 +22,22 @@ impl<T> TryMapExt<T> for Vec<T> {
...
+pub fn does_felt_overflow_from_slice(felt_hex_slice: &[u8]) -> bool {
+    if felt_hex_slice.len() > 32 {
```
ctoyan, line 27: Felt is 252 bits or 31.5 bytes.

packages/axelar-wasm-std/src/msg_id/starknet_field_element_event_index.rs
```diff
@@ -0,0 +1,284 @@
...
+            tx_hash: tx_id.into(),
+            event_index: event_index.into(),
+        })
+    }
+}
+
+// A valid field element is max 252 bits, meaning max 63 hex characters after 0x.
+// We require the hex to be 64 characters, meaning that it should be padded with zeroes in order
+// for us to consider it valid.
+const PATTERN: &str = "^(0x[0-9a-f]{64})-(0|[1-9][0-9]*)$";
```
cjcobb23, line 51: where is the 0 padding in this regex?

packages/axelar-wasm-std/Cargo.toml
```diff
@@ -50,17 +50,19 @@ serde_json = "1.0.89"
 serde_with = { version = "3.11.0", features = ["macros"] }
 sha3 = { workspace = true }
 stellar-xdr = { workspace = true }
+starknet-types-core = { workspace = true }
 strum = { workspace = true }
 sui-types = { workspace = true }
 thiserror = { workspace = true }
 valuable = { version = "0.1.0", features = ["derive"] }
+crypto-bigint = { version = "0.5.5", features = ["rand_core"] }
```
cjcobb23, line 58: I don't think we need this dep, there is already a u256 type available

packages/axelar-wasm-std/src/msg_id/starknet_field_element_event_index.rs
cjcobb23: the check is not useless, I prefer keeping it. Otherwise overflow would be handled elsewhere, and could be handled incorrectly

packages/axelar-wasm-std/src/utils.rs
```diff
@@ -19,6 +22,22 @@ impl<T> TryMapExt<T> for Vec<T> {
...
+        return true;
+    }
+    let felt_max_hex_str = format!("{:064x}", Felt::MAX);
+    U256::from_be_slice(felt_hex_slice) > U256::from_be_hex(&felt_max_hex_str)
+}
+
+/// since the `Felt` type doesn't error on overflow, we have to implement that check
+pub fn does_felt_overflow_from_str(felt_hex_str: &str) -> bool {
```
cjcobb23, line 35: can't we just check if the first nibble (4 bits) is non-zero?

packages/axelar-wasm-std/src/utils.rs
```diff
@@ -1,3 +1,6 @@
+use crypto_bigint::U256;
```
cgorenflo, line 1: I think this should be possible to express with existing types

packages/axelar-wasm-std/src/utils.rs
```diff
@@ -19,6 +22,22 @@ impl<T> TryMapExt<T> for Vec<T> {
     }
 }
 
+/// since the `Felt` type doesn't error on overflow, we have to implement that check
+pub fn does_felt_overflow_from_slice(felt_hex_slice: &[u8]) -> bool {
+    if felt_hex_slice.len() > 32 {
+        return true;
+    }
+    let felt_max_hex_str = format!("{:064x}", Felt::MAX);
```
cgorenflo, line 30: can the corresponding U256 not be defined as a const?

packages/axelar-wasm-std/src/utils.rs
```diff
@@ -19,6 +22,22 @@ impl<T> TryMapExt<T> for Vec<T> {
     }
 }
 
+/// since the `Felt` type doesn't error on overflow, we have to implement that check
+pub fn does_felt_overflow_from_slice(felt_hex_slice: &[u8]) -> bool {
+    if felt_hex_slice.len() > 32 {
+        return true;
+    }
```
cgorenflo, line 29: this case would be caught by the U256 comparison anyway right? I'd get rid of this

packages/axelar-wasm-std/src/utils.rs
cgorenflo, line 1: these function should not go into the general utils.rs, since they are specific stasknet helpers

packages/axelar-wasm-std/src/msg_id/starknet_field_element_event_index.rs
```diff
@@ -0,0 +1,284 @@
...
+use serde_with::DeserializeFromStr;
+use starknet_types_core::felt::Felt;
+
+use super::Error;
+use crate::nonempty;
+use crate::utils::does_felt_overflow_from_str;
+
+#[derive(Debug, DeserializeFromStr, Clone)]
+pub struct FieldElementAndEventIndex {
+    pub tx_hash: Felt,
```
cgorenflo, line 18: From what I've seen the `Felt` type doesn't guarantee validity. I would prefer if you'd implement something like a `ValidatedFelt` type, similar to nonempty::String, which makes sure that all checks have passed. Then you can also easily implement `TryFrom`s from byte slices and strings and don't have to do the multiple `does_felt_overflow_from_X` implementations and `FieldElementAndEventIndex` is always guaranteed to be valid

cgorenflo (changes requested):

packages/axelar-wasm-std/src/utils.rs
cgorenflo: see my other comment, I think it's a good idea to do so

packages/axelar-wasm-std/src/msg_id/starknet_field_element_event_index.rs
ctoyan: It should come padded and we just check that the length is 64 characters.
It can be padded with 1 or 2 zeroes. If it's not padded - it'll be invalid regex. If they're not zeroes (> 252 bits) the overflow check would fail.

packages/axelar-wasm-std/src/address.rs
```diff
@@ -4,9 +4,12 @@ use alloy_primitives::Address;
 use cosmwasm_schema::cw_serde;
 use cosmwasm_std::{Addr, Api};
 use error_stack::{bail, Result, ResultExt};
+use starknet_types_core::felt::Felt;
```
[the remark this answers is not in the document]
ctoyan: Same type, [re-exported](https://github.com/xJonathanLEI/starknet-rs/blob/master/starknet-core/src/types/mod.rs#L8) by the core lib. But `starknet_types_core` is CosmWASM compilable, while `starknet_core` is not and `axelar-wasm-std` is used in some contracts.

Cargo.toml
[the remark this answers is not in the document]
ctoyan: These changes came from a toml language server, on file save, so I assumed they're "more correct" 😄 
I can bring back the old formatting, if you like?

Cargo.toml
```diff
@@ -77,6 +87,11 @@ tokio-stream = "0.1.11"
 tokio-util = "0.7.11"
 voting-verifier = { version = "^1.1.0", path = "contracts/voting-verifier" }
 axelar-core-std = { version = "^1.0.0", path = "packages/axelar-core-std" }
+# Async
+futures-concurrency = "7.4"
+futures-util = "0.3"
+futures = "0.3"
+async-trait = "0"
```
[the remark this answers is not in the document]
ctoyan: I believe they are a cherry picking mistake. They're needed and used in some of the packages, but are strictly defined there.
I'll remove them.

packages/axelar-wasm-std/src/utils.rs
ctoyan: I thought that as well. 
It's a 252 bit number, but not 252 bit MAX number, 2^252-1 (252 one values).
The first bit is one, but the other ones are not all ones. This is what it looks like, represented as 256 bits:

`0000 1000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0001 0001 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000`

So checking the first 4 bits for non-zero is not enough.

packages/axelar-wasm-std/src/utils.rs
ctoyan: Note to self. Should be `>=`

packages/axelar-wasm-std/src/utils.rs
ctoyan: `U256::from_be_slice` panics and I prefer to not handle an error or return a result and just keep it a simple bool.
Does that make sense?

PS: I'd put these checks into the custom type you suggest and this might be an unneeded discussion anyway.

packages/starknet-checked-felt/Cargo.toml
ctoyan, line 1: In the ampd PR, I create a `starknet-types` folder (which we'll try to completely omit at some point), that contains some starknet specific implementation. 
The reason I have `starknet-types` in the other PR and `starknet-checked-felt` in this PR, is because otherwise I create a circular dependency of `axelar-wasm-std`.

This was the only solution I could think of and make work.

packages/axelar-wasm-std/src/utils.rs
ctoyan: Did that into a separate package because of a cyclic dependency problem.

packages/axelar-wasm-std/Cargo.toml
ctoyan: Done. I changed the test U256 randomisation and removed this dependency.

packages/axelar-wasm-std/src/utils.rs
ctoyan: You're right. I switched to `alloy-primitives` U256 type, which you already have as a depenendcy.

packages/axelar-wasm-std/src/utils.rs
ctoyan: I defined that as a const into the new package.

packages/starknet-checked-felt/Cargo.toml
ctoyan, line 1: In the ampd PR, I create a `starknet-types` folder (which we'll try to completely omit at some point), that contains some starknet specific implementation. 
The reason `starknet-checked-felt` is not part of `starknet-types`, is because I create a circular dependency on `axelar-wasm-std`. 

This was the only solution I could think of and make work.

Also this package is added into CosmWASM contracts and needs to be able to compile. I think `starknet-types` doesn't compile IIRC.

packages/axelar-wasm-std/src/msg_id/starknet_field_element_event_index.rs
cjcobb23: Based on your other comment, it looks like it will always start with a 0, and the second character will be either 0 or 1. Can we at least encode that in the regex? It doesn't hurt, and then the regex itself gives more info about what the message id should actually look like.

packages/starknet-checked-felt/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+
+        std::fmt::LowerHex::fmt(&val, f)
+    }
+}
+
+impl AsRef<Felt> for CheckedFelt {
+    fn as_ref(&self) -> &Felt {
+        self.0.as_ref()
+    }
+}
```
cgorenflo, line 136: shouldn't be necessary when Deref is defined

packages/starknet-checked-felt/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+
+#[cfg(test)]
+mod tests {
+    // use core::slice::SlicePattern;
```
cgorenflo, line 140: remove

packages/axelar-wasm-std/src/msg_id/starknet_field_element_event_index.rs
ctoyan: > it looks like it will always start with a 0, and the second character will be either 0 or 1
First one should always be`0`. Second one can be up to `8`.
This is `Felt::MAX` in 64 characters - `0x0800000000000011000000000000000000000000000000000000000000000000`

> would this be a correct(er) pattern?
Yes. I was thinking the same. I'll add it.

packages/starknet-checked-felt/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+    fn from(value: CheckedFelt) -> Self {
+        value.0
+    }
+}
+
+impl From<Felt> for CheckedFelt {
+    fn from(value: Felt) -> Self {
+        CheckedFelt(value)
+    }
+}
```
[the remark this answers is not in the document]
ctoyan: I thought about that when I was writing it, but you're right.

My reasoning was that `Felt` itself can never be invalid.
Problem is that someone can create a `Felt` using `Felt::from_hex`, which would just saturate/wrap it, and then `.into()` it to a `CheckedFelt`. It will still result in a valid `Felt`, but an overflown one.

packages/starknet-checked-felt/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+            return Err(CheckedFeltError::Overflowing);
+        }
+
+        if U256::from_be_slice(value) > FELT_MAX_U256 {
+            return Err(CheckedFeltError::Overflowing);
+        }
+
+        Ok(CheckedFelt(Felt::from_bytes_be_slice(value)))
+    }
+}
```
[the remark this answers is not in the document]
ctoyan: I accepted the change, but I have two comments on the `ensure!` macro part:

1. I believe the check should be changed to `ensure!(U256::from_be_slice(value) <= FELT_MAX_U256, CheckedFeltError::Overflowing);`
2. I get an error, which I can't seem to get around and which doesn't make much sense to me:

```
error[E0308]: mismatched types
  --> packages/starknet-checked-felt/src/lib.rs:39:9
   |
39 | /         ensure!(
40 | |             U256::from_be_slice(value) > FELT_MAX_U256,
41 | |             CheckedFeltError::Overflowing
42 | |         );
   | |_________^ expected `CheckedFeltError`, found `Error`
```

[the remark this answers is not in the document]
ctoyan: I committed it, but then decided to rework everything to work with `ensure!` for consistency.
Also the logic from the if had to be flipped, because `ensure!` negates the condition.
For example `if a > b { error } `, becomes `ensure!(a<=b, error)`

packages/axelar-wasm-std/src/address.rs
```diff
@@ -38,6 +40,38 @@ pub fn validate_address(address: &str, format: &AddressFormat) -> Result<(), Err
...
+            }
+
+            if CheckedFelt::from_str(trimmed_addr).is_err() {
+                bail!(Error::InvalidAddress(
+                    format!(
+                        "field element overflows MAX value of 2^251 + 17 * 2^192: {}",
+                        address
+                    )
+                    .to_string()
+                ))
```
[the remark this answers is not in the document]
ctoyan: Moved all those checks in the `CheckedFelt::from_str` and made it more consistent with the other code in this file.

packages/axelar-wasm-std/src/msg_id/starknet_field_element_event_index.rs
```diff
@@ -126,7 +125,8 @@ mod tests {
         let res = FieldElementAndEventIndex::from_str(
             "0x0670d1dd42a19cb229bb4378b58b9c3e76aa43edaaea46845cd8c456c1224d89-0",
         );
-        assert!(res.is_ok());
+        res.unwrap();
+        // assert!(res.is_ok());
```
ctoyan, line 129: I fixed this in the next commit 😄

packages/axelar-wasm-std/src/msg_id/starknet_field_element_event_index.rs
cgorenflo: yes, if that is guaranteed then restricting it further is even better 👍

cgorenflo (approved):

packages/axelar-wasm-std/src/msg_id/starknet_field_element_event_index.rs
```diff
@@ -0,0 +1,279 @@
...
+
+    pub fn new<T: Into<CheckedFelt> + FromStr>(
+        tx_id: T,
+        event_index: impl Into<u64>,
+    ) -> Result<Self, Error> {
+        Ok(Self {
+            tx_hash: tx_id.into(),
+            event_index: event_index.into(),
+        })
+    }
```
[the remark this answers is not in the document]
ctoyan: I accidentally left it from a previous implementation. 
Removed it now.

packages/starknet-checked-felt/src/lib.rs
ctoyan: I removed it initially, but a starknet rust lib function requires `impl AsRef<Felt>` and uses `.as_ref()` on a felt.
I would prefer to keep this impl, because it would make the ampd code a bit cleaner.
I'm bringing it back, but if you'd still prefer that it's removed, we can do that in the ampd PR.

PR #728, opened 2024-12-16
feat(ampd): add Stacks handlers

raress96:
## Description

Stacks verify msg & verify verifier set handlers.

### DIFF
--- a/ampd/src/stacks/http_client.rs
+++ b/ampd/src/stacks/http_client.rs
@@ -18,7 +18,7 @@ const LATEST_BLOCK: &str = "extended/v2/blocks/latest";
 
 const STATUS_SUCCESS: &str = "success";
 
-#[derive(Error, Debug)]
+#[derive(Error, Debug, Clone)]
 pub enum Error {
     #[error("invalid tx hash {endpoint}")]
     TxHash { endpoint: String },
@@ -28,25 +28,25 @@ pub enum Error {
     Json,
 }
 
-#[derive(Debug, Deserialize, Default)]
+#[derive(Debug, Deserialize, Default, Clone)]
 pub struct ContractLogValue {
     pub hex: String,
 }
 
-#[derive(Debug, Deserialize, Default)]
+#[derive(Debug, Deserialize, Default, Clone)]
 pub struct ContractLog {
     pub contract_id: String,
     pub topic: String,
     pub value: ContractLogValue,
 }
 
-#[derive(Debug, Deserialize, Default)]
+#[derive(Debug, Deserialize, Default, Clone)]
 pub struct TransactionEvents {
     pub event_index: u64,
     pub contract_log: Option<ContractLog>,
 }
 
-#[derive(Debug, Deserialize, Default)]
+#[derive(Debug, Deserialize, Default, Clone)]
 pub struct Transaction {
     pub tx_id: Hash,
     pub tx_status: String, // 'success'
@@ -54,7 +54,7 @@ pub struct Transaction {
     pub burn_block_height: u64,
 }
 
-#[derive(Debug, Deserialize, Default)]
+#[derive(Debug, Deserialize, Default, Clone)]
 pub struct Block {
     pub burn_block_height: u64, // Height of Bitcoin block
 }
@@ -151,7 +151,10 @@ impl Client {
 
 #[cfg(test)]
 mod tests {
+    use std::time::Duration;
+
     use super::{Block, Client, Transaction};
+    use crate::url::Url;
 
     #[test]
     fn parse_transaction() {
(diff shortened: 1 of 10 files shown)


ampd/src/stacks/contract_verifier.rs
```diff
@@ -0,0 +1,249 @@
...
+        .get_contract_info(format!("{}", reference_address.to_string()).as_str())
+        .await?;
+
+    let actual_contract_code = http_client
+        .get_contract_info(format!("{}", contract_address.to_string()).as_str())
+        .await?;
+
+    Ok(reference_contract_code.source_code == actual_contract_code.source_code)
+}
+
```
cjcobb23, line 76: I am worried about the performance implications of this. As far as I can tell, this is downloading two sets of source code and then comparing them? Is it comparing a hash of the source code or the full bytes of the source code?

ampd/src/stacks/verifier.rs
```diff
@@ -0,0 +1,649 @@
...
+
+            if &message.destination_chain != source_chain
+                && message.destination_address != message.source_address
+            {
+                return Vote::SucceededOnChain;
+            }
+
+            // In case message is from Stacks -> Stacks and the same contract to itself,
+            // try to see if we need to verify if a contract was deployed
+            let result = get_verify_contract_params(event);
```
cjcobb23, line 219: I still don't think this functionality should exist in ampd. Suddenly now ampd is doing something different than what it normally does (verify gmp messages and verify verifier set rotations). This change opens the door for ampd to start doing all kinds of custom logic, and just makes the code and system more complicated to reason about and maintain. Even from a performance perspective, we need to be aware of the overhead of attempting to verify contract deployments now. I don't think we should go down this route.

cjcobb23 (changes requested):

ampd/src/stacks/verifier.rs
raress96: The contract verification logic was removed, no custom logic for Stacks is now in ampd.

ampd/src/stacks/contract_verifier.rs
raress96: No longer applicable

ampd/src/stacks/finalizer.rs
```diff
@@ -0,0 +1,45 @@
...
+    let block = http_client.get_latest_block().await?;
+
+    let block_number: U64 = block.height.into();
+
+    // order of operations is important here when saturating, otherwise the finalization window could be cut short
+    // if we add 1 afterwards
+    Ok(block_number
+        .saturating_add(U64::from(1))
+        .saturating_sub(confirmation_height.into())
+        .as_u64())
```
eguajardo, line 22: I don't think this is enough to identify a finalized block in Stacks. If I'm understanding the [documentation](https://docs.stacks.co/concepts/block-production/bitcoin-reorgs#replaying-transactions-3) correctly, at the moment Stacks finality is tied to Bitcoin's finality. Wouldn't you need to take the anchored blocks in consideration instead of the normal blocks?

eguajardo (changes requested):

ampd/src/stacks/finalizer.rs
raress96: This was a discussion we had with the auditors & the Stacks Axelar contract developers. From my understanding, after the Stacks Nakamoto upgrade finality in Stacks should be instant, after a transaction is confirmed in a Stacks block, it should be final even if Bitcoin forks. Exceptions to this are Bitcoin-reliant transactions, which fetch state from Bitcoin, but the Stacks Axelar contract don't use such functionality:
https://docs.stacks.co/concepts/block-production/bitcoin-finality#nakamoto-transactions-and-bitcoin-reorgs

Initially we didn't even have this finalizer implemented, but we decided it would be best to add it if we want to configure a greater `confirmation_height` for Stacks other than 1 in the future. I will send this comment to the Stacks team which developed the contracts, to check if there is anything else I missed.

ampd/src/handlers/stacks_verify_msg.rs
```diff
@@ -0,0 +1,369 @@
...
+use crate::types::{Hash, TMAddress};
+
+type Result<T> = error_stack::Result<T, Error>;
+
+#[derive(Deserialize, Debug)]
+pub struct Message {
+    pub message_id: HexTxHashAndEventIndex,
+    pub destination_address: String,
+    pub destination_chain: ChainName,
+    pub source_address: String,
```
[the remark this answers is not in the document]
raress96: `StacksAddress` only includes user account addresses, not smart contracts. I tried using `PrincipalData` but unfortunately it uses the default serde derive(Deserialize) which does not handle deserialization like it should...
The address is checked here however that it is a valid Stacks address:
https://github.com/axelarnetwork/axelar-amplifier/pull/728/files#diff-09b6aab36d03d4a95649534a32886c7b058e5fa87a85415221a80934c217c03cR77

Will try to see if I can get this working with some other type, but it doesn't seem possible and we will probably have to leave this as String here.

ampd/src/handlers/stacks_verify_msg.rs
```diff
@@ -0,0 +1,369 @@
...
+                .await
+                .change_context(Error::Finalizer)?;
+
+        let tx_hashes: HashSet<Hash> = messages
+            .iter()
+            .map(|message| message.message_id.tx_hash.into())
+            .collect();
+        let transactions = self
+            .http_client
+            .get_finalized_transactions(tx_hashes, latest_finalized_block_height)
```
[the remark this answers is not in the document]
raress96: Done

ampd/src/stacks/http_client.rs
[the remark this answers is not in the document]
raress96: Done

ampd/src/lib.rs
```diff
@@ -389,6 +391,34 @@ where
...
                 ),
+                handlers::config::Config::StacksMsgVerifier {
+                    cosmwasm_contract,
+                    http_url,
+                } => self.create_handler_task(
+                    "stacks-msg-verifier",
+                    handlers::stacks_verify_msg::Handler::new(
+                        verifier.clone(),
+                        cosmwasm_contract,
+                        Client::new_http(http_url.to_string().trim_end_matches('/').into()),
```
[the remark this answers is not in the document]
raress96: Done

ampd/src/lib.rs
```diff
@@ -389,6 +391,34 @@ where
...
+                ),
+                handlers::config::Config::StacksVerifierSetVerifier {
+                    cosmwasm_contract,
+                    http_url,
+                } => self.create_handler_task(
+                    "stacks-verifier-set-verifier",
+                    handlers::stacks_verify_verifier_set::Handler::new(
+                        verifier.clone(),
+                        cosmwasm_contract,
+                        Client::new_http(http_url.to_string().trim_end_matches('/').into()),
```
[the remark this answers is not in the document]
raress96: Done

cjcobb23 (approved):

eguajardo (approved):

PR #731, opened 2024-12-18
feat: starknet GMP

ctoyan:
## Description
This PR consists of:

- call contract event verification
- signer rotation event verification
- starknet specific types and events de/serialisation implementations

## Todos

- [x] Unit tests
- [x] Manual tests
- [x] Documentation
- [ ] Connect epics/issues

## Steps to Test
You would need the Starknet relayer, in order to manually test, which is not open sourced or deployed yet.

## Expected Behaviour
Handle GMP messages from other amplifier chains

## Other Notes
The [FieldElementAndEventIndex PR](https://github.com/axelarnetwork/axelar-amplifier/pull/726) must be merged to main before this PR manages to compile successfully.


ctoyan:
Converted it back to draft, as I expect some changes here.

Cargo.toml
```diff
@@ -80,6 +80,7 @@ sui-gateway = { version = "^1.0.0", path = "packages/sui-gateway" }
 sui-types = { version = "^1.0.0", path = "packages/sui-types" }
 starknet-checked-felt = { version = "^1.0.0", path = "packages/starknet-checked-felt" }
 starknet-types-core = { version = "0.1.7" }
+starknet-types = { version = "^1.0.0", path = "packages/starknet-types" }
```
ctoyan, line 83: The types in this crate can be substituted, if we use the [cainome](https://github.com/cartridge-gg/cainome) crate. It has an `abigen!()` macro similar to the ether's one, which can auto-generate those types. 
I wasn't sure if you'd like another dependency or have custom implementations.

ampd/src/handlers/starknet_verify_msg.rs
```diff
@@ -0,0 +1,655 @@
...
+            .collect();
+
+        let mut votes = vec![];
+        for msg in unique_msgs {
+            if !events.contains_key(&msg.message_id) {
+                votes.push(Vote::NotFound);
+                continue;
+            }
+            votes.push(verify_msg(
+                &events.remove(&msg.message_id).unwrap(), // safe to unwrap, because of previous check
```
cjcobb23, line 153: let's refactor the code to avoid this unwrap

ampd/src/handlers/starknet_verify_msg.rs
```diff
@@ -0,0 +1,655 @@
...
+        let unique_msgs = messages
+            .iter()
+            .unique_by(|msg| msg.message_id.to_string())
+            .collect::<Vec<_>>();
+
+        // key is the message_id of the tx holding the event
+        let mut events: HashMap<FieldElementAndEventIndex, ContractCallEvent> =
+            try_join_all(unique_msgs.iter().map(|msg| {
+                self.rpc_client
+                    .get_events_by_hash_contract_call(msg.message_id.tx_hash.clone())
```
cjcobb23, line 138: why not pass in the event index as well, and just return the event if found? This `events` map is going to have a ton of events that the code doesn't care about, would probably be more efficient and clearer to not include those

ampd/src/starknet/verifier.rs
```diff
@@ -0,0 +1,348 @@
...
+            && axl_msg.payload_hash == self.payload_hash
+    }
+}
+
+pub fn verify_verifier_set(
+    event: &SignersRotatedEvent,
+    confirmation: &VerifierSetConfirmation,
+    source_gateway_address: &str,
+) -> Vote {
+    if event.signers.nonce != [0_u8; 32]
```
cjcobb23, line 42: what is this special case for?

ampd/src/starknet/verifier.rs
```diff
@@ -0,0 +1,348 @@
...
+        // Convert and sort expected signers
+        let mut expected_signers = expected
+            .signers
+            .values()
+            .map(|signer| {
+                if let multisig::key::PublicKey::Ecdsa(pubkey) = &signer.pub_key {
+                    (pubkey.clone(), signer.weight.u128())
+                } else {
+                    // Skip non-ECDSA keys
+                    (HexBinary::from_hex("").unwrap(), 0)
```
cjcobb23, line 65: Is this not an expected case? If so I think you should return an error from `map`, instead of this empty data.

packages/starknet-types/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
```
cjcobb23, line 1: are these types only used in ampd? or are they used in contracts as well? If they are only used in ampd, I think they should just live in a folder in ampd, close to where they are used. I see the packages folder as a place for things that are used across the entire repo.

ctoyan: The StarknetMessage type is used in the `multisig-prover`.
I can move the events and other types in ampd.

I'll use this thread to ask a question. We plan to migrate to using the `cainome` crate in our relayer. It has an `abigen!()` macro, which generates starknet types based on the ABI JSON file.

Do you prefer us to import that dependency and to maintain an ABI JSON file, or for us to leave the types in this folder? If we do - we'll delete all the types apart from `StarknetMessage` and use the generated ones. Else - we keep the ones here and only update if needed (rare to never).

ampd/src/handlers/starknet_verify_msg.rs
ctoyan: This `get_events_by_hash_contract_call ` function only returns ContractCall events and my reasoning was that fetching the ContractCall events in a tx_hash, that's pointed in the PollStarted event would be exactly what is needed.
What other ContractCall events that the code doesn't care about can originate in a single transaction 🤔

ampd/src/handlers/starknet_verify_msg.rs
ctoyan: Please, check [this commit](https://github.com/axelarnetwork/axelar-amplifier/pull/731/commits/c811313da89c88d8df0a157f6cc2f934f8bc3e3c) for the change

packages/starknet-types/Cargo.toml
ctoyan: Please check [this commit](https://github.com/axelarnetwork/axelar-amplifier/pull/731/commits/f46f33ab42a3fa5f17417e46e9f65e64d06a8152), in which I moved part of the modules to ampd and renamed `starknet-types` to `starknet-message`, keeping it in `/packages`, because it's used in the prover.

cjcobb23: yeah the prover code should live in your fork, and I'd prefer not to make changes to this repo that only really make sense for your fork. It makes things confusing. Eventually we will refactor ampd as well, so that starknet specific ampd code can also live in your fork as well.

I actually think I prefer the generated types. Can you try that and we see how it looks?

ctoyan: It will require quite a refactor. Is it ok if we leave that for the optimisations phase of the project?

What has to be done to remove the files in the new `/types` folder:
- the [cainome](https://github.com/cartridge-gg/cainome) library will be included as a dependency.
- an ABI json file needs to be kept up to date.
- we pass that file to the `abigen!()` macro.
- it generates **a lot** of stuff in memory, including an RPC client for the contract to which the ABI json corresponds (the gateway contract). In `ampd` we need at most 5% of what it will generate.
- we substitute the `ContractCall` `SignersRotated` and `ByteArray` types with the generated ones.
- remove current implementations and fix tests.

ampd/src/lib.rs
```diff
@@ -40,6 +41,7 @@ mod health_check;
 mod json_rpc;
 mod mvx;
 mod queue;
+pub(crate) mod starknet;
```
cjcobb23, line 44: why `pub(crate`, which is unlike the other modules?

ampd/src/lib.rs
```diff
@@ -389,6 +391,38 @@ where
...
+                    rpc_url,
+                } => self.create_handler_task(
+                    "starknet-msg-verifier",
+                    handlers::starknet_verify_msg::Handler::new(
+                        verifier.clone(),
+                        cosmwasm_contract,
+                        starknet::json_rpc::Client::new_with_transport(HttpTransport::new(
+                            &rpc_url,
+                        ))
+                        .unwrap(),
```
cjcobb23, line 405: should be change_context, not unwrap

ampd/src/lib.rs
```diff
@@ -389,6 +391,38 @@ where
...
+                    rpc_url,
+                } => self.create_handler_task(
+                    "starknet-verifier-set-verifier",
+                    handlers::starknet_verify_verifier_set::Handler::new(
+                        verifier.clone(),
+                        cosmwasm_contract,
+                        starknet::json_rpc::Client::new_with_transport(HttpTransport::new(
+                            &rpc_url,
+                        ))
+                        .unwrap(),
```
cjcobb23, line 421: should be change_context, not unwrap

ampd/src/starknet/json_rpc.rs
```diff
@@ -87,48 +83,32 @@ where
...
 
-        Ok(message_id_and_event_pairs)
+                ContractCallEvent::try_from(event.clone())
+                    .map(|cce| (message_id, cce))
+                    .ok()
+            }
+            TransactionReceipt::L1Handler(_) => None,
+            TransactionReceipt::Declare(_) => None,
+            TransactionReceipt::Deploy(_) => None,
+            TransactionReceipt::DeployAccount(_) => None,
```
cjcobb23, line 110: why not `_ => None`

ctoyan: AFAIK it's a best practice to exhaust all variants. Main reason is that they might add a new `TransactionReceipt` type, which we might want to handle. Applies for other enums, coming from libs, in general.
Apart from that, no other reason. 
I'll do `_ => None` if you prefer.

ampd/src/handlers/starknet_verify_msg.rs
```diff
@@ -0,0 +1,655 @@
...
+            .collect();
+
+        let mut votes = vec![];
+        for msg in unique_msgs {
+            match events.remove(&msg.message_id) {
+                Some(gateway_event) => {
+                    votes.push(verify_msg(&gateway_event, msg, &source_gateway_address));
+                }
+
+                None => votes.push(Vote::NotFound),
```
cjcobb23, line 152: why don't you just move this logic inside of the above map over unique_msgs? And then just generate the corresponding vote immediately after getting the event. There's no need to the `events` collection then. This seems a bit uneccessary to iterate twice like this, and to use mutable collections.

ampd/src/lib.rs
ctoyan: I believe we used it somewhere before, but it doesn't need to be pub(crate) anymore. 
My bad. Removing it.

ampd/src/handlers/starknet_verify_verifier_set.rs
```diff
@@ -0,0 +1,296 @@
...
+        }
+
+        if *self.latest_block_height.borrow() >= expires_at {
+            info!(poll_id = poll_id.to_string(), "skipping expired poll");
+            return Ok(vec![]);
+        }
+
+        let transaction_response = self
+            .rpc_client
+            .get_event_by_hash_signers_rotated(verifier_set.message_id.tx_hash.clone())
```
cjcobb23, line 128: correct me if I'm wrong, but it looks like the event index of the message id is being discarded? So one could put any value there, as long as the tx hash is correct. Is that right? If so, we should change that and also verify the event index

ampd/src/handlers/starknet_verify_msg.rs
ctoyan: It makes sense, yes. I mechanically changed the previous one.
It this better? ([commit](https://github.com/axelarnetwork/axelar-amplifier/pull/731/commits/3aee888d2dbba763b8e8b327490402723eef2421))

ampd/src/handlers/starknet_verify_verifier_set.rs
ctoyan: We reworked it to also accept the `message_id` and get the event by it's index.
You can check this [commit](https://github.com/axelarnetwork/axelar-amplifier/pull/731/commits/dcbb6d1883b192b65ed9ae7d500a77c880c7ff9a) for the change.

packages/axelar-wasm-std/Cargo.toml
```diff
@@ -49,6 +49,7 @@ serde_with = { version = "3.11.0", features = ["macros"] }
 sha3 = { workspace = true }
 starknet-checked-felt = { workspace = true }
 stellar-xdr = { workspace = true }
+starknet-types-core = { workspace = true }
```
cjcobb23, line 52: do you need this here?

cjcobb23 (approved):

packages/axelar-wasm-std/Cargo.toml
ctoyan: No. Removed it.

PR #743, opened 2025-01-10
refactor(minor-voting-verifier): integrate the voting-verifier with the into-event macro

fish-sammy:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues


contracts/voting-verifier/src/events.rs
```diff
@@ -635,6 +494,7 @@ mod test {
                 None,
             ],
         }
+        .non_generic()
         .into();
 
         goldie::assert_json!(json!({
```
cgorenflo, line 500: can you make sure that we're exhaustively testing all events? So the test should break if we add a new event variant without adding it here

cgorenflo (commented):
Discussion: Should this be a major change? We are breaking compatibility with older ampd versions

contracts/voting-verifier/src/events.rs
fish-sammy: Not sure how you can do that, but it's not necessary since this test is there to make sure the differences in the JSON before and after using the macro are OK. With new events, do we need them to be added here?

kyanokashi:
@cgorenflo this should be good to merge, we support this on axelarons

cgorenflo:
taken over by #952

PR #746, opened 2025-01-20
feat: add macro to inject contract version validation and setting the new version during migration

eguajardo:
Example usage:
```Rust
#[migrate_from_version(">= 1.1.0, < 1.2.0")]
pub fn migrate(
    deps: DepsMut,
    _env: Env,
    msg: MigrateMsg,
    v1_1_1::migrate(deps.storage, msg.chains_to_remove)?;
    Ok(Response::default())
}
```
which expands to:
```Rust
pub fn migrate(
    deps: DepsMut,
    _env: Env,
    msg: MigrateMsg,
    let contract_name = "router";
    let contract_version = "1.2.0";
    let old_version = semver::Version::parse(
        &cw2::get_contract_version(deps.storage)?.version,
    )?;
    let version_requirement = semver::VersionReq::parse(">= 1.1.0, < 1.2.0")?;
    if !version_requirement.matches(&old_version) {
        {
            ::core::panicking::panic_fmt(
                format_args!(
                    "{0} contract migration from version {1} to {2} is not supported",
                    contract_name,
                    old_version,
                    contract_version,
                ),
            );
        }
    }
    let result = (|| {
        {
            v1_1_1::migrate(deps.storage, msg.chains_to_remove)?;
            Ok(Response::default())
        }
    })();
    cw2::set_contract_version(deps.storage, contract_name, contract_version)?;
    result
}
```

https://axelarnetwork.atlassian.net/browse/AXE-4947

### DIFF
--- a/packages/axelar-wasm-std-derive/src/lib.rs
+++ b/packages/axelar-wasm-std-derive/src/lib.rs
@@ -309,15 +309,18 @@ pub fn migrate_from_version(input: TokenStream, item: TokenStream) -> TokenStrea
 
     let gen = quote! {
         pub fn #fn_name(#fn_inputs) #fn_output {
+            let contract_name = env!("CARGO_PKG_NAME");
+            let contract_version = env!("CARGO_PKG_VERSION");
+
             let old_version = semver::Version::parse(&cw2::get_contract_version(#deps.storage)?.version)?;
             let version_requirement = semver::VersionReq::parse(#base_version_req)?;
-            assert!(version_requirement.matches(&old_version));
+            assert!(version_requirement.matches(&old_version), "{} contract migration from version {} to {} is not supported", contract_name, old_version, contract_version);
 
             let result = (|| {
                 #fn_block
             })();
 
-            cw2::set_contract_version(#deps.storage, env!("CARGO_PKG_NAME"), env!("CARGO_PKG_VERSION"))?;
+            cw2::set_contract_version(#deps.storage, contract_name, contract_version)?;
 
             result
         }
--- a/packages/axelar-wasm-std-derive/src/lib.rs
+++ b/packages/axelar-wasm-std-derive/src/lib.rs
@@ -282,3 +282,68 @@ fn match_unit_variant(event_enum: &Ident, variant_name: &Ident) -> TokenStream2 
         #event_enum::#variant_name => cosmwasm_std::Event::new(#event_name)
     }
 }
+
+#[proc_macro_attribute]
+pub fn migrate_from_version(input: TokenStream, item: TokenStream) -> TokenStream {
+    let base_version_req = syn::parse_macro_input!(input as syn::LitStr).value();
+    let annotated_fn = syn::parse_macro_input!(item as syn::ItemFn);
+
+    let fn_name = &annotated_fn.sig.ident;
+    let fn_inputs = &annotated_fn.sig.inputs;
+    let fn_output = &annotated_fn.sig.output;
+    let fn_block = &annotated_fn.block;
+
+    if fn_name != "migrate" {
+        return syn::Error::new(
+            fn_name.span(),
+            "#[migrate_from_version] can only be applied to a 'migrate' function",
+        )
+        .to_compile_error()
+        .into();
+    }
+
+    let deps = match deps_ident(&annotated_fn.sig) {
+        Ok(deps) => deps,
+        Err(e) => return e.to_compile_error().into(),
+    };
+
+    let gen = quote! {
+        pub fn #fn_name(#fn_inputs) #fn_output {
+            let old_version = semver::Version::parse(&cw2::get_contract_version(#deps.storage)?.version)?;
+            let version_requirement = semver::VersionReq::parse(#base_version_req)?;
+            assert!(version_requirement.matches(&old_version));
+
+            let result = (|| {
+                #fn_block
+            })();
+
+            cw2::set_contract_version(#deps.storage, env!("CARGO_PKG_NAME"), env!("CARGO_PKG_VERSION"))?;
+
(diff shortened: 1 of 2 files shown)


packages/axelar-wasm-std-derive/src/lib.rs
```diff
@@ -282,3 +282,71 @@ fn match_unit_variant(event_enum: &Ident, variant_name: &Ident) -> TokenStream2
...
+            let version_requirement = semver::VersionReq::parse(#base_version_req)?;
+            assert!(version_requirement.matches(&old_version), "{} contract migration from version {} to {} is not supported", contract_name, old_version, contract_version);
+
+            let result = (|| {
+                #fn_block
+            })();
+
+            cw2::set_contract_version(#deps.storage, contract_name, contract_version)?;
+
+            result
```
fish-sammy, line 325:
```suggestion
            cw2::set_contract_version(#deps.storage, contract_name, contract_version)?;

            #fn_block
```
I see no point of executing the fn body itself as a closure call.

packages/axelar-wasm-std-derive/src/lib.rs
```diff
@@ -282,3 +282,71 @@ fn match_unit_variant(event_enum: &Ident, variant_name: &Ident) -> TokenStream2
...
+                    return Ok(pat_ident.ident.clone());
+                }
+            }
+        }
+    }
+
+    Err(syn::Error::new(
+        sig.ident.span(),
+        "first parameter must be of type DepsMut",
+    ))
```
fish-sammy, line 351:
```suggestion
    let (ty, pat) = match first_param {
        FnArg::Typed(syn::PatType { ty, pat, .. }) => (ty, pat),
        _ => {
            return Err(syn::Error::new(
                sig.ident.span(),
                "first parameter must be of type DepsMut",
            ))
        }
    };

    match (&**ty, &**pat) {
        (syn::Type::Path(syn::TypePath { path, .. }), syn::Pat::Ident(pat_ident))
            if path.is_ident("DepsMut") =>
        {
            Ok(pat_ident.ident.clone())
        }
        _ => Err(syn::Error::new(
            sig.ident.span(),
            "first parameter must be of type DepsMut",
        )),
    }
```
It's longer, but much easier to read than nested-ifs.

packages/axelar-wasm-std-derive/src/lib.rs
```diff
@@ -282,3 +282,71 @@ fn match_unit_variant(event_enum: &Ident, variant_name: &Ident) -> TokenStream2
...
+    if fn_name != "migrate" {
+        return syn::Error::new(
+            fn_name.span(),
+            "#[migrate_from_version] can only be applied to a 'migrate' function",
+        )
+        .to_compile_error()
+        .into();
+    }
+
+    let deps = match deps_ident(&annotated_fn.sig) {
```
fish-sammy, line 305: If `DepsMut` being the first param is so important, why not also check the other params to make sure the migrate function matches the right signature?

packages/axelar-wasm-std-derive/src/lib.rs
```diff
@@ -282,3 +282,71 @@ fn match_unit_variant(event_enum: &Ident, variant_name: &Ident) -> TokenStream2
         #event_enum::#variant_name => cosmwasm_std::Event::new(#event_name)
     }
 }
+
+#[proc_macro_attribute]
+pub fn migrate_from_version(input: TokenStream, item: TokenStream) -> TokenStream {
```
fish-sammy, line 287: Needs to be tested. Doctest or unit test.

packages/axelar-wasm-std-derive/src/lib.rs
eguajardo: I used the closure because that way I can execute the function body before setting the new version. We used to set the version at the end because sometimes we would have some others checks or tests inside the migration module that would depend on the previous version. Right now I'm not sure we will have something like that, but I still think it would be a good practice to leave setting the new version at the end of the execution, what do you think?

fish-sammy (changes requested):

packages/axelar-wasm-std-derive/src/lib.rs
```diff
@@ -282,3 +282,71 @@ fn match_unit_variant(event_enum: &Ident, variant_name: &Ident) -> TokenStream2
...
+    let base_version_req = syn::parse_macro_input!(input as syn::LitStr).value();
```
fish-sammy, line 288: I'd much prefer parsing two separate inputs here `from_version` and `to_version` (both have to specify only major and minor) and forming the requirement here instead of having to handwrite that requirement myself.

packages/axelar-wasm-std-derive/src/lib.rs
```diff
@@ -282,3 +283,162 @@ fn match_unit_variant(event_enum: &Ident, variant_name: &Ident) -> TokenStream2
...
+/// #[migrate_from_version("1.1")]
+/// pub fn migrate(
+///     deps: DepsMut,
+///     _env: Env,
+///     _msg: Empty,
+/// ) -> Result<Response, axelar_wasm_std::error::ContractError> {
+///     // migration logic
+///     Ok(Response::default())
+/// }
+/// ```
```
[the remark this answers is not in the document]
eguajardo: I made this intentionally, I wanted to remove clutter from here and leave just an example, the actual testing (outside of compilation errors) was done inside the `test` folder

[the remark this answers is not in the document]
fish-sammy: I prefer having the happy-path test as a doctest, but no strong opinion.

fish-sammy (approved):

PR #747, opened 2025-01-22
fix(minor-interchain-token-service): custom token reregistration

cjcobb23:
allow reregistration of custom token with same decimals

### DIFF
--- a/contracts/interchain-token-service/src/contract/execute/interceptors.rs
+++ b/contracts/interchain-token-service/src/contract/execute/interceptors.rs
@@ -299,10 +299,17 @@ pub fn register_custom_token(
         register_token.token_address.clone(),
     )
     .change_context(Error::State)?;
-    ensure!(
-        existing_token.is_none(),
-        Error::TokenAlreadyRegistered(register_token.token_address)
-    );
+
+    if let Some(registered_token) = existing_token {
+        ensure!(
+            registered_token.decimals == register_token.decimals,
+            Error::TokenDecimalsMismatch {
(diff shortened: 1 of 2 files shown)


contracts/interchain-token-service/src/contract/execute/interceptors.rs
```diff
@@ -299,10 +299,17 @@ pub fn register_custom_token(
...
     )
     .change_context(Error::State)?;
-    ensure!(
-        existing_token.is_none(),
-        Error::TokenAlreadyRegistered(register_token.token_address)
-    );
+
+    if let Some(registered_token) = existing_token {
+        ensure!(
+            registered_token.decimals == register_token.decimals,
```
milapsheth, line 305:
```suggestion
    if let Some(existing_token) = existing_token {
        ensure!(
            existing_token == register_token,
```
check the full struct for in case other fields are added later?

contracts/interchain-token-service/src/contract/execute/interceptors.rs
```diff
@@ -299,10 +299,17 @@ pub fn register_custom_token(
...
+            Error::TokenDecimalsMismatch {
+                token_address: register_token.token_address,
+                existing_decimals: registered_token.decimals,
+                new_decimals: register_token.decimals
+            }
+        );
+    }
 
     state::save_custom_token_metadata(storage, source_chain, register_token)
```
milapsheth, line 314: avoid saving if it's already present?

milapsheth (approved):

PR #748, opened 2025-01-27
feat(minor-interchain-token-service): extend its address query with chain config info

maancham:
## Description

AXE-7169

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

### DIFF
--- a/contracts/interchain-token-service/src/contract.rs
+++ b/contracts/interchain-token-service/src/contract.rs
@@ -39,8 +39,8 @@ pub enum Error {
     DisableExecution,
     #[error("failed to enable execution")]
     EnableExecution,
-    #[error("failed to query its address")]
-    QueryItsContract,
+    #[error("failed to query chain config")]
+    QueryChainConfig,
     #[error("failed to query all its addresses")]
     QueryAllItsContracts,
     #[error("failed to query a specific token instance")]
@@ -130,7 +130,7 @@ fn match_gateway(storage: &dyn Storage, _: &ExecuteMsg) -> Result<Addr, Report<E
 pub fn query(deps: Deps, _: Env, msg: QueryMsg) -> Result<Binary, ContractError> {
     match msg {
         QueryMsg::ChainConfig { chain } => {
-            query::chain_config(deps, chain).change_context(Error::QueryItsContract)
+            query::chain_config(deps, chain).change_context(Error::QueryChainConfig)
         }
         QueryMsg::AllItsContracts => {
             query::all_its_contracts(deps).change_context(Error::QueryAllItsContracts)
--- a/contracts/interchain-token-service/src/contract.rs
+++ b/contracts/interchain-token-service/src/contract.rs
@@ -129,8 +129,8 @@ fn match_gateway(storage: &dyn Storage, _: &ExecuteMsg) -> Result<Addr, Report<E
 #[cfg_attr(not(feature = "library"), entry_point)]
 pub fn query(deps: Deps, _: Env, msg: QueryMsg) -> Result<Binary, ContractError> {
     match msg {
-        QueryMsg::ChainConfig { chain } => {
-            query::chain_config(deps, chain).change_context(Error::QueryChainConfig)
+        QueryMsg::ITSChain { chain } => {
+            query::its_chain(deps, chain).change_context(Error::QueryChainConfig)
         }
         QueryMsg::AllItsContracts => {
             query::all_its_contracts(deps).change_context(Error::QueryAllItsContracts)
--- a/contracts/interchain-token-service/src/contract.rs
+++ b/contracts/interchain-token-service/src/contract.rs
@@ -129,8 +129,8 @@ fn match_gateway(storage: &dyn Storage, _: &ExecuteMsg) -> Result<Addr, Report<E
 #[cfg_attr(not(feature = "library"), entry_point)]
 pub fn query(deps: Deps, _: Env, msg: QueryMsg) -> Result<Binary, ContractError> {
     match msg {
-        QueryMsg::ItsContract { chain } => {
-            query::its_contract(deps, chain).change_context(Error::QueryItsContract)
+        QueryMsg::ChainConfig { chain } => {
+            query::chain_config(deps, chain).change_context(Error::QueryItsContract)
         }
         QueryMsg::AllItsContracts => {
             query::all_its_contracts(deps).change_context(Error::QueryAllItsContracts)
--- a/contracts/interchain-token-service/src/contract/query.rs
+++ b/contracts/interchain-token-service/src/contract/query.rs
@@ -3,11 +3,8 @@ use cosmwasm_std::{to_json_binary, Binary, Deps};
 use error_stack::{Result, ResultExt};
 use router_api::ChainNameRaw;
 
-use crate::msg::{ChainConfigResponse, TruncationConfig as MsgTruncationConfig};
-use crate::state::{
-    load_all_its_contracts, may_load_chain_config, may_load_token_config, may_load_token_instance,
-    ChainConfig as StateChainConfig,
-};
+use crate::msg;
+use crate::state;
 use crate::TokenId;
 
 #[derive(thiserror::Error, Debug, IntoContractError)]
@@ -19,12 +16,12 @@ pub enum Error {
 }
 
 pub fn chain_config(deps: Deps, chain: ChainNameRaw) -> Result<Binary, Error> {
-    let state_config: Option<StateChainConfig> =
-        may_load_chain_config(deps.storage, &chain).change_context(Error::State)?;
-    to_json_binary(&state_config.map(|config| ChainConfigResponse {
+    let state_config: Option<state::ChainConfig> =
+        state::may_load_chain_config(deps.storage, &chain).change_context(Error::State)?;
+    to_json_binary(&state_config.map(|config| msg::ChainConfigResponse {
         chain,
         its_edge_contract: config.its_address,
-        truncation: MsgTruncationConfig {
+        truncation: msg::TruncationConfig {
             max_uint: config.truncation.max_uint,
             max_decimals_when_truncating: config.truncation.max_decimals_when_truncating,
         },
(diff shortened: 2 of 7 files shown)


contracts/interchain-token-service/src/contract/query.rs
```diff
@@ -16,10 +18,18 @@ pub enum Error {
     State,
 }
 
-pub fn its_contract(deps: Deps, chain: ChainNameRaw) -> Result<Binary, Error> {
-    let contract_address =
-        may_load_its_contract(deps.storage, &chain).change_context(Error::State)?;
-    to_json_binary(&contract_address).change_context(Error::JsonSerialization)
+pub fn chain_config(deps: Deps, chain: ChainNameRaw) -> Result<Binary, Error> {
```
milapsheth, line 21: I'd rather expose the state chain config or a similar type, since it also contains the freeze status of the chain. Otherwise we'd need to add a separate query for that

milapsheth: I guess a separate freeze status query might make more sense since it should also take into account the killswitch execution status?

fish-sammy:
It shouldn't be a minor for any API changes unless the query is brand-new.

contracts/interchain-token-service/src/contract/query.rs
```diff
@@ -3,8 +3,10 @@ use cosmwasm_std::{to_json_binary, Binary, Deps};
 use error_stack::{Result, ResultExt};
 use router_api::ChainNameRaw;
 
+use crate::msg::{ChainConfig as MsgChainConfig, TruncationConfig as MsgTruncationConfig};
```
cgorenflo, line 6: is there another TruncationConfig this could get confused with?

cgorenflo (approved):

cgorenflo:
> It shouldn't be a minor for any API changes unless the query is brand-new.

It's effectively removing the old and creating a new query, so I think using a minor uptick makes sense

contracts/interchain-token-service/src/contract/query.rs
maancham: yes. We have this struct over `state.rs` file:

```
pub struct TruncationConfig {
    pub max_uint: nonempty::Uint256, // The maximum uint value that is supported by the chain's token standard
    pub max_decimals_when_truncating: u8, // The maximum number of decimals that is preserved when deploying from a chain with a larger max_uint
}
```

contracts/interchain-token-service/src/contract/query.rs
maancham: I am going to add a query to check if contract killswitch is engaged or not (`IsEnabled`). Also, I will add `AllFrozenChains` and `AllActiveChains` queries as well. They both return `#[returns(Vec<ChainNameRaw>)]`.

fish-sammy:
> > It shouldn't be a minor for any API changes unless the query is brand-new.
> 
> It's effectively removing the old and creating a new query, so I think using a minor uptick makes sense

@cgorenflo Removal is major though. Deprecation is minor.

contracts/interchain-token-service/src/contract/query.rs
maancham: Added `ChainConfigResponse` and included `frozen` status

contracts/interchain-token-service/src/contract/query.rs
```diff
@@ -16,10 +18,19 @@ pub enum Error {
...
+        may_load_chain_config(deps.storage, &chain).change_context(Error::State)?;
+    to_json_binary(&state_config.map(|config| ChainConfigResponse {
+        chain,
+        its_edge_contract: config.its_address,
+        truncation: MsgTruncationConfig {
+            max_uint: config.truncation.max_uint,
+            max_decimals_when_truncating: config.truncation.max_decimals_when_truncating,
+        },
+        frozen: config.frozen,
+    }))
```
cgorenflo, line 32: question for the team: should we implement the `From` trait for such conversions?

maancham: @cjcobb23 @milapsheth @fish-sammy

contracts/interchain-token-service/src/contract/query.rs
```diff
@@ -3,8 +3,10 @@ use cosmwasm_std::{to_json_binary, Binary, Deps};
 use error_stack::{Result, ResultExt};
 use router_api::ChainNameRaw;
 
+use crate::msg::{ChainConfigResponse, TruncationConfig as MsgTruncationConfig};
```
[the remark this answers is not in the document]
fish-sammy: Agree with Chris

contracts/interchain-token-service/src/contract/query.rs
```diff
@@ -3,8 +3,10 @@ use cosmwasm_std::{to_json_binary, Binary, Deps};
...
 use crate::state::{
-    load_all_its_contracts, may_load_its_contract, may_load_token_config, may_load_token_instance,
+    load_all_its_contracts, may_load_chain_config, may_load_token_config, may_load_token_instance,
+    ChainConfig as StateChainConfig,
```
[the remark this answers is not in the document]
fish-sammy: Agree with Chris

contracts/interchain-token-service/src/contract/query.rs
fish-sammy: I don't have a strong opinion on this.

contracts/interchain-token-service/src/contract.rs
```diff
@@ -129,8 +129,8 @@ fn match_gateway(storage: &dyn Storage, _: &ExecuteMsg) -> Result<Addr, Report<E
 #[cfg_attr(not(feature = "library"), entry_point)]
 pub fn query(deps: Deps, _: Env, msg: QueryMsg) -> Result<Binary, ContractError> {
     match msg {
-        QueryMsg::ItsContract { chain } => {
-            query::its_contract(deps, chain).change_context(Error::QueryItsContract)
+        QueryMsg::ChainConfig { chain } => {
+            query::chain_config(deps, chain).change_context(Error::QueryItsContract)
```
fish-sammy, line 133:
```suggestion
            query::chain_config(deps, chain).change_context(Error::QueryChainConfig)
```
Rename the error

contracts/interchain-token-service/src/contract/query.rs
milapsheth: I like using From trait since it'll be a common pattern for queries returning a state value. Makes the queries more readable (e.g. easy to see that it's just converting a state value and nothing more complex)

contracts/interchain-token-service/src/msg.rs
```diff
@@ -66,21 +66,32 @@ pub struct TruncationConfig {
     pub max_decimals_when_truncating: u8, // The maximum number of decimals that is preserved when deploying from a chain with a larger max_uint
 }
 
+#[cw_serde]
+pub struct ChainConfigResponse {
+    pub chain: ChainNameRaw,
+    pub its_edge_contract: Address,
+    pub truncation: TruncationConfig,
+    pub frozen: bool,
```
cjcobb23, line 74: I actually think the presence of the frozen field makes this not really a config, since a config is usually more of a static thing. I'd prefer to call the query just `Chain` or `ITSChain` or something like that

cjcobb23: plus in the future we might want to add more data to this response that could be useful, in which case it may drift even further from a config

maancham: good call. I think `ITSChain` is better

contracts/interchain-token-service/src/state.rs
```diff
@@ -36,7 +36,7 @@ pub struct Config {
 pub struct ChainConfig {
     pub truncation: TruncationConfig,
     pub its_address: Address,
-    frozen: bool,
+    pub frozen: bool,
```
cjcobb23, line 39: would prefer not to change this to `pub`. It was meant to be private so only the state module can mutate

contracts/interchain-token-service/src/msg.rs
```diff
@@ -66,21 +66,32 @@ pub struct TruncationConfig {
...
+}
+
 #[cw_serde]
 #[derive(QueryResponses)]
 pub enum QueryMsg {
-    /// Query the ITS contract address registered for a chain
-    #[returns(Option<Address>)]
-    ItsContract { chain: ChainNameRaw },
```
cjcobb23, line 74: why are we removing this? Technically this is an API breaking change, and would usually denote a new major release, since it would break any clients that rely on this. Do we know if anything is relying on this right now?

contracts/interchain-token-service/tests/utils/query.rs
```diff
@@ -43,3 +43,33 @@ pub fn query_token_config(
...
+        queried.truncation.max_decimals_when_truncating,
+        "Max decimals mismatch. Expected: {:?}, Got: {:?}",
+        original.truncation.max_decimals_when_truncating,
+        queried.truncation.max_decimals_when_truncating
+    );
+    assert_eq!(
+        original.frozen, queried.frozen,
+        "Frozen status mismatch. Expected: {:?}, Got: {:?}",
+        original.frozen, queried.frozen
+    );
```
cjcobb23, line 74: why do you need to compare the fields individually, instead of just `assert_eq!(original, queried)`. Also, do we need to specify the error message?

contracts/interchain-token-service/src/msg.rs
maancham: Since ITS is not launched yet, it's just being used by ChainOps. I ran the changes by Ayush

contracts/interchain-token-service/src/state.rs
maancham: I need to make it public to be able to create the `msg::ChainConfig` object from `state::ChainConfig` over at query file like this:

```
pub fn its_chain(deps: Deps, chain: ChainNameRaw) -> Result<Binary, Error> {
    let state_config: Option<state::ChainConfig> =
        state::may_load_chain_config(deps.storage, &chain).change_context(Error::State)?;
    to_json_binary(&state_config.map(|config| msg::ChainConfigResponse {
        chain,
        its_edge_contract: config.its_address,
        truncation: msg::TruncationConfig {
            max_uint: config.truncation.max_uint,
            max_decimals_when_truncating: config.truncation.max_decimals_when_truncating,
        },
        frozen: config.frozen,
    }))
    .change_context(Error::JsonSerialization)
}
```

contracts/interchain-token-service/tests/utils/query.rs
maancham: I wanted it to be thorough and if only one field is different, just be able to get a readable message

contracts/interchain-token-service/src/state.rs
cjcobb23: if you implement `From` you don't. You could also define a getter. That's how it's read in other modules right now, via a getter

contracts/interchain-token-service/tests/utils/query.rs
cjcobb23: I think this is unnecessary. Pretty sure it will spit out the whole object if there is a mismatch, and it's fine for the programmer to take a minute or two to debug if there is a mismatch. The test will pass unless the code is broken, I don't think we need to add extra code just to help someone debug a broken unit test.

contracts/interchain-token-service/src/msg.rs
cgorenflo: `ChainState`?

contracts/interchain-token-service/src/state.rs
cgorenflo: Houmaan and I talked about it. It needs to be public because the query needs to map it to the response type's field. Because it's a boolean that doesn't depend on any other logic to be set I'm ok making it public instead of creating a getter function. It shouldn't be exported outside the crate though (i.e. make sure the `state` module is not exported from the `lib.rs`)

contracts/interchain-token-service/tests/utils/query.rs
cgorenflo: should we replace this with a goldie test? That would also catch a field addition, which would be missed by a comparison test

maancham: Removing this utility function for now as Its only being used in one place and also CJ has a point. @cgorenflo let me know if you want to convert to a goldie test later.

PR #753, opened 2025-02-05
feat: add contracts migration remover tool

eguajardo:
This tool assumes the contrac's migration code follows the same directory structure as the router contract on this PR, i.e. all migration-related code must be inside `src/contract/migrations`

More info in README

How to test:
```
cargo run --bin migration-remover -- -c router
```

https://axelarnetwork.atlassian.net/browse/AXE-7389

### DIFF
--- a/contracts/router/src/contract.rs
+++ b/contracts/router/src/contract.rs
@@ -12,7 +12,7 @@ use crate::state;
 use crate::state::{load_chain_by_gateway, load_config, Config};
 
 mod execute;
-mod migrations;
+pub(crate) mod migrations;
 mod query;
 
 pub use migrations::migrate;
--- a/contracts/router/src/contract.rs
+++ b/contracts/router/src/contract.rs
@@ -12,10 +12,10 @@ use crate::state;
 use crate::state::{load_chain_by_gateway, load_config, Config};
 
 mod execute;
-pub(crate) mod migrations;
+mod migrations;
 mod query;
 
-pub use migrations::migrate;
+pub use migrations::{migrate, MigrateMsg};
 
 pub const CONTRACT_NAME: &str = env!("CARGO_PKG_NAME");
 pub const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
--- a/contracts/router/src/contract.rs
+++ b/contracts/router/src/contract.rs
@@ -1,4 +1,4 @@
-use axelar_wasm_std::{address, killswitch, migrate_from_version, permission_control, FnExt};
+use axelar_wasm_std::{address, killswitch, permission_control, FnExt};
 #[cfg(not(feature = "library"))]
 use cosmwasm_std::entry_point;
 use cosmwasm_std::{
@@ -6,9 +6,8 @@ use cosmwasm_std::{
 };
 use router_api::error::Error;
 
-use crate::contract::migrations::v1_1_1;
 use crate::events::RouterInstantiated;
-use crate::msg::{ExecuteMsg, InstantiateMsg, MigrateMsg, QueryMsg};
+use crate::msg::{ExecuteMsg, InstantiateMsg, QueryMsg};
 use crate::state;
 use crate::state::{load_chain_by_gateway, load_config, Config};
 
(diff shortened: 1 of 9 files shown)


tools/migration-remover/src/template.rs
```diff
@@ -0,0 +1,14 @@
...
+
+#[cfg_attr(not(feature = "library"), entry_point)]
+#[migrate_from_version("0.0")]
+pub fn migrate(
+    deps: DepsMut,
+    _env: Env,
+    _msg: Empty,
+) -> Result<Response, axelar_wasm_std::error::ContractError> {
+    Ok(Response::default())
+}
```
cjcobb23, line 14: don't we still want to set the version?

eguajardo: its done with the macro `#[migrate_from_version("0.0")]`, the actual version is replaced from contract's cargo.toml

contracts/router/src/msg.rs
```diff
@@ -10,10 +10,5 @@ pub struct InstantiateMsg {
     pub axelarnet_gateway: String,
 }
 
-#[cw_serde]
-pub struct MigrateMsg {
-    pub chains_to_remove: Vec<String>,
-}
-
 // these messages are extracted into a separate package to avoid circular dependencies
 pub use router_api::msg::{ExecuteMsg, QueryMsg};
```
cgorenflo, line 14: add a re-export for the migration msg here

cgorenflo: is that maintainable with the tool?

tools/migration-remover/src/main.rs
```diff
@@ -0,0 +1,84 @@
...
+        .replace(
+            r#"#[migrate_from_version("0.0")]"#,
+            &format!(r#"#[migrate_from_version("{}")]"#, from_version),
+        );
+
+    Ok(template)
+}
+
+fn reset_migration_mod(contract_root_dir: &Path, content: &str) -> Result<(), Error> {
+    let migrations_mod_dir = contract_root_dir.join("src/contract/migrations");
```
cgorenflo, line 56: make this relative path a const as well?

tools/migration-remover/src/template.rs
cgorenflo, line 1: to make the re-export in the msg.rs possible, add a MigrateMsg enum without variants?

eguajardo: tried to re-export `MigrateMsg`, but the compiler will warn of unused code and I'm not sure is a good idea to add `#[warn(dead_code)]` to the template

eguajardo: used `pub type MigrateMsg = Empty;` instead and used it in `migrate` function signature to remove dead code warning. I had to add re-exports in tool's `main.rs` gated by `#[cfg(test)]` to remove dead code warning in `template.rs` as well. Let me know what you think

tools/migration-remover/src/main.rs
```diff
@@ -0,0 +1,89 @@
...
+    Ok(())
+}
+
+// tests that `template.rs` compiles correctly and content can be re-exported
+#[cfg(test)]
+mod template;
+#[cfg(test)]
+pub use template::migrate;
+#[cfg(test)]
+pub use template::MigrateMsg;
```
cgorenflo, line 89: add a comment that these exports are used to make the dead code warning go away?

contracts/router/src/contract.rs
```diff
@@ -1,35 +1,25 @@
...
-use crate::contract::migrations::v1_1_1;
 use crate::events::RouterInstantiated;
-use crate::msg::{ExecuteMsg, InstantiateMsg, MigrateMsg, QueryMsg};
+use crate::msg::{ExecuteMsg, InstantiateMsg, QueryMsg};
 use crate::state;
 use crate::state::{load_chain_by_gateway, load_config, Config};
 
 mod execute;
-mod migrations;
+pub(crate) mod migrations;
```
cgorenflo, line 15: why does this need to be pub(crate)?

cgorenflo: can it be restricted to this?
```suggestion
mod migrations;

pub(crate) use migrations::MigrateMsg
```

eguajardo: it needs to be pub so that it can be re-exported in `msg.rs`

PR #754, opened 2025-02-05
feat(minor-interchain-token-service): all chains with optional filter query message 

maancham:
## Description
AXE-7369

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros

### DIFF
--- a/contracts/interchain-token-service/src/contract.rs
+++ b/contracts/interchain-token-service/src/contract.rs
@@ -139,9 +139,12 @@ pub fn query(deps: Deps, _: Env, msg: QueryMsg) -> Result<Binary, ContractError>
         QueryMsg::AllItsContracts => {
             query::all_its_contracts(deps).change_context(Error::QueryAllItsContracts)
         }
-        QueryMsg::ItsChains { filter } => {
-            query::its_chains(deps, filter).change_context(Error::QueryAllChainConfigs)
-        }
+        QueryMsg::ItsChains {
+            filter,
+            start_after,
+            limit,
+        } => query::its_chains(deps, filter, start_after, limit)
+            .change_context(Error::QueryAllChainConfigs),
         QueryMsg::TokenInstance { chain, token_id } => {
             query::token_instance(deps, chain, token_id).change_context(Error::QueryTokenInstance)
         }
--- a/contracts/interchain-token-service/src/contract.rs
+++ b/contracts/interchain-token-service/src/contract.rs
@@ -49,6 +49,8 @@ pub enum Error {
     QueryTokenConfig,
     #[error("failed to query the status of contract")]
     QueryContractStatus,
+    #[error("failed to query chain configs")]
+    QueryAllChainConfigs,
 }
 
 #[cfg_attr(not(feature = "library"), entry_point)]
@@ -136,6 +138,9 @@ pub fn query(deps: Deps, _: Env, msg: QueryMsg) -> Result<Binary, ContractError>
         }
         QueryMsg::AllItsContracts => {
             query::all_its_contracts(deps).change_context(Error::QueryAllItsContracts)
+        }
+        QueryMsg::ItsChains { filter } => {
+            query::its_chains(deps, filter).change_context(Error::QueryAllChainConfigs)
         }
         QueryMsg::TokenInstance { chain, token_id } => {
             query::token_instance(deps, chain, token_id).change_context(Error::QueryTokenInstance)
--- a/contracts/interchain-token-service/src/contract/query.rs
+++ b/contracts/interchain-token-service/src/contract/query.rs
@@ -1,14 +1,10 @@
 use axelar_wasm_std::{killswitch, IntoContractError};
-use cosmwasm_std::{to_json_binary, Binary, Deps, Order};
-use cw_storage_plus::Bound;
+use cosmwasm_std::{to_json_binary, Binary, Deps};
 use error_stack::{Result, ResultExt};
 use itertools::Itertools;
 use router_api::ChainNameRaw;
 
 use crate::{msg, state, TokenId};
-
-// Pagination limit
-const DEFAULT_LIMIT: u32 = u32::MAX;
 
 #[derive(thiserror::Error, Debug, IntoContractError)]
 pub enum Error {
@@ -45,26 +41,29 @@ pub fn its_chains(
     start_after: Option<ChainNameRaw>,
     limit: Option<u32>,
 ) -> Result<Binary, Error> {
-    let state_chain_configs = state::load_chain_configs();
-
-    let limit = limit.unwrap_or(DEFAULT_LIMIT) as usize;
-    let start = start_after.as_ref().map(Bound::exclusive);
-
-    let filtered_chain_configs: Vec<_> = state_chain_configs
-        .range(deps.storage, start, None, Order::Ascending)
-        .map(|r| r.change_context(Error::State))
-        .map_ok(|(chain, config)| msg::ChainConfigResponse {
-            chain,
-            its_edge_contract: config.its_address,
-            truncation: msg::TruncationConfig {
-                max_uint: config.truncation.max_uint,
-                max_decimals_when_truncating: config.truncation.max_decimals_when_truncating,
+    let filter_fn = |config: &state::ChainConfig| {
+        filter.as_ref().map_or(true, |f| match &f.status {
+            Some(status) => match status {
+                msg::ChainStatusFilter::Frozen => config.frozen,
+                msg::ChainStatusFilter::Active => !config.frozen,
             },
-            frozen: config.frozen,
+            None => true,
         })
-        .filter_ok(|config| !filter.clone().is_some_and(|f| !f.matches(config)))
-        .take(limit)
-        .try_collect()?;
+    };
+
+    let filtered_chain_configs: Vec<_> =
+        state::load_chain_configs(deps.storage, filter_fn, start_after, limit)
+            .map(|r| r.change_context(Error::State))
+            .map_ok(|(chain, config)| msg::ChainConfigResponse {
+                chain,
+                its_edge_contract: config.its_address,
+                truncation: msg::TruncationConfig {
+                    max_uint: config.truncation.max_uint,
+                    max_decimals_when_truncating: config.truncation.max_decimals_when_truncating,
+                },
+                frozen: config.frozen,
+            })
+            .try_collect()?;
 
     to_json_binary(&filtered_chain_configs).change_context(Error::JsonSerialization)
 }
--- a/contracts/interchain-token-service/src/contract/query.rs
+++ b/contracts/interchain-token-service/src/contract/query.rs
@@ -1,10 +1,14 @@
 use axelar_wasm_std::{killswitch, IntoContractError};
 use cosmwasm_std::{to_json_binary, Binary, Deps, Order};
+use cw_storage_plus::Bound;
 use error_stack::{Result, ResultExt};
 use itertools::Itertools;
 use router_api::ChainNameRaw;
 
 use crate::{msg, state, TokenId};
+
+// Pagination limit
+const DEFAULT_LIMIT: u32 = u32::MAX;
 
 #[derive(thiserror::Error, Debug, IntoContractError)]
 pub enum Error {
@@ -35,11 +39,19 @@ pub fn all_its_contracts(deps: Deps) -> Result<Binary, Error> {
     to_json_binary(&contract_addresses).change_context(Error::JsonSerialization)
 }
 
-pub fn its_chains(deps: Deps, filter: Option<msg::ChainFilter>) -> Result<Binary, Error> {
+pub fn its_chains(
+    deps: Deps,
+    filter: Option<msg::ChainFilter>,
+    start_after: Option<ChainNameRaw>,
+    limit: Option<u32>,
+) -> Result<Binary, Error> {
     let state_chain_configs = state::load_chain_configs();
 
-    let chain_config_responses: Vec<msg::ChainConfigResponse> = state_chain_configs
-        .range(deps.storage, None, None, Order::Ascending)
+    let limit = limit.unwrap_or(DEFAULT_LIMIT) as usize;
+    let start = start_after.as_ref().map(Bound::exclusive);
+
+    let filtered_chain_configs: Vec<_> = state_chain_configs
+        .range(deps.storage, start, None, Order::Ascending)
         .map(|r| r.change_context(Error::State))
         .map_ok(|(chain, config)| msg::ChainConfigResponse {
             chain,
@@ -50,15 +62,10 @@ pub fn its_chains(deps: Deps, filter: Option<msg::ChainFilter>) -> Result<Binary
             },
             frozen: config.frozen,
         })
+        .filter_ok(|config| !filter.clone().is_some_and(|f| !f.matches(config)))
+        .take(limit)
         .try_collect()?;
 
-    let filtered_chain_configs = match &filter {
-        Some(filter) => chain_config_responses
-            .into_iter()
-            .filter(|config| filter.matches(config))
-            .collect(),
-        None => chain_config_responses,
-    };
     to_json_binary(&filtered_chain_configs).change_context(Error::JsonSerialization)
 }
 
--- a/contracts/interchain-token-service/src/contract/query.rs
+++ b/contracts/interchain-token-service/src/contract/query.rs
@@ -35,23 +35,10 @@ pub fn all_its_contracts(deps: Deps) -> Result<Binary, Error> {
     to_json_binary(&contract_addresses).change_context(Error::JsonSerialization)
 }
 
-fn convert_chain_filter_option(
-    message_chain_filter: Option<msg::ChainFilter>,
-) -> impl Fn(&state::ChainConfig) -> bool + 'static {
-    move |config| match &message_chain_filter {
-        None => true,
-        Some(filter) => convert_chain_filter(filter)(config),
-    }
-}
-
-fn convert_chain_filter(
-    message_chain_filter: &msg::ChainFilter,
-) -> impl Fn(&state::ChainConfig) -> bool + '_ {
-    move |config| match &message_chain_filter.status {
-        Some(status) => match status {
-            msg::ChainStatusFilter::Frozen => config.frozen,
-            msg::ChainStatusFilter::Active => !config.frozen,
-        },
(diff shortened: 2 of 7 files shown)


contracts/interchain-token-service/src/msg.rs
```diff
@@ -85,6 +96,10 @@ pub enum QueryMsg {
     #[returns(HashMap<ChainNameRaw, Address>)]
     AllItsContracts,
 
+    /// Query all chain configs with optional frozen filter
+    #[returns(Vec<ChainConfigResponse>)]
+    ItsChains { filter: Option<ChainConfigFilter> },
```
cjcobb23, line 101: do we want to paginate this now?

contracts/interchain-token-service/src/msg.rs
```diff
@@ -53,6 +53,17 @@ pub enum ExecuteMsg {
...
 
+#[cw_serde]
+pub enum ChainStatusFilter {
+    Frozen,
+    Active,
+}
+
+#[cw_serde]
+pub struct ChainConfigFilter {
+    pub status: Option<ChainStatusFilter>,
```
cjcobb23, line 64:
```suggestion
    pub frozen_status: Option<ChainStatusFilter>,
```

contracts/interchain-token-service/src/msg.rs
```diff
@@ -53,6 +53,17 @@ pub enum ExecuteMsg {
...
 }
 
+#[cw_serde]
+pub enum ChainStatusFilter {
+    Frozen,
+    Active,
+}
+
+#[cw_serde]
+pub struct ChainConfigFilter {
```
cjcobb23, line 63:
```suggestion
pub struct ChainFilter {
```

might not necessarily be filtering on the chain config, could filter on something else in the future

contracts/interchain-token-service/src/contract/query.rs
```diff
@@ -34,6 +34,12 @@ pub fn all_its_contracts(deps: Deps) -> Result<Binary, Error> {
     to_json_binary(&contract_addresses).change_context(Error::JsonSerialization)
 }
 
+pub fn its_chains(deps: Deps, filter: Option<msg::ChainConfigFilter>) -> Result<Binary, Error> {
+    let chain_configs =
+        state::load_filtered_chain_configs(deps.storage, filter).change_context(Error::State)?;
```
cjcobb23, line 39: I would just filter here and not add the filter logic into the state module. The implementation in the state module loads all of the chains anyways, and then filters

contracts/interchain-token-service/src/msg.rs
maancham: @cgorenflo Do we have an estimate on the rate of chain registration OR the current number of registered chains on ITS?

contracts/interchain-token-service/src/contract/query.rs
```diff
@@ -34,6 +34,30 @@ pub fn all_its_contracts(deps: Deps) -> Result<Binary, Error> {
...
+    to_json_binary(&chain_configs).change_context(Error::JsonSerialization)
+}
+
+fn matches_filter(
+    config: &msg::ChainConfigResponse,
+    status: Option<&msg::ChainStatusFilter>,
+) -> bool {
+    match status {
+        Some(msg::ChainStatusFilter::Frozen) => config.frozen,
+        Some(msg::ChainStatusFilter::Active) => !config.frozen,
```
cgorenflo, line 56: I would suggest to implement a `matches` function directly on the ChainStatusFilter

contracts/interchain-token-service/src/contract.rs
```diff
@@ -137,6 +139,9 @@ pub fn query(deps: Deps, _: Env, msg: QueryMsg) -> Result<Binary, ContractError>
         QueryMsg::AllItsContracts => {
             query::all_its_contracts(deps).change_context(Error::QueryAllItsContracts)
         }
+        QueryMsg::ItsChains { filter } => {
```
cgorenflo, line 142: this also needs pagination to avoid gas overflow

contracts/interchain-token-service/src/contract/query.rs
```diff
@@ -34,6 +34,30 @@ pub fn all_its_contracts(deps: Deps) -> Result<Binary, Error> {
     to_json_binary(&contract_addresses).change_context(Error::JsonSerialization)
 }
 
+pub fn its_chains(deps: Deps, filter: Option<msg::ChainFilter>) -> Result<Binary, Error> {
+    let state_configs = state::load_chain_configs(deps.storage).change_context(Error::State)?;
```
cgorenflo, line 38: the load_chain_configs should make use of pagination as well. Loading all configs from storage could lead to gas overflow later on

contracts/interchain-token-service/src/contract/query.rs
```diff
@@ -34,6 +34,30 @@ pub fn all_its_contracts(deps: Deps) -> Result<Binary, Error> {
...
+
+    let chain_configs = match filter {
+        Some(filter) if filter.frozen_status.is_some() => state_configs
+            .into_iter()
+            .filter(|config| matches_filter(config, filter.frozen_status.as_ref()))
+            .collect(),
+        _ => state_configs,
```
cgorenflo, line 45:
```suggestion
    let chain_configs = match filter {
        Some(msg::ChainFilter {
            frozen_status: Some(frozen_status),
        }) => state_configs
            .into_iter()
            .filter(|config| frozen_status.matches(config))
            .collect(),
        _ => state_configs,
```

contracts/interchain-token-service/src/state.rs
```diff
@@ -170,6 +170,27 @@ pub fn load_chain_config(
         .ok_or_else(|| report!(Error::ChainNotFound(chain.to_owned())))
 }
 
+pub fn load_chain_configs(storage: &dyn Storage) -> Result<Vec<msg::ChainConfigResponse>, Error> {
+    let configs = CHAIN_CONFIGS
+        .range(storage, None, None, Order::Ascending)
```
cgorenflo, line 175: as mentioned before, this shouldn't be an unbounded range query

contracts/interchain-token-service/src/state.rs
```diff
@@ -170,6 +170,27 @@ pub fn load_chain_config(
...
+                    truncation: msg::TruncationConfig {
+                        max_uint: config.truncation.max_uint,
+                        max_decimals_when_truncating: config
+                            .truncation
+                            .max_decimals_when_truncating,
+                    },
+                    frozen: config.frozen,
+                })
+        })
+        .collect::<Result<Vec<_>, _>>()?;
```
cgorenflo, line 190: use `map_ok(...).try_collect()` instead

contracts/interchain-token-service/src/state.rs
```diff
@@ -170,6 +170,27 @@ pub fn load_chain_config(
...
+                .map(|(chain, config)| msg::ChainConfigResponse {
+                    chain,
+                    its_edge_contract: config.its_address,
+                    truncation: msg::TruncationConfig {
+                        max_uint: config.truncation.max_uint,
+                        max_decimals_when_truncating: config
+                            .truncation
+                            .max_decimals_when_truncating,
+                    },
+                    frozen: config.frozen,
```
cgorenflo, line 187: I would suggest to extract the mapping into it's own function, it's a separate concern to convert it into a msg type from loading it from storage. In fact, I think it's cleaner to return the actual chain configs here, and convert it in a separate step, at the call site

contracts/interchain-token-service/src/contract/query.rs
maancham: Added these two to also handle the `None => true` path:
```
impl ChainFilter {
    pub fn matches(&self, config: &ChainConfigResponse) -> bool {
        match &self.frozen_status {
            Some(frozen_status) => frozen_status.matches(config),
            None => true,
        }
    }
}

impl ChainStatusFilter {
    pub fn matches(&self, config: &ChainConfigResponse) -> bool {
        match self {
            ChainStatusFilter::Frozen => config.frozen,
            ChainStatusFilter::Active => !config.frozen,
        }
    }
}

File: contracts/interchain-token-service/src/msg.rs [hunk 12]
```diff
@@ -53,6 +53,35 @@ pub enum ExecuteMsg {
...
+        }
+    }
+}
+
+impl ChainFilter {
+    pub fn matches(&self, config: &ChainConfigResponse) -> bool {
+        match &self.frozen_status {
+            Some(frozen_status) => frozen_status.matches(config),
+            None => true,
+        }
```
fish-sammy, line 81:
```suggestion
        self.frozen_status.is_some_and(|frozen_status| frozen_status.matches(config))
```

fish-sammy, line 81:
```suggestion
        !self.frozen_status.is_some_and(|frozen_status| !frozen_status.matches(config))
```

contracts/interchain-token-service/src/msg.rs
```diff
@@ -53,6 +53,35 @@ pub enum ExecuteMsg {
...
 
+#[cw_serde]
+pub enum ChainStatusFilter {
+    Frozen,
+    Active,
+}
+
+#[cw_serde]
+pub struct ChainFilter {
+    pub frozen_status: Option<ChainStatusFilter>,
```
fish-sammy, line 64:
```suggestion
    pub status: Option<ChainStatusFilter>,
```

contracts/interchain-token-service/src/contract/query.rs
```diff
@@ -34,6 +35,33 @@ pub fn all_its_contracts(deps: Deps) -> Result<Binary, Error> {
     to_json_binary(&contract_addresses).change_context(Error::JsonSerialization)
 }
 
+pub fn its_chains(deps: Deps, filter: Option<msg::ChainFilter>) -> Result<Binary, Error> {
+    let state_chain_configs = state::load_chain_configs();
+
+    let chain_config_responses: Vec<msg::ChainConfigResponse> = state_chain_configs
```
fish-sammy, line 41:
```suggestion
    let chain_config_responses: Vec<_> = state_chain_configs
```

contracts/interchain-token-service/src/state.rs
```diff
@@ -170,6 +170,10 @@ pub fn load_chain_config(
         .ok_or_else(|| report!(Error::ChainNotFound(chain.to_owned())))
 }
 
+pub fn load_chain_configs() -> Map<&'static ChainNameRaw, ChainConfig> {
+    CHAIN_CONFIGS
```
fish-sammy, line 174: Don't like this. There's a reason this mod exist and internal knowledge isn't exposed. I'd much prefer
1. Apply the `ChainFilter` filtering on `ChainConfig` intead of `ChainConfigResponse`.
2. Create a new function `load_chain_configs(storage: &dyn Storage, filter: ChainFilter, pagination: Pagination)` in state.rs

fish-sammy (changes requested):

contracts/interchain-token-service/src/msg.rs
maancham: we might need other fields that could be used for filtering later on. Isn't it better to keep `frozen_status`? I was originally using `status`

fish-sammy: `status` is better imo as the type name `ChainStatusFilter` suggests nothing related to "frozen". Adding other fields wouldn't matter unless the nature of `ChainStatusFilter` also changes.

contracts/interchain-token-service/src/state.rs
maancham: Right now, the filter is defined over `msg.rs` for `ChainConfigResponse`.  If we want to do filtering over `state.rs`, we need to define the filter for `ChainConfig` of state. But the filter is also coming from the query message. Wouldn't this couple msg and state? Also, the new suggested `load_chain_configs` function requires a type conversion from the `msg::filter` received by the query to `state::filter`. At the end of the day, for both approaches, someone can just read the entire state if they don't provide filtering or pagination

contracts/interchain-token-service/src/msg.rs
maancham: Ok, changing it back to `status`

contracts/interchain-token-service/src/state.rs
cjcobb23: irrespective of state and msg separation, I actually sort of prefer the filter to be in the query module. I think state module should just return the chain configs, and query module can do whatever filtering necessary. We don't really gain anything by filtering in the state module, we still need to load all of the chain configs anyways. And the filter function only exists for the query, and isn't used in any other module, so why not just let filtering happen there?

That said, I don't think the way it's being done here is correct. state module should return some collection of chain configs, not direct access to the map

cgorenflo: We have two conflicting interests here:
1. I agree that the filter/pagination types make more sense in the msg context than the state context and it's awkward let the state do all the conversion
2. Exposing the raw state object defeats the purpose of encapsulation as @fish-sammy mentioned. However, if we return a collection without filtering/paginating we have to load the entire state, potentially leading to gas issues.

I suggest we create a layer of abstraction to make it work together:
```
pub fn load_chain_configs<'a>(storage: &'a dyn Storage, filter: impl Fn(&ChainConfig)->bool + 'a, limit: usize) -> impl Iterator<Item=Result<(ChainNameRaw, ChainConfig), Error>> + 'a {
    CHAIN_CONFIGS
        .range(storage, None, None, cosmwasm_std::Order::Ascending)
        .map(|r| r.change_context(Error::Storage))
        .filter_ok(move |(_, config)|filter(config))
        .take(limit)
}
```
This way, the pagination is easily converted into the `limit` parameter, and we can pass a filter function that can be defined in the query. Because this still returns an iterator, even the map to the `Response` type can be done in the query file without allocating the collection twice

contracts/interchain-token-service/src/state.rs
```diff
@@ -2,13 +2,17 @@ use std::collections::HashMap;
...
+use cosmwasm_std::{Addr, Order, OverflowError, StdError, Storage, Uint256};
+use cw_storage_plus::{Bound, Item, Map};
 use error_stack::{report, Result, ResultExt};
+use itertools::Itertools;
 use router_api::{Address, ChainNameRaw};
 
 use crate::{msg, RegisterTokenMetadata, TokenId};
 
+// Pagination limit
+const DEFAULT_LIMIT: u32 = u32::MAX;
```
cgorenflo, line 14: the default should be something like 30 or 50, this is effectively unbounded 😆

contracts/interchain-token-service/src/msg.rs
```diff
@@ -83,6 +94,17 @@ pub enum QueryMsg {
...
 
+    /// Query all chain configs with optional frozen filter
+    // The list is paginated by:
+    // - start_after: the chain name to start after, which the next page of results should start.
+    // - limit: limit the number of chains returned, default is u32::MAX.
+    #[returns(Vec<ChainConfigResponse>)]
+    ItsChains {
+        filter: Option<ChainFilter>,
+        start_after: Option<ChainNameRaw>,
+        limit: Option<u32>,
```
cgorenflo, line 105:
```suggestion
        #[serde(default = "default_pagination_limit"]
        limit: u32,
```

with 
```
const fn default_pagination_limit() -> u32{
    DEFAULT_LIMIT
}
```

this makes it clearer that we never do an unbounded search

contracts/interchain-token-service/src/msg.rs
```diff
@@ -9,6 +9,13 @@ use router_api::{Address, ChainNameRaw};
 use crate::state::{TokenConfig, TokenInstance};
 use crate::TokenId;
 
+// Pagination limit
+const DEFAULT_LIMIT: u32 = 30;
```
cgorenflo, line 13: just call the parameter DEFAULT_PAGINATION_LIMIT, then the comment isn't necessary

cgorenflo (commented):
I'd like to see one more test that takes an empty json encoded `ItsChains` and you make sure you get after deserialization
```
ItsChains{ filter: None, start_after: None, limit:30 }
```

contracts/interchain-token-service/src/contract/query.rs
```diff
@@ -34,6 +35,54 @@ pub fn all_its_contracts(deps: Deps) -> Result<Binary, Error> {
...
 }
 
+fn convert_chain_filter_option(
+    message_chain_filter: Option<msg::ChainFilter>,
+) -> impl Fn(&state::ChainConfig) -> bool + 'static {
+    move |config| match &message_chain_filter {
+        None => true,
+        Some(filter) => convert_chain_filter(filter)(config),
+    }
+}
```
fish-sammy, line 45:
```suggestion
fn convert_chain_filter_option(
    message_chain_filter: Option<msg::ChainFilter>,
) -> impl Fn(&state::ChainConfig) -> bool + 'static {
    let chain_filter = message_chain_filter.unwrap_or_default();

    move |config| convert_chain_filter(&chain_filter)(config)
}
```
And add `#[derive(Default)]` to `ChainFilter`. The idea is that the default filter should be NOT filtering anything.

fish-sammy (approved):

contracts/interchain-token-service/src/contract/query.rs
```diff
@@ -34,6 +35,54 @@ pub fn all_its_contracts(deps: Deps) -> Result<Binary, Error> {
...
+
+pub fn its_chains(
+    deps: Deps,
+    filter: Option<msg::ChainFilter>,
+    start_after: Option<ChainNameRaw>,
+    limit: u32,
+) -> Result<Binary, Error> {
+    let filtered_chain_configs: Vec<_> = state::load_chain_configs(
+        deps.storage,
+        convert_chain_filter_option(filter),
```
fish-sammy, line 67:
```suggestion
fn convert_chain_filter(chain_filter: msg::ChainFilter) -> impl Fn(&state::ChainConfig) -> bool {
    move |config| match &chain_filter.status {
        Some(status) => match status {
            msg::ChainStatusFilter::Frozen => config.frozen,
            msg::ChainStatusFilter::Active => !config.frozen,
        },
        None => true,
    }
}

pub fn its_chains(
    deps: Deps,
    filter: Option<msg::ChainFilter>,
    start_after: Option<ChainNameRaw>,
    limit: u32,
) -> Result<Binary, Error> {
    let filtered_chain_configs: Vec<_> = state::load_chain_configs(
        deps.storage,
        convert_chain_filter(filter.unwrap_or_default()),
```
And add `#[derive(Default)]` to `ChainFilter`.

PR #755, opened 2025-02-06
chore: update to rust version 1.81.0

cgorenflo:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros


haiyizxx (approved):

contracts/axelarnet-gateway/src/contract/execute.rs
```diff
@@ -139,7 +139,7 @@ pub fn route_messages(
     };
 
     msgs.into_iter()
-        .group_by(|msg| msg.destination_chain.to_owned())
+        .chunk_by(|msg| msg.destination_chain.to_owned())
```
cgorenflo, line 142: group_by has been deprecated

contracts/multisig/src/contract/execute.rs
```diff
@@ -188,10 +188,9 @@ pub fn authorize_callers(
 ) -> Result<Response, ContractError> {
     contracts
         .iter()
-        .map(|(contract_address, chain_name)| {
+        .try_for_each(|(contract_address, chain_name)| {
             AUTHORIZED_CALLERS.save(deps.storage, contract_address, chain_name)
-        })
-        .try_collect()?;
```
cgorenflo, line 194: the 2024 edition compiler is stricter with `()` and `!` types and complains about potentially unsupported type conversions. We don't need to collect the output anyway, so replacing it with `try_for_each` which matches the actual intent better

fish-sammy (approved):

PR #759, opened 2025-02-11
feat: update dockerfile to build only ampd

eguajardo:
Reduces docker build time by about 4 minutes by only building ampd instead of the whole workspace. Also copies over the whole workspace so there is no need to update every time there is a change in the directory structure.

It seems that the image weights the same (131mb) as the image created before this changes

### DIFF
--- a/ampd/Dockerfile
+++ b/ampd/Dockerfile
@@ -4,7 +4,10 @@ RUN apt-get update && apt-get install -y clang protobuf-compiler cmake
 WORKDIR /ampd
 COPY . .
 
-RUN cargo install --locked --path ./ampd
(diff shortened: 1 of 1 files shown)


fish-sammy (changes requested):
I like the idea of building ampd only and excluding unrelated packages and contracts. However, the multi-stage build of ampd itself shouldn't be removed.

maancham (approved):

eguajardo:
After some analysis, it seems the best way to optimize the Dockerfile would be using cache mounts, however they don't work with github actions because the cache does not persist. There might be plugins or tools to make them persistent, but we will make a proper analysis and testing of those tools at a later stage and evaluate if the addition of those is worth it.

PR #760, opened 2025-02-11
refactor(minor-interchain-token-service): change max uint to max bits

cjcobb23:
## Todos
need to migrate the existing chain configs

https://axelarnetwork.atlassian.net/browse/AXE-7486

accompanying PR in contract deployment repo: https://github.com/axelarnetwork/axelar-contract-deployments/pull/563

### DIFF
--- a/contracts/interchain-token-service/src/contract/execute/interceptors.rs
+++ b/contracts/interchain-token-service/src/contract/execute/interceptors.rs
@@ -1,5 +1,5 @@
 use axelar_wasm_std::{nonempty, FnExt};
-use cosmwasm_std::{Storage, Uint256};
+use cosmwasm_std::{OverflowError, Storage, Uint256};
 use error_stack::{bail, ensure, report, Result, ResultExt};
 use router_api::ChainNameRaw;
 
@@ -204,8 +204,8 @@ fn destination_token_decimals(
 
     if source_chain_config
         .truncation
-        .max_uint
-        .le(&destination_chain_config.truncation.max_uint)
+        .max_uint_bits
+        .le(&destination_chain_config.truncation.max_uint_bits)
     {
         source_token_decimals
     } else {
@@ -242,10 +242,10 @@ fn destination_amount(
         return Ok(source_amount);
     }
 
-    let destination_max_uint = state::load_chain_config(storage, destination_chain)
+    let destination_max_uint_bits = state::load_chain_config(storage, destination_chain)
         .change_context(Error::State)?
         .truncation
-        .max_uint;
+        .max_uint_bits;
 
     // It's intentionally written in this way since the end result may still be fine even if
     //     1) amount * (10 ^ (dest_chain_decimals)) overflows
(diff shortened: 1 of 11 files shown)


fish-sammy:
What's the reason for making this change?

contracts/interchain-token-service/src/contract/execute/interceptors.rs
```diff
@@ -288,6 +288,16 @@ fn destination_amount(
     })
 }
 
+fn amount_overflows(amount: Uint256, target_chain_max_bits: u32) -> bool {
+    match amount.checked_shr(target_chain_max_bits) {
+        Ok(res) => res.gt(&Uint256::zero()),
+        // this overflow error occurs when trying to shift 256 bits or more.
+        // But this can only happen if max_bits is >= 256, and amount itself is only 256 bits.
+        // So in this can, amount cannot possibly overflow the max uint of the target chain
```
fish-sammy, line 296:
```suggestion
        // So in this case, amount cannot possibly overflow the max uint of the target chain
```

contracts/interchain-token-service/src/msg.rs
```diff
@@ -60,8 +59,8 @@ pub struct ChainConfig {
 
 #[cw_serde]
 pub struct TruncationConfig {
-    pub max_uint: nonempty::Uint256, // The maximum uint value that is supported by the chain's token standard
-    pub max_decimals_when_truncating: u8, // The maximum number of decimals that is preserved when deploying from a chain with a larger max_uint
+    pub max_uint_bits: u32, // The maximum number of bits used by the chain to represent unsigned integers
```
fish-sammy, line 62: Do we have a `nonempty::u32`? zero wouldn't make sense here.

contracts/interchain-token-service/src/state.rs
```diff
@@ -41,15 +41,15 @@ pub struct ChainConfig {
 
 #[cw_serde]
 pub struct TruncationConfig {
-    pub max_uint: nonempty::Uint256, // The maximum uint value that is supported by the chain's token standard
-    pub max_decimals_when_truncating: u8, // The maximum number of decimals that is preserved when deploying from a chain with a larger max_uint
+    pub max_uint_bits: u32, // The maximum number of bits used to represent unsigned integer values that is supported by the chain's token standard
```
fish-sammy, line 44: Same here.

contracts/interchain-token-service/src/msg.rs
cjcobb23: if we are going to change the type, I'd rather add more restrictions than just non zero. While zero is wrong, I think many other u32 values are also wrong, such as 1 or other small values, odd values probably, values that are not a power of 2, and values greater than 256.

cjcobb23:
> What's the reason for making this change? Easier to configure?

yeah, someone registered a bunch of chains on devnet with a max uint of 256, because they thought it was supposed to be bits. It's harder to work with when you need to look up what 2^256-1 is

contracts/interchain-token-service/src/msg.rs
cjcobb23: i created a new type for this, lmk what you think

haiyizxx (approved):

PR #763, opened 2025-02-18
fix(minor-interchain-token-service): migrate max uint to bits

cjcobb23:
## Description
https://axelarnetwork.atlassian.net/browse/AXE-7515
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros


contracts/interchain-token-service/src/contract.rs
```diff
@@ -55,9 +56,7 @@ pub enum Error {
 
 #[cfg_attr(not(feature = "library"), entry_point)]
 pub fn migrate(deps: DepsMut, _env: Env, _msg: Empty) -> Result<Response, ContractError> {
-    // Implement migration logic if needed
-
-    // TODO migrate max uint to max uint bits
+    migrations::v1_1_0::migrate(deps.storage)?;
```
eguajardo, line 59: could you use the new format for migrations. The router is already using that format, basically move the `migrate` endpoint and everything related to the migration to the mod and re-export as necessary

eguajardo: note the use of the macro `migrate_from_version`

fish-sammy: convention checklist failure lol

contracts/interchain-token-service/src/contract/migrations/v1_1_0.rs
```diff
@@ -0,0 +1,134 @@
...
+fn convert_max_uint_to_max_bits(max_uint: nonempty::Uint256) -> Result<NumBits, ContractError> {
+    let fixed_cases = [256, 128, 64, 32];
+    for case in fixed_cases {
+        // some chains were incorrectly registered with the max number of bits as the max uint
+        if *max_uint == Uint256::from_u128(case as u128) {
+            return NumBits::try_from(case).map_err(|err| err.into());
+        }
+    }
+    // Need to add one to correctly get the number of bits. This will round down if the result is not a power of two
+    NumBits::try_from((Uint512::from(*max_uint) + Uint512::one()).ilog2()).map_err(|err| err.into())
```
fish-sammy, line 70: Not related to this PR, but if only certain values are allowed for `NumBits`, wouldn't it be more efficient and safer to do it as an enum than wrapping a `u32`?

contracts/interchain-token-service/src/contract/migrations/v1_1_0.rs
```diff
@@ -0,0 +1,134 @@
...
+        NEW_CHAIN_CONFIGS.save(storage, &chain, &config)?;
+    }
+
+    Ok(())
+}
+
+fn convert_max_uint_to_max_bits(max_uint: nonempty::Uint256) -> Result<NumBits, ContractError> {
+    let fixed_cases = [256, 128, 64, 32];
+    for case in fixed_cases {
+        // some chains were incorrectly registered with the max number of bits as the max uint
```
fish-sammy, line 64: Is this migration going to be needed for mainnet?

contracts/interchain-token-service/src/contract/migrations/v1_1_0.rs
cjcobb23: i thought of this for the other PR, but I found it a bit clunky. The conversion functions are more verbose with an enum, and the default serialization is not ideal.

contracts/interchain-token-service/src/contract/migrations/v1_1_0.rs
cjcobb23: the migration in general or the incorrect number case? In general, yes since the contract was deployed to mainnet. But I don't think any chains were registered incorrectly on mainnet

eguajardo (approved):

PR #764, opened 2025-02-19
feat(ampd): add XRPL verification & signing

themicp:
## Description
Enable XRPL verification and signing on ampd.


ampd/src/xrpl/verifier.rs
```diff
@@ -0,0 +1,191 @@
...
+pub fn is_successful_tx(tx: &Transaction) -> bool {
+    if let Some(meta) = &tx.common().meta {
+        return meta.transaction_result.category() == ResultCategory::Tes;
+    }
+    false
+}
+
+fn remove_0x_prefix(s: String) -> String {
+    s.strip_prefix("0x").unwrap_or(&s).to_string()
+}
```
cjcobb23, line 101: this is a little risky, because this strips the prefix if it exists, and otherwise returns the original string. I don't like using something like this in verification, because it's hard to make sure that there aren't multiple different representations of the same message (which we should not allow). Can you change this to error if there is no prefix?

ampd/src/xrpl/verifier.rs
```diff
@@ -0,0 +1,191 @@
...
+        };
+
+        Some(amount == message.amount)
+    }()
+    .unwrap_or(false)
+}
+
+pub fn verify_memos(memos: HashMap<String, String>, message: &XRPLUserMessage) -> bool {
+    let destination_address =
+        remove_0x_prefix(message.destination_address.to_string()).to_uppercase();
```
cjcobb23, line 123: I think this is unsafe, there are effectively two different representations of the same message. Because of the optional stripping, and because of the `to_uppercase`

ampd/src/xrpl/verifier.rs
```diff
@@ -0,0 +1,191 @@
...
+                }
+            }
+            None => return false,
+        }
+
+        return payment_tx.destination == multisig_address.to_string()
+            && message.source_address.to_string() == tx.common().account
+            && verify_amount(tx_amount, message)
+            && verify_memos(memos, message)
+            && payment_tx.flags.is_empty(); // TODO: whitelist specific flags
```
cjcobb23, line 87: why do we need to check the flags are empty?

ampd/src/handlers/xrpl_verify_msg.rs
```diff
@@ -0,0 +1,215 @@
...
+#[try_from("wasm-messages_poll_started")]
+struct PollStartedEvent {
+    poll_id: PollId,
+    #[serde(with = "xrpl_account_id_string")]
+    source_gateway_address: XRPLAccountId,
+    confirmation_height: u32,
+    expires_at: u64,
+    messages: Vec<XRPLMessage>,
+    participants: Vec<TMAddress>,
+}
```
cjcobb23, line 43: there should probably be a source chain in this message, which we should check

ampd/src/handlers/multisig.rs
```diff
@@ -120,6 +124,14 @@ where
             result => result.change_context(DeserializeEvent)?,
         };
 
+        if chain.eq(&ChainName::from_str("XRPL").unwrap()) {
```
cjcobb23, line 127: shouldn't have an unwrap here. Also, I am not sure this is the best long term solution, to just hardcode cases here. But doesn't need to be solved in this PR

packages/xrpl-types/src/types.rs
```diff
@@ -0,0 +1,1001 @@
...
+const XRPL_TX_HASH_LENGTH: usize = 32;
+
+pub const XRP_DECIMALS: u8 = 6;
+pub const XRPL_ISSUED_TOKEN_DECIMALS: u8 = 15;
+pub const XRP_MAX_UINT: u64 = 100_000_000_000_000_000u64;
+
+const UNSIGNED_TRANSACTION_MULTI_SIGNING_HASH_PREFIX: [u8; 4] = [0x53, 0x4D, 0x54, 0x00];
+
+const ALLOWED_CURRENCY_CHARS: &str =
+    "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789?!@#$%^&*<>(){}[]|";
```
cjcobb23, line 28: surely there is a cleaner way to write this regex?

packages/xrpl-types/src/types.rs
```diff
@@ -0,0 +1,1001 @@
...
+
+const XRPL_ACCOUNT_ID_LENGTH: usize = 20;
+const XRPL_CURRENCY_LENGTH: usize = 20;
+const XRPL_TX_HASH_LENGTH: usize = 32;
+
+pub const XRP_DECIMALS: u8 = 6;
+pub const XRPL_ISSUED_TOKEN_DECIMALS: u8 = 15;
+pub const XRP_MAX_UINT: u64 = 100_000_000_000_000_000u64;
+
+const UNSIGNED_TRANSACTION_MULTI_SIGNING_HASH_PREFIX: [u8; 4] = [0x53, 0x4D, 0x54, 0x00];
```
cjcobb23, line 25: a comment here could be useful, maybe with a link to documentation

packages/xrpl-types/src/types.rs
```diff
@@ -0,0 +1,1001 @@
...
+    }
+}
+
+// HASHING LOGIC
+
+fn xrpl_hash(prefix: [u8; 4], tx_blob: &[u8]) -> [u8; 32] {
+    let mut hasher = Sha512::new_with_prefix(prefix);
+    hasher.update(tx_blob);
+    let digest: [u8; 64] = hasher.finalize().into();
+    digest[..32].try_into().unwrap()
```
cjcobb23, line 328: shouldn't unwrap

ampd/src/handlers/errors.rs
```diff
@@ -1,13 +1,20 @@
...
+    #[error("failed to get the latest finalized block")]
+    Finalizer,
+    #[error("failed to prepare message for signing")]
+    MessageToSign,
+    #[error("failed to parse public key")]
+    PublicKey,
     #[error("failed to get signature from tofnd")]
     Sign,
+    #[error(transparent)]
+    Std(#[from] StdError),
```
cjcobb23, line 17: I generally don't like this, it results in a very cryptic error message. Better to just change_context into an error that makes sense

ampd/src/handlers/config.rs
```diff
@@ -25,13 +24,13 @@ pub enum Config {
     EvmMsgVerifier {
         cosmwasm_contract: TMAddress,
         #[serde(flatten, with = "chain")]
-        chain: Chain,
+        chain: Chain<crate::evm::finalizer::Finalization>,
```
cjcobb23, line 27: what is this change?

ampd/src/xrpl/finalizer.rs
```diff
@@ -0,0 +1,118 @@
...
+#[async_trait]
+pub trait Finalizer: Send + Sync {
+    async fn latest_validated_ledger_index(&self) -> Result<u32>;
+}
+
+#[derive(Debug, Deserialize, Serialize, PartialEq, Default)]
+pub enum Finalization {
+    #[default]
+    ValidatedLedger,
+    ConfirmationHeight,
```
cjcobb23, line 21: why do we need two finalization mechanisms for XRPL?

cjcobb23: should just be the validated ledger option right?

ampd/src/handlers/multisig.rs
cjcobb23: actually, this needs to be configurable right? the chain might not be named "xrpl", especially in different environments (devnet, etc)

ampd/src/handlers/multisig.rs
```diff
@@ -138,8 +150,8 @@ where
         match pub_keys.get(&self.verifier) {
             Some(pub_key) => {
                 let key_type = match pub_key {
-                    PublicKey::Secp256k1(_) => tofnd::Algorithm::Ecdsa,
-                    PublicKey::Ed25519(_) => tofnd::Algorithm::Ed25519,
+                    PublicKey::Secp256k1(_) => tofnd::Algorithm::Ed25519,
+                    PublicKey::Ed25519(_) => tofnd::Algorithm::Ecdsa,
```
cjcobb23, line 154: this is not right, this is what bricked ampd signing sometime ago. these lines shouldn't be changed

ampd/src/xrpl/verifier.rs
```diff
@@ -0,0 +1,191 @@
...
+    memos.get("destination_address") == Some(&destination_address)
+        && memos.get("destination_chain") == Some(&destination_chain)
+        && memos
+            .get("payload_hash")
+            .and_then(|payload_hash_hex| {
+                hex::decode(remove_0x_prefix(payload_hash_hex.to_owned())).ok()
+            })
+            .map_or(false, |payload_hash| {
+                message.payload_hash == payload_hash.try_into().ok()
+            })
```
cjcobb23, line 135: this is going to return false if there is no payload hash, but isn't this supposed to be a supported case? I.e. pure token transfer, no contract call. There would be no payload hash in that case right?

ampd/src/xrpl/verifier.rs
```diff
@@ -0,0 +1,191 @@
...
+    tx: &Transaction,
+    multisig_address: &XRPLAccountId,
+    message: &XRPLUserMessage,
+    memos: HashMap<String, String>,
+) -> bool {
+    if let Payment(payment_tx) = &tx {
+        let tx_amount = payment_tx.amount.clone();
+        match tx.common().meta.clone() {
+            Some(tx_meta) => {
+                if tx_meta.delivered_amount != Some(tx_amount.clone()) {
```
cjcobb23, line 76: when could this case be hit? We should add a comment

packages/xrpl-types/src/msg.rs
```diff
@@ -0,0 +1,204 @@
...
+#[cw_serde]
+#[derive(Eq, Hash)]
+pub struct XRPLProverMessage {
+    #[serde(with = "tx_hash_hex")]
+    #[schemars(with = "String")]
+    pub tx_id: TxHash,
+    #[serde(with = "tx_hash_hex")]
+    #[schemars(with = "String")]
+    pub unsigned_tx_hash: TxHash,
+}
```
cjcobb23, line 61: what is this struct for? should add a comment

ampd/src/xrpl/verifier.rs
```diff
@@ -0,0 +1,191 @@
...
+                XRPLToken {
+                    issuer: a.issuer.try_into().ok()?,
+                    currency: a.currency.try_into().ok()?,
+                },
+                a.value.try_into().ok()?,
+            ),
+            Amount::Drops(a) => XRPLPaymentAmount::Drops(a.parse().ok()?),
+        };
+
+        Some(amount == message.amount)
```
cjcobb23, line 116: how are we handling the fee? Isn't some part of the amount supposed to be a fee paid to the gas service? How is that specified?

ampd/src/xrpl/verifier.rs
```diff
@@ -0,0 +1,191 @@
...
+
+pub fn is_valid_prover_message(
+    tx: &Transaction,
+    multisig_address: &XRPLAccountId,
+    message: &XRPLProverMessage,
+    memos: HashMap<String, String>,
+) -> bool {
+    tx.common().account == multisig_address.to_string()
+        && memos.get("unsigned_tx_hash")
+            == Some(&message.unsigned_tx_hash.to_string().to_uppercase())
```
cjcobb23, line 63: can you remind me what this is for? Why is there a hash in the memo, what is it a hash of, etc. Should add a comment as well.

packages/axelar-wasm-std/src/msg_id/tx_hash.rs
```diff
@@ -65,6 +65,15 @@ impl Display for HexTxHash {
     }
 }
 
+impl From<HexTxHash> for nonempty::String {
+    fn from(msg_id: HexTxHash) -> Self {
+        msg_id
+            .to_string()
+            .try_into()
+            .expect("failed to convert msg id to non-empty string")
```
cjcobb23, line 73: no expect

cjcobb23: actually nevermind, this one is probably fine

packages/xrpl-types/src/msg.rs
```diff
@@ -0,0 +1,204 @@
...
+            XRPLMessage::ProverMessage(msg) => {
+                let mut res: Vec<Attribute> = msg.into();
+                res.push(("type", "prover_message").into());
+                res
+            }
+            XRPLMessage::UserMessage(msg) => {
+                let mut res: Vec<Attribute> = msg.into();
+                res.push(("type", "user_message").into());
+                res
+            }
```
cjcobb23, line 106: can you do this without mut?

packages/xrpl-types/src/msg.rs
```diff
@@ -0,0 +1,204 @@
...
+    }
+}
+
+impl CrossChainMessage for TxHash {
+    fn cc_id(&self, source_chain: ChainNameRaw) -> CrossChainId {
+        CrossChainId {
+            source_chain,
+            message_id: format!("0x{}", HexBinary::from(self.clone()).to_hex())
+                .try_into()
+                .unwrap(),
```
cjcobb23, line 179: no unwrap

packages/xrpl-types/src/msg.rs
```diff
@@ -0,0 +1,204 @@
...
+    }
+}
+
+impl CrossChainMessage for XRPLProverMessage {
+    fn cc_id(&self, source_chain: ChainNameRaw) -> CrossChainId {
+        CrossChainId {
+            source_chain,
+            message_id: format!("0x{}", HexBinary::from(self.tx_id.clone()).to_hex())
+                .try_into()
+                .unwrap(),
```
cjcobb23, line 190: no unwrap

packages/xrpl-types/src/msg.rs
```diff
@@ -0,0 +1,204 @@
...
+    }
+}
+
+impl CrossChainMessage for XRPLUserMessage {
+    fn cc_id(&self, source_chain: ChainNameRaw) -> CrossChainId {
+        CrossChainId {
+            source_chain,
+            message_id: format!("0x{}", HexBinary::from(self.tx_id.clone()).to_hex())
+                .try_into()
+                .unwrap(),
```
cjcobb23, line 201: no unwrap

packages/xrpl-types/src/msg.rs
```diff
@@ -0,0 +1,204 @@
...
+        }
+    }
+}
```
cjcobb23, line 204: what do you need all of this for? Creating the cross chain id

packages/xrpl-types/src/types.rs
```diff
@@ -0,0 +1,1001 @@
...
+pub const XRPL_TOKEN_MAX_EXPONENT: i64 = MAX_EXPONENT;
+
+const MAX_XRPL_TOKEN_AMOUNT: XRPLTokenAmount = XRPLTokenAmount {
+    mantissa: MAX_MANTISSA,
+    exponent: MAX_EXPONENT,
+};
+
+#[cw_serde]
+#[derive(Eq, Hash)]
+pub struct TxHash([u8; XRPL_TX_HASH_LENGTH]);
```
cjcobb23, line 47: Why not just use `HexTxHash` for this type?

ampd/src/handlers/multisig.rs
milapsheth, line 154:
```suggestion
                    PublicKey::Secp256k1(_) => tofnd::Algorithm::Ecdsa,
                    PublicKey::Ed25519(_) => tofnd::Algorithm::Ed25519,
```

ampd/src/xrpl/verifier.rs
themicp: Looks like we don't need it at all, so I just removed it.

ampd/src/xrpl/verifier.rs
themicp: Removed the stripping. I guess the uppercase should be safe, since the XRPL API returns everything in uppercase.

packages/xrpl-types/src/types.rs
themicp: Yep, did another pass on it.

ampd/src/xrpl/verifier.rs
cjcobb23: i think the uppercasing is also unsafe because you can submit two different messages to the voting verifier that diff only in casing, and both will be confirmed

packages/xrpl-types/src/types.rs
cjcobb23: i guess this one is ok, but let's make it an `expect` instead

ampd/src/xrpl/verifier.rs
cjcobb23: nevermind, I figured this out

ampd/src/xrpl/verifier.rs
themicp: Right, I see. I removed the uppercasing from the message fields and did lowercasing on the XRPL fields.

ampd/src/xrpl/verifier.rs
themicp: Added a comment in the definition of this field.

packages/xrpl-types/src/msg.rs
themicp: Turned this (and the following unwraps) to expect, since this is not expected to fail.

ampd/src/handlers/errors.rs
themicp: Thankfully it wasn't used, so I just removed it.

ampd/src/xrpl/verifier.rs
themicp: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/108fdd5fe2040003c9a6369228dcac9423ac4598

packages/xrpl-types/src/types.rs
themicp: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/fa0e5d42f043c9c1497fabd3baf14007d018008e

packages/xrpl-types/src/types.rs
themicp: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/b5d39c7cac7d34f1cdb6a4a74edabde12b7f293f

packages/xrpl-types/src/types.rs
themicp: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/64b079c47e07601b5c0a753c13bb53361afb76de

ampd/src/handlers/errors.rs
themicp: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/c5a75f810942fa2f85cf3381e31b58bdf139cc90

ampd/src/xrpl/verifier.rs
themicp: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/1065e24f24a73c48ee0b47faef3eaf11b2f2db47

ampd/src/xrpl/verifier.rs
themicp: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/13162b9d59106ef5c595dce8a718af452d987fce

packages/xrpl-types/src/msg.rs
themicp: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/f5ee3136bad58e003f1cfff855b5187ebed5b1f8

ampd/src/xrpl/verifier.rs
themicp: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/a042b51d65e08f31b8f7e3caad32c78ed5e4f755

packages/xrpl-types/src/msg.rs
themicp: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/1b728f4aef41c29e173616912fc36cdf0a03a5ea

ampd/src/handlers/config.rs
themicp: Because a new finalizer is introduced [here](https://github.com/axelarnetwork/axelar-amplifier/pull/764/files#diff-3b3be62fb2dae0a6eee787cdfa79c275a449c30f968a111e1fe2f77dd09c0a48R18),  and used [here](https://github.com/axelarnetwork/axelar-amplifier/pull/764/files#diff-6b67bfec76817a5392e38dcf36fcb2903b1f9fe435ad4f868b8a696a21207dc4R173).

packages/xrpl-types/src/msg.rs
```diff
@@ -0,0 +1,204 @@
...
+#[derive(Eq, Hash)]
+pub struct XRPLUserMessage {
+    #[serde(with = "tx_hash_hex")]
+    #[schemars(with = "String")] // necessary attribute in conjunction with #[serde(with ...)]
+    pub tx_id: TxHash,
+    #[serde(with = "xrpl_account_id_string")]
+    #[schemars(with = "String")]
+    pub source_address: XRPLAccountId,
+    pub destination_chain: ChainName,
+    pub destination_address: nonempty::HexBinary,
```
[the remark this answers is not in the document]
themicp: Right, updating this to be `nonempty::String`. However, [InterchainTransfer](https://github.com/axelarnetwork/axelar-amplifier/blob/fed53af55c02d7195d2981f0e364d443309d0479/contracts/interchain-token-service/src/primitives.rs#L46) should also probably be updated to support non-hex addresses.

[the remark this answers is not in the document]
themicp: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/604c0722af84d8439ec90a78cae5ea521db8cfde

packages/xrpl-types/src/types.rs
themicp: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/3e52a1fdd178348c7530ecaf5c523abc2838c15c

packages/xrpl-types/src/msg.rs
themicp: Turns out it is not needed in the context of ampd, so I got rid of the trait completely.

ampd/src/handlers/xrpl_verify_msg.rs
themicp: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/c63d25bfd554c2f2112eaa63738385ead7864b1f

ampd/src/xrpl/verifier.rs
themicp: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/0a179dfc7bd0aef8f0be0315ab89b71de16ed3b9

packages/xrpl-types/src/types.rs
```diff
@@ -0,0 +1,1018 @@
...
+    }
+}
+
+impl fmt::Display for XRPLAccountId {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let mut payload = Vec::<u8>::with_capacity(25);
+        payload.extend(&[0x00]);
+        payload.extend_from_slice(self.as_ref());
+
+        let checksum_hash1 = Sha256::digest(payload.clone());
```
maancham, line 193: For calculation of `checksum_hash1`, is it possible to just pass the reference to the payload instead of cloning it for optimization?

packages/xrpl-types/src/types.rs
```diff
@@ -0,0 +1,1018 @@
...
+impl From<[u8; XRPL_ACCOUNT_ID_LENGTH]> for XRPLAccountId {
+    fn from(bytes: [u8; XRPL_ACCOUNT_ID_LENGTH]) -> Self {
+        XRPLAccountId(bytes)
+    }
+}
+
+impl fmt::Display for XRPLAccountId {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let mut payload = Vec::<u8>::with_capacity(25);
+        payload.extend(&[0x00]);
```
maancham, line 190: Wouldn't it be better to use `push` instead of extend for single bytes to be explicit?

ampd/Cargo.toml
```diff
@@ -83,6 +84,8 @@ url = "2.3.1"
 valuable = { version = "0.1.0", features = ["derive"] }
 valuable-serde = { version = "0.1.0", features = ["std"] }
 voting-verifier = { workspace = true }
+xrpl_http_client = { version = "0.15.0" }
+xrpl-types = { workspace = true }
```
milapsheth, line 88: To avoid duplication of the utils, can you import the `xrpl-types` from the `xrpl` branch of your repo instead of checking it in. CJ mentioned that some changes were made in `xrpl-types` in this PR that aren't there in the original fork

ampd/src/xrpl/verifier.rs
```diff
@@ -0,0 +1,271 @@
...
+        };
+
+        Some(gas_fee_amount == message.gas_fee_amount && gas_fee_amount <= message.amount)
+    }()
+    .unwrap_or(false)
+}
+
+pub fn verify_memos(memos: HashMap<String, String>, message: &XRPLUserMessage) -> bool {
+    let expected_destination_address = message.destination_address.to_string();
+    let expected_destination_chain = hex::encode(message.destination_chain.to_string());
```
cjcobb23, line 147: why is the destination chain being hex encoded but not the destination address? Everything in the memo is hex encoded right? The address in the message may not be hex, it could be bech32 or base58 or something else, so we need to translate from what's in the memo.

ampd/src/xrpl/verifier.rs
```diff
@@ -0,0 +1,271 @@
...
+            Some(amount) => amount.to_string(),
+        };
+
+        let gas_fee_amount = match message.amount.clone() {
+            XRPLPaymentAmount::Issued(token, _) => XRPLPaymentAmount::Issued(
+                XRPLToken {
+                    issuer: token.issuer.try_into().ok()?,
+                    currency: token.currency.try_into().ok()?,
+                },
+                gas_fee_amount_str.try_into().ok()?,
```
cjcobb23, line 133: is this going to work? gas_fee_amount_str is hex, but the `FromStr` impl for XRPLTokenAmount expects non hex characters.

ampd/src/xrpl/verifier.rs
themicp: There was some inconsistency between the `memo_data` fields, some were hex-encoded and some were not. Made it consistent now, expecting all of them to be hex-encoded.
https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/73767841cb7b3f58b034032d97d61fdff7b900d1

ampd/src/xrpl/verifier.rs
```diff
@@ -0,0 +1,271 @@
...
+            ),
+            XRPLPaymentAmount::Drops(_) => {
+                XRPLPaymentAmount::Drops(gas_fee_amount_str.parse().ok()?)
```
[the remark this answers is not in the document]
themicp: It wasn't hex, but now made all the `memo_data` fields to be hex: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/73767841cb7b3f58b034032d97d61fdff7b900d1

packages/xrpl-types/src/types.rs
themicp: Yep, good point: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/228372c575915821946f340108d26b200e5de1ad

ampd/src/xrpl/verifier.rs
cjcobb23: also I think if we treat the destination address as a string, and then hex encode it, the data will show better in the xrpl explorer

packages/axelar-core-std/src/nexus/query.rs
```diff
@@ -15,7 +15,7 @@ pub enum QueryMsg {
     IsChainRegistered { chain: String },
 }
 
-#[derive(Clone, Debug, PartialEq, Deserialize)]
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
```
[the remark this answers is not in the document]
k4m4: Doesn't seem to be needed anymore: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/c4c5a0bff1fc0079a8109f07215cdfdfa8beb555

ampd/src/handlers/xrpl_verify_msg.rs
```diff
@@ -0,0 +1,215 @@
...
+
+type Result<T> = error_stack::Result<T, Error>;
+
+#[derive(Deserialize, Debug)]
+#[try_from("wasm-messages_poll_started")]
+struct PollStartedEvent {
+    poll_id: PollId,
+    #[serde(with = "xrpl_account_id_string")]
+    source_gateway_address: XRPLAccountId,
+    confirmation_height: u32,
```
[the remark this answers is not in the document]
k4m4: Kept the validation finality only: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/fa47622daa080344c1503c315d9313537c810a16

ampd/src/handlers/multisig.rs
k4m4: Fixed: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/a10002baa461518dc073732e26291dec17f0f42a

ampd/src/xrpl/verifier.rs
k4m4: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/73767841cb7b3f58b034032d97d61fdff7b900d1

ampd/src/handlers/xrpl_verify_msg.rs
```diff
@@ -0,0 +1,215 @@
...
+                            Some((tx_hash, tx_res.tx))
+                        } else {
+                            None
+                        }
+                    } else {
+                        None
+                    },
+                )
+            })()
+            .unwrap_or(None)
```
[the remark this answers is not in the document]
k4m4: Simplified: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/fa47622daa080344c1503c315d9313537c810a16

ampd/src/xrpl/finalizer.rs
k4m4: Yeah, kept just the `ValidatedLedger` option: https://github.com/axelarnetwork/axelar-amplifier/commit/fa47622daa080344c1503c315d9313537c810a16

ampd/src/handlers/xrpl_verify_msg.rs
```diff
@@ -0,0 +1,215 @@
...
+
+            (|| -> Result<_> {
+                Ok(
+                    if ledger_index
+                        .unwrap_or(u32::MAX)
+                        .le(&latest_validated_ledger_index)
+                        && tx_hash.is_some()
+                    {
+                        let tx_hash = TxHash::try_from(
+                            HexBinary::from_hex(tx_hash.unwrap().as_str()).map_err(Error::from)?,
```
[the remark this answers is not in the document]
k4m4: Addressed here: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/3e52a1fdd178348c7530ecaf5c523abc2838c15c

packages/xrpl-types/src/msg.rs
k4m4: This snippet was removed.

packages/xrpl-types/src/msg.rs
k4m4: This snippet was removed as well.

ampd/Cargo.toml
themicp: Rust would not be happy with this because the same types would be referenced by the `xrpl` branch and the current branch, but on different versions. For example, `HexTxHash` would be defined twice; once in the current branch and once in the `xrpl` branch, and the compiler doesn't know which one to use. Maybe we could export `xrpl-types` to a separate crate? Happy to hear any other ideas.

packages/axelar-wasm-std/src/msg_id/tx_hash.rs
```diff
@@ -11,15 +12,23 @@ use super::Error;
...
+    #[serde(flatten)]
     pub tx_hash: Hash,
 }
 
 impl HexTxHash {
-    pub fn tx_hash_as_hex(&self) -> nonempty::String {
-        format!("0x{}", HexBinary::from(self.tx_hash).to_hex())
-            .try_into()
-            .expect("failed to convert tx hash to non-empty string")
+    pub fn tx_hash_as_hex(&self, hex_prefix: bool) -> nonempty::String {
```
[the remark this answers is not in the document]
themicp: Right, https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/376137a96b7191698f4621ed24815f5453c8a2b0

packages/xrpl-types/src/types.rs
```diff
@@ -0,0 +1,1961 @@
...
+}
+
+// XRPLTokenAmount always in canonicalized XRPL mantissa-exponent format,
+// such that MIN_MANTISSA <= mantissa <= MAX_MANTISSA (or equal to zero), MIN_EXPONENT <= exponent <= MAX_EXPONENT,
+// In XRPL generally it can be decimal and even negative (!) but in our case that doesn't apply.
+#[cw_serde]
+#[derive(Eq, Hash)]
+pub struct XRPLTokenAmount {
+    mantissa: u64,
+    exponent: i64,
```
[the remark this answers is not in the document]
themicp: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/f5f7ab80d31bd89e5bb0136cf22c097e87d80451

packages/xrpl-types/src/types.rs
```diff
@@ -0,0 +1,1961 @@
...
+pub struct XRPLCurrency([u8; XRPL_CURRENCY_LENGTH]);
+
+impl XRPLCurrency {
+    pub fn new(s: &str) -> Result<Self, XRPLError> {
+        if s == "XRP" || !CURRENCY_CODE_REGEX.is_match(s) {
+            return Err(XRPLError::InvalidCurrency);
+        }
+
+        let mut buffer = [0u8; XRPL_CURRENCY_LENGTH];
+        buffer[12..15].copy_from_slice(s.as_bytes());
```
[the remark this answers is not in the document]
themicp: Updated the encoding to match the spec: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/67f39015d26a17dbdfabf7ef6f7717320672c83a

packages/xrpl-types/src/types.rs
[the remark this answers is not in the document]
themicp: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/f5f7ab80d31bd89e5bb0136cf22c097e87d80451

packages/xrpl-types/src/types.rs
```diff
@@ -0,0 +1,1961 @@
...
+
+lazy_static! {
+    static ref CURRENCY_CODE_REGEX: Regex =
+        Regex::new(r"^[A-Za-z0-9\?\!@#\$%\^&\*<>\(\)\{\}\[\]\|]{3}$").expect("valid regex");
+}
+
+const MIN_MANTISSA: u64 = 1_000_000_000_000_000;
+const MAX_MANTISSA: u64 = 10_000_000_000_000_000 - 1;
+const MIN_EXPONENT: i64 = -96;
+const MAX_EXPONENT: i64 = 80;
```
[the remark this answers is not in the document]
themicp: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/baaad5a592c29d222420a4a136f5d83d06d167cc

packages/axelar-wasm-std/src/msg_id/tx_hash.rs
```diff
@@ -11,13 +12,24 @@ use super::Error;
 use crate::hash::Hash;
 use crate::nonempty;
 
+#[cw_serde]
+#[derive(Eq, Hash)]
+#[serde(transparent)]
 pub struct HexTxHash {
+    #[serde(flatten)]
```
cjcobb23, line 19: why is this necessary?

ampd/src/handlers/xrpl_verify_msg.rs
```diff
@@ -0,0 +1,195 @@
...
+        .await
+        .into_iter()
+        .filter_map(std::result::Result::unwrap_or_default)
+        .filter_map(|tx_res| {
+            let tx_common = tx_res.tx.common();
+            let (ledger_index, tx_hash) = (tx_common.ledger_index?, tx_common.hash.clone()?);
+
+            if ledger_index.ge(&latest_validated_ledger_index) {
+                return None;
+            }
```
cjcobb23, line 100: you don't need this check. You can just check for the `validated` field in the `tx` response. Will save you an rpc call per message.

packages/xrpl-types/src/msg.rs
```diff
@@ -0,0 +1,423 @@
...
+
+/// Represents an XRPL `Payment` transaction towards the XRPL multisig,
+/// performed by an XRPL user to initiate an interchain transfer call.
+/// Such messages are verified by the XRPL Voting Verifier and routed by the Router.
+#[cw_serde]
+#[derive(Eq, Hash)]
+pub struct XRPLInterchainTransferMessage {
+    #[serde(with = "hex_tx_hash")]
+    #[schemars(with = "String")] // necessary attribute in conjunction with #[serde(with ...)]
+    pub tx_id: HexTxHash,
```
cjcobb23, line 116:
```suggestion
    pub msg_id: HexTxHash,
```

I would apply this change to all of the messages

packages/axelar-wasm-std/src/msg_id/tx_hash.rs
themicp: It is redundant for this struct when combined with `#[serde(transparent)]`, removed it.

packages/axelar-wasm-std/src/msg_id/tx_hash.rs
```diff
@@ -11,13 +12,24 @@ use super::Error;
 use crate::hash::Hash;
 use crate::nonempty;
 
+#[cw_serde]
+#[derive(Eq, Hash)]
+#[serde(transparent)]
```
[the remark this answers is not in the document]
themicp: In most places where we use `HexTxHash`, we'd prefer to flatten the structure and serialize it as a string, like for the `tx_id` of a message for example.

packages/xrpl-types/src/msg.rs
```diff
@@ -0,0 +1,428 @@
...
+    pub source_address: XRPLAccountId,
+    pub destination_chain: ChainName,
+    pub destination_address: nonempty::String,
+    /// for better user experience, the payload hash gets encoded into hex at the edges (input/output),
+    /// but internally, we treat it as raw bytes to enforce its format.
+    #[serde(with = "hex_option")]
+    #[schemars(with = "String")]
+    pub payload_hash: Option<[u8; 32]>,
+    /// The total amount of tokens sent to the XRPL multisig,
+    /// including the gas fee.
```
themicp, line 127: Bringing over a comment from @cjcobb23:
> I feel like this is a bit confusing, to include the gas fee in the amount. Is there a good reason for this? I feel it would be more readable to just have this be the transfer amount. Maybe could rename to transfer_amount.

ampd/src/handlers/xrpl_verify_msg.rs
k4m4: yep, removed: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/fe41627e7bad6c7c35250db5504945452fd86e83

ampd/src/handlers/config.rs
```diff
@@ -25,13 +24,13 @@ pub enum Config {
...
         rpc_timeout: Option<Duration>,
     },
     EvmVerifierSetVerifier {
         cosmwasm_contract: TMAddress,
         #[serde(flatten, with = "chain")]
-        chain: Chain,
+        chain: Chain<crate::evm::finalizer::Finalization>,
```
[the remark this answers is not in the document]
k4m4: XRPL finalizer was redundant, so we removed it: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/fe41627e7bad6c7c35250db5504945452fd86e83

ampd/src/handlers/config.rs
```diff
@@ -47,6 +46,16 @@ pub enum Config {
         rpc_url: Url,
         rpc_timeout: Option<Duration>,
     },
+    XRPLMsgVerifier {
+        cosmwasm_contract: TMAddress,
+        #[serde(flatten, with = "chain")]
+        chain: Chain<crate::xrpl::finalizer::Finalization>,
```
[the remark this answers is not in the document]
k4m4: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/fe41627e7bad6c7c35250db5504945452fd86e83

ampd/src/handlers/config.rs
k4m4: The XRPL finalizer was redundant, so we removed it: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/fe41627e7bad6c7c35250db5504945452fd86e83

ampd/src/xrpl/finalizer.rs
k4m4: Removed the XRPL finalizer module altogether. Just checking if the TX is `validated`.

ampd/src/handlers/xrpl_multisig.rs
```diff
@@ -0,0 +1,437 @@
...
+        } = match event.try_into() as error_stack::Result<_, _> {
+            Err(report) if matches!(report.current_context(), EventTypeMismatch(_)) => {
+                return Ok(vec![]);
+            }
+            result => result.change_context(DeserializeEvent)?,
+        };
+
+        info!(
+            session_id = session_id,
+            msg = encode(&unsigned_tx),
```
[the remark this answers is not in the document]
k4m4: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/14629b45f13c878de41bd65d1502440cc732bf9e

packages/axelar-wasm-std/src/msg_id/tx_hash.rs
themicp: https://github.com/commonprefix/axelar-amplifier/commit/24265ca3f85a93c4e7f4edb4da6914fc815a79db

packages/xrpl-types/src/msg.rs
themicp: https://github.com/commonprefix/axelar-amplifier/commit/28c5db6bd97ca14ecd86330984e877953050fea6

ampd/Cargo.toml
themicp: Linking to the relevant discussion on Slack: https://commonprefix.slack.com/archives/C087CDX0AG7/p1741980290993409

ampd/src/handlers/multisig.rs
```diff
@@ -54,6 +53,7 @@ where
 pub struct Handler<S> {
     verifier: TMAddress,
     multisig: TMAddress,
+    chain: ChainName,
```
cjcobb23, line 56: I was actually thinking the handler would take a list of chains, instead of one handler per chain. This is interesting though, because this effectively changes the execution model, because now signing could run in parallel for several chains. Though this will dramatically increase the number of tasks. I actually prefer this model, since each chain has a dedicated voting handler, and so it makes sense to do the same for signing. 

However, I'd prefer to do this as a separate PR, so that way we don't force verifiers to update their configs as part of this release. There's already enough moving parts.

ampd/src/xrpl/verifier.rs
```diff
@@ -0,0 +1,423 @@
...
+    multisig_address: &XRPLAccountId,
+    message: &XRPLAddGasMessage,
+    memos: &HashMap<String, String>,
+) -> bool {
+    if let Payment(payment_tx) = &tx {
+        payment_tx.destination == multisig_address.to_string()
+            && verify_delivered_full_amount(tx, payment_tx.amount.clone())
+            && verify_amount(message.amount.clone(), payment_tx.amount.clone())
+            && verify_memo(memos, "type", XRPLMessageType::AddGas.to_string())
+            && verify_memo(memos, "tx_id", message.msg_tx_id.to_string())
```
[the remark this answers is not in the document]
themicp: https://github.com/commonprefix/axelar-amplifier/commit/e5b88ef63558295648fd1c5d560dae083ef0a9c3

ampd/src/handlers/multisig.rs
k4m4: Created [a separate PR](https://github.com/axelarnetwork/axelar-amplifier/pull/772) and [reverted the commit](https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/849e724931bf8c6a4676fa83b8c1729c73b7e02d) from here (`xrpl-ampd`).

packages/xrpl-types/src/msg.rs
k4m4: We're using `msg_id` in AddGas to refer to the TX hash of the message that's being topped up: https://github.com/axelarnetwork/axelar-amplifier/pull/764/commits/e5b88ef63558295648fd1c5d560dae083ef0a9c3

But we're still using `tx_id` to refer to the TX hash of the current message everywhere

packages/xrpl-types/src/types.rs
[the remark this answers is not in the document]
themicp: Did another pass: https://github.com/commonprefix/axelar-amplifier/commit/85d179fd5afc4ffc1253b69effab636e5426d559

packages/xrpl-types/src/msg.rs
themicp: Relevant commit from `xrpl` branch: https://github.com/commonprefix/axelar-amplifier/commit/42acb5b65104ddd7d4b2af83a3bc28a735913a3f

PR #767, opened 2025-03-14
fix(minor-ampd)!: ignore fee estimation failures

cjcobb23:


ampd/src/event_processor.rs
```diff
@@ -122,10 +120,12 @@ where
     match future::with_retry(|| handler.handle(event), retry_policy).await {
         Ok(msgs) => {
             for msg in msgs {
-                broadcaster
-                    .broadcast(msg)
-                    .await
-                    .change_context(Error::Broadcaster)?;
+                if let Err(err) = broadcaster.broadcast(msg.clone()).await {
+                    warn!(
```
milapsheth, line 124:
```suggestion
                    error!(
```
Since it's not really expected, and probably means there's a bug?

milapsheth (approved):

ampd/src/event_processor.rs
cjcobb23: eh usually errors crash the process, i hesitate to escalate this to error when it's not crashing the process now

fish-sammy: agree with CJ. warn is more suitable than error here

fish-sammy (approved):

PR #774, opened 2025-03-21
feat(minor-interchain-token-service): register p2p token

cjcobb23:
* add an execute msg to register a p2p token for a particular chain

https://axelarnetwork.atlassian.net/browse/AXE-8087


Foivos (approved):

contracts/interchain-token-service/src/msg.rs
```diff
@@ -31,6 +31,18 @@ pub enum ExecuteMsg {
...
     Execute(AxelarExecutableMsg),
 
+    /// Registers an existing ITS token with the hub. This is useful for tokens that were deployed
+    /// before the hub existed and have operated in p2p mode. Both instance_chain and origin_chain
+    /// must be registered with the hub.
+    #[permission(Elevated)]
+    RegisterP2pTokenInstance {
+        instance_chain: ChainNameRaw,
+        origin_chain: ChainNameRaw,
+        token_id: TokenId,
```
milapsheth, line 41:
```suggestion
        chain: ChainNameRaw,
        token_id: TokenId,
        origin_chain: ChainNameRaw,
```
perhaps

contracts/interchain-token-service/src/contract/execute/mod.rs
```diff
@@ -408,6 +424,52 @@ fn update_chain(storage: &mut dyn Storage, config: msg::ChainConfig) -> Result<(
     Ok(())
 }
 
+pub fn register_p2p_token_instance(
+    deps: DepsMut,
```
milapsheth, line 428:
```suggestion
    storage: &mut dyn Storage,
```
for consistency

contracts/interchain-token-service/src/contract/execute/mod.rs
```diff
@@ -408,6 +424,52 @@ fn update_chain(storage: &mut dyn Storage, config: msg::ChainConfig) -> Result<(
...
+        instance_chain.clone(),
+        token_id,
+        &TokenInstance { supply, decimals },
+    )
+    .change_context(Error::State)?;
+
+    match state::may_load_token_config(deps.storage, &token_id).change_context(Error::State)? {
+        Some(token_config) => {
+            ensure!(
+                token_config.origin_chain == origin_chain,
```
milapsheth, line 459: add a comment explaining why this doesn't fail if the config is already exists

contracts/interchain-token-service/src/contract/execute/mod.rs
```diff
@@ -408,6 +424,52 @@ fn update_chain(storage: &mut dyn Storage, config: msg::ChainConfig) -> Result<(
...
+    state::save_token_instance(
+        deps.storage,
+        instance_chain.clone(),
+        token_id,
+        &TokenInstance { supply, decimals },
+    )
+    .change_context(Error::State)?;
+
+    match state::may_load_token_config(deps.storage, &token_id).change_context(Error::State)? {
+        Some(token_config) => {
```
milapsheth, line 457:
```suggestion
        Some(TokenConfig { origin_chain: stored_origin_chain }) => {
```
could also destructure

contracts/interchain-token-service/src/contract/execute/mod.rs
```diff
@@ -1189,6 +1246,317 @@ mod tests {
...
+    }
+
+    fn cc_id(source_chain: ChainNameRaw) -> CrossChainId {
+        CrossChainId {
+            source_chain,
+            message_id: HexTxHashAndEventIndex::new([1u8; 32], 0u32).into(),
+        }
+    }
+
+    fn do_transfer(
```
milapsheth, line 1485:
```suggestion
    fn transfer_token(
```
`do_*` naming feels inconsistent with other functions

contracts/interchain-token-service/src/contract/execute/mod.rs
```diff
@@ -1189,6 +1246,317 @@ mod tests {
...
+        };
+        execute_message(
+            deps,
+            cc_id(from),
+            ITS_ADDRESS.to_string().try_into().unwrap(),
+            msg.abi_encode(),
+        )
+    }
+
+    fn do_deploy(
```
milapsheth, line 1511:
```suggestion
    fn deploy_token(
```

contracts/interchain-token-service/src/contract/execute/mod.rs
```diff
@@ -1189,6 +1246,317 @@ mod tests {
...
+            token_id,
+            "Test".parse().unwrap(),
+            "TEST".parse().unwrap(),
+            18,
+            None,
+        )
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    fn do_deploy_helper(
```
milapsheth, line 1530:
```suggestion
    fn deploy_token_with_metadata(
```

contracts/interchain-token-service/src/contract/execute/mod.rs
```diff
@@ -1189,6 +1246,317 @@ mod tests {
...
+    #[test]
+    fn should_register_p2p_tokens() {
+        let mut deps = mock_dependencies();
+        init(&mut deps);
+
+        let origin_chain = ethereum();
+        let instance_chains: Vec<ChainNameRaw> = vec![solana(), ethereum()];
+        let decimals = 18;
+        let supply = TokenSupply::Tracked(Uint256::one());
+        for chain in instance_chains {
```
milapsheth, line 1258:
```suggestion
        let supply = TokenSupply::Tracked(Uint256::one());

        for chain in instance_chains {
```

contracts/interchain-token-service/src/msg.rs
```diff
@@ -31,6 +31,18 @@ pub enum ExecuteMsg {
     #[permission(Specific(gateway))]
     Execute(AxelarExecutableMsg),
 
+    /// Registers an existing ITS token with the hub. This is useful for tokens that were deployed
+    /// before the hub existed and have operated in p2p mode. Both instance_chain and origin_chain
+    /// must be registered with the hub.
+    #[permission(Elevated)]
```
milapsheth, line 37: We should consider adding another role for these methods that's separate from the admin role, to make the migration easier to perform. It doesn't modify existing token instances so the risk is lower

contracts/interchain-token-service/src/contract/execute/mod.rs
cjcobb23: the pub functions in this file actually take `DepsMut` and only the private functions accept `Storage`, so it's actually consistent as is

contracts/interchain-token-service/src/msg.rs
cjcobb23: the roles are part of the permission control macro, and there's only two right now, admin and governance. What would the third role be? Operator?

Also, I'd like to push such a change to a separate PR.

contracts/interchain-token-service/src/contract/execute/mod.rs
```diff
@@ -408,6 +424,52 @@ fn update_chain(storage: &mut dyn Storage, config: msg::ChainConfig) -> Result<(
...
+    .change_context(Error::State)?;
+
+    match state::may_load_token_config(deps.storage, &token_id).change_context(Error::State)? {
+        Some(token_config) => {
+            ensure!(
+                token_config.origin_chain == origin_chain,
+                report!(Error::WrongOriginChain {
+                    token_id,
+                    origin_chain: token_config.origin_chain
+                })
```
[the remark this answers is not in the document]
cjcobb23: it wasnt

contracts/interchain-token-service/src/msg.rs
```diff
@@ -31,6 +31,18 @@ pub enum ExecuteMsg {
...
+    RegisterP2pTokenInstance {
```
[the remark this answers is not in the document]
cjcobb23: would prefer to keep it as is, the consensus during spec review was to name it something very specific to its usecase

contracts/interchain-token-service/src/msg.rs
cgorenflo: we can use the `Specific` modifier for custom permission checks, but as @cjcobb23 said, that should be done in a separate PR if necessary

contracts/interchain-token-service/src/msg.rs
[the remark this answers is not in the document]
cgorenflo: agreed. This endpoint is specifically created to migrate the p2p tokens, nothing else

cgorenflo (approved):

contracts/interchain-token-service/src/msg.rs
milapsheth: yeah, we should add a `Specific(operator)`. This operator role will be useful for other things like setting a rate limit in the future. And yes, it should be added in a separate PR

milapsheth (approved):

PR #777, opened 2025-03-24
refactor(minor-ampd): refactor the event sub to support the gRPC function Subscribe

fish-sammy:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros

### DIFF
--- a/ampd/src/event_sub.rs
+++ b/ampd/src/event_sub.rs
@@ -3,8 +3,9 @@ use std::time::Duration;
 
 use error_stack::{FutureExt, Report, Result, ResultExt};
 use events::Event;
-use futures::TryStreamExt;
+use futures::{StreamExt, TryStreamExt};
 use mockall::automock;
+use report::LoggableError;
 use tendermint::block;
 use thiserror::Error;
 use tokio::sync::broadcast::{self, Sender};
@@ -13,36 +14,53 @@ use tokio_stream::wrappers::errors::BroadcastStreamRecvError;
 use tokio_stream::wrappers::BroadcastStream;
 use tokio_stream::Stream;
 use tokio_util::sync::CancellationToken;
-use tracing::info;
+use tracing::{error, info};
+use valuable::Valuable;
 
 use crate::tm_client::TmClient;
 
+#[derive(Error, Debug, Clone)]
+pub enum Error {
+    #[error("failed querying the block results for block {block}")]
+    BlockResultsQuery { block: block::Height },
+    #[error("failed querying the latest block")]
+    LatestBlockQuery,
+    #[error("failed decoding event in block {block}")]
+    EventDecoding { block: block::Height },
+    #[error("failed receiving event from broadcast stream")]
+    BroadcastStreamRecv(#[from] BroadcastStreamRecvError),
+}
+
 #[automock]
 pub trait EventSub {
-    fn subscribe(
-        &self,
-    ) -> impl Stream<Item = Result<Event, BroadcastStreamRecvError>> + Send + 'static;
+    fn subscribe(&self) -> impl Stream<Item = Result<Event, Error>> + Send + 'static;
 }
 
 pub struct EventSubscriber {
-    tx: Sender<Event>,
+    tx: Sender<std::result::Result<Event, Error>>,
 }
 
 impl EventSub for EventSubscriber {
-    fn subscribe(&self) -> impl Stream<Item = Result<Event, BroadcastStreamRecvError>> + 'static {
-        BroadcastStream::new(self.tx.subscribe()).map_err(Report::from)
+    fn subscribe(&self) -> impl Stream<Item = Result<Event, Error>> + 'static {
+        BroadcastStream::new(self.tx.subscribe())
+            .map(|event| match event {
+                Ok(Ok(event)) => Ok(event),
+                Ok(Err(err)) => Err(err),
+                Err(err) => Err(err.into()),
+            })
+            .map_err(Report::from)
     }
 }
 
 pub struct EventPublisher<T: TmClient + Sync> {
     tm_client: T,
     poll_interval: Duration,
-    tx: Sender<Event>,
+    tx: Sender<std::result::Result<Event, Error>>,
 }
 
 impl<T: TmClient + Sync> EventPublisher<T> {
     pub fn new(client: T, capacity: usize) -> (Self, EventSubscriber) {
-        let (tx, _) = broadcast::channel::<Event>(capacity);
+        let (tx, _) = broadcast::channel(capacity);
         let publisher = EventPublisher {
             tm_client: client,
             poll_interval: Duration::new(5, 0),
@@ -53,14 +71,14 @@ impl<T: TmClient + Sync> EventPublisher<T> {
         (publisher, subscriber)
     }
 
-    pub async fn run(mut self, token: CancellationToken) -> Result<(), EventSubError> {
-        let mut curr_block_height = self.latest_block_height().await?;
+    pub async fn run(mut self, token: CancellationToken) -> Result<(), Error> {
+        let mut curr_block_height = latest_block_height(&self.tm_client).await?;
         let mut interval = time::interval(self.poll_interval);
 
         loop {
             select! {
                 _ = interval.tick() => {
-                    curr_block_height = self.process_blocks_from(curr_block_height, &token).await?.increment();
+                    curr_block_height = self.process_blocks_from(curr_block_height, &token).await;
                 },
                 _ = token.cancelled() => {
                     info!("exiting event sub");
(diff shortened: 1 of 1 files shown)


ampd/src/event_sub.rs
```diff
@@ -71,91 +89,104 @@ impl<T: TmClient + Sync> EventPublisher<T> {
...
             return Ok(());
         }
 
         let events = iter::once(Event::BlockBegin(height))
-            .chain(self.events(height).await?.into_iter())
+            .chain(events(&self.tm_client, height).await?.into_iter())
             .chain(iter::once(Event::BlockEnd(height)));
 
         for event in events {
             // ignore the error if there are no subscribers
```
cgorenflo, line 146:
```suggestion
            // send fails only when a subscriber has dropped unexpectedly. This should not affect block processing, so ignore the error
```

ampd/src/event_sub/mod.rs
```diff
@@ -53,109 +72,38 @@ impl<T: TmClient + Sync> EventPublisher<T> {
         (publisher, subscriber)
     }
 
-    pub async fn run(mut self, token: CancellationToken) -> Result<(), EventSubError> {
-        let mut curr_block_height = self.latest_block_height().await?;
-        let mut interval = time::interval(self.poll_interval);
+    pub async fn run(self, token: CancellationToken) -> Result<(), Error> {
+        let mut event_stream =
+            event_stream::event_stream(&self.tm_client, self.poll_interval, 10000, || {
```
maancham, line 77: can we consider making the buffer size configureable?

ampd/src/event_sub/event_stream.rs
```diff
@@ -0,0 +1,272 @@
...
+
+    fn stream<'a, T>(
+        self,
+        tm_client: &'a T,
+        interval: Interval,
+    ) -> impl Stream<Item = Result<block::Height>> + 'a
+    where
+        T: TmClient,
+    {
+        futures::stream::unfold(
```
maancham, line 137: Is it possible to have excessive memory use here? If the downstream consumer processes items slower than they're produced (can this happen?), memory could accumulate in buffers, especially with buffer size being set to 10k.

ampd/src/event_sub/mod.rs
fish-sammy: We can, but it's a weird one. I made a constant for it now. It isn't the capacity of the stream, but pending futures that are allowed to be polled at any given time. Not sure how would anyone not familiar with the code reason and configure it. @cgorenflo thought?

ampd/src/event_sub/event_stream.rs
fish-sammy: Again, the buffer is the max number of pending futures to be polled at the same time. It can't be overflown no matter what.

ampd/src/event_sub/event_stream.rs
```diff
@@ -0,0 +1,272 @@
...
+    skip_block_fn: impl Fn() -> bool,
+) -> Result<Vec<Event>>
+where
+    T: TmClient,
+{
+    match (skip_block_fn(), block_height) {
+        (true, _) => Ok(vec![]),
+        (_, Ok(block_height)) => events(tm_client, block_height).await,
+        (_, Err(err)) => Err(err),
+    }
```
cgorenflo:
```suggestion
    if skip_block(){
        Ok(vec![])
    else{
        events(tm_client, block_height?).await,
    }
```
might be cleaner

ampd/src/event_sub/event_stream.rs
```diff
@@ -0,0 +1,272 @@
...
+where
+    T: TmClient,
+{
+    latest_block_height(tm_client)
+        .await
+        .map(|latest| BlockState {
+            current: latest,
+            latest,
+        })
+        .map(|block_state| block_state.stream(tm_client, interval(poll_interval)))
```
cgorenflo: I'm not sure this works as you expect it to. This will create a new interval for each block height, instead of reusing the same one

ampd/src/event_sub/event_stream.rs
```diff
@@ -0,0 +1,272 @@
...
+    where
+        T: TmClient,
+    {
+        futures::stream::unfold(
+            (self, tm_client, interval),
+            |(block_state, tm_client, mut interval)| async move {
+                let current = block_state.current;
+
+                match block_state.update(tm_client, &mut interval).await {
+                    Ok(block_state) => Some((Ok(current), (block_state, tm_client, interval))),
```
cgorenflo: this is going to emit a block 1 in the future

ampd/src/event_sub/mod.rs
```diff
@@ -53,109 +72,38 @@ impl<T: TmClient + Sync> EventPublisher<T> {
...
-                _ = interval.tick() => {
-                    curr_block_height = self.process_blocks_from(curr_block_height, &token).await?.increment();
+                event = event_stream.next() => match event {
+                    Some(event) => {
+                        let event = event
+                            .inspect_err(|err| {
+                                error!(err = LoggableError::from(err).as_value(), "failed to subscribe to events");
+                            })
+                            .map_err(|err| err.current_context().clone());
+                        let _ = self.tx.send(event);
```
cgorenflo, line 98: I think we should log this error, too

ampd/src/event_sub/mod.rs
```diff
@@ -53,109 +72,38 @@ impl<T: TmClient + Sync> EventPublisher<T> {
...
 
         loop {
             select! {
-                _ = interval.tick() => {
-                    curr_block_height = self.process_blocks_from(curr_block_height, &token).await?.increment();
+                event = event_stream.next() => match event {
+                    Some(event) => {
+                        let event = event
+                            .inspect_err(|err| {
+                                error!(err = LoggableError::from(err).as_value(), "failed to subscribe to events");
```
cgorenflo, line 95: this is less a subscription error and more of a "couldn't get specific event" error, isn't it?

cgorenflo (changes requested):
the solution looks much better structured now

ampd/src/event_sub/stream.rs
```diff
@@ -0,0 +1,479 @@
...
+use tokio::time::{interval, Interval};
+use tokio_util::sync::CancellationToken;
+
+use crate::asyncutil::future::{with_retry, RetryPolicy};
+use crate::tm_client::TmClient;
+
+type Error = super::Error;
+type Result<T> = error_stack::Result<T, Error>;
+
+const BLOCK_PROCESSING_BUFFER: usize = 10;
```
cgorenflo, line 18: give a brief explanation for why this is 10

ampd/src/event_sub/stream.rs
```diff
@@ -0,0 +1,479 @@
...
+}
+
+#[derive(Copy, Clone)]
+struct BlockState {
+    current: block::Height,
+    latest: block::Height,
+}
+
+impl BlockState {
+    async fn update<T>(mut self, tm_client: &T, interval: &mut Interval) -> Result<Self>
```
cgorenflo, line 117: pass in cancellation? This could hang ampd on a chain halt (e.g. upgrade) otherwise

ampd/src/event_sub/stream.rs
```diff
@@ -0,0 +1,479 @@
...
+                }
+            },
+        )
+    }
+}
+
+async fn latest_block_height<T: TmClient>(tm_client: &T) -> Result<block::Height> {
+    tm_client
+        .latest_block()
+        .await
```
cgorenflo, line 165: does this have a timeout?

ampd/src/event_sub/stream.rs
```diff
@@ -0,0 +1,479 @@
...
+        let retry_pocily = RetryPolicy::RepeatConstant {
+            sleep: Duration::from_millis(100),
+            max_attempts: 3,
+        };
+        let block_stream = stream::iter(vec![
+            Err(report!(Error::LatestBlockQuery)),
+            Err(report!(Error::LatestBlockQuery)),
+            Err(report!(Error::LatestBlockQuery)),
+        ]);
+        let mut stream = events(&tm_client, block_stream, retry_pocily);
```
maancham, line 194: we have a typo here and on line 185.

ampd/src/event_sub/stream.rs
fish-sammy: Not sure, It probably has a default one that's used by the underlying library. We can do a separate PR to apply a timeout ourselves.

ampd/src/event_sub/event_stream.rs
fish-sammy: How does it happen?

ampd/src/event_sub/event_stream.rs
fish-sammy: This `.map` is applied on a `Result` not a stream, so I'm not sure how does it "create a new interval for each block height"?

cgorenflo (approved):

PR #780, opened 2025-03-26
fix(xrpl-types)!: arithmetic operations with XRPLTokenAmount

themicp:
## Description
When the orders of magnitude difference between two amounts was large (> 20), adjusting mantissas to a common exponent would overflow `u64`. For example, adding "2000000000000000e-21" and "1e1" required scaling the second amount down by 21 orders of magnitude, exceeding `u64`'s capacity.

Solution: Represent mantissas as `Uint256` during intermediate calculations to handle the large scale differences. Scale mantissas to a common exponent, perform arithmetic operations, then canonicalize the result back to XRPL's mantissa-exponent format.

### DIFF
--- a/packages/xrpl-types/src/types.rs
+++ b/packages/xrpl-types/src/types.rs
@@ -939,77 +939,13 @@ impl XRPLTokenAmount {
 
 impl PartialOrd for XRPLTokenAmount {
     fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
-        // If either amount is zero, we can compare directly
-        if self.mantissa == 0 {
-            return if other.mantissa == 0 {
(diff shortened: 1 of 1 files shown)


packages/xrpl-types/src/types.rs
```diff
@@ -1376,6 +1260,45 @@ pub fn scale_to_decimals(
     }
 }
 
+pub fn scale_to_common_exponent(
+    left: &XRPLTokenAmount,
+    right: &XRPLTokenAmount,
+    common_exponent: i64,
```
cjcobb23, line 1266: I would just compute this within the function, and return it

cjcobb23 (approved):

PR #782, opened 2025-03-26
feat(minor-interchain-token-service): modify supply

cjcobb23:
* Add an execute msg to modify the supply for a deployed token. This includes increasing or decreasing the supply, as well as setting the supply for a token where the supply is currently untracked.

https://axelarnetwork.atlassian.net/browse/AXE-8089

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros


contracts/interchain-token-service/src/contract/execute/mod.rs
```diff
@@ -457,11 +499,200 @@ mod tests {
...
+                deps.as_mut(),
+                solana(),
+                ethereum(),
+                token_id(),
+                Uint256::one().try_into().unwrap()
+            ),
+            Error,
+            Error::TokenSupplyInvariantViolated { .. }
+        );
+    }
```
cgorenflo, line 536: looking at these tests I have no idea why the first one should succeed and the second one shouldn't

contracts/interchain-token-service/src/contract/execute/mod.rs
cgorenflo, line 1: I don't think we have a query to get the supply for a token, do we? I think we should add that and also use it in tests to make sure the state is what we expected

Foivos (approved):

contracts/interchain-token-service/src/msg.rs
```diff
@@ -38,6 +39,17 @@ pub enum ExecuteMsg {
...
+    // If the supply is untracked, this command will attempt to set it.
+    // Errors if the token is not deployed to the specified chain, or if
+    // the supply modification overflows or underflows
+    #[permission(Elevated)]
+    ModifySupply {
+        chain: ChainNameRaw,
+        token_id: TokenId,
+        supply_modifier: SupplyModifier,
+    },
+
```
milapsheth, line 52: I'd prefer having two entrypoints `IncreaseTokenSupply` and `DecreaseTokenSupply` for simplicity. The implementation could use a modify/update supply internally for reuse

contracts/interchain-token-service/src/contract/execute/mod.rs
```diff
@@ -408,6 +411,43 @@ fn update_chain(storage: &mut dyn Storage, config: msg::ChainConfig) -> Result<(
...
+            .checked_add(amount)
+            .change_context(Error::ModifySupplyOverflow(token_instance.supply))?,
+        SupplyModifier::DecreaseSupply(amount) => token_instance
+            .supply
+            .clone()
+            .checked_sub(amount)
+            .change_context(Error::ModifySupplyOverflow(token_instance.supply))?,
+    };
+
+    state::save_token_instance(deps.storage, chain.clone(), token_id, &token_instance)
```
milapsheth, line 445: emit an event for visibility

contracts/interchain-token-service/src/contract/execute/mod.rs
cjcobb23: we expose a query to get the token instance, which includes the supply. It's in the query module though and just returns a blob, so not the best to use here. So I just queried the state directly (which is what the query module does anyways), and asserted the supply is what we expect.

contracts/interchain-token-service/src/contract/execute/mod.rs
cjcobb23: added some comments and generally tried to clean these up, lmk what you think

contracts/interchain-token-service/src/msg.rs
cjcobb23: What would be the purpose of that? It will be more code and more complexity in the contract if we have two endpoints, so without a good reason to have two, I prefer one endpoint. A user isn't any more likely to mess something up as it is now versus if there are two endpoints.

cgorenflo: I don't see any benefit in splitting the endpoint either. Could you elaborate if you feel strongly about it @milapsheth?

milapsheth: Not a dealbreaker, but I find it more readable to have two simpler entrypoints for other teams looking at it, and I don't like exposing the complexity to the user due to internal implementation reuse details.
```
IncreaseSupply {
        chain: ChainNameRaw,
        token_id: TokenId,
        amount: nonempty::Uint256,
    },
```
and the JSON looks like
`{"increase_supply": { "chain": "sui", "token_id": "...", "amount": "123" } }`

instead of
`{"modify_supply": { "chain": "sui", "token_id": "...", "supply_modifier": { "increase_supply": "123" } } }`

It's common to see this in other [APIs](https://docs.openzeppelin.com/contracts/2.x/api/token/erc20#ERC20-increaseAllowance-address-uint256-) too, granted it's not as bad as using a bool for the modifier

milapsheth (approved):

contracts/interchain-token-service/src/msg.rs
cjcobb23: hmm you make a good point. However, the users of this are just the operators of the contract, which is just our devops team, and the commands will likely be wrapped in a script. And this contract is already overly complex imo, so I prefer to err on the side of keeping the code simple here. If this API was more user facing, i.e. called by many third parties, external to our org, I would feel differently. Keeping it as is for now.

PR #785, opened 2025-03-28
fix(ampd): chain name comparison

eguajardo:
## Description

fixes chain name comparison to make it case insensitive

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros

### DIFF
--- a/ampd/src/evm/verifier.rs
+++ b/ampd/src/evm/verifier.rs
@@ -21,7 +21,7 @@ impl PartialEq<IAxelarGatewayEventsWithLog<'_>> for &Message {
                 {
                     Ok(chain) => chain,
                     Err(e) => {
-                        debug!(error = e, "failed to parse destination chain");
+                        debug!(error = ?e, "failed to parse destination chain");
                         return false;
                     }
                 };
--- a/ampd/src/evm/verifier.rs
+++ b/ampd/src/evm/verifier.rs
@@ -22,7 +22,7 @@ impl PartialEq<IAxelarGatewayEventsWithLog<'_>> for &Message {
                         Ok(chain) => self.destination_chain == chain,
                         Err(e) => {
                             debug!(error = ?e, "failed to parse destination chain");
-                            return false;
+                            false
                         }
                     };
 
--- a/ampd/src/evm/verifier.rs
+++ b/ampd/src/evm/verifier.rs
@@ -3,6 +3,7 @@ use ethers_contract::EthLogDecode;
 use ethers_core::types::{Log, TransactionReceipt, H256};
 use evm_gateway::{IAxelarAmplifierGatewayEvents, WeightedSigners};
 use router_api::ChainName;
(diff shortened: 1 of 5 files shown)


ampd/src/evm/verifier.rs
```diff
@@ -17,7 +18,9 @@ impl PartialEq<IAxelarGatewayEventsWithLog<'_>> for &Message {
             IAxelarAmplifierGatewayEvents::ContractCallFilter(event) => {
                 log.transaction_hash == Some(self.message_id.tx_hash.into())
                     && event.sender == self.source_address
-                    && self.destination_chain == event.destination_chain
+                    && self.destination_chain
+                        == ChainName::try_from(event.destination_chain.as_ref())
+                            .expect("failed to parse to ChainName")
```
milapsheth, line 23: user provides this value in the event, so the conversion to ChainName can fail. If the conversion, then this should automatically return false. It might be useful to have a debug log and print the conversion error too

cgorenflo: agree with both points (no expect + add logging in case of conversion failure)

cjcobb23 (commented):
I think we should fix this for XRPL as well

ampd/src/evm/verifier.rs
```diff
@@ -15,9 +17,18 @@ impl PartialEq<IAxelarGatewayEventsWithLog<'_>> for &Message {
...
         match event {
             IAxelarAmplifierGatewayEvents::ContractCallFilter(event) => {
+                let destination_chain = match ChainName::try_from(event.destination_chain.as_ref())
+                {
+                    Ok(chain) => chain,
+                    Err(e) => {
+                        debug!("failed to parse destination chain {:?}", e);
+                        return false;
+                    }
+                };
```
cgorenflo, line 27:
```suggestion
                let destination_chain = ChainName::try_from(event.destination_chain.as_ref())
                    .inspect_err(|e|  debug!(error=e, "failed to parse destination chain")?;
```

ampd/src/evm/verifier.rs
```diff
@@ -15,9 +17,18 @@ impl PartialEq<IAxelarGatewayEventsWithLog<'_>> for &Message {
 
         match event {
             IAxelarAmplifierGatewayEvents::ContractCallFilter(event) => {
-                log.transaction_hash == Some(self.message_id.tx_hash.into())
+                let matches_destination_chain =
+                    match ChainName::try_from(event.destination_chain.as_ref()) {
+                        Ok(chain) => self.destination_chain == chain,
+                        Err(e) => {
+                            debug!(error = ?e, "failed to parse destination chain");
+                            return false;
```
cgorenflo, line 25: don't return here

cgorenflo (approved):

PR #797, opened 2025-04-04
fix(xrpl-types): parsing of nonstandard XRPL currencies

themicp:
## Description
Fixes the way nonstandard currency strings are parsed by properly decoding hex-encoded strings to bytes using `hex::decode`. Also fixes the way nonstandard currency strings are being serialized.


packages/xrpl-types/src/types.rs
```diff
@@ -802,7 +802,12 @@ impl XRPLCurrency {
     }
 
     fn from_nonstandard_str(s: &str) -> Self {
-        XRPLCurrency(s.as_bytes().try_into().expect("should be 20 bytes"))
+        XRPLCurrency(
+            hex::decode(s)
+                .expect("invalid nonstandard currency hex")
+                .try_into()
+                .expect("should be 20 bytes"),
+        )
```
cjcobb23, line 810: can this be triggered by user input at all? I'm wondering if it should return an error

packages/xrpl-types/src/types.rs
```diff
@@ -824,9 +829,18 @@ impl From<XRPLCurrency> for [u8; XRPL_CURRENCY_LENGTH] {
...
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        let str = std::str::from_utf8(&self.0[12..15])
-            .expect("currency code should always be valid UTF-8")
-            .to_string();
+        let str = match self.0[0] {
+            0x00 => {
+                // standard currency
+                std::str::from_utf8(&self.0[12..15])
+                    .expect("currency code should always be valid UTF-8")
+                    .to_string()
```
cjcobb23, line 837: same question here, can this be triggered by user input?

cjcobb23 (approved):

PR #798, opened 2025-04-04
feat(minor-interchain-token-service): add operator role

cjcobb23:
* operator can call register_p2p_token and modify_supply

https://axelarnetwork.atlassian.net/browse/AXE-8493

### DIFF
--- a/contracts/interchain-token-service/src/contract/migrations/mod.rs
+++ b/contracts/interchain-token-service/src/contract/migrations/mod.rs
@@ -1,16 +1,20 @@
 use axelar_wasm_std::error::ContractError;
 use axelar_wasm_std::migrate_from_version;
+use cosmwasm_schema::cw_serde;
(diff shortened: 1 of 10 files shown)


contracts/interchain-token-service/src/contract/migrations/mod.rs
```diff
@@ -1,16 +1,20 @@
...
 use cosmwasm_std::entry_point;
-use cosmwasm_std::{DepsMut, Empty, Env, Response};
+use cosmwasm_std::{DepsMut, Env, Response};
 
 pub mod v1_1_0;
 
-pub type MigrateMsg = Empty;
+#[cw_serde]
+pub struct MigrateMsg {
+    operator_address: String,
```
milapsheth, line 12: This should be set to the same address as the prover admin fyi for creating the release doc

milapsheth (approved):

PR #803, opened 2025-04-09
feat(minor-coordinator): add ChainContracts query endpoint

SDavidson1177:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros


integration-tests/tests/chain_contracts_details.rs
```diff
@@ -0,0 +1,33 @@
...
+
+#[test]
+fn chain_contracts_information_should_be_consistent_in_coordinator() {
+    let test_utils::TestCase {
+        mut protocol,
+        chain1: ethereum,
+        ..
+    } = test_utils::setup_test_case();
+
+    let ethereum_by = test_utils::chain_contracts_info_from_coordinator(
```
maancham, line 13: I believe the name of the variable can be set to better convey that this info is coming from coordinator (or the fact that it is query response)

contracts/coordinator/src/contract.rs
```diff
@@ -322,23 +323,52 @@ mod tests {
         )
         .unwrap();
 
-        let record_response_by_chain = contracts_by_chain(
-            test_setup.deps.as_ref().storage,
-            test_setup.chain_name.clone(),
+        let record_response_by_chain: StdResult<ChainContractsRecord> = cosmwasm_std::from_json(
```
maancham, line 326: could you add a check here to make sure the response is ok and then do the unwrap?

contracts/coordinator/src/contract.rs
```diff
@@ -322,23 +323,52 @@ mod tests {
...
-        assert_eq!(record_response_by_chain.unwrap(), test_setup.chain_record);
 
-        let record_response_by_prover =
-            contracts_by_prover(test_setup.deps.as_ref().storage, test_setup.prover.clone());
-        assert_eq!(record_response_by_prover.unwrap(), test_setup.chain_record);
+        assert_eq!(record_response_by_chain.unwrap(), test_setup.chain_record);
 
-        let record_response_by_gateway =
-            contracts_by_gateway(test_setup.deps.as_ref().storage, test_setup.gateway.clone());
+        let record_response_by_gateway: StdResult<ChainContractsRecord> = cosmwasm_std::from_json(
```
maancham, line 339: same here.

contracts/coordinator/src/contract.rs
```diff
@@ -322,23 +323,52 @@ mod tests {
...
+            )
+            .unwrap(),
+        );
         assert_eq!(record_response_by_gateway.unwrap(), test_setup.chain_record);
 
-        let record_response_by_verifier = contracts_by_verifier(
-            test_setup.deps.as_ref().storage,
-            test_setup.verifier.clone(),
+        let record_response_by_prover: StdResult<ChainContractsRecord> = cosmwasm_std::from_json(
+            query(
```
maancham, line 352: and here.

contracts/coordinator/src/contract.rs
```diff
@@ -322,23 +323,52 @@ mod tests {
...
+                test_setup.env.clone(),
+                QueryMsg::ChainContractsInfo(ChainContractsKey::ProverAddress(
+                    test_setup.prover.clone(),
+                )),
+            )
+            .unwrap(),
+        );
+        assert_eq!(record_response_by_prover.unwrap(), test_setup.chain_record);
+
+        let record_response_by_verifier: StdResult<ChainContractsRecord> = cosmwasm_std::from_json(
```
maancham, line 363: I believe this should be the last instance.

contracts/coordinator/src/contract/query.rs
```diff
@@ -60,3 +63,26 @@ fn get_provers_for_verifier(
...
     provers.change_context(ContractError::FailedToGetProversForVerifier)
 }
+
+pub fn get_chain_contracts_info(
+    deps: Deps,
+    chain_contracts_key: ChainContractsKey,
+) -> Result<ChainContractsRecord, ContractError> {
+    match chain_contracts_key {
+        ChainContractsKey::ChainName(chain_name) => {
+            Ok(contracts_by_chain(deps.storage, chain_name.clone())?.into())
```
maancham, line 73: I am not sure about this, but is there a way to change the state helper function to accept reference? It allows us to get rid of all the clones here.

contracts/coordinator/src/msg.rs
```diff
@@ -55,3 +62,30 @@ pub struct VerifierInfo {
...
+#[cw_serde]
+pub enum ChainContractsKey {
+    ChainName(ChainName),
+    ProverAddress(ProverAddress),
+    GatewayAddress(GatewayAddress),
+    VerifierAddress(VerifierAddress),
+}
+
+#[cw_serde]
+pub struct ChainContractsRecord {
```
maancham, line 75: Can we change the name here so it is not the same as the state struct? For example, we have `ChainConfigResponse` for msg.rs and `CHAIN_CONFIGS` for state.rs over interchain token service.

contracts/coordinator/src/state.rs
```diff
@@ -46,7 +46,7 @@ impl IndexList<ChainContractsRecord> for ChainContractsIndexes<'_> {
     }
 }
 
-const CHAIN_CONTRACTS_MAP: IndexedMap<ChainName, ChainContractsRecord, ChainContractsIndexes> =
+pub const CHAIN_CONTRACTS_MAP: IndexedMap<ChainName, ChainContractsRecord, ChainContractsIndexes> =
```
maancham, line 49: Wht was this made public here? Are we directly using the IndexedMap somewhere else?

maancham (changes requested):

contracts/coordinator/src/msg.rs
SDavidson1177: Yep, just changed it to ChainContractsResponse.

contracts/coordinator/src/state.rs
SDavidson1177: My mistake. It was public when I was accessing it directly, but I forgot to change it back when I switched the code to use the helper functions.

contracts/coordinator/src/client.rs
```diff
@@ -1,10 +1,10 @@
...
-use cosmwasm_std::CosmosMsg;
+use cosmwasm_std::{Addr, CosmosMsg};
 use error_stack::{Result, ResultExt};
 use router_api::ChainName;
 
-use crate::msg::{ExecuteMsg, QueryMsg};
+use crate::msg::{ChainContractsKey, ChainContractsResponse, ExecuteMsg, QueryMsg};
 
 #[derive(thiserror::Error, Debug, PartialEq)]
 pub enum Error {
```
cgorenflo, line 10: These errors should be refactored. Essentially anything we implement should describe only things in the the context they own, and if possible one thing at a time. These errors seem to be aware that they are run in the context of a query failure, as well as what the underlying cause for failure is. With error-stack we can easily separate those concerns without losing the usefulness of the response:
- top level example: "failed to retrieve chain contracts information"
- lower level error (cause for top level): "chain name {0} is not registered", "gateway {0} is not registered" etc

That way, a developer is not surprised by the output of the error when using it in code and doesn't have to look up the error string every time. E.g. I would not expect `Error::ChainNameNotRegistered` to return something about a query, but for a `Error::ChainContractsInfoQuery` I would expect that.

cgorenflo: this can be handled in a separate PR though

contracts/coordinator/src/client.rs
```diff
@@ -20,6 +20,18 @@ pub enum Error {
...
+    GatewayNotRegistered(Addr),
+
+    #[error("failed to execute ChainContractsInfo query at coordinator contract. prover {0}")]
+    ProverNotRegistered(Addr),
+
+    #[error("failed to execute ChainContractsInfo query at coordinator contract. verifier {0}")]
+    VerifierNotRegistered(Addr),
 }
 
 impl From<QueryMsg> for Error {
```
cgorenflo, line 37: Similarly to the error types, a conversion from query message to error feels very surprising (even though the mapping aligns nicely). I'd get rid of this entirely and rather build the errors where needed. Alternatively, it could be a `err_from_msg_params` function instead of a `From` trait implementation to make it clearer what the purpose of the conversion is.

contracts/coordinator/src/msg.rs
```diff
@@ -7,6 +7,12 @@ use msgs_derive::EnsurePermissions;
 use router_api::ChainName;
 use service_registry_api::Verifier;
 
+use crate::state::ChainContractsRecord;
```
cgorenflo, line 10: see last point on the PR checklist. I believe you already have the appropriate type to map to, but the return type still points to the `state` type

SDavidson1177: Thanks, I have changed the return type to the `msg` equivalent ChainContractsResponse

contracts/coordinator/src/client.rs
SDavidson1177: That's a good point. Since the `From` trait implementation was present prior to these changes, I will address it in another PR

contracts/coordinator/src/msg.rs
```diff
@@ -55,3 +64,30 @@ pub struct VerifierInfo {
...
+impl From<ChainContractsRecord> for ChainContractsResponse {
+    fn from(chain_contracts_record: ChainContractsRecord) -> Self {
+        ChainContractsResponse {
+            chain_name: chain_contracts_record.chain_name,
+            prover_address: chain_contracts_record.prover_address,
+            gateway_address: chain_contracts_record.gateway_address,
+            verifier_address: chain_contracts_record.verifier_address,
+        }
+    }
+}
```
cgorenflo, line 93: In cases where a msg <-> state conversion is necessary we decided to do the conversion in the state.rs, so the external interface of the contract doesn't have to change if the internal structure changes. So please move this into state.rs and make sure there is no longer any dependency on `crate::state::*` in this file

cgorenflo (approved):

contracts/coordinator/src/state.rs
```diff
@@ -79,53 +79,51 @@ pub fn save_chain_contracts(
...
     CHAIN_CONTRACTS_MAP
         .idx
         .by_verifier
         .item(storage, verifier_address.clone())?
         .map(|(_, record)| record)
-        .ok_or(ContractError::VerifierNotRegistered(verifier_address))
+        .ok_or(ContractError::VerifierNotRegistered(
+            verifier_address.clone(),
+        ))
 }
```
cgorenflo, line 127: in most cases it's best to propagate ownership concerns up the chain. In particular, try not to make decisions about cloning for the caller. They might not need the value after the call to these functions anymore and could just relent ownership so the data wouldn't need to be cloned

contracts/coordinator/src/msg.rs
SDavidson1177: Yep, just moved the conversion to state.rs

maancham (approved):

PR #805, opened 2025-04-09
refactor(minor-ampd): merge auth, bank and service gRPC clients into CosmosGRpcClient

fish-sammy:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros


ampd/src/cosmos.rs
```diff
@@ -0,0 +1,100 @@
...
+use thiserror::Error;
+use tonic::transport::Channel;
+use tonic::Response;
+
+type Result<T> = error_stack::Result<T, Error>;
+
+#[derive(Error, Debug)]
+pub enum Error {
+    #[error("failed to connect to the grpc endpoint")]
+    GRpcConnection(#[from] tonic::transport::Error),
```
maancham, line 22: Can we use `GrpcRequest` instead here? To be consistent with what we currently have on ampd-sdk?

maancham, line 22: same here.

ampd/src/broadcaster/confirm_tx.rs
```diff
@@ -47,23 +46,23 @@ impl From<cosmrs::proto::cosmos::base::abci::v1beta1::TxResponse> for TxResponse
 pub enum Error {
     #[error("failed confirming tx due to tx not found: {tx_hash}")]
     Confirmation { tx_hash: String },
-    #[error("failed confirming tx due to grpc error {status}: {tx_hash}")]
-    Grpc { status: Status, tx_hash: String },
+    #[error("failed confirming tx due to grpc error: {tx_hash}")]
```
cgorenflo, line 49: I'm not a fan of using colons in the error string, because that's how the error stacks get concatenated. Can these change to `failed confirming tx {tx_hash}...`?

ampd/src/commands/mod.rs
```diff
@@ -130,43 +126,33 @@ async fn broadcast_tx(
 async fn instantiate_broadcaster(
     config: Config,
     pub_key: CosmosPublicKey,
-) -> Result<(impl Broadcaster, TxConfirmer<ServiceClient<Channel>>), Error> {
+) -> Result<(impl Broadcaster, TxConfirmer<cosmos::CosmosGRpcClient>), Error> {
```
cgorenflo, line 129:
```suggestion
) -> Result<(impl Broadcaster, TxConfirmer<cosmos::CosmosGrpcClient>), Error> {
```

ampd/src/cosmos.rs
```diff
@@ -0,0 +1,100 @@
...
+        let endpoint: tonic::transport::Endpoint = url.parse().map_err(ErrorExt::into_report)?;
+        let conn = endpoint.connect().await.map_err(ErrorExt::into_report)?;
+
+        Ok(Self {
+            auth: AuthQueryClient::new(conn.clone()),
+            bank: BankQueryClient::new(conn.clone()),
+            service: ServiceClient::new(conn),
+        })
+    }
+}
```
cgorenflo, line 57: I like this!

ampd/src/cosmos.rs
```diff
@@ -0,0 +1,100 @@
...
+#[derive(Clone)]
+pub struct CosmosGRpcClient {
+    auth: AuthQueryClient<Channel>,
+    bank: BankQueryClient<Channel>,
+    service: ServiceClient<Channel>,
+}
+
+impl CosmosGRpcClient {
+    pub async fn new(url: &str) -> Result<Self> {
+        let endpoint: tonic::transport::Endpoint = url.parse().map_err(ErrorExt::into_report)?;
```
cgorenflo, line 48: conceptual question: If we keep the client a slim facade that essentially just converts errors I'm ok keeping it untested in unit tests. Otherwise, we should think about mocking the endpoint instead of the client itself

ampd/src/cosmos.rs
```diff
@@ -0,0 +1,100 @@
...
+    async fn simulate(&mut self, req: SimulateRequest) -> Result<SimulateResponse>;
+    async fn tx(&mut self, req: GetTxRequest) -> Result<GetTxResponse>;
+
+    async fn account(&mut self, address: QueryAccountRequest) -> Result<QueryAccountResponse>;
+
+    async fn balance(&mut self, request: QueryBalanceRequest) -> Result<QueryBalanceResponse>;
+}
+
+#[derive(Clone)]
+pub struct CosmosGRpcClient {
```
cgorenflo, line 40: add documentation what it means for the connection to clone this?

packages/report/src/error_ext.rs
```diff
@@ -0,0 +1,52 @@
...
+    /// use report::ErrorExt;
+    ///
+    /// #[derive(Error, Debug)]
+    /// enum InnerError {
+    ///     #[error("An inner error occurred")]
+    ///     Inner,
+    /// }
+    ///
+    /// #[derive(Error, Debug)]
+    /// enum OutterError {
```
cgorenflo, line 28:
```suggestion
    /// enum OuterError {
```

packages/report/src/error_ext.rs
```diff
@@ -0,0 +1,52 @@
...
+    ///     #[error("An outter error occurred")]
+    ///     Outter(#[from] InnerError),
```
cgorenflo, line 30:
```suggestion
    ///     Outer(#[from] InnerError),
```

packages/report/src/error_ext.rs
```diff
@@ -0,0 +1,52 @@
...
+    ///     #[error("Another outter error occurred")]
+    ///     AnotherOutter
+    /// }
```
cgorenflo, line 33: Since you are not actually wrapping the errors but converting them, I think it makes more sense to call them `OriginalError` and `ConvertedError`

cgorenflo (approved):

ampd/src/cosmos.rs
fish-sammy: This client is gonna be kept as simple as it is now. You'll see what I did in my next PR and how things are tested.

PR #807, opened 2025-04-14
feat(minor-ampd): create new tx broadcaster with todo

fish-sammy:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros

### DIFF
--- a/ampd/src/tx_broadcaster/msg_queue.rs
+++ b/ampd/src/tx_broadcaster/msg_queue.rs
@@ -1,0 +1,512 @@
+use core::pin::Pin;
+use core::task::{ready, Context, Poll};
+use std::future::Future;
+
+use cosmrs::{Any, Gas};
+use error_stack::{report, ResultExt};
+use futures::Stream;
+use pin_project_lite::pin_project;
+use report::ErrorExt;
+use tokio::sync::{mpsc, oneshot};
+use tokio::time;
+use tokio_stream::adapters::Fuse;
+use tokio_stream::wrappers::ReceiverStream;
+use tokio_stream::StreamExt;
+
+use super::{Error, Result};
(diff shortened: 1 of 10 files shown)


ampd/src/tx_broadcaster/msg_queue.rs
```diff
@@ -0,0 +1,512 @@
...
+impl Queue {
+    pub fn push(&mut self, msg: Msg) -> Result<()> {
+        self.gas_cost = self
+            .gas_cost
+            .checked_add(msg.gas)
+            .ok_or(report!(Error::IntegerOverflow))?;
+        self.msgs.push(msg);
+
+        Ok(())
+    }
```
cgorenflo, line 198: could we incorporate the `pop_all` here? I.e. if the gas would overflow, pop and return existing messages, else just psuh the current one. Then the calling code doesn't have to worry about the queues internal consistency and can just handle the two cases where it either gets nothing or the previous messages back

ampd/src/types/key.rs
```diff
@@ -125,6 +125,17 @@ impl From<CosmosPublicKey> for PublicKey {
...
+#[cfg(test)]
+pub mod test_utils {
+    use rand::rngs::OsRng;
+
+    use super::CosmosPublicKey;
+
+    pub fn random_cosmos_public_key() -> CosmosPublicKey {
+        CosmosPublicKey::from(k256::ecdsa::SigningKey::random(&mut OsRng).verifying_key())
+    }
+}
```
cgorenflo, line 137: nice

cgorenflo (changes requested):

PR #811, opened 2025-04-21
fix(minor-coordinator): refactor coordinator tests to use goldie

SDavidson1177:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros

## Notes
The purpose of this PR is to address the comments from #803. Furthermore, tests for the coordinator have been modified to better adhere to best practices. Goldie patches are used to compare result structs, and tests use the coordinator's public interface wherever possible. I have also refactored the tests to use cw_multi_test instead of cosmwasm_std::testing. The cw_multi_test crate is more robust, and I believe is therefore better suited for testing upcoming complex features.


contracts/coordinator/src/contract.rs
```diff
@@ -140,137 +140,155 @@ mod tests {
...
+            chain_name: chain_name.clone(),
+            prover: prover.clone(),
+            gateway: gateway.clone(),
+            verifier: verifier.clone(),
         }
     }
 
     #[test]
     #[allow(clippy::arithmetic_side_effects)]
     fn test_instantiation() {
```
cgorenflo, line 215: give this test a better name like "only_allow_governance_to_register_prover"

cgorenflo: actually, by giving this test a proper name I realized that it's doing the same as add_prover_from_governance_succeeds and add_prover_from_random_address_fails, so we can remove it

contracts/coordinator/src/contract.rs
```diff
@@ -283,144 +301,135 @@ mod tests {
...
-        ));
+            .contains(
+                &axelar_wasm_std::error::ContractError::from(
+                    permission_control::Error::WhitelistNotFound {
+                        sender: test_setup.prover
+                    }
+                )
+                .to_string()
+            ));
     }
```
cgorenflo, line 406: these two tests couple the internal logic of the prover to the coordinator. I don't think this is a good test here. The coordinator doesn't care that SetActiveVerifiers works on the prover. At the very least, these tests should get moved to the prover

contracts/coordinator/src/client.rs
```diff
@@ -9,57 +9,33 @@ use crate::msg::{ChainContractsKey, ChainContractsResponse, ExecuteMsg, QueryMsg
...
-                }
-                ChainContractsKey::ProverAddress(prover_addr) => {
-                    Error::ProverNotRegistered(prover_addr)
-                }
-                ChainContractsKey::VerifierAddress(verifier_addr) => {
-                    Error::VerifierNotRegistered(verifier_addr)
-                }
-                ChainContractsKey::ChainName(chain_name) => {
-                    Error::ChainNameNotRegistered(chain_name.into())
-                }
```
cgorenflo, line 60: I think keeping different error messages is still valuable to make it easier to debug what's going on

cgorenflo (changes requested):

contracts/coordinator/src/contract.rs
SDavidson1177: Yes, you're right. I'll remove it.

contracts/coordinator/src/contract.rs
SDavidson1177: I see, good point. I was reading through contracts/multisig-prover/contract.rs, and it looks like there is a test for this. I'll check that it properly covers this case, and I will make the required modifications if necessary.

contracts/coordinator/src/client.rs
SDavidson1177: Sounds good, I'll add them back

contracts/coordinator/src/contract.rs
SDavidson1177: There wasn't a test that checked that only the prover contract could update the verifier set. Since this involves several cross-contract calls, I used cw_multi_test rather than just cosmwasm_std::testing. However, this now means that there are two different test setups for the prover. I can consolidate all the tests to use the same setup but, since this would be a larger change, I'm not sure if it should have its own PR. Either way, contracts/multisig-prover/contract.rs:test_update_verifier_set_from_non_prover_should_fail() demonstrates the new test.

contracts/coordinator/src/contract.rs
```diff
@@ -3,39 +3,58 @@ mod migrations;
...
+    let governance = address::validate_cosmwasm_address(deps.api, &msg.governance_address)
+        .change_context(Error::Instantiate)?;
 
-    let governance = address::validate_cosmwasm_address(deps.api, &msg.governance_address)?;
-    permission_control::set_governance(deps.storage, &governance)?;
+    permission_control::set_governance(deps.storage, &governance)
+        .map_err(|err| error_stack::Report::new(err))
+        .change_context(Error::Instantiate)?;
 
     Ok(Response::default())
```
cgorenflo, line 59: use `report!` instead of `error_stack::Report::new`

contracts/coordinator/src/state.rs
cgorenflo, line 1: use the `report!` macro here as well

contracts/coordinator/src/state.rs
```diff
@@ -80,60 +101,73 @@ pub fn save_chain_contracts(
...
         .idx
         .by_gateway
-        .item(storage, gateway_address.clone())?
-        .map(|(_, record)| record)
-        .ok_or(ContractError::GatewayNotRegistered(gateway_address))
+        .item(storage, gateway_address.clone())
+        .map_err(|err| Report::new(err))
+        .change_context(Error::StateParseFailed)?
+        .ok_or_else(|| Report::new(Error::GatewayNotRegistered(gateway_address)))?
+        .1)
```
cgorenflo, line 156: what's the `.1`?

cgorenflo (approved):

PR #812, opened 2025-04-22
fix(interchain-token-service): fix migration to handle all values

cjcobb23:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros

### DIFF
--- a/contracts/interchain-token-service/src/contract/migrations/v1_1_0.rs
+++ b/contracts/interchain-token-service/src/contract/migrations/v1_1_0.rs
@@ -51,23 +51,23 @@ pub fn migrate(deps: DepsMut, msg: MigrateMsg) -> Result<(), ContractError> {
     let transformed_configs: Vec<(ChainNameRaw, state::ChainConfig)> = old_configs
         .into_iter()
         .map(|(chain_name, old_config)| {
-            convert_max_uint_to_max_bits(old_config.truncation.max_uint).map(|max_uint_bits| {
-                (
-                    chain_name,
-                    state::ChainConfig {
-                        truncation: state::TruncationConfig {
-                            max_decimals_when_truncating: old_config
-                                .truncation
-                                .max_decimals_when_truncating,
-                            max_uint_bits,
-                        },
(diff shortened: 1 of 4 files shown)


contracts/interchain-token-service/src/contract/migrations/v1_1_0.rs
```diff
@@ -76,17 +76,14 @@ pub fn migrate(deps: DepsMut, msg: MigrateMsg) -> Result<(), ContractError> {
...
 
-fn convert_max_uint_to_max_bits(max_uint: nonempty::Uint256) -> Result<NumBits, ContractError> {
-    let fixed_cases = [256, 128, 64, 32];
-    for case in fixed_cases {
-        // some chains were incorrectly registered with the max number of bits as the max uint
-        if *max_uint == Uint256::from_u128(case as u128) {
-            return NumBits::try_from(case).map_err(|err| err.into());
-        }
+fn convert_max_uint_to_max_bits(max_uint: Uint256) -> NumBits {
+    if max_uint <= Uint256::from_u128(256) {
```
milapsheth, line 80: Don't we allow overwriting the max unit now? Can that be used instead of having a devnet specific migration? It's low risk in production so I'm still fine with having it

contracts/interchain-token-service/src/shared.rs
```diff
@@ -25,6 +26,22 @@ impl TryFrom<u32> for NumBits {
...
 }
 
+impl NumBits {
+    // rounds up to the nearest supported number of bits.
+    // Anything greater than 256 is rounded to 256
+    pub fn round_to_nearest<T>(value: T) -> Self
+    where
+        T: Into<Uint256>,
+    {
+        match value.into() {
```
milapsheth, line 36: Why are we only allowing these specific values for max_uint_bits? E.g. stellar uses uint127

milapsheth (approved):

PR #819, opened 2025-04-25
feat(ampd-sdk): event handler trait

maancham:
## Description

AXE-8587

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros


packages/ampd-sdk/src/utils/mod.rs
fish-sammy, line 1: I'd create a sub-module named `future` to contain the retry code here. But nice job! It's cleaner than the original implementation.

ampd-handlers/src/main.rs
```diff
@@ -1,17 +1,7 @@
-fn get_greeting() -> &'static str {
-    "Hello from AMPD handler!"
-}
-
-fn main() {
-    println!("{}", get_greeting());
+#[tokio::main(flavor = "current_thread")]
```
fish-sammy, line 1: Why `flavor = "current_thread")`?

packages/ampd-sdk/src/event/event_handler.rs
```diff
@@ -0,0 +1,563 @@
...
+)]
+#[async_trait]
+pub trait EventHandler: Send + Sync {
+    type Err: Context;
+    type Event: TryFrom<Event>;
+
+    async fn handle(
+        &self,
+        event: &Self::Event,
+        token: &CancellationToken,
```
fish-sammy, line 38: Thought we decided to wrap this in the `Context` struct?

packages/ampd-sdk/src/event/event_handler.rs
```diff
@@ -0,0 +1,563 @@
...
+    ) -> Result<Vec<Any>, Self::Err>;
+
+    async fn subscription_params(&self) -> SubscriptionParams;
+}
+
+pub struct SubscriptionParams {
+    pub event_filters: Vec<AbciEventTypeFilter>,
+    pub include_block_begin_end: bool,
```
fish-sammy, line 46: Why `pub` fields?

packages/ampd-sdk/src/event/event_handler.rs
```diff
@@ -0,0 +1,563 @@
...
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct Config {
+    #[serde(with = "humantime_serde")]
+    pub stream_timeout: Duration,
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            stream_timeout: Duration::from_secs(30),
```
fish-sammy, line 58: Think this is way too long. I'd go for 10s.

packages/ampd-sdk/src/event/event_handler.rs
```diff
@@ -0,0 +1,563 @@
...
+        }
+    }
+}
+
+#[derive(Error, Debug)]
+pub enum Error {
+    #[error("failed to retreive events stream from the client")]
+    ClientEventStream,
```
fish-sammy, line 66:
```suggestion
    EventStream,
```

packages/ampd-sdk/src/event/event_handler.rs
```diff
@@ -0,0 +1,563 @@
...
+            .unwrap_or(vec![]);
+
+        error!(
+            err = LoggableError::from(&report!(err)).as_value(),
+            msgs = msgs.as_value(),
+            "failed to broadcast messages"
+        );
+
+        HandlerTaskAction::Continue
+    }
```
fish-sammy, line 155: Don't think these need to be methods in the handler task itself but individual functions. Also, I'd move them to some other file.

packages/ampd-sdk/src/event/event_handler.rs
```diff
@@ -0,0 +1,563 @@
...
+
+    fn default_handler_error_cb(event: &H::Event, err: H::Err) -> HandlerTaskAction
+    where
+        <H::Event as TryFrom<Event>>::Error: Context,
+        H::Event: Display,
+    {
+        error!(
+            err = LoggableError::from(&report!(err)).as_value(),
+            event = format!("{event}").as_value(),
+            "failed to process events"
```
fish-sammy, line 135:
```suggestion
            "failed to handle events"
```

packages/ampd-sdk/src/event/event_handler.rs
```diff
@@ -0,0 +1,563 @@
...
+
+        HandlerTaskAction::Continue
+    }
+
+    fn default_broadcast_error_cb(msgs: &[Any], err: H::Err) -> HandlerTaskAction {
+        let msgs = msgs
+            .iter()
+            .map(serde_json::to_string)
+            .try_collect()
+            .unwrap_or(vec![]);
```
fish-sammy, line 146: I prefer `.expect(...)` over `unwrap_or`. Also, you should use `unwrap_or_default` instead.

packages/ampd-sdk/src/event/event_handler.rs
```diff
@@ -0,0 +1,563 @@
...
+            msgs = msgs.as_value(),
+            "failed to broadcast messages"
+        );
+
+        HandlerTaskAction::Continue
+    }
+    pub async fn run(
+        self,
+        task_label: impl Into<String>,
+        client: &mut impl client::Client,
```
fish-sammy, line 159: Why aren't these fields part of `HandlerTask`?

packages/ampd-sdk/src/event/event_handler.rs
```diff
@@ -0,0 +1,563 @@
...
+    type Err: Context;
+    type Event: TryFrom<Event>;
+
+    async fn handle(
+        &self,
+        event: &Self::Event,
+        token: &CancellationToken,
+    ) -> Result<Vec<Any>, Self::Err>;
+
+    async fn subscription_params(&self) -> SubscriptionParams;
```
fish-sammy, line 41: Why is this `async`?

fish-sammy (changes requested):

ampd-handlers/src/main.rs
maancham: added `rt-multi-thread` feature to tokio and using `#[tokio::main]` now.

packages/ampd-sdk/src/event/event_handler.rs
maancham: removed for now.

packages/ampd-sdk/src/event/event_handler.rs
cgorenflo: oh yeah, didn't realize I could make the associated types generics on the functions to move them out

packages/ampd-sdk/src/event/event_handler.rs
cgorenflo: definitely not here. We don't know what kind of messages the handler gave us and I don't want to panic the whole process over logging

packages/ampd-sdk/src/event/event_handler.rs
cgorenflo: the label can be, I just thought it's nicer to have that as part of the run instead of having to define yet another parameter on the builder. The client is a mutable reference and becomes a huge pain to deal with, because it's borrowed by the stream, which also needs immutable references to `self`. So the alternative to taking it here is to completely deconstruct self inside run. That looked way worse to me, so I think passing the client in separately is a good solution

packages/ampd-sdk/src/event/event_handler.rs
```diff
@@ -0,0 +1,563 @@
...
+
+        let stream = client
+            .subscribe(
+                subscription_params.event_filters,
+                subscription_params.include_block_begin_end,
+            )
+            .await
+            .change_context(Error::ClientEventStream)?
+            .take_while(|_| !token.is_cancelled())
+            .timeout_repeating(interval(self.config.stream_timeout));
```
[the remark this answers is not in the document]
cgorenflo: good point, I thought it was awkward to have the flatten function be the first thing when parsing the event. I like moving it here better, too

packages/ampd-sdk/src/event/event_handler.rs
fish-sammy: You could use `.clone` to work around the mutability thing. Tonic actually encourages doing it.

packages/ampd-sdk/src/event/event_handler.rs
fish-sammy: They shouldn't need to. They should only know what these callbacks are for and what their signatures are. Good interface design should eliminate the need for the user to understand what's going on behind the scenes.

packages/ampd-sdk/src/event/event_handler.rs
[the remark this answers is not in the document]
fish-sammy: Would suggest also have `StreamTimeout #[from] Elapsed` so that you can use `ErrorExt::into_report` instead.

packages/ampd-sdk/src/event/event_handler.rs
cgorenflo: true, I forgot about that

packages/ampd-sdk/src/event/event_handler.rs
cgorenflo: I do like the separation between immutable data and mutable client though. Is it an issue to keep the client separate?

maancham: added `task_label` to the `HandlerTask`. I also think keeping the client as input argument is a good idea.

packages/ampd-sdk/src/future/mod.rs
fish-sammy, line 1: Unit tests still are missing.

packages/ampd-sdk/src/event/event_handler.rs
```diff
@@ -0,0 +1,520 @@
...
+        .await
+        .map_err(|err| (self.handler_error_cb)(&event, err))
+    }
+
+    async fn broadcast_msgs(
+        &self,
+        msgs: Vec<Any>,
+        _token: &CancellationToken,
+    ) -> Result<(), HandlerTaskAction> {
+        with_retry(|| async { Ok(()) }, self.broadcast_retry_policy) // Placeholder for actual broadcast
```
fish-sammy, line 230: Add a TODO here.

packages/ampd-sdk/src/event/event_handler.rs
```diff
@@ -0,0 +1,520 @@
...
+    #[builder(default = RetryPolicy::NoRetry)]
+    handler_retry_policy: RetryPolicy,
+    #[builder(default = RetryPolicy::NoRetry)]
+    broadcast_retry_policy: RetryPolicy,
+    #[builder(default = default_event_subscription_error_cb)]
+    event_subscription_error_cb: fn(Report<Error>) -> HandlerTaskAction,
+    #[builder(default = default_handler_error_cb)]
+    handler_error_cb: fn(&H::Event, H::Err) -> HandlerTaskAction,
+    #[builder(default = default_broadcast_error_cb)]
+    broadcast_error_cb: fn(&[Any], H::Err) -> HandlerTaskAction,
```
fish-sammy, line 112:
```suggestion
    broadcast_error_cb: fn(&Any, Report<Error>) -> HandlerTaskAction,
```
Broadcaster happens with the via the gRPC client you built and error has nothing to do with `H:Err`. Also, the broadcast grpc method is per msg instead of a batch of msgs.

packages/ampd-sdk/src/event/event_handler.rs
```diff
@@ -0,0 +1,520 @@
...
+#[async_trait]
+pub trait EventHandler: Send + Sync {
+    type Err: Context;
+    type Event: TryFrom<Event>;
+
+    async fn handle(
+        &self,
+        event: &Self::Event,
+        token: &CancellationToken,
+    ) -> Result<Vec<Any>, Self::Err>;
```
fish-sammy, line 38: This should be `error_stack::Result`.
```suggestion
    use error_stack::Result;
    
    ...

    ) -> Result<Vec<Any>, Self::Err>;
```

packages/ampd-sdk/src/event/event_handler.rs
```diff
@@ -0,0 +1,520 @@
...
+    #[builder(default = String::from("default_task"))]
+    task_label: String,
+    #[builder(default = RetryPolicy::NoRetry)]
+    handler_retry_policy: RetryPolicy,
+    #[builder(default = RetryPolicy::NoRetry)]
+    broadcast_retry_policy: RetryPolicy,
+    #[builder(default = default_event_subscription_error_cb)]
+    event_subscription_error_cb: fn(Report<Error>) -> HandlerTaskAction,
+    #[builder(default = default_handler_error_cb)]
+    handler_error_cb: fn(&H::Event, H::Err) -> HandlerTaskAction,
```
fish-sammy, line 110:
```suggestion
    handler_error_cb: fn(&H::Event, Report<H::Err>) -> HandlerTaskAction,
```

packages/ampd-sdk/src/event/event_handler.rs
```diff
@@ -0,0 +1,520 @@
...
+    #[error("error when broadcasting message")]
+    BroadcastFailed,
+
+    #[error("{0} task aborted")]
+    TaskAborted(String),
+}
+
+#[allow(dead_code)]
+pub enum HandlerTaskAction {
+    Abort,
```
fish-sammy, line 86: I have had this thought of 
```suggestion
    Abort(Error),
```
so that you can return the actual error in the end instead of using something so generic like `TaskAborted`.

fish-sammy (changes requested):

packages/ampd-sdk/src/event/callbacks.rs
fish-sammy, line 1: Add unit tests for all these default callbacks. It should be as simple as making sure they return `Continue` and that's it.

packages/ampd-sdk/src/event/event_handler.rs
```diff
@@ -0,0 +1,520 @@
...
+    async fn parse_event(
+        &self,
+        element: error_stack::Result<Event, Error>,
+    ) -> Result<H::Event, HandlerTaskAction> {
+        element
+            .inspect(Self::log_block_boundary)
+            .and_then(|event| {
+                H::Event::try_from(event.clone())
+                    .change_context(Error::EventConversion)
+                    .attach_printable(format!("failed to create handler event from: {event}"))
```
fish-sammy, line 195:
```suggestion
                    .attach_printable(json!({ "event": event }))
```
Add `Serialize` to the Event derive.

cgorenflo (approved):

fish-sammy (approved):

PR #822, opened 2025-05-01
feat(minor-ampd): implement gRPC method subscribe for the blockchain service

fish-sammy:
## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros

### DIFF
--- a/ampd/src/grpc/blockchain_service.rs
+++ b/ampd/src/grpc/blockchain_service.rs
@@ -1,5 +1,5 @@
 use std::pin::Pin;
-use std::time::Duration;
+use std::str::FromStr;
 
 use ampd_proto::blockchain_service_server::BlockchainService;
 use ampd_proto::{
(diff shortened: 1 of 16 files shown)


ampd/src/grpc/blockchain_service.rs
```diff
@@ -92,3 +104,409 @@ impl BlockchainService for Service {
...
+                events::Event::Abci { event_type, .. } => {
+                    filter_abci_event(&filters, event_type, contract)
+                }
+            }
+        }
+        Err(_) => true,
+    }
+}
+
+fn filter_abci_event(
```
maancham, line 149: Is there a way to make this simpler? Something like 
```
filters.iter().any(|filter| {
    let type_matches = filter.r#type.is_empty() || filter.r#type == event_type;
    let contract_matches = filter.contract.is_empty() || 
        filter.contract == contract.as_ref().map(ToString::to_string).unwrap_or_default();
    
    type_matches && contract_matches
})
```

fish-sammy: I'm refactoring this now.

ampd/src/grpc/blockchain_service.rs
```diff
@@ -1,70 +1,90 @@
...
+            include_block_begin_end,
+        } = req.into_inner();
+
+        let filters: event_filters::EventFilters = (filters, include_block_begin_end)
+            .try_into()
+            .map_err(|err| {
+                error!(
+                    err = LoggableError::from(&err).as_value(),
+                    "invalid event filters provided for event subscription"
+                );
```
cgorenflo, line 57: I don't like logging and returning the error, because that makes it confusing who should be responsible for it. Wherever the error is finally handled should the logging happen

ampd/src/grpc/blockchain_service.rs
```diff
@@ -1,70 +1,90 @@
...
+            self.event_sub
+                .subscribe()
+                .filter(move |event| filters.filter(event))
+                .map_ok(Into::into)
+                .map_ok(|event| ampd_proto::SubscribeResponse { event: Some(event) })
+                .map_err(|err| {
+                    error!(
+                        err = LoggableError::from(&err).as_value(),
+                        "event subscription error"
+                    );
```
cgorenflo, line 72: same here

ampd/src/grpc/event_filters.rs
```diff
@@ -0,0 +1,380 @@
...
+            None
+        } else {
+            let contract = AccountId::from_str(&event_filter.contract)
+                .change_context(Error::InvalidContractAddress(event_filter.contract.clone()))
+                .and_then(|contract| {
+                    if contract.prefix() != PREFIX {
+                        return Err(report!(Error::InvalidContractAddress(
+                            event_filter.contract
+                        )));
+                    }
```
cgorenflo, line 54: `ensure!`

ampd/src/grpc/event_filters.rs
```diff
@@ -0,0 +1,380 @@
...
+                    Ok(contract)
+                })
+                .map(Into::into)?;
+
+            Some(contract)
+        };
+
+        if event_type.is_none() && contract.is_none() {
+            return Err(report!(Error::EmptyFilter));
+        }
```
cgorenflo, line 65: should this be an error? It could mean "give me all events", couldn't it?
if we keep it use `ensure!`

ampd/src/grpc/event_filters.rs
```diff
@@ -0,0 +1,380 @@
...
+                "invalid contract address {} provided in filters",
+                contract
+            )),
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct EventFilter {
+    event_type: Option<nonempty::String>,
```
cgorenflo, line 35: nice, I like the nonempty

ampd/src/grpc/event_filters.rs
```diff
@@ -0,0 +1,380 @@
...
+            contract,
+        })
+    }
+}
+
+impl EventFilter {
+    pub fn filter(&self, event_type: &str, contract: &Option<TMAddress>) -> bool {
+        if self
+            .event_type
+            .as_ref()
```
cgorenflo, line 78: why is as_ref necessary?

ampd/src/grpc/event_filters.rs
```diff
@@ -0,0 +1,380 @@
...
+            .is_some_and(|filter| *filter != event_type)
+        {
+            return false;
+        }
+
+        if self.contract.is_some() && self.contract != *contract {
+            return false;
+        }
+
+        true
```
cgorenflo, line 88:
```suggestion
    self.matches_event_type(event_type) && self.matches_contract(contract)
```
and move the checks into their own little functions. They are complicated enough that they would benefit from being named

ampd/src/grpc/event_filters.rs
```diff
@@ -0,0 +1,380 @@
...
+}
+
+#[derive(Debug)]
+pub struct EventFilters {
+    filters: Vec<EventFilter>,
+    include_block_begin_end: bool,
+}
+
+impl EventFilters {
+    pub fn filter(&self, event: &Result<events::Event, event_sub::Error>) -> bool {
```
cgorenflo, line 99: why filter on a result?

ampd/src/grpc/event_filters.rs
```diff
@@ -0,0 +1,380 @@
...
+
+        Ok(EventFilter {
+            event_type,
+            contract,
+        })
+    }
+}
+
+impl EventFilter {
+    pub fn filter(&self, event_type: &str, contract: &Option<TMAddress>) -> bool {
```
cgorenflo, line 75: `&Option` should be avoided. Do `Option<&TMAddress>` instead. Option has an `as_ref()` for that

Can you also rename to `TmAddress`? (likely separate PR)

ampd/src/lib.rs
```diff
@@ -188,18 +190,15 @@ where
 {
     #[allow(clippy::too_many_arguments)]
     fn new(
-        tm_client: tendermint_rpc::HttpClient,
+        event_publisher: event_sub::EventPublisher<tendermint_rpc::HttpClient>,
+        event_subscriber: event_sub::EventSubscriber,
```
cgorenflo, line 194: is this necessary?

packages/events/src/event.rs
```diff
@@ -141,27 +176,22 @@ impl TryFrom<ampd_proto::subscribe_response::Event> for Event {
...
+                    .attributes
+                    .into_iter()
+                    .map(|(key, value)| {
+                        (
+                            key,
+                            serde_json::from_str(&value)
+                                .unwrap_or(serde_json::Value::String(value)),
+                        )
+                    })
+                    .collect(),
```
cgorenflo, line 189: nice, much nicer than the previous version

ampd/src/grpc/event_filters.rs
fish-sammy: In that case, they should provide an empty filters Vec instead of having an empty filter in the Vec.

ampd/src/grpc/event_filters.rs
fish-sammy: becuz `is_some_and` consumes `self` for some reason.

ampd/src/lib.rs
fish-sammy: The grpc server needs a clone of the `event_subscriber`, so I had to initialize the event sub before this `new` is called. Any other suggestions?

ampd/src/grpc/event_filters.rs
cgorenflo: ok that's dumb

ampd/src/grpc/event_filters.rs
```diff
@@ -0,0 +1,348 @@
...
+                .change_context(Error::InvalidContractAddress(event_filter.contract.clone()))
+                .and_then(|contract| {
+                    ensure!(
+                        contract.prefix() == PREFIX,
+                        Error::InvalidContractAddress(event_filter.contract)
+                    );
+
+                    Ok(contract)
+                })
+                .map(Into::into)?;
```
cgorenflo, line 44: define TryFrom<&str> for TmAddress?

ampd/src/lib.rs
cgorenflo: oh I searched for the usage and didn't find a hit, but it was just collapsed in unchanged code

cgorenflo (approved):

PR #825, opened 2025-05-05
feat(minor-ampd): implement gRPC method broadcast for the blockchain service

fish-sammy:
## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros

### DIFF
--- a/ampd/src/broadcaster_v2/msg_queue.rs
+++ b/ampd/src/broadcaster_v2/msg_queue.rs
@@ -100,9 +100,9 @@ where
         &mut self,
         msg: Any,
     ) -> Result<impl Future<Output = Result<(String, u64)>> + Send> {
-        let attachment = json!({ "msg": msg });
+        let attachment = json!({ "msg": &msg });
         let rx = self
-            .enqueue_with_channel(msg.clone())
+            .enqueue_with_channel(msg)
             .await
             .map_err(|err| err.attach_printable(attachment.clone()))?;
 
--- a/ampd/src/broadcaster_v2/msg_queue.rs
+++ b/ampd/src/broadcaster_v2/msg_queue.rs
@@ -5,7 +5,7 @@ use std::future::Future;
 use axelar_wasm_std::nonempty;
 use cosmrs::{Any, Gas};
 use error_stack::{report, Report, ResultExt};
-use futures::{FutureExt, Stream};
+use futures::{FutureExt, Stream, TryFutureExt};
 use pin_project_lite::pin_project;
 use report::{ErrorExt, LoggableError};
 use serde_json::json;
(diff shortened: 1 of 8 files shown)


ampd/src/broadcaster_v2/msg_queue.rs
```diff
@@ -100,13 +100,19 @@ where
...
-            Err(err) => Err(err.into_report()),
-        }))
+        Ok(rx
+            .map(|result| match result {
+                Ok(Ok(result)) => Ok(result),
+                Ok(Err(err)) => Err(err),
+                Err(err) => Err(err.into_report()),
+            })
+            .map_err(move |err| err.attach_printable(attachment)))
     }
```
cgorenflo, line 116: is this possible so we don't need to clone all over the place?
```
        let with_attachment = |err| err.attach_printable(json!({ "msg": msg }));
        let rx = self
            .enqueue_with_channel(msg.clone())
            .await
            .map_err(with_attachment)?;

        Ok(rx
            .map(|result| match result {
                Ok(Ok(result)) => Ok(result),
                Ok(Err(err)) => Err(err),
                Err(err) => Err(err.into_report()),
            })
            .map_err(with_attachment)
```

ampd/src/broadcaster_v2/msg_queue.rs
```diff
@@ -225,24 +231,21 @@ impl MsgQueue {
...
-        MsgQueueClient<T>,
-    )>
+    ) -> (Pin<Box<MsgQueue>>, MsgQueueClient<T>)
     where
         T: cosmos::CosmosClient,
     {
         let (tx, rx) = mpsc::channel(msg_cap);
 
-        Ok((
+        (
```
cgorenflo, line 240: nice

ampd/src/grpc/blockchain_service.rs
```diff
@@ -6,58 +6,39 @@ use ampd_proto::{
...
-            .try_into()
-            .inspect_err(|err| {
-                error!(
-                    err = LoggableError::from(err).as_value(),
-                    "invalid event filters provided for event subscription"
-                );
-            })
-            .map_err(ErrorExt::into_status)?;
+        let filters = reqs::validate_subscribe(req)
+            .inspect_err(error::log("invalid subscribe request"))
```
cgorenflo, line 40: lol

ampd/src/grpc/blockchain_service.rs
```diff
@@ -67,22 +48,35 @@ where
...
+            .clone()
+            .enqueue(msg)
+            .map(|rx| async {
+                match rx {
+                    Ok(rx) => rx.await,
+                    Err(err) => Err(err),
+                }
+            })
+            .await
+            .await
```
cgorenflo, line 75: double await for first enqueuing and then callback result?

cgorenflo: use and_then in that case?

ampd/src/lib.rs
cgorenflo, line 1: is it possible/does it make sense to split the prepare_app function a bit?

ampd/src/broadcaster_v2/msg_queue.rs
fish-sammy: That doesn't work, but I figured something out.

ampd/src/lib.rs
fish-sammy: I prefer doing this later. Broadcaster v1 should be removed eventually, and then we can revisit `prepare_app`.

cgorenflo (approved):

PR #827, opened 2025-05-06
feat: updated to solana deps 2.1.21

ICavlek:
## Description
Updated `solana` dependencies to the latest version `2.1.21` which is suitable for mainnet.
Additionally, `solana-axelar` repository has been updated to the latest stage.


ICavlek:
@cgorenflo Thank you for reviewing it. Could you please also check the failed test? After quick look into it, it seems to me that it is only due to the network and that after restart, it should be ok. Unfortunately, I cannot restart it

ICavlek:
@cgorenflo PR is still failing with the same code coverage issue. After initial [googling](https://stackoverflow.com/questions/67861379/codecov-fails-in-github-actions), rotating of token is needed and/or force retry,

ICavlek:
@cgorenflo @eguajardo Possible [fix](https://github.com/keaven/gsDesign/pull/179#issuecomment-2634502479), needs to be updated in environment

eguajardo:
@ICavlek @eloylp we do have a [codecov token](https://github.com/axelarnetwork/axelar-amplifier/blob/main/.github/workflows/codecov.yaml#L61), but the workflow doesn't seem to be loading it when it's running on PRs coming from a fork. This seems consistent with what was discussed in the github issue you shared. Could you try getting a token and add it to your github environment and see if this way it loads?

ICavlek:
@eguajardo codecov token is added. I have synced the fork and workflows have been refreshed automatically. I suppose that this should be sufficient, that token can be read also from your side

eguajardo:
@ICavlek I updated codecov to version v4, it seems that some fix was added to be able to use the token when a PR comes from a fork https://github.com/codecov/codecov-action/pull/1471/files

We can try again after you sync with main. Is there a way to temporarily allow me to push changes to Eiger's branch? It's a small window we have from when you sync with main and us running the workflows before main becomes unsync again, it would be faster if I could do the syncing myself

ICavlek:
@eguajardo thank you for your help and support :)

PR #830, opened 2025-05-07
feat(ampd-handlers): base handler client connection and image creation process

maancham:
## Description
AXE-9294

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros

### DIFF
--- a/.github/workflows/ampd-handlers-docker-ecr.yaml
+++ b/.github/workflows/ampd-handlers-docker-ecr.yaml
@@ -4,9 +4,6 @@ on:
   push:
     branches:
       - feat/dummy-handler-client-connection # TODO: replace with main before merge
-    paths:
-      - 'ampd-handlers/**'
-      - '.github/workflows/ampd-handlers-docker-ecr.yaml'
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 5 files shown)


ampd-handlers/Dockerfile
```diff
@@ -1,10 +1,13 @@
...
-WORKDIR /ampd-handlers
+WORKDIR /app
 COPY . .
-RUN cargo build --release --bin ampd_handler
 
-FROM debian:bullseye-slim
+# Build the ampd-handlers binary
+RUN cargo build --release --bin ampd-handlers
+
+FROM debian:bookworm-slim
```
fish-sammy, line 10:
```suggestion
FROM debian:bookworm-slim as runner
```
I don't it doesn't do anything, but just better for readability

.github/workflows/ampd-handlers-docker-ecr.yaml
```diff
@@ -0,0 +1,57 @@
+name: ampd-handlers (push to main) - Build and push image to ECR
+
+on:
+  push:
+    branches:
+      - feat/dummy-handler-client-connection # TODO: replace with main before merge
```
fish-sammy, line 6: Don't forget doing this

fish-sammy (approved):

PR #839, opened 2025-05-13
feat(minor-multisig): accept arbitrary length message for signing

cjcobb23:


contracts/multisig/src/types.rs
```diff
@@ -33,42 +32,10 @@ pub enum MultisigState {
...
-        if other.len() != MESSAGE_HASH_LEN {
-            return Err(ContractError::InvalidMessageFormat {
-                reason: "Invalid input length".into(),
-            });
-        }
-
-        Ok(MsgToSign(other))
+impl From<HexBinary> for MsgToSign {
+    fn from(value: HexBinary) -> Self {
+        MsgToSign(value)
```
eguajardo, line 38: if I understand correctly the msg is now arbitrary length, but ampd is still signing on 32 bytes, right? in that case wouldn't you need to change how the verification works as well since now it will pass the whole message to the verification function? https://github.com/axelarnetwork/axelar-amplifier/blob/ba3d3c5117b2a89f527ff9c84f588ef9097ac51b/contracts/multisig/src/signing.rs#L85

When this rolls out, what would happen to messages that are currently being signed/verified?

cjcobb23: hmm this is a good point, the evm chains will still just pass 32 byte message hashes to be signed and it will work as is, and I would only think XRPL would pass larger messages. However, maybe we should only allow greater than 32 bytes when a sig verifier callback is attached

cgorenflo, line 38: this circumvents previous assumptions about MsgToSign, I want to discuss this in more detail before approving this change

cgorenflo (approved):
approved with the understanding that this is a temporary fix and will get cleaned up with the next release

eguajardo (approved):
Looks safe to me, my observations are:
- MsgToSign semantics might have inconsistencies between chain implementations, some will work as always and some not. I think those will show in Axelarscan
- This change could add more gas/load because now events for chains that not limit the msg to 32 bytes will include the whole message. 
- Probably in the future we could pass the whole message to the sig_verifier, but keep the current MsgToSign as 32 bytes or something like that?

PR #842, opened 2025-05-14
feat(minor-ampd): implement blockchain service smart_contract_state

fish-sammy:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros


ampd/src/config.rs
cgorenflo, line 1: why is it hardcoded in most cases and formatted in others?

ampd/src/grpc/blockchain_service.rs
```diff
@@ -120,27 +138,30 @@ mod tests {
 
     async fn setup(
         mock_event_sub: MockEventSub,
-        mut mock_cosmos_client: MockCosmosClient,
+        mut broadcaster_mock_cosmos_client: MockCosmosClient,
+        mock_cosmos_client: MockCosmosClient,
```
cgorenflo, line 142: why two separate mocks?

ampd/src/grpc/error.rs
```diff
@@ -115,6 +115,27 @@ impl From<&broadcaster_v2::Error> for Error {
     }
 }
 
+impl From<&cosmos::Error> for Error {
+    fn from(err: &cosmos::Error) -> Self {
+        match err {
+            cosmos::Error::GrpcConnection(_) | cosmos::Error::GrpcRequest(_) => {
```
cgorenflo, line 121: is the inner error not useful?

ampd/src/grpc/error.rs
```diff
@@ -115,6 +115,27 @@ impl From<&broadcaster_v2::Error> for Error {
...
+            cosmos::Error::QuerySmartContractState(reason) => Status::unknown(format!(
+                "failed to query smart contract state with error {reason}"
+            )),
+            cosmos::Error::GasInfoMissing
+            | cosmos::Error::AccountMissing
+            | cosmos::Error::TxResponseMissing
+            | cosmos::Error::MalformedResponse
+            | cosmos::Error::TxBuilding => {
+                Status::internal("server encountered an error processing request")
+            }
```
cgorenflo, line 133: why do you obfuscate all of these?

ampd/src/grpc/reqs.rs
```diff
@@ -24,12 +26,40 @@ pub fn validate_broadcast(req: Request<BroadcastRequest>) -> Result<Any, Error>
...
+    Ok(address)
+}
+
 #[derive(Error, Debug)]
 pub enum Error {
     #[error("empty filter")]
     EmptyFilter,
-    #[error("invalid contract address in filter")]
+    #[error("invalid contract address")]
     InvalidContractAddress(String),
```
cgorenflo, line 60: what happens to this address string?

ampd/src/grpc/reqs.rs
```diff
@@ -24,12 +26,40 @@ pub fn validate_broadcast(req: Request<BroadcastRequest>) -> Result<Any, Error>
...
+) -> Result<(TMAddress, Vec<u8>), Error> {
+    let SmartContractStateRequest { contract, query } = req.into_inner();
+
+    ensure!(!query.is_empty(), Error::InvalidQuery);
+    let query: serde_json::Value =
+        serde_json::from_slice(&query).change_context(Error::InvalidQuery)?;
+
+    Ok((
+        validate_address(&contract)?,
+        serde_json::to_vec(&query).expect("must be valid JSON"),
```
cgorenflo, line 40: is there a case where you can deserialize it but not serialize it again? Why not take the original value?

ampd/src/cosmos.rs
```diff
@@ -603,4 +721,31 @@ mod tests {
...
+            }))
+            .unwrap(),
+        };
+
+        goldie::assert_json!(json!({
+            "req": format!("{:?}", req),
+            "res": format!("{:?}", res),
+            "req_bytes": req.to_bytes().unwrap(),
+            "res_bytes": res.to_bytes().unwrap()
+        }));
```
cgorenflo, line 749: what's the point of goldie here?

ampd/src/config.rs
fish-sammy: What are you refering here?

ampd/src/grpc/blockchain_service.rs
fish-sammy: One is for the broadcaster, and one is for the grpc server itself. In production, they are just a clone of each other. In tests, I wanna ensure separate behaviour.

ampd/src/grpc/error.rs
fish-sammy: They are logged elsewhere. I don't think they are useful to the client as we want to avoid blindly passing error from cosmos gRPC to ampd SDK.

ampd/src/grpc/error.rs
fish-sammy: I like having all of them here instead of using `_` so that anyone adding new errors must come here and think about the correct mapping.

ampd/src/grpc/reqs.rs
fish-sammy: It's used here https://github.com/axelarnetwork/axelar-amplifier/blob/704e1f04de2a76cb55bdc6e11d0ef6cae489de67/ampd/src/grpc/error.rs#L66

ampd/src/cosmos.rs
fish-sammy: Ensuring the request types we use don't change serialization when upgrading cosmrs version, and if they do, we catch it.

ampd/src/grpc/error.rs
fish-sammy: These are all internal implementations IMO that the client should not be aware of, but we can discuss if you think some of them worth exposing.

ampd/src/config.rs
cgorenflo: `reqs_timeout = '30s'`

ampd/src/grpc/error.rs
cgorenflo: does the logging happen in the trace layer?

ampd/src/grpc/error.rs
cgorenflo: do these get logged, too? I agree that the clients don't need to know why these failed, but for the operator of ampd it definitely is useful information

ampd/src/grpc/error.rs
fish-sammy: The inner gets logged here https://github.com/axelarnetwork/axelar-amplifier/blob/26548a69b40cdfce4d988b382d3c7b10b9d21358/ampd/src/grpc/blockchain_service.rs#L90, and the Status we return in the end gets logged in the trace layer.

ampd/src/grpc/error.rs
fish-sammy: Ya they all get logged before turning into the Status at https://github.com/axelarnetwork/axelar-amplifier/blob/26548a69b40cdfce4d988b382d3c7b10b9d21358/ampd/src/grpc/blockchain_service.rs#L90.

ampd/src/cosmos.rs
```diff
@@ -44,6 +48,8 @@ pub enum Error {
     MalformedResponse,
     #[error("failed to build tx")]
     TxBuilding,
+    #[error("failed to query the smart contract state with error {0}")]
+    QuerySmartContractState(String),
```
cgorenflo, line 52:
```suggestion
    QueryContractState(String),
```

ampd/src/grpc/mod.rs
```diff
@@ -35,6 +36,8 @@ pub struct Config {
     pub port: u16,
     pub concurrency_limit: nonempty::Usize,
     pub concurrency_limit_per_connection: nonempty::Usize,
+    #[serde(with = "humantime_serde")]
+    pub reqs_timeout: Duration,
```
cgorenflo, line 40:
```suggestion
    pub request_timeout: Duration,
```
as function parameter it's relatively easy to understand, here it should have a proper name because it's going to be converted into toml

ampd/src/grpc/mod.rs
```diff
@@ -35,6 +36,8 @@ pub struct Config {
     pub port: u16,
     pub concurrency_limit: nonempty::Usize,
     pub concurrency_limit_per_connection: nonempty::Usize,
```
cgorenflo, line 38: what's the difference between the two?

ampd/src/grpc/status.rs
```diff
@@ -17,98 +17,107 @@ pub fn log<Err>(msg: &str) -> impl Fn(&Report<Err>) + '_ {
...
+    fn into_status(self) -> tonic::Status {
         self.into().0
     }
 }
 
-impl<'a, Err> ErrorExt for &Report<Err>
+impl<'a, Err> From<&'a Report<Err>> for Status
 where
-    Error: From<&'a Err>,
+    for<'b> Status: From<&'b Err>,
```
cgorenflo, line 38: that's crazy 🤯

cgorenflo (approved):

ampd/src/grpc/mod.rs
fish-sammy: One is the global limit of concurrent requests, and the other is the limit per HTTP2 connection.

ampd/src/grpc/status.rs
fish-sammy: They have different lifetime lol

PR #843, opened 2025-05-14
feat(minor-coordinator): coordinator instantiates gateway, verifier and prover

SDavidson1177:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros


contracts/coordinator/src/contract/execute.rs
```diff
@@ -55,3 +87,278 @@ pub fn set_active_verifier_set(
...
+    deployment_name: String,
+    params: &DeploymentParams,
+) -> error_stack::Result<Response, Error> {
+    let mut response = Response::new();
+    deployment_name_is_free(&deps.as_ref(), &deployment_name)
+        .change_context(Error::InstantiateContracts)?;
+
+    let config = load_config(deps.storage);
+
+    let salt = instantiate2_salt(&mut deps).change_context(Error::InstantiateContracts)?;
```
eguajardo, line 277: I think salts should be coming from the user, maybe using the `deployment_name` since that's already unique? The reason is that sometimes we want the same contract addresses for multiple environments (e.g. stagenet, testnet, mainnet), but if storage is not synchronized then the salt would be different for each env. Another reason why we probably shouldn't allow any user to deploy contracts, so that they cannot burn deployment names

contracts/coordinator/src/msg.rs
```diff
@@ -41,6 +46,55 @@ pub enum ExecuteMsg {
...
+    #[permission(Any)]
+    InstantiateChainContracts {
+        deployment_name: String,
+        params: Box<DeploymentParams>,
+    },
+}
+
+#[cw_serde]
+pub enum DeploymentParams {
+    Manual {
```
eguajardo, line 59: what does `Manual` mean in this context? could you maybe add a comment explaining why this is an enum and the meaning of manual?

contracts/coordinator/src/msg.rs
```diff
@@ -41,6 +46,55 @@ pub enum ExecuteMsg {
...
+        gateway_code_id: u64,
+        gateway_label: String,
+        verifier_code_id: u64,
+        verifier_label: String,
+        verifier_msg: VerifierMsg,
+        prover_code_id: u64,
+        prover_label: String,
+        prover_msg: ProverMsg,
+    },
```
eguajardo, line 68: depending of the future variants, this enum could have a lot of content, maybe extract all those fields into a struct? Something like this:

```suggestion
pub enum DeploymentParams {
    Manual(ManualDeploymentParams),
}
```

eguajardo: you could abstract even further by creating a struct that holds code_id, label and params, then reuse that struct for gateway, verifier, prover and whatever other future contract we might want to instantiate

contracts/coordinator/src/contract/execute.rs
```diff
@@ -55,3 +87,278 @@ pub fn set_active_verifier_set(
...
+    verifier_code_id: u64,
+    prover_code_id: u64,
+}
+
+pub fn instantiate_chain_contracts(
+    mut deps: DepsMut,
+    env: Env,
+    info: MessageInfo,
+    deployment_name: String,
+    params: &DeploymentParams,
```
eguajardo, line 269: I don't think this should be a reference, this function should own the params. It would eliminate some cloning and dereferencing

contracts/coordinator/Cargo.toml
```diff
@@ -34,21 +34,25 @@ optimize = """docker run --rm -v "$(pwd)":/code \
 axelar-wasm-std = { workspace = true, features = ["derive"] }
 client = { workspace = true }
 cosmwasm-schema = { workspace = true }
-cosmwasm-std = { workspace = true }
+cosmwasm-std = { workspace = true, features = ["cosmwasm_1_2"] }
```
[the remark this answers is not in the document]
SDavidson1177: This feature is required for QueryCodeInfo and Instantiate2. QueryCodeInfo is used to get the bytecode of the contract, which is then used to construct a predictable address. This address is used as one of the instantiation arguments (verifier address used to instantiate gateway). Instantiate2 just instantiates using this predictable addressing

contracts/coordinator/src/contract/execute.rs
```diff
@@ -55,3 +87,278 @@ pub fn set_active_verifier_set(
...
+                    .change_context(Error::Instantiate2Address)?,
+                salt,
+            )
+            .change_context(Error::Instantiate2Address)?,
+        )
+        .change_context(Error::Instantiate2Address)
+}
+
+fn deployment_name_is_free(deps: &Deps, deployment_name: &str) -> error_stack::Result<(), Error> {
+    let deployments = DEPLOYED_CHAINS
```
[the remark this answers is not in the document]
SDavidson1177: Makes sense, I'll do that

contracts/coordinator/src/contract/execute.rs
```diff
@@ -55,3 +87,278 @@ pub fn set_active_verifier_set(
...
+                    .addr_canonicalize(&env.contract.address.to_string().clone())
+                    .change_context(Error::Instantiate2Address)?,
+                salt,
+            )
+            .change_context(Error::Instantiate2Address)?,
+        )
+        .change_context(Error::Instantiate2Address)
+}
+
+fn deployment_name_is_free(deps: &Deps, deployment_name: &str) -> error_stack::Result<(), Error> {
```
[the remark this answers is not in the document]
SDavidson1177: Sounds good, I'll rename it

contracts/coordinator/src/contract/execute.rs
```diff
@@ -55,3 +87,278 @@ pub fn set_active_verifier_set(
...
+            )?;
+
+            response = response
+                .add_message(msg)
+                .add_event(ContractInstantiated::MultisigProver {
+                    address: multisig_prover_address.clone(),
+                    code_id: *prover_code_id,
+                });
+
+            DEPLOYED_CHAINS
```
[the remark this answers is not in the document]
SDavidson1177: I'll also do the same with the instantiate2 salt counter

contracts/coordinator/src/contract/execute.rs
SDavidson1177: Ahh okay I understand. I'll make the salt user provided then

contracts/coordinator/src/events.rs
```diff
@@ -0,0 +1,9 @@
+use axelar_wasm_std::IntoEvent;
+use cosmwasm_std::Addr;
+
+#[derive(IntoEvent)]
+pub enum ContractInstantiated {
+    Gateway { address: Addr, code_id: u64 },
+    VotingVerifier { address: Addr, code_id: u64 },
+    MultisigProver { address: Addr, code_id: u64 },
+}
```
eguajardo, line 9: should this include the deployment_name as well? could this also be arranged something like this:
```suggestion
pub struct ContractInstantiated {
    address: Addr,
    code_id: u64,
    deployment_name: String,
    contract: Contract, // enum
}
```

contracts/coordinator/src/msg.rs
```diff
@@ -41,6 +46,63 @@ pub enum ExecuteMsg {
     },
     #[permission(Specific(prover))]
     SetActiveVerifiers { verifiers: HashSet<String> },
+
+    #[permission(Any)]
+    InstantiateChainContracts {
+        deployment_name: String,
```
eguajardo, line 52:
```suggestion
        deployment_name: nonempty::String,
```

contracts/coordinator/src/contract/execute.rs
```diff
@@ -55,3 +78,243 @@ pub fn set_active_verifier_set(
...
+        ctx.salt.clone(),
+        ctx.gateway_code_id,
+        cosmwasm_std::to_json_binary(&gateway_api::msg::InstantiateMsg {
+            verifier_address: verifier_address.to_string().clone(),
+            router_address: router_address.to_string().clone(),
+        })
+        .change_context(Error::InstantiateGateway)?,
+        label,
+    )
+}
```
eguajardo, line 149: I think these instantiation functions and the way they are called could be improved, but I need to think a bit more about it

cgorenflo (approved):
comments will be addressed in follow-up PRs

contracts/coordinator/src/events.rs
SDavidson1177: Will resolve in another PR

PR #850, opened 2025-05-16
feat(minor-ampd): change XRPL handler to listen to multisig contract

cjcobb23:


ampd/src/handlers/multisig.rs
```diff
@@ -127,11 +127,6 @@ where
         };
 
         if !chain.eq(&self.chain) {
-            info!(
```
milapsheth, line 130: removing to avoid unnecessary verbose logging?

ampd/src/handlers/xrpl_multisig.rs
```diff
@@ -103,25 +106,34 @@ where
...
+            chain,
         } = match event.try_into() as error_stack::Result<_, _> {
             Err(report) if matches!(report.current_context(), EventTypeMismatch(_)) => {
                 return Ok(vec![]);
             }
             result => result.change_context(DeserializeEvent)?,
         };
 
+        if !chain.eq(&self.chain) {
+            info!(
```
milapsheth, line 127: remove this log as well?

milapsheth (approved):

eguajardo (approved):

PR #856, opened 2025-05-16
refactor(minor-multisig): update module visibility

yidan233:
**changes:** 
- Marked internal modules as `mod` instead of `pub mod` where possible
- Left some modules public if they are still accessed in tests 
- checked: `User` type and similar internal types are not risky , no validation added
- only refactored imports in other contracts to use `exported.rs` 

AXE-7324
**Description**

We created the following convention:
types used in `msg.rs` must be one of the following

> types defined in the msg module
> imported types from external crates
> imported types from this crate's exported  module

the exported module gets re-exported in the top level `lib.rs`, to allow for more ergonomic use:

```
mod exported;
pub use exported::*;
```
other contract modules can import types from anywhere, but must map those types to validated internal types as soon as possible if necessary

**IMPORTANT**: As we modify the message types, we also need to check for missed validations. For example, the PoolId in the rewards contract currently uses Addr as a field type, which then goes unchecked. In these cases, we need a separate msg object that gets converted (+validation) into the corresponding state object at the beginning of the msg execution

PoolId [usage](https://github.com/axelarnetwork/axelar-amplifier/blob/1a8a4ccf5fc7ac202138342020d4619dce145e24/contracts/rewards/src/msg.rs#L74)
PoolId [definition](https://github.com/axelarnetwork/axelar-amplifier/blob/1a8a4ccf5fc7ac202138342020d4619dce145e24/contracts/rewards/src/state.rs#L46)


cgorenflo:
closing this PR, there are some major issues to be solved with the multisig contract message types and this current approach is turned out to not be useful.

PR #859, opened 2025-05-17
feat(xrpl-types): implement TryFrom<String> for XRPLMessageType

themicp:
## Description

Implements the `TryFrom<String>` trait for `XRPLMessageType`. This is currently used only by the relayer.

### DIFF
--- a/packages/xrpl-types/src/msg.rs
+++ b/packages/xrpl-types/src/msg.rs
@@ -56,6 +56,21 @@ impl std::fmt::Display for XRPLMessageType {
     }
 }
 
+impl TryFrom<String> for XRPLMessageType {
+    type Error = String;
+
+    fn try_from(s: String) -> Result<Self, Self::Error> {
+        match s.as_str() {
+            "proof" => Ok(XRPLMessageType::Proof),
+            "interchain_transfer" => Ok(XRPLMessageType::InterchainTransfer),
+            "call_contract" => Ok(XRPLMessageType::CallContract),
+            "add_gas" => Ok(XRPLMessageType::AddGas),
+            "add_reserves" => Ok(XRPLMessageType::AddReserves),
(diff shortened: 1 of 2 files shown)


packages/xrpl-types/src/msg.rs
```diff
@@ -56,6 +56,21 @@ impl std::fmt::Display for XRPLMessageType {
     }
 }
 
+impl TryFrom<String> for XRPLMessageType {
```
fish-sammy, line 59: Implement FromStr https://doc.rust-lang.org/std/str/trait.FromStr.html instead.

fish-sammy (changes requested):

packages/xrpl-types/src/msg.rs
```diff
@@ -56,6 +56,21 @@ impl std::fmt::Display for XRPLMessageType {
...
+    type Error = String;
```
fish-sammy, line 60: Create or use one of those errors from `XRPLError` instead of hacking it with a String error here.

cgorenflo:
@themicp could you address the comments?

packages/xrpl-types/src/msg.rs
themicp: Good point: https://github.com/axelarnetwork/axelar-amplifier/pull/859/commits/540ff9ef7eb91c6809918d2a0103303f1611a018

packages/xrpl-types/src/msg.rs
themicp: https://github.com/commonprefix/axelar-amplifier/commit/cf8ac14d1bb39e390281f74bb710cb7029aa5ce8

themicp:
> @themicp could you address the comments?

Apologies for the delay, just pushed the changes.

PR #860, opened 2025-05-19
minor(ampd): add configurable delay before block processing

cjcobb23:
## Description
Add a configurable delay before block processing. Defaults to 1 second.

### DIFF
--- a/ampd/src/event_sub/stream.rs
+++ b/ampd/src/event_sub/stream.rs
@@ -95,6 +95,8 @@ where
         })
         .collect::<Result<Vec<_>>>()?;
 
+    tokio::time::sleep(Duration::new(1, 0)).await;
+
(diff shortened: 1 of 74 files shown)


ampd/src/event_sub/stream.rs
```diff
@@ -95,6 +95,8 @@ where
         })
         .collect::<Result<Vec<_>>>()?;
 
+    tokio::time::sleep(Duration::new(1, 0)).await;
```
fish-sammy, line 98:
```suggestion
    tokio::time::sleep(Duration::from_secs(1)).await;
```

fish-sammy: This also isn't the best place to add the delay. This function is wrapped in a `with_retry` and makes no sense to force every failure take >1 secs. It should be added in the `update` method of `BlockState` after the while loop.

PR #861, opened 2025-05-19
feat(minor-ampd): implement crypto service method key

fish-sammy:
## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros


ampd/src/grpc/reqs.rs
```diff
@@ -49,6 +52,40 @@ fn validate_address(address: &str) -> Result<TMAddress, Error> {
     Ok(address)
 }
 
+pub fn validate_key(
+    req: Request<KeyRequest>,
+) -> Result<(nonempty::String, tofnd::Algorithm), Error> {
+    let KeyRequest { key_id } = req.into_inner();
+
+    validate_key_id(key_id.unwrap_or_default())
+}
```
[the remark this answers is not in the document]
maancham: Do we have other instances of validation in the scope of this task? I mean validating something other than `key_id`. I am asking since it makes sense to me to use trait if there are at least two implementations

ampd/src/grpc/crypto_service.rs
```diff
@@ -1,23 +1,324 @@
...
+            .inspect_err(status::log("invalid sign request"))
+            .map_err(status::StatusExt::into_status)?;
+        // TODO: memoize the key
+        let pub_key = self
+            .0
+            .keygen(&id, algorithm)
+            .await
+            .inspect_err(status::log("query key error"))
+            .map_err(status::StatusExt::into_status)?;
+        let msg_hash: [u8; 32] = Keccak256::digest(msg.as_ref()).into();
```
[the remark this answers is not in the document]
maancham: Nice catch. So, we should pass the `msg` directly I guess?

eguajardo:
bugbot run

ampd/src/grpc/reqs.rs
```diff
@@ -1,41 +1,60 @@
...
-    ensure!(!query.is_empty(), Error::InvalidQuery);
-    let _: serde_json::Value =
-        serde_json::from_slice(&query).change_context(Error::InvalidQuery)?;
+    fn validate(self) -> Result<Self::Output, Error> {
+        let ContractStateRequest { contract, query } = self.into_inner();
 
-    Ok((validate_address(&contract)?, query))
+        ensure!(!query.is_empty(), Error::InvalidQuery);
+        let _: serde_json::Value =
+            serde_json::from_slice(&query).change_context(Error::InvalidQuery)?;
```
cgorenflo, line 54: this needs to become generic, right?

ampd/src/grpc/reqs.rs
```diff
@@ -1,41 +1,60 @@
...
+            .ok_or(report!(Error::EmptyBroadcastMsg))
+    }
 }
 
-pub fn validate_contract_state(
-    req: Request<ContractStateRequest>,
-) -> Result<(TMAddress, Vec<u8>), Error> {
-    let ContractStateRequest { contract, query } = req.into_inner();
+impl Validate for Request<ContractStateRequest> {
+    type Output = (TMAddress, Vec<u8>);
```
cgorenflo, line 47: create a type for this so it's obvious what this output means?

ampd/src/grpc/reqs.rs
```diff
@@ -49,6 +68,46 @@ fn validate_address(address: &str) -> Result<TMAddress, Error> {
     Ok(address)
 }
 
+impl Validate for Request<KeyRequest> {
+    type Output = (nonempty::String, tofnd::Algorithm);
```
cgorenflo, line 72: create a type for this so it's obvious what this output means?

ampd/src/grpc/reqs.rs
```diff
@@ -49,6 +68,46 @@ fn validate_address(address: &str) -> Result<TMAddress, Error> {
...
+
+    fn validate(self) -> Result<Self::Output, Error> {
+        let KeyRequest { key_id } = self.into_inner();
+
+        validate_key_id(key_id.unwrap_or_default())
+    }
+}
+
+impl Validate for Request<SignRequest> {
+    type Output = (nonempty::String, tofnd::Algorithm, [u8; 32]);
```
cgorenflo, line 82: create a type for this so it's obvious what this output means?

ampd/src/grpc/reqs.rs
```diff
@@ -483,7 +548,138 @@ mod tests {
...
+        let invalid_algorithm = 999; // some invalid algorithm value
+
+        let key_id_obj = KeyId {
+            id: key_id.to_string(),
+            algorithm: invalid_algorithm,
+        };
+
+        let result = validate_key_id(key_id_obj);
+        assert_err_contains!(result, Error, Error::InvalidCryptoAlgorithm(_));
+    }
```
cgorenflo, line 618: try to create tests that don't rely on/test private functions

ampd/src/grpc/status.rs
```diff
@@ -270,4 +287,28 @@ mod tests {
...
+            tofnd::Error::InvalidSignResponse.into_status().code(),
+            Code::Internal
+        );
+        assert_eq!(
+            tofnd::Error::ExecutionFailed("key generation failed".to_string())
+                .into_status()
+                .code(),
+            Code::Internal
+        );
+    }
```
cgorenflo, line 313: seems like this is just a slightly different format than the action `from` implementation. Use goldie instead?

cgorenflo: (seems like existing tests followed this format already, so would be good to change those as well)

ampd/src/tofnd/mod.rs
```diff
@@ -1,20 +1,47 @@
...
+pub use proto::Algorithm;
+
+pub fn install_debug_hooks() {
+    Report::install_debug_hook::<KeygenResponse>(|value, context| {
+        context.push_body(format!("KeygenResponse: {:?}", value));
+    });
+    Report::install_debug_hook::<SignResponse>(|value, context| {
+        context.push_body(format!("SignResponse: {:?}", value));
+    });
+}
```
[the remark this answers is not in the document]
cgorenflo: yes we aborted that. Prefer to use `#[instrument]` to annotate functions. That will capture the context (function name, arguments) instead

ampd/src/tofnd/grpc.rs
```diff
@@ -68,33 +66,32 @@ impl Multisig for MultisigClient {
...
-                response
-                    .into_inner()
+            .map_err(ErrorExt::into_report)
+            .and_then(|res| {
+                let res = res.into_inner();
+
+                res.clone()
                     .keygen_response
-                    .ok_or_else(|| Status::internal("keygen response is empty"))
+                    .ok_or(report!(Error::InvalidKeygenResponse).attach(res))
```
cgorenflo, line 75: see comment about debug hooks. You can also add additional values to `instrument`, see https://docs.rs/tracing/latest/tracing/attr.instrument.html#adding-fields

ampd/src/tofnd/grpc.rs
```diff
@@ -68,33 +66,32 @@ impl Multisig for MultisigClient {
...
-                }
-                .change_context(Error::ParsingFailed)
-                .attach_printable(format!("{{ invalid_value = {:?} }}", pub_key)),
-                KeygenResponse::Error(error_msg) => {
-                    Err(TofndError::ExecutionFailed(error_msg)).change_context(Error::KeygenFailed)
+                    Algorithm::Ecdsa => PublicKey::new_secp256k1(pub_key),
+                    Algorithm::Ed25519 => PublicKey::new_ed25519(pub_key),
                 }
+                .change_context(Error::InvalidKeygenResponse)
+                .attach(res),
```
cgorenflo, line 83: see other comment about attach

ampd/src/tofnd/grpc.rs
cgorenflo, line 1: there is a lot of cloning going on in both keygen and sign, is that necessary?

ampd/src/grpc/crypto_service.rs
```diff
@@ -3,21 +3,322 @@ use ampd_proto::{KeyRequest, KeyResponse, SignRequest, SignResponse};
...
+
+        self.0
+            .keygen(&id, algorithm)
+            .await
+            .map(|pub_key| KeyResponse {
+                pub_key: pub_key.to_bytes(),
+            })
+            .map(Response::new)
+            .inspect_err(status::log("query key error"))
+            .map_err(status::StatusExt::into_status)
```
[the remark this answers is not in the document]
eguajardo: there is not much to extract, these lines are part of the `key` function but not of the `sign`
```
            .map(|pub_key| KeyResponse {
                pub_key: pub_key.to_bytes(),
            })
            .map(Response::new)
```
also I think the logged error should be slightly different

ampd/src/tofnd/grpc.rs
eguajardo: can't remove the cloning, the reason is that the auto generated proto functions use `&mut self`. We would need to either use mutable everywhere or clone here

ampd/src/grpc/reqs.rs
```diff
@@ -49,6 +52,40 @@ fn validate_address(address: &str) -> Result<TMAddress, Error> {
...
+fn validate_key_id(key_id: KeyId) -> Result<(nonempty::String, tofnd::Algorithm), Error> {
+    let KeyId { id, algorithm } = key_id;
+
+    let id = nonempty::String::try_from(id).change_context(Error::EmptyKeyId)?;
+    let algorithm = match algorithm.try_into() {
+        Ok(Algorithm::Ecdsa) => tofnd::Algorithm::Ecdsa,
+        Ok(Algorithm::Ed25519) => tofnd::Algorithm::Ed25519,
+        _ => {
+            return Err(report!(Error::InvalidCryptoAlgorithm(algorithm)));
+        }
```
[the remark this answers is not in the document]
cgorenflo: bail! is a good addition, but otherwise this will not have any change in behaviour

ampd/src/grpc/crypto_service.rs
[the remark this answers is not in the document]
cgorenflo: fair enough it looked more with all those codecov notifications in between

cgorenflo (approved):

ampd/src/grpc/reqs.rs
eguajardo: yeah, validate that the query bytes can be deserialized to valid json

PR #866, opened 2025-05-20
feat(metrics): add a prometheus client

yidan233:
## Description
- add a prometheus_client that tracks 
- number of times calling verifier address
- number of times calling bond verifier 
- time to load config file 
- accessible on http://localhost:9898/metrics by default 
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros


ampd/src/metrics.rs
```diff
@@ -0,0 +1,108 @@
...
+
+pub fn gather_metrics() -> String {
+    let encoder = TextEncoder::new();
+    let mut buffer = Vec::new();
+    let metric_families = prometheus::gather();
+    encoder.encode(&metric_families, &mut buffer).unwrap();
+    String::from_utf8(buffer).unwrap()
+}
+
+pub async fn start_metrics_server(address: Option<SocketAddr>) -> JoinHandle<()> {
```
fish-sammy, line 96: Have a look at `health_check.rs`. Ampd should only run one HTTP server instead of another one dedicated for the metrics. The metrics should be exposed via a path `/metrics`. Rename the `health_check` to be something more generic like `monitor` perhaps.

ampd/src/metrics.rs
```diff
@@ -0,0 +1,108 @@
...
+    let addr = address.unwrap_or_else(|| std::net::SocketAddr::from(([127u8, 0, 0, 1], 9898)));
```
fish-sammy, line 97: Should use the config to provide default instead of this kind of hardcoding. It also makes not sense to use `unwrap_or_else` but `wrap_or` in this situation.

ampd/src/main.rs
```diff
@@ -44,11 +46,12 @@ enum Output {
 
 #[tokio::main]
 async fn main() -> ExitCode {
+    let metrics_server = start_metrics_server(None).await;
```
fish-sammy, line 49: If you read the warp doc, the `.run` methods says `Run this Server forever on the current thread.`. You are now blocking the entire ampd process to run the server now, which breaks the whole thing completely.

ampd/src/metrics.rs
```diff
@@ -0,0 +1,108 @@
...
+
+pub fn increment_verifiers_bonded() {
+    VERIFIERS_BONDED.inc();
+}
+
+pub fn increment_verifier_address_calls() {
+    VERIFIER_ADDRESS_CALLS.inc();
+}
+
+pub struct Timer {
```
fish-sammy, line 68: There should be a metrics client struct, that abstract this away. You shouldn't have to expose a Timer struct, and other code has to understand how to best use it.

ampd/src/metrics.rs
```diff
@@ -0,0 +1,108 @@
...
+            start: Instant::now(),
+        }
+    }
+
+    pub fn elapsed_seconds(&self) -> f64 {
+        self.start.elapsed().as_secs_f64()
+    }
+}
+
+pub fn observe_config_load_time(duration_seconds: f64) {
```
fish-sammy, line 84: This should be something like `client::observe_config_load_time<F>(F: load_config) where F: Fn() { ... }` instead of the duration being passed in.

ampd/src/metrics.rs
```diff
@@ -0,0 +1,108 @@
...
+    // 2-secnd updates
+    pub static ref UPTIME_SECONDS: IntGauge = register_int_gauge!(
+        "ampd_uptime_seconds",
+        "Number of seconds the application has been running"
+    ).unwrap();
+}
+
+lazy_static! {
+    static ref START_TIME: Instant = Instant::now();
+}
```
fish-sammy, line 42: None of these should be done with `lazy_static`, but as part of the server initialization. The server should wrap the metrics, and then other code update the metrics via a `MetricsClient`, which should be clonable. `MetricsClient` update the `Metrics` via a `mpsc::channel`.

ampd/Cargo.toml
```diff
@@ -7,6 +7,9 @@ license = "MIT OR Apache-2.0"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+prometheus = "0.14"
+lazy_static = "1.4"
+warp = "0.3"
```
fish-sammy, line 12: No need of `warp` when we already have `axum` as one of the dependencies.

ampd/src/metrics.rs
```diff
@@ -0,0 +1,108 @@
...
+pub fn observe_config_load_time(duration_seconds: f64) {
+    CONFIG_LOAD_TIME.observe(duration_seconds);
+}
+
+pub fn gather_metrics() -> String {
+    let encoder = TextEncoder::new();
+    let mut buffer = Vec::new();
+    let metric_families = prometheus::gather();
+    encoder.encode(&metric_families, &mut buffer).unwrap();
+    String::from_utf8(buffer).unwrap()
```
fish-sammy, line 93: No unwrap.

PR #893, opened 2025-05-29
refactor(interchain-token-service): improve space efficiency of NumBits

cgorenflo:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros

### DIFF
--- a/contracts/interchain-token-service/src/shared.rs
+++ b/contracts/interchain-token-service/src/shared.rs
@@ -1,5 +1,4 @@
 use core::fmt;
-use std::ops::Deref;
 
 use axelar_wasm_std::IntoContractError;
 use cosmwasm_schema::cw_serde;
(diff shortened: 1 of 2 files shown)


contracts/interchain-token-service/src/shared.rs
```diff
@@ -1,35 +1,36 @@
...
+    #[error("number of bits must be between 1 and 256 (given: {0})")]
     InvalidNumberOfBits(u32),
 }
 
 #[cw_serde]
-#[serde(try_from = "u32")]
-pub struct NumBits(u32);
+#[derive(PartialOrd)]
+#[serde(try_from = "u32", into = "u32")]
+pub struct NumBits(u8);
```
milapsheth, line 16: Migration will be added in a separate PR?

milapsheth: We thought about doing this but it was potentially confusing. You'll need to round up when exposing this in a query to avoid backwards compatibility

contracts/interchain-token-service/src/contract/migrations/v1_1_1.rs
```diff
@@ -79,7 +79,7 @@ pub fn migrate(deps: DepsMut, msg: MigrateMsg) -> Result<(), ContractError> {
 fn convert_max_uint_to_max_bits(max_uint: nonempty::Uint256) -> Result<NumBits, ContractError> {
     // Need to add one to correctly get the number of bits. This will round down if the result is not a power of two
     #[allow(clippy::arithmetic_side_effects)] // can't possibly overflow
-    NumBits::try_from(Uint512::from(*max_uint).ilog2() + 1).map_err(|err| err.into())
```
milapsheth, line 82: This max uint to max bits migration was already done on mainnet, so you would require another migration

cgorenflo: removed

cjcobb23 (approved):

PR #894, opened 2025-05-30
feat(minor-service-registry): add service overrides

eguajardo:
## Description

https://axelarnetwork.atlassian.net/browse/AXE-9306

## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros


packages/service-registry-api/src/msg.rs
```diff
@@ -83,6 +96,12 @@ pub enum QueryMsg {
     #[returns(Service)]
     Service { service_name: String },
 
+    #[returns(Service)]
+    ServiceParams {
```
maancham, line 100: what is the difference between this newly added message and existing `Service { service_name: String },`? Are they both going to be used?

packages/service-registry-api/src/primitives.rs
```diff
@@ -62,3 +62,21 @@ pub enum AuthorizationState {
...
+
+// Represents any modifiable fields of the Service struct
+// Any non-None field overwrites the value currently stored in the Service object
+#[cw_serde]
+pub struct UpdatedServiceParams {
+    pub min_num_verifiers: Option<u16>,
+    pub max_num_verifiers: Option<Option<u16>>,
+    pub min_verifier_bond: Option<nonempty::Uint128>,
+    pub unbonding_period_days: Option<u16>,
+}
```
cjcobb23, line 74: this doesn't feel like it should be in primitives, it's pretty specific to the api

packages/service-registry-api/src/primitives.rs
```diff
@@ -62,3 +62,21 @@ pub enum AuthorizationState {
...
+
+// Represents any overrideable fields of the Service struct
+// Any non-None field overrides the value currently stored in the Service object
+#[cw_serde]
+pub struct ServiceParamsOverride {
+    pub min_num_verifiers: Option<u16>,
+    pub max_num_verifiers: Option<Option<u16>>,
```
cjcobb23, line 81: why double wrapped option?

contracts/service-registry/src/state.rs
```diff
@@ -46,10 +50,111 @@ pub const VERIFIERS_PER_CHAIN: IndexedMap<
...
+            max_num_verifiers: params_override
+                .max_num_verifiers
+                .unwrap_or(service.max_num_verifiers),
+            ..service
+        }),
+        None => Ok(service),
+    }
+}
+
+pub fn default_service_params(
```
cjcobb23, line 84: naming is a little confusing, I feel like this should be called `service` and the above `service_with_overrides`. Or maybe call this `base_service`.

contracts/service-registry/src/state.rs
```diff
@@ -46,10 +50,111 @@ pub const VERIFIERS_PER_CHAIN: IndexedMap<
...
+    storage: &dyn Storage,
+    service_name: &ServiceName,
+) -> error_stack::Result<Service, ContractError> {
+    SERVICES
+        .may_load(storage, service_name)
+        .change_context(ContractError::StorageError)?
+        .ok_or(report!(ContractError::ServiceNotFound))
```
maancham, line 91: Are we using both error types here (`StorageError` and `ServiceNotFound`) at the same time?

contracts/service-registry/src/contract/query.rs
```diff
@@ -72,8 +73,14 @@ pub fn verifier(
...
+    state::default_service_params(deps.storage, &service_name)
+}
+
+pub fn service_params(
+    deps: Deps,
+    service_name: String,
+    chain_name: ChainName,
+) -> error_stack::Result<Service, ContractError> {
+    state::service(deps.storage, &service_name, &chain_name)
 }
```
cjcobb23, line 86: these names are confusing to me as well, i'd call it `service` and `service_with_overrides`. or `base_service` and `service_with_overrides`

packages/service-registry-api/src/primitives.rs
cjcobb23, line 81: can we add some logic to actually enforce the max? doesn't necessarily need to be in this PR, but a followup. I don't like that we are adding stuff to the API that actually does nothing.

cjcobb23 (commented):
couple of comments, looks good mostly

contracts/service-registry/src/state.rs
eguajardo: `StorageError` is only thrown if the `may_load` fails, `ServiceNotFound` is thrown when the return of `may_load` is `None`. So they are not thrown at the same time

contracts/service-registry/src/contract/query.rs
eguajardo: sounds good, I like the base_service/service_with_overrides combo. If some function is just `service` I would prefer it to be the most restrictive one, so in the future, we don't call `service` by accident when the intention was to return with overrides. However, the `QueryMsg` will still need to stay as just `Service` for the base service so we don't change the API and don't introduce a breaking change, what do you think?

contracts/service-registry/src/contract/execute.rs
```diff
@@ -256,3 +244,15 @@ pub fn claim_stake(
...
 }
+
+fn ensure_service_exists(
+    storage: &dyn Storage,
+    service_name: &String,
+) -> Result<(), ContractError> {
+    ensure!(
+        state::has_service(storage, service_name),
+        ContractError::ServiceNotFound
+    );
```
cgorenflo, line 255: I like this change to call has_service, but I don't think extracting the outer function has a lot of benefits, it adds more lines than it saves 😄

contracts/service-registry/src/contract/query.rs
```diff
@@ -72,8 +73,17 @@ pub fn verifier(
...
+) -> error_stack::Result<Service, ContractError> {
+    state::base_service(deps.storage, &service_name)
+}
+
+pub fn service_with_overrides(
+    deps: Deps,
+    service_name: String,
+    chain_name: ChainName,
+) -> error_stack::Result<Service, ContractError> {
+    state::service_with_overrides(deps.storage, &service_name, &chain_name)
```
cgorenflo, line 88: combine with optional chain name

contracts/service-registry/src/state.rs
```diff
@@ -46,10 +51,110 @@ pub const VERIFIERS_PER_CHAIN: IndexedMap<
...
+            max_num_verifiers: params_override
+                .max_num_verifiers
+                .unwrap_or(service.max_num_verifiers),
+            ..service
+        }),
+        None => Ok(service),
+    }
+}
+
+pub fn base_service(
```
cgorenflo, line 85: I'm not sure if anything outside of the state module needs to be concerned with calling explicitly into base or override. Make the ChainName optional and deal with the difference here

contracts/service-registry/src/state.rs
```diff
@@ -46,10 +51,110 @@ pub const VERIFIERS_PER_CHAIN: IndexedMap<
...
+
+pub fn save_service_override(
+    storage: &mut dyn Storage,
+    service_name: &ServiceName,
+    chain: &ChainName,
+    service_params_override: &ServiceParamsOverride,
+) -> error_stack::Result<(), ContractError> {
+    SERVICE_OVERRIDES
+        .save(storage, (service_name, chain), service_params_override)
+        .into_report()
```
cgorenflo, line 147: this will allow an override for a nonexistent service

packages/service-registry-api/src/client.rs
```diff
@@ -71,11 +84,22 @@ impl Client<'_> {
...
 
+    #[deprecated(
+        note = "Use service_params instead, which supports specifying a chain name and returns the service parameters considering chain overrides"
+    )]
     pub fn service(&self, service_name: String) -> Result<Service> {
         let msg = QueryMsg::Service { service_name };
         self.client.query(&msg).change_context_lazy(|| msg.into())
     }
 
+    pub fn service_params(&self, service_name: String, chain_name: ChainName) -> Result<Service> {
```
cgorenflo, line 95: adjust according to the merger suggestion

contracts/service-registry/src/contract/execute.rs
eguajardo: I decided to do it this way, because I noticed some previous inconsistencies between all functions when checking for service and reporting errors. This way enforces a standard error

packages/service-registry-api/src/primitives.rs
eguajardo: if I remember correctly, one of the reasons why we kept it that way is because we already have a precedence with `UpdatedServiceParams`, I think consistency is better in this case unless we are ok introducing a breaking change. Is it really a breaking change considering we don't use `max_num_verifiers` yet and it's only callable by governance?

cgorenflo (approved):

PR #906, opened 2025-06-06
feat: add a macro to create a valid Address from a str literal

cgorenflo:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros


packages/router-api/src/primitives.rs
```diff
@@ -114,12 +114,33 @@ impl TryFrom<String> for Address {
     }
 }
 
+impl Address {
+    pub const fn is_address(value: &str) -> bool {
+        !value.is_empty() && !contains!(value, FIELD_DELIMITER)
```
milapsheth, line 119:
```suggestion
        nonempty::String::is_nonempty(value) && !contains!(value, FIELD_DELIMITER)
```
if we want to follow this pattern recursively to encapsulate the logic

milapsheth: Could also start defining `*::is_valid` so usage becomes `Address::is_valid` instead `Address::is_address` although it doesn't feel as good since the argument itself isn't an Address but a string

milapsheth (approved):

packages/router-api/src/primitives.rs
```diff
@@ -114,12 +114,33 @@ impl TryFrom<String> for Address {
...
+}
+
 impl std::fmt::Display for Address {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "{}", *self.0)
     }
 }
 
+#[macro_export]
+macro_rules! address {
```
milapsheth, line 130: update tests to use this?

cgorenflo: yeah, that's a massive change though, so I didn't want to do that at the same time as introducing it

packages/router-api/src/primitives.rs
cgorenflo: this check only affects the static compiler error, the actual parsing is still defined recursively. So I think I prefer to keep it simple, but I'll think about it

PR #917, opened 2025-06-16
feat(packages): macro for external execute msg permissions

SDavidson1177:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros


contracts/coordinator/src/contract/execute.rs
```diff
@@ -320,16 +320,16 @@ pub fn register_deployment(
 
     Ok(Response::new().add_message(WasmMsg::Execute {
         contract_addr: protocol_contracts.router.to_string(),
-        msg: to_json_binary(&router_api::msg::ExecuteMsg::ExecuteFromCoordinator {
-            original_sender: sender,
-            msg: Box::new(router_api::msg::ExecuteMsg::RegisterChain {
+        msg: to_json_binary(&router_api::msg::ExecuteMsgFromContract::Relay {
```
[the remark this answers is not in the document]
SDavidson1177: Yes, will do. I will do this in a separate PR

integration-tests/src/router_contract.rs
```diff
@@ -19,7 +19,8 @@ impl RouterContract {
         axelarnet: Addr,
         coordinator: Addr,
     ) -> Self {
-        let code = ContractWrapper::new_with_empty(execute, instantiate, query);
+        let code: ContractWrapper<router_api::msg::ExecuteMsgFromContract, _, _, _, _, _, _, _> =
+            ContractWrapper::new_with_empty(execute, instantiate, query);
```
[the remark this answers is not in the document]
SDavidson1177: Unfortunately it cannot. This is because the `app` argument sets one of the generic types to be `AxelarQueryMsg`. The ContractWrapper has a default value of `Empty` for that generic.  When we call `app.store_code` afterwards, `app.store_code` will assume that generic to be `AxelarQueryMsg`, where it is actually `Empty`. So, we need to either specify `AxelarQueryMsg` directly when creating the ContractWrapper, or set it to `_`. The compiler will then infer the correct type `AxelarQueryMsg`.

packages/msgs-derive/tests/macro.rs
```diff
@@ -155,158 +159,166 @@ fn ensure_specific_permissions() {
...
     };
 
+    // Given a set of specific permissions, the EnsurePermissions macro will now determine
+    // a deterministic order of how the permission checking functions should appear as
+    // arguments in 'ensure_permissions'. Previously, the developer was responsible for
+    // calling ensure_permissions. Since the call to ensure premissions is now generated,
+    // this deterministic ordering helps to ensure arguments are inserted in the correct
+    // order.
+    // The order in this test is: gateway3, gateway1, gateway2
+
```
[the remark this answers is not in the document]
SDavidson1177: No worries. Please see my response to your last comment for a better explanation.

contracts/coordinator/src/contract/execute.rs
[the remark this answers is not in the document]
SDavidson1177: Please see #923 for the fix

packages/msgs-derive/src/lib.rs
```diff
@@ -357,3 +442,276 @@ fn build_full_check_function(
...
+        .sorted_by(|a, b| sort_permissions(&a.0, &b.0))
+        .unzip();
+
+    let contract_names_literals: Vec<_> = contract_names
+        .iter()
+        .map(|cn| Literal::string(cn.to_string().as_str()))
+        .collect();
+
+    // Replace ExecuteMsg with ExternalExecute trait
+    // Both ExecuteMsg and ExecuteMsg2 implement this trait
```
cgorenflo, line 655: I believe it's called ExecuteMsgFromContract now

packages/msgs-derive/src/lib.rs
cgorenflo, line 1: replace unwraps with expects. While it's code that's used to generate other code, it's still used in a production context and unwrap panics without giving any context what went wrong.

packages/msgs-derive/src/lib.rs
```diff
@@ -357,3 +442,276 @@ fn build_full_check_function(
...
+        .inputs
+        .push(parse_quote! {msg: #new_msg_ident});
+
+    let validate_fn = validate_external_contract_function(contract_names.clone());
+    let validate_fn = syn::parse_macro_input!(validate_fn as ItemFn);
+
+    let statements = execute_fn.block.stmts;
+    execute_fn.block = parse_quote!(
+        {
+            let msg_old = msg.clone();
```
cgorenflo, line 677: what's this for?

packages/msgs-derive/src/lib.rs
```diff
@@ -357,3 +442,276 @@ fn build_full_check_function(
...
+
+            let (msg, info) = match msg {
```
cgorenflo, line 679:
```suggestion
    execute_fn
        .sig
        .inputs
        .push(parse_quote! {msg: Into<#new_msg_ident>});

    let validate_fn = validate_external_contract_function(contract_names.clone());
    let validate_fn = syn::parse_macro_input!(validate_fn as ItemFn);

    let statements = execute_fn.block.stmts;
    execute_fn.block = parse_quote!(
        {
            let msg_old = msg.clone();

            let (msg, info) = match msg.into() {
```

cgorenflo: this will allow for automatic conversion of ExecuteMsg into ExecuteMsgFromContract, therefore all the tests don't have to define `.into()` when calling execute. Plus, we want the generated execute function to behave as much as possible as written, so there are no unexpected compilation errors for someone who is not familiar with what the macro does

packages/router-api/src/msg.rs
```diff
@@ -10,7 +10,7 @@ use crate::primitives::*;
 #[derive(EnsurePermissions)]
 pub enum ExecuteMsg {
     /// Registers a new chain with the router
-    #[permission(Governance, Specific(coordinator))]
+    #[permission(Governance, External(coordinator))]
```
cgorenflo, line 13: I suggest `Proxy` instead, because really any call to execute is external

packages/msgs-derive/tests/macro.rs
cgorenflo, line 1: add a test for the new proxy case

packages/msgs-derive/src/lib.rs
SDavidson1177: When I used `Into`, I got an error that the cosmwasm `entry_point` macro could not infer the message type. After some digging, it looks like this must be a concrete type. So, I had to keep it as is
Screenshot 2025-06-27 at 3 57 01 PM

integration-tests/src/router_contract.rs
```diff
@@ -19,7 +19,8 @@ impl RouterContract {
         axelarnet: Addr,
         coordinator: Addr,
     ) -> Self {
-        let code = ContractWrapper::new_with_empty(execute, instantiate, query);
+        let code: ContractWrapper<router_api::msg::ExecuteMsgFromProxy, _, _, _, _, _, _, _> =
```
cgorenflo, line 22:
```suggestion
        let code =
```

packages/msgs-derive/src/lib.rs
```diff
@@ -357,3 +457,274 @@ fn build_full_check_function(
...
+        .inputs
+        .push(parse_quote! {msg: #new_msg_ident});
+
+    let validate_fn = validate_external_contract_function(contract_names.clone());
+    let validate_fn = syn::parse_macro_input!(validate_fn as ItemFn);
+
+    let statements = execute_fn.block.stmts;
+    execute_fn.block = parse_quote!(
+        {
+            let (msg, info) = match msg {
```
cgorenflo, line 692:
```suggestion
            let (msg, info) = match #new_msg_ident::from(msg) {
```
this should solve the `into` ambiguity

cgorenflo (dismissed):

packages/msgs-derive/tests/macro.rs
```diff
@@ -36,6 +40,53 @@ enum TestMsg2 {
...
+pub enum TestMsg3 {
+    #[permission(Any, Proxy(gateway1))]
+    Any,
+    #[permission(Elevated, Proxy(gateway1))]
+    Proxy1,
+    #[permission(Specific(gateway1), Proxy(gateway1))]
+    Proxy2,
+    #[permission(Any, Proxy(gateway1))]
+    Proxy3,
+    #[permission(Any, Proxy(gateway1, gateway2, gateway3))]
```
cgorenflo, line 60: keep at least one of them as `Specific`, `Proxy2` only checks what happens when you have the same name for both roles

packages/msgs-derive/src/lib.rs
cgorenflo: I read the type annotation error incorrectly, this might not be needed (`msg.into()` should be enough)

packages/msgs-derive/src/lib.rs
cgorenflo: this is not resolved yet

packages/msgs-derive/src/lib.rs
cgorenflo: ok, looks like we need to go with a slightly more sophisticated approach. Essentially, instead of simply defining `execute(deps: DepsMut, env: Env, info: MessageInfo, msg: Into<ExecuteMsgFromProxy>) -> Result<Response, ContractError>`, it must be `execute(deps: DepsMut, env: Env, info: MessageInfo, msg: impl Into<ExecuteMsgFromProxy> + DeserializedOwned) -> Result<Response, ContractError>`

We can tackle this in a separate PR though, I don't want this to hold up the merge of this work any longer

cgorenflo (approved):

PR #918, opened 2025-06-16
feat(service-registry): enforce max verifier set limit in service registry

yidan233:
## Description
AXE-9631
Added counter storage: AUTHORIZED_VERIFIER_COUNT map tracks authorized verifiers per service
prevent AuthorizeVerifiers if that exceeds the allowable number of verifiers
prevent UpdateService if a new lower max is set and that is below the current number of authorized verifiers

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros

### DIFF
--- a/contracts/service-registry/src/contract.rs
+++ b/contracts/service-registry/src/contract.rs
@@ -262,7 +262,7 @@ mod test {
                 .expect("Failed to get authorized verifier count");
 
         let actual_count = crate::state::VERIFIERS
-            .prefix(&service_name.to_string()) 
+            .prefix(&service_name.to_string())
             .range(&deps.storage, None, None, cosmwasm_std::Order::Ascending)
             .filter_map(|item| item.ok().map(|(_, verifier)| verifier))
             .filter(|verifier| verifier.authorization_state == AuthorizationState::Authorized)
--- a/contracts/service-registry/src/contract.rs
+++ b/contracts/service-registry/src/contract.rs
@@ -254,16 +254,15 @@ mod test {
     }
     fn check_authorized_verifier_count(
         deps: &OwnedDeps<MockStorage, MockApi, MockQuerier, Empty>,
-        service_name: &str,
+        service_name: &String,
         expected: u16,
     ) {
-        let stored_count = crate::state::AUTHORIZED_VERIFIER_COUNT
-            .may_load(&deps.storage, &service_name.to_string())
-            .unwrap()
-            .expect("service should exist");
+        let stored_count =
+            crate::state::count_authorized_verifiers(&deps.storage, &service_name.to_string())
+                .expect("Failed to get authorized verifier count");
 
         let actual_count = crate::state::VERIFIERS
-            .prefix(&service_name.to_string()) // ← Use your better approach
+            .prefix(&service_name.to_string()) // No need to convert to string again
             .range(&deps.storage, None, None, cosmwasm_std::Order::Ascending)
             .filter_map(|item| item.ok().map(|(_, verifier)| verifier))
             .filter(|verifier| verifier.authorization_state == AuthorizationState::Authorized)
@@ -345,7 +344,61 @@ mod test {
         assert!(res.is_ok());
         service_params_override
     }
+    // Helper function to set up test environment with 5 authorized verifiers
+    fn setup_and_authorize_5() -> (
+        OwnedDeps<MockStorage, MockApi, MockQuerier, Empty>,
+        MockApi,
+        String,
+        Vec<String>,
+    ) {
+        let mut deps = setup();
+        let api = deps.api.clone();
+        let service_name = "validators";
 
+        // Register service
+        execute(
+            deps.as_mut(),
+            mock_env(),
+            message_info(&api.addr_make(GOVERNANCE_ADDRESS), &[]),
+            ExecuteMsg::RegisterService {
+                service_name: service_name.into(),
+                coordinator_contract: api.addr_make(COORDINATOR_ADDRESS).to_string(),
+                min_num_verifiers: 0,
+                max_num_verifiers: Some(10),
+                min_verifier_bond: Uint128::new(100).try_into().unwrap(),
+                bond_denom: AXL_DENOMINATION.into(),
+                unbonding_period_days: 10,
+                description: "Some service".into(),
+            },
+        )
+        .expect("Failed to register service");
+
+        // Create verifier addresses
+        let verifiers = vec![
+            api.addr_make("verifier1").to_string(),
+            api.addr_make("verifier2").to_string(),
+            api.addr_make("verifier3").to_string(),
+            api.addr_make("verifier4").to_string(),
+            api.addr_make("verifier5").to_string(),
+        ];
+
+        // Authorize all verifiers
+        execute(
+            deps.as_mut(),
+            mock_env(),
+            message_info(&api.addr_make(GOVERNANCE_ADDRESS), &[]),
+            ExecuteMsg::AuthorizeVerifiers {
+                verifiers: verifiers.clone(),
+                service_name: service_name.into(),
+            },
+        )
+        .expect("Failed to authorize verifiers");
+
+        // Verify count is correct
+        check_authorized_verifier_count(&deps, &service_name.into(), 5);
+
+        (deps, api, service_name.into(), verifiers)
+    }
     #[test]
     fn register_service() {
         let mut deps = setup();
@@ -2848,7 +2901,7 @@ mod test {
         assert!(err_contains!(
             err.report,
             ContractError,
-            ContractError::MaxVerifiersExceeded(4, 1)
+            ContractError::VerifierLimitExceed(4, 1)
         ));
 
         // Third batch: authorize 1 more verifier  -> 4 verifiers authorized -> should success
@@ -2950,12 +3003,10 @@ mod test {
         assert!(res.is_ok());
     }
     #[test]
-    fn authorized_verifier_count_stays_in_sync() {
+    fn register_service_initializes_with_zero_authorized_verifiers() {
         let mut deps = setup();
         let api = deps.api;
-
         let service_name = "validators";
-        let min_verifier_bond: nonempty::Uint128 = Uint128::new(100).try_into().unwrap();
 
         let res = execute(
             deps.as_mut(),
@@ -2966,7 +3017,7 @@ mod test {
                 coordinator_contract: api.addr_make(COORDINATOR_ADDRESS).to_string(),
                 min_num_verifiers: 0,
                 max_num_verifiers: Some(10),
-                min_verifier_bond,
+                min_verifier_bond: Uint128::new(100).try_into().unwrap(),
                 bond_denom: AXL_DENOMINATION.into(),
                 unbonding_period_days: 10,
                 description: "Some service".into(),
@@ -2974,33 +3025,14 @@ mod test {
         );
         assert!(res.is_ok());
 
-        check_authorized_verifier_count(&deps, &service_name, 0);
+        check_authorized_verifier_count(&deps, &service_name.into(), 0);
+    }
 
-        // authorized 5 verifiers (None -> Authorized)
-        // expected: 5
-        let verifiers_batch_1 = vec![
-            api.addr_make("verifier1").to_string(),
-            api.addr_make("verifier2").to_string(),
-            api.addr_make("verifier3").to_string(),
-            api.addr_make("verifier4").to_string(),
-            api.addr_make("verifier5").to_string(),
-        ];
+    #[test]
+    fn re_authorizing_same_verifiers_does_not_increase_count() {
+        let (mut deps, api, service_name, _verifiers) = setup_and_authorize_5();
 
-        let res = execute(
-            deps.as_mut(),
-            mock_env(),
-            message_info(&api.addr_make(GOVERNANCE_ADDRESS), &[]),
-            ExecuteMsg::AuthorizeVerifiers {
-                verifiers: verifiers_batch_1.clone(),
-                service_name: service_name.into(),
-            },
-        );
-        assert!(res.is_ok());
-
-        check_authorized_verifier_count(&deps, &service_name, 5);
-
-        // Re-authorize already authorized verifiers (should not change count) authroized -> authroized
-        // expected: 5
+        check_authorized_verifier_count(&deps, &service_name.clone().into(), 5);
         let res = execute(
             deps.as_mut(),
             mock_env(),
@@ -3010,15 +3042,16 @@ mod test {
                     api.addr_make("verifier1").to_string(),
                     api.addr_make("verifier2").to_string(),
                 ],
-                service_name: service_name.into(),
+                service_name: service_name.clone().into(),
             },
         );
         assert!(res.is_ok());
+        check_authorized_verifier_count(&deps, &service_name.into(), 5);
+    }
+    #[test]
+    fn unauthorize_verifiers_reduces_count() {
+        let (mut deps, api, service_name, _verifiers) = setup_and_authorize_5();
 
-        check_authorized_verifier_count(&deps, &service_name, 5);
-
-        // Unauthorize 2 verifiers (Authorized -> NotAuthorized)
-        // expected: 3
         let res = execute(
             deps.as_mut(),
             mock_env(),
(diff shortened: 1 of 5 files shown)


contracts/service-registry/src/state.rs
```diff
@@ -89,6 +88,16 @@ const SERVICES: Map<&ServiceName, Service> = Map::new("services");
 const SERVICE_OVERRIDES: Map<(&ServiceName, &ChainName), ServiceParamsOverride> =
     Map::new("service_overrides");
 
+const AUTHORIZED_VERIFIER_COUNT: Map<&ServiceName, u16> = Map::new("authorized_verifier_count");
```
maancham, line 91: Would you please remind me again why we are using `u16` for this and not something like `u64` or `u32`?

contracts/service-registry/src/state.rs
```diff
@@ -122,6 +131,42 @@ pub fn service(
...
+                        authorization_state: auth_state,
+                        service_name: service_name.clone(),
+                    }),
+                }
+            },
+        )
+        .change_context(ContractError::StorageError)
+}
+
+pub fn save_servicename_in_counter(
```
maancham, line 162: `register_new_service_counter` or something similar is better I guess

contracts/service-registry/src/state.rs
```diff
@@ -122,6 +131,42 @@ pub fn service(
...
+    storage: &mut dyn Storage,
+    service_name: &ServiceName,
+) -> error_stack::Result<(), ContractError> {
+    AUTHORIZED_VERIFIER_COUNT
+        .save(storage, service_name, &0u16)
+        .into_report()
```
maancham, line 168: what will this `into_report` return in terms of error type?

contracts/service-registry/src/state.rs
```diff
@@ -139,11 +184,96 @@ pub fn has_service(storage: &dyn Storage, service_name: &ServiceName) -> bool {
...
+    Ok(())
+}
+
+pub fn count_authorized_verifiers(
+    storage: &dyn Storage,
+    service_name: &ServiceName,
+) -> error_stack::Result<u16, ContractError> {
+    let count = AUTHORIZED_VERIFIER_COUNT
+        .may_load(storage, service_name)
+        .change_context(ContractError::StorageError)?
```
maancham, line 257: which error are we going to see first?  `StorageError` if there is a problem loading from the storage (like not found) or `ServiceNotFound`?

contracts/service-registry/src/state.rs
```diff
@@ -139,11 +184,96 @@ pub fn has_service(storage: &dyn Storage, service_name: &ServiceName) -> bool {
...
+    Ok(count)
+}
+
 pub fn update_service(
     storage: &mut dyn Storage,
     service_name: &ServiceName,
     updated_service_params: UpdatedServiceParams,
 ) -> error_stack::Result<Service, ContractError> {
+    if let Some(new_max) = updated_service_params.max_num_verifiers {
+        let current_authorized = count_authorized_verifiers(storage, service_name)?;
```
maancham, line 268: `current_authorized_count` or `number_of_authorized_verifiers`

contracts/service-registry/src/contract.rs
```diff
@@ -314,7 +344,59 @@ mod test {
         assert!(res.is_ok());
         service_params_override
     }
+    // Helper function to set up test environment with 5 authorized verifiers
+    fn setup_and_authorize_5() -> (
```
maancham, line 348: can we have a better name for this function? Also, please remove the comments. 
Can we also remove the use of `.expect` as much as possible? If there are no other options, then I guess its ok.

contracts/service-registry/src/contract.rs
```diff
@@ -2754,4 +2836,298 @@ mod test {
...
+            deps.as_mut(),
+            mock_env(),
+            message_info(&api.addr_make(GOVERNANCE_ADDRESS), &[]),
+            ExecuteMsg::AuthorizeVerifiers {
+                verifiers: verifiers_2.clone(),
+                service_name: service_name.into(),
+            },
+        );
+        let err = res.unwrap_err();
+        assert!(err_contains!(
```
maancham, line 2899: can you please look into how we are using `assert!(matches!...` and using `err.current_context()` to check for these?

contracts/service-registry/src/state.rs
yidan233: I think we will first see `ContractError::StorageError` if the storage access (loading) failed,  and then if the item loaded is` None`, `ok_or(report!(ContractError::ServiceNotFound))` will be returned..?

contracts/service-registry/src/state.rs
```diff
@@ -89,6 +88,16 @@ const SERVICES: Map<&ServiceName, Service> = Map::new("services");
 const SERVICE_OVERRIDES: Map<(&ServiceName, &ChainName), ServiceParamsOverride> =
     Map::new("service_overrides");
 
+const AUTHORIZED_VERIFIER_COUNT: Map<&ServiceName, u16> = Map::new("authorized_verifier_count");
+
+pub fn load_max_verifiers(
+    storage: &dyn Storage,
+    service_name: &ServiceName,
+) -> error_stack::Result<Option<u16>, ContractError> {
+    let service = service(storage, service_name, None)?;
```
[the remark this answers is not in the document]
yidan233: I think `service()` is already returning ` error_stack::Result<Service, ContractError> `https://github.com/axelarnetwork/axelar-amplifier/blob/main/contracts/service-registry/src/state.rs#L96 and the all errors are already contextualized there

contracts/service-registry/src/state.rs
```diff
@@ -7,10 +7,12 @@ use report::ResultExt;
 use router_api::ChainName;
 use service_registry_api::error::ContractError;
 use service_registry_api::{AuthorizationState, BondingState, Service, Verifier};
-
 type ServiceName = String;
 type VerifierAddress = Addr;
-
+pub enum CountOperation {
```
maancham, line 12: `VerifierCountOperation`

contracts/service-registry/src/contract/execute.rs
```diff
@@ -34,9 +34,38 @@ pub fn register_service(
             description,
         },
     )?;
-
+    state::register_new_service_counter(deps.storage, &service_name)?;
     Ok(Response::new())
 }
```
maancham, line 39: there should be a line between the end of this function and start of the next one.

contracts/service-registry/src/contract/execute.rs
```diff
@@ -34,9 +34,38 @@ pub fn register_service(
...
+fn ensure_authorization_max_limit_respected(
+    deps: &DepsMut,
+    service_name: &String,
+    verifiers: &[Addr],
+) -> Result<(), ContractError> {
+    let max_limit = state::get_service_verifier_max_limit(deps.storage, service_name)?;
```
maancham, line 45: use this function in the file imports at the beginning `use crate::state::{self, CountOperation, ...`
You could also do the same for `state::register_new_service_counter` in the previous function.

contracts/service-registry/src/state.rs
```diff
@@ -139,11 +188,85 @@ pub fn has_service(storage: &dyn Storage, service_name: &ServiceName) -> bool {
...
+            None => {
+                count = count
+                    .checked_add(1)
+                    .ok_or(ContractError::AuthorizedVerifiersExceedu16)?;
+            }
+            _ => {}
+        }
+    }
+    Ok(count)
+}
```
maancham, line 218: we need a new line here as well.

contracts/service-registry/src/contract/execute.rs
```diff
@@ -46,30 +75,54 @@ pub fn update_verifier_authorization_status(
 ) -> Result<Response, ContractError> {
     ensure_service_exists(deps.storage, &service_name)?;
 
+    if auth_state == AuthorizationState::Authorized {
+        ensure_authorization_max_limit_respected(&deps, &service_name, &verifiers)?;
+    }
     for verifier in verifiers {
-        VERIFIERS.update(
+        let old_state = state::get_verifier_auth_state(deps.storage, &service_name, &verifier)?;
```
maancham, line 82: `previous_auth_state`

contracts/service-registry/src/contract/execute.rs
```diff
@@ -46,30 +75,54 @@ pub fn update_verifier_authorization_status(
...
+            &service_name,
+            &auth_state,
+            old_state,
         )?;
     }
 
     Ok(Response::new())
 }
 
+fn update_count_based_on_state_transition(
```
maancham, line 100: this can be refactored into a cleaner version. How about this?

```
let operation = match (old_state, auth_state) {
        (Some(NotAuthorized | Jailed), Authorized) | (None, Authorized) => CountOperation::Increment,
        (Some(Authorized), NotAuthorized | Jailed) => CountOperation::Decrement,
        _ => return Ok(()),
    };
    
state::update_authorized_count(storage, service_name, operation)
  ```
  
 Also, the function name should specify which `count` we are updating. Lastly, I think this function should be over state.rs not here.

File: contracts/service-registry/src/state.rs [hunk 15]
```diff
@@ -139,11 +188,85 @@ pub fn has_service(storage: &dyn Storage, service_name: &ServiceName) -> bool {
...
+    storage: &mut dyn Storage,
+    service_name: &ServiceName,
+    operation: CountOperation,
+) -> error_stack::Result<(), ContractError> {
+    AUTHORIZED_VERIFIER_COUNT
+        .update(
+            storage,
+            service_name,
+            |count| -> std::result::Result<u16, ContractError> {
+                let current = count.ok_or(ContractError::ServiceNotFound)?;
```
maancham, line 229: why is this an option here?

contracts/service-registry/src/state.rs
maancham: Do you have a unit test for these two paths?

contracts/service-registry/src/state.rs
```diff
@@ -139,11 +188,85 @@ pub fn has_service(storage: &dyn Storage, service_name: &ServiceName) -> bool {
...
+        .ok_or(report!(ContractError::ServiceNotFound))?;
+    Ok(count)
+}
+
 pub fn update_service(
     storage: &mut dyn Storage,
     service_name: &ServiceName,
     updated_service_params: UpdatedServiceParams,
 ) -> error_stack::Result<Service, ContractError> {
+    if let Some(new_max) = updated_service_params.max_num_verifiers {
```
maancham, line 260: The nested if let statements can be flattened. You could use two `Some` or `.flatten()`

contracts/service-registry/src/state.rs
yidan233:
```
/// Loads the data, perform the specified action, and store the result
    /// in the database. This is shorthand for some common sequences, which may be useful.
    ///
    /// If the data exists, `action(Some(value))` is called. Otherwise `action(None)` is called.
    ///
    /// This is a bit more customized than needed to only read "old" value 1 time, not 2 per naive approach
    pub fn update<A, E>(
        &self,
        store: &mut dyn Storage,
        k: K,
        height: u64,
        action: A,
    ) -> Result<T, E>
    where
```
 I guess is converted by update()?

contracts/service-registry/src/state.rs
yidan233: I do have a test that checks for the `ServiceNotFound `case, but I’m not sure how to mock` a StorageError (`e.g., making the storage access fail or panic). I looked around the codebase but couldn’t find an example of how to simulate that kind of error..

contracts/service-registry/src/contract/execute.rs
```diff
@@ -34,10 +37,35 @@ pub fn register_service(
...
+        ensure!(
+            total_after_update <= max_limit,
+            ContractError::VerifierLimitExceed
+        );
+    }
+
+    Ok(())
+}
+
 pub fn update_verifier_authorization_status(
```
cgorenflo, line 69: suggestion for a slightly modified approach:
- let `state` fully control how the verifier state is modelled, i.e. `execute` gives the command to change the auth state, but doesn't care how that is done
- after the authorization call get the new number
- ensure the new number is below or equal to the limit

you don't have to implement a "simulation mode", on error everything gets rolled back

packages/service-registry-api/src/error.rs
```diff
@@ -11,10 +11,16 @@ pub enum ContractError {
...
-
     #[error(transparent)]
     NonEmpty(#[from] nonempty::Error),
-
+    #[error("verifiers count exceeds u16 limit")]
+    AuthorizedVerifiersExceedu16,
+    #[error("verifiers count is negative")]
+    AuthorizedVerifiersNegative,
+    #[error("max verifiers limit exceeded")]
+    VerifierLimitExceed,
```
cgorenflo, line 21:
```suggestion
    VerifierLimitExceeded,
```

packages/service-registry-api/src/error.rs
```diff
@@ -11,10 +11,16 @@ pub enum ContractError {
 
     #[error(transparent)]
     Overflow(#[from] OverflowError),
-
     #[error(transparent)]
     NonEmpty(#[from] nonempty::Error),
-
+    #[error("verifiers count exceeds u16 limit")]
+    AuthorizedVerifiersExceedu16,
```
cgorenflo, line 17:
```suggestion
    #[error("not more than 65535 authorized verifiers allowed")]
    AuthorizedVerifiersIntegerOverflow,
```

packages/service-registry-api/src/error.rs
```diff
@@ -11,10 +11,16 @@ pub enum ContractError {
...
+    #[error("verifiers count is negative")]
+    AuthorizedVerifiersNegative,
```
cgorenflo, line 19: with the changes I proposed the count will be modified immediately where the verifiers state changes, so this becomes even more of an invariant (this can never happen as long as the program runs normally). At that point, I would prefer removing the error and replacing it with an `.expect()`, because if we manage to get to a potentially negative number here something has gone horribly wrong and the call parameters of the contract shouldn't have any influence on it

contracts/service-registry/src/state.rs
```diff
@@ -195,6 +219,61 @@ pub fn remove_service_override(
...
+                    let (new_verifier, count_change) = match existing_verifier {
+                        Some(mut verifier) => {
+                            let old_state = Some(verifier.authorization_state.clone());
+                            let operation = calculate_auth_count_change(&old_state, auth_state);
+                            verifier.authorization_state = auth_state.clone();
+                            (verifier, operation)
+                        }
+
+                        None => {
+                            let operation = calculate_auth_count_change(&None, auth_state);
```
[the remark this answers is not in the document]
maancham: also, I guess using `count_operation` is better that `operation`

contracts/service-registry/src/state.rs
```diff
@@ -195,6 +219,61 @@ pub fn remove_service_override(
...
+                            let verifier = Verifier {
+                                address: verifier_addr.clone(),
+                                bonding_state: BondingState::Unbonded,
+                                authorization_state: auth_state.clone(),
+                                service_name: service_name.clone(),
+                            };
+                            (verifier, operation)
+                        }
+                    };
+                    authorized_count_delta =
```
maancham, line 255: I am not sure about this, I wanted to say it would be good to pass this count as mutable argument to the function and change it over there rather than returning and overwriting. @cgorenflo any preference?

contracts/service-registry/src/state.rs
```diff
@@ -300,6 +379,70 @@ pub fn deregister_chains_support(
...
+            .checked_add(1)
+            .ok_or(ContractError::AuthorizedVerifiersIntegerOverflow),
+        Some(VerifierCountOperation::Decrement) => Ok(current
+            .checked_sub(1)
+            .expect("authorized verifiers count should not underflow")),
+        None => Ok(current),
+    }
+}
+
+fn apply_authorized_count_delta(
```
maancham, line 412: maybe avoid using delta? Something like `update` or `change` would be easier to understand.

contracts/service-registry/src/state.rs
```diff
@@ -300,6 +379,70 @@ pub fn deregister_chains_support(
...
+    }
+
+    AUTHORIZED_VERIFIER_COUNT
+        .update(
+            storage,
+            service_name,
+            |count| -> std::result::Result<u16, ContractError> {
+                let current = count.ok_or(ContractError::ServiceNotFound)?;
+
+                let delta_magnitude = u16::try_from(delta.abs())
```
maancham, line 428: again, you could use `change` instead of delta if you're ok with it

contracts/service-registry/src/contract/execute.rs
```diff
@@ -20,11 +20,11 @@ pub fn register_service(
     unbonding_period_days: u16,
     description: String,
 ) -> Result<Response, ContractError> {
-    state::save_new_service(
+    save_new_service(
```
cgorenflo, line 23: prefer to keep the module name, so it's clear this affects the state

contracts/service-registry/src/contract/execute.rs
```diff
@@ -20,11 +20,11 @@ pub fn register_service(
...
         deps.storage,
-        &service_name.clone(),
+        &service_name,
         Service {
-            name: service_name,
+            name: service_name.clone(),
```
cgorenflo, line 27:
```suggestion
            name: service_name,
```

contracts/service-registry/src/state.rs
```diff
@@ -195,6 +214,64 @@ pub fn remove_service_override(
...
+
+    for verifier_addr in verifiers.iter() {
+        VERIFIERS
+            .update(
+                storage,
+                (service_name, verifier_addr),
+                |existing_verifier| -> std::result::Result<Verifier, ContractError> {
+                    let previous_state = existing_verifier
+                        .as_ref()
+                        .map(|verifier| verifier.authorization_state.clone());
```
cgorenflo, line 233: do you need to clone here?

contracts/service-registry/src/state.rs
```diff
@@ -195,6 +214,64 @@ pub fn remove_service_override(
...
+
+                    update_auth_verifier_count_change_tracker(
+                        &mut authorized_count_change,
+                        &previous_state,
+                        auth_state,
+                    )?;
+
+                    let new_verifier = match existing_verifier {
+                        Some(mut verifier) => {
+                            verifier.authorization_state = auth_state.clone();
```
cgorenflo, line 243: don't borrow a type just to clone it later, prefer taking ownership of the object (update_verifier_authorization_status should take a reference to auth_state)

contracts/service-registry/src/state.rs
```diff
@@ -195,6 +214,64 @@ pub fn remove_service_override(
...
+                            verifier
+                        }
+
+                        None => Verifier {
+                            address: verifier_addr.clone(),
+                            bonding_state: BondingState::Unbonded,
+                            authorization_state: auth_state.clone(),
```
cgorenflo, line 250: same here

contracts/service-registry/src/state.rs
```diff
@@ -195,6 +214,64 @@ pub fn remove_service_override(
...
+                    )?;
+
+                    let new_verifier = match existing_verifier {
+                        Some(mut verifier) => {
+                            verifier.authorization_state = auth_state.clone();
+                            verifier
+                        }
+
+                        None => Verifier {
+                            address: verifier_addr.clone(),
```
cgorenflo, line 248: same here

contracts/service-registry/src/state.rs
```diff
@@ -300,6 +377,61 @@ pub fn deregister_chains_support(
...
+        }
+        (Some(Authorized), NotAuthorized) | (Some(Authorized), Jailed) => {
+            *current = current
+                .checked_sub(1)
+                .expect("authorized verifiers count should not underflow");
+        }
+        _ => {}
+    }
+    Ok(())
+}
```
cgorenflo, line 399:
```suggestion
fn calculate_auth_verifier_count_change(
    previous_state: Option<&AuthorizationState>,
    auth_state: &AuthorizationState,
) -> i16 {
    match (previous_state, auth_state) {
        (Some(Authorized), Authorized) => 0,
        (Some(Authorized), _) => -1,
        (_, Authorized) => 1,
        _ => 0
    }
}
```
Try to avoid passing mutable state around. Also &Option<T> clones the underlying type, whereas Option<&T> just copies the reference

contracts/service-registry/src/state.rs
```diff
@@ -195,6 +214,64 @@ pub fn remove_service_override(
...
+                |existing_verifier| -> std::result::Result<Verifier, ContractError> {
+                    let previous_state = existing_verifier
+                        .as_ref()
+                        .map(|verifier| verifier.authorization_state.clone());
+
+                    update_auth_verifier_count_change_tracker(
+                        &mut authorized_count_change,
+                        &previous_state,
+                        auth_state,
+                    )?;
```
cgorenflo, line 239:
```suggestion
                    authorized_count_change = authorized_count_change.checked_add(calculate_auth_verifier_count_change(previous_state.as_ref(), &auth_state)).ok_or(ContractError::AuthorizedVerifiersIntegerOverflow);
```
changes according to my other suggestion.

integration-tests/tests/test_utils/mod.rs
```diff
@@ -976,7 +976,7 @@ pub fn register_in_service_registry(
             service_name: protocol.service_name.to_string(),
         },
     );
-    assert!(response.is_ok());
+    assert!(response.is_ok()); // failed here
```
cgorenflo, line 979: what does this comment mean?

contracts/service-registry/src/state.rs
```diff
@@ -195,6 +214,64 @@ pub fn remove_service_override(
     Ok(())
 }
 
+pub fn update_verifier_authorization_status(
+    storage: &mut dyn Storage,
+    service_name: &ServiceName,
+    auth_state: &AuthorizationState,
+    verifiers: &[Addr],
+) -> error_stack::Result<(), ContractError> {
+    let mut authorized_count_change = 0i32;
```
cgorenflo, line 223: use `i16`, it will make it way easier to deal with and it's plenty for verifiers

contracts/service-registry/src/state.rs
```diff
@@ -300,6 +377,61 @@ pub fn deregister_chains_support(
...
+                } else {
+                    Ok(current
+                        .checked_sub(change_magnitude)
+                        .expect("authorized verifiers count should not underflow"))
+                }
+            },
+        )
+        .change_context(ContractError::StorageError)?;
+    Ok(())
+}
```
cgorenflo, line 433:
```suggestion
fn apply_authorized_count_change(
    storage: &mut dyn Storage,
    service_name: &ServiceName,
    change: i16,
) -> error_stack::Result<(), ContractError> {
    if change == 0 {
        return Ok(());
    }

    AUTHORIZED_VERIFIER_COUNT
        .update(
            storage,
            service_name,
            |count| -> std::result::Result<u16, ContractError> {
                let current = count.ok_or(ContractError::ServiceNotFound)?;

               current
                        .checked_add_signed(change)
                        .ok_or(ContractError::AuthorizedVerifiersIntegerOverflow)
            })
}
```

cgorenflo (changes requested):

contracts/service-registry/src/state.rs
cgorenflo: agreed, see my suggestions

contracts/service-registry/src/contract/execute.rs
yidan233: I have borrowed the` service_name` at  the second argument of `save_new_service,` then rust doesn't allow me to move a value when its borrowed and force me to clone it

contracts/service-registry/src/state.rs
```diff
@@ -300,6 +379,42 @@ pub fn deregister_chains_support(
...
 }
 
+fn calculate_auth_verifier_count_change(
+    previous_state: Option<&AuthorizationState>,
+    auth_state: &AuthorizationState,
+) -> i16 {
+    match (previous_state, auth_state) {
+        (Some(NotAuthorized) | Some(Jailed), Authorized) | (None, Authorized) => 1,
+        (Some(Authorized), NotAuthorized) | (Some(Authorized), Jailed) => -1,
+        _ => 0,
```
cgorenflo, line 389: this is very brittle, if we add a new state the transition will not be covered and it will falsely fall into the 0 case. Please go with the suggestion I showed previously, explicitly stating `(Authorized, Authorized)`, `(_, Authorized)`, and `(Authorized, _)`. This will guarantee that future extensions wont cause an error

cgorenflo (approved):

PR #923, opened 2025-06-19
feat(minor-coordinator): coordinator can register deployment with router

SDavidson1177:
## Description

Allow the coordinator to use the router's client when registering chains.

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros

## Notes

I have also made changes to the router's client so that it no longer panics. Previously, trying to route unserializable messages would cause the contract to panic. In order to improve UX and bring the code in better alignment with best practice, an error is now propagated. The resulting behaviour is the same, and the transaction is aborted.


contracts/coordinator/src/contract/execute.rs
```diff
@@ -302,3 +304,29 @@ pub fn instantiate_chain_contracts(
...
+    Ok(Response::new().add_message(
+        router
+            .register_chain(
+                original_sender,
+                deployed_contracts.chain_name,
+                router_api::Address::try_from(deployed_contracts.gateway.to_string())
+                    .change_context(Error::ChainContractsInfo)?,
+                deployed_contracts.msg_id_format,
+            )
+            .change_context(Error::RouterClient)?,
```
cgorenflo, line 330: this error should be improved, there is no execution happening until after the coordinator execution has returned. It failed to create the message to register the chain

contracts/coordinator/src/contract/migrations/mod.rs
```diff
@@ -46,18 +87,61 @@ pub fn migrate(
...
+    let mut new_contract_details: HashMap<String, ChainNameAndMsgFormat> = HashMap::new();
+    for contracts in msg.chain_contracts {
+        if let Some { .. } =
+            new_contract_details.insert(contracts.deployment_name.to_string(), contracts.clone())
+        {
+            return Err(
+                MigrationError::DuplicateDeployment(contracts.deployment_name.to_string()).into(),
+            );
+        }
+    }
```
cgorenflo, line 99:
```suggestion
    let new_contract_details: HashMap<String, ChainNameAndMsgFormat> = 
    msg.chain_contracts.map(|contract|(contracts.deployment_name.to_string(), contract)).collect();
```

contracts/coordinator/src/contract/migrations/mod.rs
```diff
@@ -1,24 +1,65 @@
...
+    pub deployment_name: nonempty::String,
+    pub chain_name: ChainName,
+    pub msg_id_format: MessageIdFormat,
+}
+
 #[cw_serde]
 pub struct MigrateMsg {
     pub router: String,
     pub multisig: String,
+    pub chain_contracts: Vec<ChainNameAndMsgFormat>,
```
cgorenflo, line 62: require a `HashSet` and you won't have to check for duplicates

contracts/coordinator/src/contract/migrations/mod.rs
```diff
@@ -46,18 +87,61 @@ pub fn migrate(
...
+        .collect::<Result<Vec<_>, axelar_wasm_std::error::ContractError>>()?
+    {
+        let deployment_details = new_contract_details.get(&chain_contracts.0).ok_or(report!(
+            MigrationError::MissingDeploymentFromMsg(chain_contracts.0.clone())
+        ))?;
+
+        state::save_deployed_contracts(
+            deps.storage,
+            axelar_wasm_std::nonempty::String::try_from(chain_contracts.0.clone())
+                .change_context(Error::OldConfigNotFound)?,
```
cgorenflo, line 116: this error doesn't make a lot of sense

contracts/coordinator/src/contract/migrations/mod.rs
```diff
@@ -1,24 +1,65 @@
...
+}
+
+#[cw_serde]
+pub struct ChainContracts {
+    pub chain_name: ChainName,
+    pub msg_id_format: MessageIdFormat,
+    pub gateway: Addr,
+    pub voting_verifier: Addr,
+    pub multisig_prover: Addr,
+}
```
cgorenflo, line 47: what are these used for?

contracts/coordinator/src/contract/migrations/mod.rs
cgorenflo, line 1: have there been any chain contracts registered this way with the coordinator? Would it be an issue if we just deleted all entries instead of doing this complicated modification?

packages/router-api/src/client.rs
cgorenflo, line 1: why make these changes?

contracts/coordinator/src/contract/migrations/mod.rs
SDavidson1177: As far as I know, there has not been

contracts/coordinator/src/contract/execute.rs
SDavidson1177: Good point. I have added specific errors in the router client that tell us when it failed to create a message. The coordinator just says that an error occurred with the router client, and the "failed to create message..." error can be seen by inspecting the error stack

contracts/coordinator/src/contract/migrations/mod.rs
SDavidson1177: Yes, I see what you're saying. I have implemented and tested this using a HashSet, and it looks like this approach will take the last `chain_contracts` item if duplicates are provided. Could this lead to ambiguity if the user does not expect that behaviour, or if they accidentally duplicated an entry? Realistically, this should not happen. However, by checking for duplicates, we could mitigate this edge case

contracts/coordinator/src/contract/migrations/mod.rs
```diff
@@ -46,25 +76,62 @@ pub fn migrate(
...
+    let mut chain_contracts: HashMap<nonempty::String, ChainContractsDetails> = HashMap::new();
+
+    for cc in msg.chain_contracts {
+        if chain_contracts
+            .insert(cc.deployment_name.clone(), cc.clone())
+            .is_some()
+        {
+            return Err(MigrationError::DuplicateDeployment(cc.deployment_name.clone()).into());
+        }
+    }
```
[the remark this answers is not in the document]
cgorenflo: prefer immutable state when possible (mutation might sometimes be necessary due to performance considerations or other constraints)

packages/router-api/src/client.rs
```diff
@@ -19,17 +24,55 @@ impl<T> Router<T> {
...
+        Ok(WasmMsg::Execute {
             contract_addr: self.address.to_string(),
-            msg: to_json_binary(&msg).expect("msg should always be serializable"),
+            msg: to_json_binary(&ExecuteMsgFromProxy::Direct(msg.clone()))
+                .map_err(|_| report!(Error::Serialize))?,
             funds: vec![],
         }
-        .into()
+        .into())
     }
```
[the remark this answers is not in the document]
SDavidson1177: Yes, you make a good point. I'll change them back to expects

contracts/gateway/src/contract/execute.rs
```diff
@@ -99,13 +101,13 @@ fn verify(
...
         .then(flat_unzip)
-        .then(|(msgs, events)| (verifier.verify_messages(msgs), events))
+        .then(|(msgs, events)| (verifier.verify_messages(msgs), events)))
 }
 
 fn route(
     router: &Router,
     msgs_by_status: Vec<(VerificationStatus, Vec<Message>)>,
-) -> (Option<CosmosMsg>, Vec<Event>) {
+) -> Result<(Option<CosmosMsg>, Vec<Event>), Error> {
```
cgorenflo, line 110: no result needed

contracts/gateway/src/contract/execute.rs
```diff
@@ -89,8 +91,8 @@ fn group_by_status(
 fn verify(
     verifier: &voting_verifier::Client,
     msgs_by_status: Vec<(VerificationStatus, Vec<Message>)>,
-) -> (Option<CosmosMsg>, Vec<Event>) {
-    msgs_by_status
+) -> Result<(Option<CosmosMsg>, Vec<Event>), Error> {
```
cgorenflo, line 94: no result needed

contracts/gateway/src/contract/execute.rs
```diff
@@ -47,13 +47,15 @@ pub fn route_outgoing_messages(
 fn apply(
     verifier: &voting_verifier::Client,
     msgs: Vec<Message>,
-    action: impl Fn(Vec<(VerificationStatus, Vec<Message>)>) -> (Option<CosmosMsg>, Vec<Event>),
+    action: impl Fn(
+        Vec<(VerificationStatus, Vec<Message>)>,
+    ) -> Result<(Option<CosmosMsg>, Vec<Event>), Error>,
```
cgorenflo, line 52: change not needed

cgorenflo (approved):

PR #926, opened 2025-06-24
refactor(coordinator): remove usage of register prover contract message

SDavidson1177:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros

## Notes

In this PR, I remove all internal usage of RegisterProverContract. The RegisterProverContract endpoint is kept, however, since external processes still depend on it. It will be removed later.

### DIFF
--- a/contracts/coordinator/src/contract/execute.rs
+++ b/contracts/coordinator/src/contract/execute.rs
@@ -47,11 +47,15 @@ pub fn register_chain(
     state::save_chain_contracts(
         deps.storage,
         chain_name.clone(),
-        prover_addr,
+        prover_addr.clone(),
         gateway_addr,
         voting_verifier_address,
     )
-    .change_context(Error::ChainNotRegistered(chain_name))?;
+    .change_context(Error::ChainNotRegistered(chain_name.clone()))?;
+
+    state::save_prover_for_chain(deps.storage, chain_name, prover_addr.clone())
+        .change_context(Error::ProverNotRegistered(prover_addr))?;
+
     Ok(Response::new())
 }
 
--- a/contracts/coordinator/src/contract/execute.rs
+++ b/contracts/coordinator/src/contract/execute.rs
@@ -53,9 +53,6 @@ pub fn register_chain(
     )
     .change_context(Error::ChainNotRegistered(chain_name.clone()))?;
 
-    state::save_prover_for_chain(deps.storage, chain_name, prover_addr.clone())
-        .change_context(Error::ProverNotRegistered(prover_addr))?;
-
     Ok(Response::new())
 }
 
--- a/contracts/coordinator/src/state.rs
+++ b/contracts/coordinator/src/state.rs
@@ -250,10 +250,8 @@ pub fn is_prover_registered(
     match contracts_by_prover(storage, prover_address) {
         Ok(..) => Ok(true),
         Err(e)
-            if e.downcast_ref::<Error>().is_some_and(|e| match e {
-                Error::ProverNotRegistered(..) => true,
-                _ => false,
-            }) =>
+            if e.downcast_ref::<Error>()
+                .is_some_and(|e| matches!(e, Error::ProverNotRegistered(..))) =>
         {
             Ok(false)
         }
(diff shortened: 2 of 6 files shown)


contracts/coordinator/src/contract/execute.rs
```diff
@@ -47,11 +47,12 @@ pub fn register_chain(
     state::save_chain_contracts(
         deps.storage,
         chain_name.clone(),
-        prover_addr,
+        prover_addr.clone(),
```
cgorenflo, line 50: why this clone now?

contracts/coordinator/src/state.rs
```diff
@@ -243,23 +243,24 @@ pub fn is_prover_registered(
...
+    match contracts_by_prover(storage, prover_address) {
+        Ok(..) => Ok(true),
+        Err(e)
+            if e.downcast_ref::<Error>()
+                .is_some_and(|e| matches!(e, Error::ProverNotRegistered(..))) =>
+        {
+            Ok(false)
+        }
+        Err(e) => Err(e),
+    }
```
cgorenflo, line 255: why change this? Now it's casting the option into an error and then tries to dynamically cast it back into an option

contracts/coordinator/src/state.rs
```diff
@@ -243,23 +243,24 @@ pub fn is_prover_registered(
...
 pub fn load_prover_by_chain(
     storage: &dyn Storage,
     chain_name: ChainName,
 ) -> Result<ProverAddress, Error> {
-    Ok(CHAIN_PROVER_INDEXED_MAP
-        .may_load(storage, chain_name.clone())
-        .change_context(Error::StateParseFailed)?
-        .ok_or(Error::ChainNotRegistered(chain_name))?)
+    Ok(contracts_by_chain(storage, chain_name)?.prover_address)
 }
```
cgorenflo, line 264: get rid of the whole function. I'm not a fan of building test specific code into production. Sometimes it's necessary, but getting the prover from all chain contracts is now trivial

contracts/coordinator/src/state.rs
SDavidson1177: I was removing reliance on `CHAIN_PROVER_INDEXED_MAP`, since it is a part of the deprecated/legacy setup. I will clean this up though

contracts/coordinator/src/contract/execute.rs
```diff
@@ -32,8 +32,14 @@ pub fn register_prover(
...
-    state::save_prover_for_chain(deps.storage, chain_name, new_prover_addr.clone())
-        .change_context(Error::ProverNotRegistered(new_prover_addr))?;
+    state::save_chain_contracts(
+        deps.storage,
+        chain_name.clone(),
+        new_prover_addr.clone(),
+        Addr::unchecked(""),
+        Addr::unchecked(""),
+    )
+    .change_context(Error::ProverNotRegistered(new_prover_addr))?;
```
cgorenflo, line 42: This breaks the assumption that we have validated state, and could potentially be quite dangerous (bricking all multisig-provers for example because a coordinator query panics). We'll have to find an alternative

cgorenflo (changes requested):

contracts/coordinator/src/contract/execute.rs
SDavidson1177: Sounds good, I'll revert that commit for now

contracts/coordinator/src/contract/migrations/mod.rs
cgorenflo, line 1: one thing I had forgotten to integrate in my suggestion was the check for duplicates in the contracts vector, which was the reason to use vec over hashmap to begin with. You can simply check that the length of chain_contracts and provers_by_chain is the same. If they are due to a duplicate, then the loop will fail to find some missing entry (it's impossible to have all chains, the same number of elements before deduplication and duplicates at the same time)

cgorenflo (approved):

contracts/coordinator/src/contract/migrations/mod.rs
SDavidson1177: Alright thanks, I will check for that

PR #927, opened 2025-06-25
feat(metrics): make Prometheus endpoint configurable and optional for AMPD

yidan233:
## Description
AXE-9552
Introduced a new` prometheus_monitor config` section with:

> - enabled (bool): toggle metrics endpoint on/off.
> - bind_address (optional SocketAddrV4): specify the server bind address.

the default setting: enabled: False, address = 127.0.0.1:3000
-  `monitor::Server::new(...) `now uses `get_bind_addr() `to check if Prometheus should start. 

> - it returns the address if enabled, to start the live server 
> - it returns `None` if disabled, to start the dummy server 

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros


ampd/src/config.rs
```diff
@@ -45,37 +42,18 @@ impl Default for Config {
...
             service_registry: ServiceRegistryConfig::default(),
             rewards: RewardsConfig::default(),
-            prometheus_monitor_bind_addr: Some(SocketAddrV4::new(Ipv4Addr::UNSPECIFIED, 3000)),
             grpc: grpc::Config::default(),
+            prometheus_monitor: prometheus_metrics::Config::default(),
         }
     }
 }
 
-impl Debug for Config {
```
maancham, line 54: why is this part deleted?

ampd/src/monitoring/mod.rs
```diff
@@ -0,0 +1,68 @@
+mod endpoints;
+pub mod server;
+use std::net::{Ipv4Addr, SocketAddrV4};
```
maancham, line 3: I believe `mod.rs` file should only have other modules and it is better to shift the types and methods to other files

ampd/src/monitoring/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+            self.bind_address
+        } else {
+            None
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::net::{Ipv4Addr, SocketAddrV4};
```
maancham, line 39: aren't these two included based on `use super::*;`?

ampd/src/monitoring/endpoints/metrics.rs
```diff
@@ -63,3 +68,36 @@ pub fn gather(registry: &Registry) -> Result<String, MetricsError> {
 
     String::from_utf8(buffer).change_context(MetricsError::Utf8Error)
 }
+
+#[cfg(test)]
+mod tests {
```
maancham, line 73: can we also have tests for `Err(e) => (StatusCode::INTERNAL_SERVER_ERROR` here please or are they handled somewhere else?

ampd/src/monitoring/mod.rs
yidan233: I was reviewing the structure in` ampd/src/config.rs `and noticed that most components define their configuration in their own` mod.rs `files. 

**grpc**: https://github.com/axelarnetwork/axelar-amplifier/blob/main/ampd/src/grpc/mod.rs#L37
**tofnd:** https://github.com/axelarnetwork/axelar-amplifier/blob/main/ampd/src/tofnd/mod.rs#L19
**broadcaster**: https://github.com/axelarnetwork/axelar-amplifier/blob/main/ampd/src/broadcaster/mod.rs#L67

not sure if I should split them into`config.rs` and `type.rs `

maancham: in that case, I'd suggest you follow the same order of imports used by `mod.rs` of broadcaster (`use` statements, new line, `mod`s)

ampd/src/monitoring/endpoints/metrics.rs
yidan233: I think Prometheus's TextEncoder doesn't easily fail when encoding metrics with data, its quite unexpected to happen..
I could try mocking the render or gather behavior to force it to return an EncodeError, but would that still be helpful..?

ampd/src/monitoring/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+
+#[derive(Clone)]
+pub enum MetricsMsg {
+    IncBlockReceived,
+}
+
+#[derive(Serialize, Deserialize, Clone, PartialEq, Debug)]
+pub struct Config {
+    pub enabled: bool,
+    pub bind_address: Option<SocketAddrV4>,
```
cgorenflo, line 15: this allows the representation of an invalid state `enabled: true, bind_address: None`. Remove the boolean, it's already covered by the `Option`

ampd/src/monitoring/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+
+use serde::{Deserialize, Serialize};
+
+mod endpoints;
+pub mod server;
+
+#[derive(Clone)]
+pub enum MetricsMsg {
+    IncBlockReceived,
+}
```
cgorenflo, line 11: this is part of metrics

ampd/src/monitoring/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+    fn default() -> Self {
+        Self {
+            enabled: false,
+            bind_address: SocketAddrV4::new(Ipv4Addr::new(127, 0, 0, 1), 3000),
+        }
+    }
+}
+
+impl Config {
+    pub fn get_bind_addr(&self) -> Option<SocketAddrV4> {
```
cgorenflo, line 29:
```suggestion
    pub fn bind_addr(&self) -> Option<SocketAddrV4> {
```
Rust convention is not to use `get_`

ampd/src/monitoring/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+
+#[derive(Clone)]
+pub enum MetricsMsg {
+    IncBlockReceived,
+}
+
+#[derive(Serialize, Deserialize, Clone, PartialEq, Debug)]
+pub struct Config {
+    pub enabled: bool,
+    pub bind_address: SocketAddrV4,
```
cgorenflo, line 16:
```suggestion
    enabled: bool,
    bind_address: SocketAddrV4,
```
we want the usage to go through the bind_addr() method

ampd/src/monitoring/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+    #[test]
+    fn serde_some_address_serializes_to_string() {
+        let config = Config::default();
+        let serialized = toml::to_string(&config).unwrap();
+        assert!(serialized.contains("enabled = false"));
+        assert!(serialized.contains("bind_address = \"127.0.0.1:3000\""));
+        let expected = "enabled = false\nbind_address = \"127.0.0.1:3000\"";
+        assert_eq!(serialized.trim(), expected);
+    }
+}
```
cgorenflo, line 69: move to server.rs

ampd/src/config.rs
```diff
@@ -591,4 +568,22 @@ mod tests {
...
+            bind_address = '{bind_address}'
+            ",
+        );
+        let cfg: Config = toml::from_str(&config_str).unwrap();
+        assert!(cfg.monitoring_server.enabled);
+        assert_eq!(
+            cfg.monitoring_server.bind_address,
+            SocketAddrV4::new(Ipv4Addr::new(0, 0, 0, 0), 3001)
+        );
+    }
```
cgorenflo, line 588: can you also add a test where the bind_address is not defined, so we can make sure it won't break deserialization?

ampd/src/lib.rs
```diff
@@ -44,8 +43,8 @@ mod evm;
 mod grpc;
 mod handlers;
 mod json_rpc;
+pub mod monitoring;
```
cgorenflo, line 46:
```suggestion
mod monitoring;
```

PR #930, opened 2025-06-27
feat(minor-interchain-token-service): implement translation hooks for Interchain Token Service (ITS) Hub

cjcobb23:
… Interchain Token Service (ITS) Hub

- Add abi-translation-contract for message translation logic
- Extend ChainConfig with translation_contract field
- Update ITS Hub to use translation hooks for message encoding/decoding
- Add migration for existing chains
- Move ITS types to package

## Description

implements https://github.com/axelarnetwork/arcs/pull/12


contracts/abi-translation-contract/schema.rs
```diff
@@ -0,0 +1,18 @@
...
+fn main() {
+    let mut out_dir = current_dir().unwrap();
+    out_dir.push("schema");
+    create_dir_all(&out_dir).unwrap();
+    remove_schemas(&out_dir).unwrap();
+
+    export_schema(&schema_for!(InstantiateMsg), &out_dir);
+    export_schema(&schema_for!(ExecuteMsg), &out_dir);
+    export_schema(&schema_for!(QueryMsg), &out_dir);
+}
```
cgorenflo, line 18:
```suggestion
} 

```

cgorenflo: does this need to be different than any other bin/schema.rs we have?
```
fn main() {
    write_api! {
        instantiate: InstantiateMsg,
        execute: ExecuteMsg,
        query: QueryMsg,
    }
}

```

contracts/interchain-token-service/src/contract/migrations/mod.rs
```diff
@@ -1,16 +1,507 @@
...
+        if !migration_chains.contains(existing_chain) {
+            return Err(Error::ChainNotFound(existing_chain.clone()).into());
+        }
+    }
+
+    // Migrate each chain's config by loading old state and saving new state
+    for chain_config in msg.chain_translation_configs {
+        // Load the old chain config (without translation_contract)
+        let old_chain_state = load_old_chain_config(deps.storage, &chain_config.chain)
+            .change_context(Error::ChainNotFound(chain_config.chain.clone()))?;
```
cgorenflo, line 99: why do you have to load them all again?

contracts/interchain-token-service/src/contract/execute/mod.rs
```diff
@@ -391,39 +410,46 @@ pub fn enable_execution(deps: DepsMut) -> Result<Response, Error> {
...
 
     Ok(Response::new())
 }
 
-fn register_chain(storage: &mut dyn Storage, config: msg::ChainConfig) -> Result<(), Error> {
-    match state::may_load_chain_config(storage, &config.chain).change_context(Error::State)? {
-        Some(_) => bail!(Error::ChainAlreadyRegistered(config.chain)),
-        None => state::save_chain_config(storage, &config.chain.clone(), &config.into())
+fn register_chain(deps: &mut DepsMut, config: msg::ChainConfig) -> Result<(), Error> {
+    let chain = config.chain.clone();
```
cgorenflo, line 425: you could just use the chain from the validated_config instead of cloning it

contracts/interchain-token-service/tests/execute.rs
```diff
@@ -58,27 +66,33 @@ fn register_update_its_contract_succeeds() {
...
-
     let chain: ChainNameRaw = "ethereum".parse().unwrap();
     let address: Address = "0x1234567890123456789012345678901234567890"
         .parse()
         .unwrap();
+    let translation_contract: Address = MockApi::default()
+        .addr_make("translation_contract")
+        .to_string()
+        .parse()
+        .unwrap();
```
cgorenflo, line 77: use the new `address!` macro

packages/interchain-token-api/src/payload_translation/client.rs
```diff
@@ -0,0 +1,52 @@
...
+    /// Query the translation contract to decode a chain-specific payload into a HubMessage
+    pub fn from_bytes(&self, payload: HexBinary) -> Result<HubMessage, Error> {
+        self.query(&TranslationQueryMsg::FromBytes { payload })
+    }
+
+    /// Query the translation contract to encode a HubMessage into a chain-specific payload
+    pub fn to_bytes(&self, message: HubMessage) -> Result<HexBinary, Error> {
+        self.query(&TranslationQueryMsg::ToBytes { message })
+    }
+}
```
cgorenflo, line 46: is it not possible to make use of the existing packages/client implementation, like for example the verifier contract does?

packages/interchain-token-api/src/payload_translation/mod.rs
cgorenflo, line 1: what's the reason for introducing a submodule here?

cjcobb23: i feel like the payload translation api is a bit of a self contained thing. I wanted to have some separation. It could be it's own package, but I'm not sure that extra layer of misdirection is necessary or helpful.

contracts/interchain-token-service/tests/execute.rs
cjcobb23: i don't think the macro works here because it needs a string literal? But we need an actual valid cosmos address, so it needs to be constructed via this addr_make call

packages/interchain-token-api/src/payload_translation/mod.rs
cjcobb23: moved the api to its own package

contracts/interchain-token-service/src/msg.rs
```diff
@@ -141,6 +142,7 @@ pub struct ChainConfigResponse {
     pub its_edge_contract: Address,
     pub truncation: TruncationConfig,
     pub frozen: bool,
+    pub translation_contract: Addr,
```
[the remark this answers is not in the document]
milapsheth: We should consider adding a wrapper type/alias `ContractAddress` so fields don't need to be suffixed with `_contract`

contracts/interchain-token-service/src/msg.rs
```diff
@@ -127,6 +127,7 @@ pub struct ChainConfig {
     pub chain: ChainNameRaw,
     pub its_edge_contract: Address,
     pub truncation: TruncationConfig,
+    pub translation_contract: Address,
```
milapsheth, line 130:
```suggestion
    pub msg_translator_contract: Addr,
```

milapsheth: Should it be `Addr` since it's an axelar address? Or because we need to additional validation?

contracts/interchain-token-service/src/state.rs
```diff
@@ -40,27 +43,35 @@ pub struct ChainConfig {
...
-pub struct TruncationConfig {
-    pub max_uint_bits: NumBits, // The maximum number of bits used to represent unsigned integer values that is supported by the chain's token standard
-    pub max_decimals_when_truncating: u8, // The maximum number of decimals that is preserved when deploying from a chain with a larger max unsigned integer
-}
-
-impl From<msg::ChainConfig> for ChainConfig {
-    fn from(value: msg::ChainConfig) -> Self {
-        Self {
+impl ChainConfig {
+    pub fn from_input(input: msg::ChainConfig, api: &dyn cosmwasm_std::Api) -> Result<Self, Error> {
```
milapsheth, line 50:
```suggestion
    pub fn from_msg(input: msg::ChainConfig, api: &dyn cosmwasm_std::Api) -> Result<Self, Error> {
```
or `from_config`, or even just calling it `new`

milapsheth: why not change the `From` to `TryFrom` instead?

contracts/interchain-token-service/src/contract/migrations/mod.rs
```diff
@@ -1,16 +1,442 @@
...
+#[cw_serde]
+pub struct OldChainConfig {
+    pub truncation: OldTruncationConfig,
+    pub its_address: Address,
+    pub frozen: bool,
+    // Note: no translation_contract field in old state
+}
+
+// Storage map for old chain configs
+const OLD_CHAIN_CONFIGS: Map<&ChainNameRaw, OldChainConfig> = Map::new("chain_configs");
```
milapsheth, line 32: wrap all of the old types and methods inside a `mod legacy { ... }` so you can access it via `legacy::ChainConfig` instead of having to prefix it with `Old`

example
https://github.com/axelarnetwork/axelar-amplifier-stellar/blob/c86b31621b4b1022f33e6bb5e80358ca6658e8d7/contracts/stellar-interchain-token-service/src/migrate.rs#L19

milapsheth: We could even create a `legacy_state.rs` file with the types and methods. Easier to see the diff and safer to keep it separate

contracts/interchain-token-service/src/contract/migrations/mod.rs
```diff
@@ -1,16 +1,442 @@
...
 #[cfg_attr(not(feature = "library"), entry_point)]
 #[migrate_from_version("1.2")]
 pub fn migrate(
     deps: DepsMut,
     _env: Env,
-    _msg: MigrateMsg,
+    msg: MigrateMsg,
 ) -> Result<Response, axelar_wasm_std::error::ContractError> {
-    Ok(Response::default())
+    // Get all existing chains from old state format
```
milapsheth, line 70: unnecessary comments

contracts/interchain-token-service/src/contract/migrations/mod.rs
```diff
@@ -1,16 +1,442 @@
...
+                    .max_decimals_when_truncating,
+            },
+            its_address: old_chain_state.its_address,
+            frozen: old_chain_state.frozen,
+            translation_contract: validate_cosmwasm_address(
+                deps.api,
+                &translation_contract.to_string(),
+            )?,
+        };
+        save_chain_config(deps.storage, &chain, &new_chain_state)?;
```
milapsheth, line 99:
```suggestion

        save_chain_config(deps.storage, &chain, &new_chain_state)?;
```

contracts/interchain-token-service/src/contract/migrations/mod.rs
```diff
@@ -1,16 +1,442 @@
...
+    let translation_contracts: HashMap<ChainNameRaw, Address> = msg
+        .chain_translation_configs
+        .iter()
+        .map(|config| (config.chain.clone(), config.translation_contract.clone()))
+        .collect();
+
+    for (chain, old_chain_state) in existing_chains {
+        let translation_contract = translation_contracts
+            .get(&chain)
+            .ok_or_else(|| Error::ChainNotFound(chain.clone()))?;
```
milapsheth, line 83:
```suggestion
            .ok_or_else(|| Error::ChainNotFound(chain.clone()))?;

```

contracts/interchain-token-service/src/contract/migrations/mod.rs
```diff
@@ -1,16 +1,442 @@
...
+#[cw_serde]
+pub struct ChainTranslationConfig {
+    pub chain: ChainNameRaw,
+    pub translation_contract: Address,
+}
+
+#[cw_serde]
+pub struct MigrateMsg {
+    /// List of chain and translation contract address pairs
+    pub chain_translation_configs: Vec<ChainTranslationConfig>,
```
milapsheth, line 43: Since all chains use ABI translator at the moment, the migrate msg should specify a single abi_translator contract to use for all chains. It can then be upgraded via UpdateConfig msg if a chain decides to switch

contracts/interchain-token-service/src/contract/execute/mod.rs
```diff
@@ -166,30 +175,40 @@ fn execute_message_on_hub(
...
-fn apply_to_hub(
-    storage: &mut dyn Storage,
-    source_chain: ChainNameRaw,
-    destination_chain: ChainNameRaw,
-    message: Message,
-) -> Result<Message, Error> {
-    ensure_chain_not_frozen(storage, &source_chain)?;
-    ensure_chain_not_frozen(storage, &destination_chain)?;
+/// Translate chain-specific payload to standardized ITS Hub Message format using translation contract
+fn translate_from_bytes(
```
milapsheth, line 179:
```suggestion
fn hub_message_from_bytes(
```
maybe

contracts/interchain-token-service/src/contract/execute/mod.rs
```diff
@@ -166,30 +175,40 @@ fn execute_message_on_hub(
...
+    let translation_client: TranslationClient =
+        client::ContractClient::new(querier, &chain_config.translation_contract).into();
+
+    translation_client
+        .from_bytes(payload.clone())
+        .change_context(Error::InvalidPayload)
+}
+
+/// Translate standardized ITS Hub Message to chain-specific payload format using translation contract
+fn translate_to_bytes(
```
milapsheth, line 197: move these to a separate file, since this one is already too big?

contracts/interchain-token-service/src/contract/execute/interceptors.rs
```diff
@@ -525,15 +542,23 @@ mod test {
...
-                    max_decimals_when_truncating: 6,
+            &ChainConfig::from_input(
+                msg::ChainConfig {
+                    chain: destination_chain.clone(),
+                    its_edge_contract: "itsedgecontract".to_string().try_into().unwrap(),
+                    truncation: TruncationConfig {
+                        max_uint_bits: 64.try_into().unwrap(),
+                        max_decimals_when_truncating: 6,
+                    },
+                    translation_contract: MockApi::default()
```
milapsheth, line 553: use `address!()` since it creates object of `Address` type

contracts/interchain-token-service/src/contract/execute/mod.rs
```diff
@@ -391,39 +410,46 @@ pub fn enable_execution(deps: DepsMut) -> Result<Response, Error> {
...
 
-fn register_chain(storage: &mut dyn Storage, config: msg::ChainConfig) -> Result<(), Error> {
-    match state::may_load_chain_config(storage, &config.chain).change_context(Error::State)? {
-        Some(_) => bail!(Error::ChainAlreadyRegistered(config.chain)),
-        None => state::save_chain_config(storage, &config.chain.clone(), &config.into())
+fn register_chain(deps: &mut DepsMut, config: msg::ChainConfig) -> Result<(), Error> {
+    let chain = config.chain.clone();
+    let validated_config =
+        state::ChainConfig::from_input(config, deps.api).change_context(Error::State)?;
+    match state::may_load_chain_config(deps.storage, &chain).change_context(Error::State)? {
```
milapsheth, line 428:
```suggestion

    match state::may_load_chain_config(deps.storage, &chain).change_context(Error::State)? {
```

contracts/interchain-token-service/src/contract/execute/mod.rs
```diff
@@ -391,39 +410,46 @@ pub fn enable_execution(deps: DepsMut) -> Result<Response, Error> {
...
     Ok(Response::new())
 }
 
-fn register_chain(storage: &mut dyn Storage, config: msg::ChainConfig) -> Result<(), Error> {
-    match state::may_load_chain_config(storage, &config.chain).change_context(Error::State)? {
-        Some(_) => bail!(Error::ChainAlreadyRegistered(config.chain)),
-        None => state::save_chain_config(storage, &config.chain.clone(), &config.into())
+fn register_chain(deps: &mut DepsMut, config: msg::ChainConfig) -> Result<(), Error> {
+    let chain = config.chain.clone();
+    let validated_config =
```
milapsheth, line 426:
```suggestion
    let config =
```
shadow the old one so it can't be accessed anymore? also the more idiomatic pattern in Rust

contracts/interchain-token-service/src/contract/execute/mod.rs
```diff
@@ -391,39 +410,46 @@ pub fn enable_execution(deps: DepsMut) -> Result<Response, Error> {
...
 }
 
-fn update_chain(storage: &mut dyn Storage, config: msg::ChainConfig) -> Result<(), Error> {
-    match state::may_load_chain_config(storage, &config.chain).change_context(Error::State)? {
-        None => bail!(Error::ChainNotRegistered(config.chain)),
-        Some(_) => state::save_chain_config(storage, &config.chain.clone(), &config.into())
+fn update_chain(deps: &mut DepsMut, config: msg::ChainConfig) -> Result<(), Error> {
+    let chain = config.chain.clone();
+    let validated_config =
+        state::ChainConfig::from_input(config, deps.api).change_context(Error::State)?;
```
milapsheth, line 447: same comment

contracts/interchain-token-service/src/contract/execute/mod.rs
```diff
@@ -391,39 +410,46 @@ pub fn enable_execution(deps: DepsMut) -> Result<Response, Error> {
...
+    match state::may_load_chain_config(deps.storage, &chain).change_context(Error::State)? {
```
milapsheth, line 448:
```suggestion

    match state::may_load_chain_config(deps.storage, &chain).change_context(Error::State)? {
```

contracts/interchain-token-service/tests/execute.rs
milapsheth: add a `cosmos_address!` macro instead that does this?

contracts/interchain-token-service/src/state.rs
cjcobb23: it can't be TryFrom because we need the additional api argument

Cargo.toml
```diff
@@ -72,8 +72,11 @@ heck = "0.5.0"
 hex = "0.4.3"
 humantime-serde = "1.1.1"
 integration-tests = { version = "^1.0.0", path = "integration-tests" }
+interchain-token = { version = "^1.0.0", path = "packages/interchain-token" }
```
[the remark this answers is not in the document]
cjcobb23: it could contain more than just types, we could put utility functions here. I'm leaning towards `interchain-token-service-std`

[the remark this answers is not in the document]
cjcobb23: went with `interchain-token-service-std`

contracts/interchain-token-service/src/contract/execute/interceptors.rs
cjcobb23: it will be validated later on, so still need to use the mock api to create

cjcobb23: created a new macro and used it here

contracts/interchain-token-service/tests/execute.rs
cjcobb23: created the new macro

Cargo.toml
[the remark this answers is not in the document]
milapsheth: yeah, sounds good

PR #933, opened 2025-06-30
feat(Monitoring): track total verifier votes cast and vote success rate

yidan233:
## Description
[AXE-9554](https://axelarnetwork.atlassian.net/browse/AXE-9554)

This document outlines a proposed set of Prometheus metrics 
https://www.notion.so/bright-ambert-2bd/Amplifier-Proposed-Prometheus-Metrics-1e6c53fccb7780afa9d0f7a0ff338e6d
**Metrics Tracking** 
This PR adds two new Prometheus metrics to monitor verifier voting behavior in the Amplifier module:
- `verifier_votes_cast_total `(Counter):
  Tracks the total number of votes cast by each verifier per chain.
  **Labels:**` verifier_id`, `chain`
- `verifier_vote_success_rate` (Gauge):
  Measures the success rate of votes cast by each verifier, ranging from 0 to 1.
  **Labels:**` verifier_id,` `chain`
**HandlerInfo:**
- Metrics record the broadcast result after Handler broadcast voting messages.
- To get the label, A new `HandlerInfo `struct is introduced to provide each handler's` verifier_id`, `chain_name`, and `cast_votes` flag.

**Handler:**
- Handlers like `*_verify_msg` and `*_verify_verifier_set` emit vote messages by calling the ` vote_msg` function
(e.g., [evm_verify_msg.rs#L217](https://github.com/axelarnetwork/axelar-amplifier/blob/main/ampd/src/handlers/evm_verify_msg.rs#L217) → [vote_msg()](https://github.com/axelarnetwork/axelar-amplifier/blob/main/ampd/src/handlers/evm_verify_msg.rs#L124)).
- The `multisig handler` does not cast votes, only signs messages.
Therefore, `cast_votes` is explicitly set to `false` for them 
(see [multisig.rs#L171](https://github.com/axelarnetwork/axelar-amplifier/blob/main/ampd/src/handlers/multisig.rs#L171)).

**Metrics**
Three new Prometheus metrics are introduced under the `verifier_votes_casted_ prefix`:
`verifier_votes_casted_successful{verifier_id, chain_name} (IntCounterVec)`
Cumulative count of successfully broadcasted vote messages.
`verifier_votes_casted_failed{verifier_id, chain_name} (IntCounterVec)`
Cumulative count of failed vote message broadcasts.
`verifier_votes_casted_success_rate{verifier_id, chain_name} (GaugeVec)`
Real-time success rate of vote casting per verifier.
success rate = (success cast + failed cast) / total casts
The value ranges from 0.0 to 1.0, and is updated after every vote cast.

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros


ampd/src/event_processor.rs
```diff
@@ -94,6 +96,8 @@ where
                     sleep: event_processor_config.retry_delay,
                     max_attempts: event_processor_config.retry_max_attempts,
                 },
+                monitoring_client.clone(),
```
maancham, line 99: is there a better way to pass this client and not to do clone for each iteration? It could become costly

ampd/src/handlers/config.rs
```diff
@@ -410,4 +574,223 @@ mod tests {
         assert!(!debug_output.contains("API_KEY"));
         assert!(debug_output.contains("RPCFinalizedBlock"));
     }
+
+    #[test]
+    fn handler_info_provides_correct_info_for_all_chain_config() {
```
maancham, line 579: I'd say its best to break down this test into multiple smaller tests and use helper functions (for instance, for doing`assert`) and setup builder to simplify and clean this up a bit

ampd/src/monitoring/endpoints/metrics.rs
```diff
@@ -1,11 +1,20 @@
 use axum::http::StatusCode;
-use error_stack::{Result, ResultExt};
-use prometheus::{Encoder, IntCounter, Registry, TextEncoder};
+use error_stack::{ensure, Result, ResultExt};
+use prometheus::{Encoder, GaugeVec, IntCounter, IntCounterVec, Opts, Registry, TextEncoder};
 use thiserror::Error;
 
 #[derive(Clone)]
+#[allow(clippy::enum_variant_names)] // will add other msg enum variants that dont start with 'inc' when implementing other metrics in the future
```
maancham, line 7: can we change the current two variants not to have this issue?

ampd/src/monitoring/endpoints/metrics.rs
```diff
@@ -24,30 +33,196 @@ pub enum MetricsError {
...
+    block_received: BlockReceivedMetrics,
+    votes_casted: VotesCastedMetrics,
+}
+
+struct BlockReceivedMetrics {
+    total: IntCounter,
+}
+
+struct VotesCastedMetrics {
+    succeed: IntCounterVec,
```
maancham, line 52: `succeeded`?

ampd/src/monitoring/endpoints/metrics.rs
```diff
@@ -24,30 +33,196 @@ pub enum MetricsError {
     MetricRegisterFailed,
     #[error("failed to spawn metric")]
     MetricSpawnFailed,
+    #[error("failed to increment metric counter, value is larger than 1.7976931348623157e+308")]
```
maancham, line 36: I am not sure we will ever need this error type, also, I'd suggest you only mention f64:max rather than the actual value

ampd/src/monitoring/endpoints/metrics.rs
```diff
@@ -1,11 +1,20 @@
...
 
 #[derive(Clone)]
+#[allow(clippy::enum_variant_names)] // will add other msg enum variants that dont start with 'inc' when implementing other metrics in the future
 pub enum MetricsMsg {
     IncBlockReceived,
+    IncSuccessVoteCasted {
+        verifier_id: String,
+        chain_name: String,
+    },
+    IncFailedVoteCasted {
```
maancham, line 14: the naming of these two variants does not make sense. What does it mean to say `FailedVoteCasted`? Is it the same as increasing number of failed casted votes?

ampd/src/monitoring/endpoints/metrics.rs
```diff
@@ -24,30 +33,196 @@ pub enum MetricsError {
...
+    failed: IntCounterVec,
+    success_rate: GaugeVec,
 }
 
 impl Metrics {
     pub fn new(registry: &Registry) -> Result<Self, MetricsError> {
-        let block_received = IntCounter::new("blocks_received", "number of blocks received")
-            .change_context(MetricsError::MetricSpawnFailed)?;
+        let block_received = BlockReceivedMetrics::new()?;
+        let votes_casted = VotesCastedMetrics::new()?;
```
maancham, line 60: maybe better to use `CastedVotesMetrics`?

ampd/src/monitoring/endpoints/metrics.rs
```diff
@@ -24,30 +33,196 @@ pub enum MetricsError {
...
+    }
+
+    fn register(&self, registry: &Registry) -> Result<(), MetricsError> {
+        registry
+            .register(Box::new(self.total.clone()))
+            .change_context(MetricsError::MetricRegisterFailed)?;
+        Ok(())
+    }
+
+    fn increment(&self) -> Result<(), MetricsError> {
```
maancham, line 109: Maybe a race condition here? Between checking the value and incrementing it, another thread could increment the counter. I guess Prometheus counters are atomic and handle overflow internally. could you check it out?

ampd/src/monitoring/endpoints/metrics.rs
```diff
@@ -24,30 +33,196 @@ pub enum MetricsError {
...
+            .failed
+            .with_label_values(&[verifier_id, chain_name])
+            .get();
+
+        let total_votes = successful_votes
+            .checked_add(failed_votes)
+            .ok_or(MetricsError::MetricValueOverflow)?;
+
+        let success_rate = match total_votes {
+            0 => {
```
maancham, line 216: how is this an invalid state?

ampd/src/monitoring/endpoints/metrics.rs
yidan233: yes.. you are right this could cause a race problem. I think when overflow occurs, Prometheus counters just start over around 0 again, doesn't report any error. Then should I get rid of the overflow error? it seems quite unlikely to overflow

ampd/src/monitoring/server.rs
```diff
@@ -137,15 +137,18 @@ impl Server {
...
+        F: Fn(MetricsMsg) -> Result<(), MetricsError> + Send + 'static,
     {
         tokio::spawn(async move {
             loop {
                 tokio::select! {
                     msg = metrics_rx.recv() => {
                         match msg {
                             Some(msg) => {
-                                message_handler(msg)
+                                if let Err(e) = message_handler(msg) {
```
maancham, line 148: I am not sure about this. Breaking the processing loop on any metric error seems too harsh to me. I think they should be best effort. What are the errors that we must break the loop for?

ampd/src/monitoring/endpoints/metrics.rs
```diff
@@ -24,30 +33,196 @@ pub enum MetricsError {
...
+        Ok(Self {
+            succeed,
+            failed,
+            success_rate,
+        })
+    }
+
+    fn register(&self, registry: &Registry) -> Result<(), MetricsError> {
+        registry
+            .register(Box::new(self.succeed.clone()))
```
maancham, line 158: `succeeded` as opposed to `failed`?
or you could move to `success` and `failure` if you want

ampd/src/monitoring/server.rs
yidan233: the `overflow  error` and `zero count success rate`. should I change it back , and `warn!` if those error exists?

ampd/src/monitoring/endpoints/metrics.rs
yidan233: I have got rid of the redundant overflow check, but since there is only one stream handling the messages, I think there might not be a race condition? I am still using `get()` to calculate the success rate, just want to confirm...
pub fn run(self, cancel: CancellationToken) -> JoinHandle<()> {
        tokio::spawn(async move {
            let stream = self.stream.take_until(cancel.cancelled());

            tokio::pin!(stream);
            while let Some(msg) = stream.next().await {
                self.metrics.handle_message(msg);
            }

            if !cancel.is_cancelled() {
                warn!("all metrics clients disconnected, metrics processing stopped");
            }
        })
    }
}
```

File: ampd/src/monitoring/endpoints/metrics.rs (hunk 9 above)
Comment (yidan233) [reply ampd/src/monitoring/endpoints/metrics.rs]: I first increment either `succeeded votes` or `failed votes`, then calculate success rate. So the `total_votes `is at least 1

File: ampd/src/monitoring/endpoints/metrics.rs [hunk 12]
```diff
@@ -32,7 +36,17 @@ const OPENMETRICS_CONTENT_TYPE: &str = "application/openmetrics-text; version=1.
 #[derive(Clone)]
 pub enum Msg {
     /// Increment the count of blocks received
-    IncBlockReceived,
+    BlockReceived,
+    // Increment the count of succeeded votes casts by verifier
```
maancham, line 40: casted instead of cast? Also, you could just name the message `VoteSucceeded` to match `VoteFailed` and its simpler

ampd/src/monitoring/endpoints/metrics.rs
```diff
@@ -188,31 +203,166 @@ async fn serve_metrics(
...
+
+        Self {
+            succeeded,
+            failed,
+            success_rate,
+        }
+    }
+
+    fn register(&self, registry: &mut Registry) {
+        registry.register(
```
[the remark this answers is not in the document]
yidan233: I updated the code so that vote failures are recorded for both` broadcasting errors` and `handler errors`, not just failures during broadcasting...

ampd/src/handlers/config.rs
```diff
@@ -11,6 +11,14 @@ use crate::evm::finalizer::Finalization;
 use crate::types::TMAddress;
 use crate::url::Url;
 
+#[derive(Clone, Debug)]
```
maancham, line 14: Can we have some documentation here?  like this?

```
/// Information about a blockchain handler for metrics and identification
#[derive(Clone, Debug, PartialEq, Eq)]
pub struct HandlerInfo {
    /// The name of the blockchain this handler operates on
    pub chain_name: String,
    /// The cosmwasm contract address serving as the verifier ID
    pub verifier_id: String,
    /// ,,.
    pub cast_votes: bool,
    /// ,,,,
    pub label: String,
}
```

ampd/src/event_processor.rs
```diff
@@ -574,4 +651,249 @@ mod tests {
...
+        assert!(result_with_timeout.unwrap().is_ok());
+
+        tokio::time::sleep(Duration::from_millis(100)).await;
+
+        let base_url = Url::parse(&format!("http://{}", bind_addr.unwrap())).unwrap();
+        let metrics_url = base_url.join("metrics").unwrap();
+        let response = reqwest::get(metrics_url).await.unwrap();
+        let metrics_text = response.text().await.unwrap();
+
+        assert!(metrics_text.contains("verifier_votes_cast_failed_total{verifier_id=\"verifier_id\",chain_name=\"chain_name\"} 1"));
```
maancham, line 889: same comment as the one on a test of `metrics.rs` file, is there a way to clean these assertions up?

ampd/src/event_processor.rs
```diff
@@ -181,6 +193,56 @@ fn should_task_stop(stream_status: StreamStatus, token: &CancellationToken) -> b
...
+        };
+        warn!(
+            err = %metric_error,
+             "failed to record {} vote metrics",
+             metric_type,
+        );
+    }
+}
+
+fn record_vote_failure(monitoring_client: &monitoring::Client, handler_info: HandlerInfo) {
```
maancham, line 230: so if this function handles `msgs` errors, why is it named `record_vote_failure`? Because we already have a `Msg::VoteFailed` on `record_vote_metrics` above. The naming of these two functions are a bit confusing.

ampd/src/monitoring/endpoints/metrics.rs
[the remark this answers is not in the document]
yidan233: `prometheus_client` automatically add` total `to the metrics name if its a counter

ampd/src/event_processor.rs
```diff
@@ -181,6 +205,43 @@ fn should_task_stop(stream_status: StreamStatus, token: &CancellationToken) -> b
...
+) {
+    if !handler_info.cast_votes {
+        return;
+    }
+
+    let metric_msg = match result {
+        VoteResult::Success => Msg::VoteSucceeded {
+            verifier_id: handler_info.verifier_id.clone(),
+            chain_name: handler_info.chain_name.clone(),
+        },
```
yidan233, line 221: I passed `handler_info` by reference and used `.clone() `on the fields because I was thinking the cloning would need to happen regardless, even if I passed `handler_info` by value, since `warn!` need access. please let me know if this breaks any conventions or if there's a preferred way to handle this.

ampd/src/event_processor.rs
```diff
@@ -181,6 +193,56 @@ fn should_task_stop(stream_status: StreamStatus, token: &CancellationToken) -> b
...
+            verifier_id: handler_info.verifier_id,
+            chain_name: handler_info.chain_name,
+        },
+        Err(_) => Msg::VoteFailed {
+            verifier_id: handler_info.verifier_id,
+            chain_name: handler_info.chain_name,
+        },
+    };
+
+    if let Err(metric_error) = monitoring_client.metrics().record_metric(metric_msg) {
```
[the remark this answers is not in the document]
yidan233: https://github.com/axelarnetwork/axelar-amplifier/blob/main/ampd/src/monitoring/endpoints/metrics.rs#L86   `record_metric` in the monitoring crate would only fail if the `processor` shuts down or the internal `channel` is full. And thats kind of hard to simulate in test..

Do you think it's worth adding a separate Client variant (`Client::Failing`) that always returns an error for testing purposes?

ampd/src/event_processor.rs
```diff
@@ -572,6 +664,253 @@ mod tests {
...
+
+        let base_url = Url::parse(&format!("http://{}", bind_addr.unwrap())).unwrap();
+        let metrics_url = base_url.join("metrics").unwrap();
+        let response = reqwest::get(metrics_url).await.unwrap();
+        let metrics_text = response.text().await.unwrap();
+
+        let sorted_metrics = sort_metrics_text(metrics_text);
+        goldie::assert!(sorted_metrics);
+
+        let _ = monitoring_client;
```
yidan233, line 730: when running the test case, it would sometimes fail with the following error from one of the `event processor` tests:
```

'event_processor::tests::should_not_record_vote_metrics_when_non_voting_handler_broadcast' panicked at ampd/src/event_processor.rs:722:56:
`called `Result::unwrap()` on an `Err` value: reqwest::Error { kind: Request, url: Url { scheme: "http", cannot_be_a_base: false, username: "", password: None, host: Some(Ipv4(127.0.0.1)), port: Some(98), path: "/metrics", query: None, fragment: None }, source: hyper::Error(Connect, ConnectError("tcp connect error", Os { code: 111, kind: ConnectionRefused, message: "Connection refused" })) }` 
```
but some times everything passes

I suspect this is due to the `monitoring_clien`t being dropped prematurely, causing the background task (`endpoint_handles`) to end before request is made, so server shut down earlier. 
 To fix this, I added `let _ = monitoring_client`; to ensure `monitoring_client `stays alive for the duration of the test. 
But I am not sure if the problem is solved by this...

ampd/src/event_processor.rs
```diff
@@ -541,37 +637,288 @@ mod tests {
...
         let metrics_text = response.text().await.unwrap();
         assert!(metrics_text.contains(&format!("blocks_received_total {}", num_block_ends)));
 
         cancel_token.cancel();
+        let _ = server_handle.await;
+    }
+
+    #[tokio::test(start_paused = true)]
+    async fn should_not_record_vote_metrics_when_non_voting_handler_broadcast() {
+        let events: Vec<Result<Event, event_processor::Error>> =
```
yidan233, line 676: sometimes the test cases fail on one of the `event_processor/server` test cases (when accessing the endpoint), but I am not sure why this happen...
`reqwest::Error { kind: Request, url: Url { scheme: "http", cannot_be_a_base: false, username: "", password: None, host: Some(Ipv4(127.0.0.1)), port: Some(107), path: "/metrics", query: None, fragment: None }, source: hyper::Error(Connect, ConnectError("tcp connect error", Os { code: 111, kind: ConnectionRefused, message: "Connection refused" })) }`

ampd/src/event_processor.rs
```diff
@@ -541,37 +637,288 @@ mod tests {
...
+
+        let response = reqwest::get(metrics_url).await.unwrap();
+        let metrics_text = response.text().await.unwrap();
+
+        let sorted_metrics = sort_metrics_text(metrics_text);
+        goldie::assert!(sorted_metrics);
+
+        cancel_token.cancel();
+        let _ = server_handle.await;
+    }
```
cgorenflo, line 908: this is testing stuff that is already tested in the monitoring module and that this shouldn't have to know about.
Instead, make a test server available in the monitoring module that gives you direct access to messages sent by the client. Then you can test the output of the handler directly without having to go through all the monitoring setup that the handler shouldn't be concerned with

cgorenflo (changes requested):

PR #934, opened 2025-07-01
feat(service-registry): add query to get service params override

eguajardo:
## Description

Adds a query to the service registry to be able to retrieve the parameters override. It might be useful so we can easily check if a given chain is using some override instead of the base parameters

https://axelarnetwork.atlassian.net/browse/AXE-9584

## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros


packages/service-registry-api/src/msg.rs
```diff
@@ -99,6 +99,12 @@ pub enum QueryMsg {
         chain_name: Option<ChainName>,
     },
 
+    #[returns(ServiceParamsOverride)]
```
cgorenflo, line 102: I'd prefer to return an Option instead of failing the query if there is no override

packages/service-registry-api/src/client.rs
```diff
@@ -38,6 +44,13 @@ impl From<QueryMsg> for Error {
                 service_name,
                 chain_name: _,
             } => Error::Service(service_name),
+            QueryMsg::ServiceParamsOverride {
+                service_name,
+                chain_name,
+            } => Error::ServiceParamsOverride {
+                service_name,
+                chain_name,
+            },
```
cgorenflo, line 53: I find this trait implementation kind of strange. Can we convert it into a regular function? It's so odd to see `msg.into()` to create an error

packages/service-registry-api/src/client.rs
```diff
@@ -17,16 +17,22 @@ pub enum Error {
...
         service_name: String,
         verifier: String,
     },
 }
 
-impl From<QueryMsg> for Error {
-    fn from(value: QueryMsg) -> Self {
-        match value {
+impl Error {
+    pub(crate) fn from_msg(msg: QueryMsg) -> Self {
```
cgorenflo, line 34:
```suggestion
    pub(crate) fn for_msg(msg: QueryMsg) -> Self {
```

cgorenflo (approved):

packages/service-registry-api/src/client.rs
cgorenflo: from still sounds a bit odd to me, but if you prefer it I'm ok with it

PR #935, opened 2025-07-02
chore: replace `cargo-run-script` for `just`

eguajardo:
## Description
`cargo-run-script` doesn't  work when trying to pass arguments to the script, also it hasn't been maintained in a while. Switched to [just](https://crates.io/crates/just) crate, which is more widely used and better maintained.

Also added some additional useful scripts to `just` and removed broken scripts

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros


justfile
```diff
@@ -0,0 +1,26 @@
...
+
+# Run project checks.
+check:
+    cargo clippy --all-targets -- -D warnings -A deprecated
+    cargo +nightly fmt --all --check
+    cargo sort --workspace --check --check-format
+
+# Run migration remover for a specific contract.
+remove-migration contract:
+    cargo run --bin migration-remover -- -c {{contract}}
```
milapsheth, line 26: missing EOF

justfile
```diff
@@ -0,0 +1,26 @@
...
+    docker run --rm -v "$(pwd)":/code \
+      --mount type=volume,source="$(basename "$(pwd)")_cache",target=/code/target \
+      --mount type=volume,source=registry_cache,target=/usr/local/cargo/registry \
+      cosmwasm/optimizer-arm64:0.16.1
+
+# Run project checks.
+check:
+    cargo clippy --all-targets -- -D warnings -A deprecated
+    cargo +nightly fmt --all --check
+    cargo sort --workspace --check --check-format
```
milapsheth, line 22: can we add cargo aliases for each of these so they can be reused in both the workflow and with just instead of duplicating?

milapsheth (approved):

PR #940, opened 2025-07-07
fix(packages): ensure permissions can accept empty direct attribute

SDavidson1177:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros

## Notes

There was a bug in the previous ensure permissions implementation where `direct` needed to be present in the attribute. This is undesirable, as not all contract will allow another proxy contract to execute message on them. So, they should be allowed to omit the `direct` expression if needed.


contracts/multisig/src/contract.rs
```diff
@@ -56,18 +57,15 @@ pub fn instantiate(
     Ok(Response::default())
 }
 
+#[ensure_permissions(direct(authorized = can_start_signing_session(&info.sender)))]
```
SDavidson1177, line 60: I added the macro here to improve code coverage, and because we may allow the coordinator to execute `authorize_callers` in the future

packages/msgs-derive/src/lib.rs
```diff
@@ -568,50 +568,64 @@ impl Parse for AllPermissions {
...
+                    storage: &dyn cosmwasm_std::Storage,
+                    contract_addr: Addr,
+                ) -> error_stack::Result<String, axelar_wasm_std::permission_control::Error>
+                    {
+                // This is only called when a relay message is executed. Since no proxy contract has
+                // permission to execute a message, this will always be an error.
+                Err(error_stack::report!(axelar_wasm_std::permission_control::Error::Unauthorized))
+            }
+        })
+    }
```
cgorenflo, line 628: try to nest as little as possible, and when you have large branch cases like this it's preferable to extract functions from them so it's more readable and you don't have to keep as much context in mind

PR #941, opened 2025-07-07
chore(lints): integrate axelar-lints

andrew-interop:
## Description

- Integrating lints from [`axelar-lints`](https://github.com/axelarnetwork/axelar-lints), ensuring our code meets the standard.

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros


contracts/axelarnet-gateway/src/clients/external.rs
```diff
@@ -13,6 +13,7 @@ pub struct AxelarExecutableMsg {
 
 /// By convention, amplifier-compatible contracts must expose this `Execute` variant.
 /// Due to identical json serialization, we can imitate it here so the gateway can call it.
+#[cfg_attr(dylint_lib = "amplifier_lints", allow(ensure_msg_has_permissions))]
```
cgorenflo, line 16: call the lint something like `msg_without_explicit_permissions`, so it reads better with the allow

integration-tests/src/lib.rs
```diff
@@ -1,3 +1,10 @@
+#![cfg_attr(dylint_lib = "amplifier_lints", allow(ensure_msg_has_permissions))]
+#![cfg_attr(
+    dylint_lib = "amplifier_lints",
+    allow(restrict_cosmwasm_addr_in_msg_struct)
```
cgorenflo, line 4:
```suggestion
    allow(cosmwasm_addr_in_msg_struct)
```
same reason as for the other lint

packages/axelar-core-std/src/nexus/execute.rs
```diff
@@ -40,8 +40,10 @@ fn parse_message_id(message_id: &str) -> Result<(nonempty::Vec<u8>, u64), Error>
 impl From<router_api::Message> for Message {
     fn from(msg: router_api::Message) -> Self {
         // fallback to using all 0's as the tx ID if it's not in the expected format
-        let (source_tx_id, source_tx_index) =
-            parse_message_id(&msg.cc_id.message_id).unwrap_or((vec![0; 32].try_into().unwrap(), 0));
+        let (source_tx_id, source_tx_index) = parse_message_id(&msg.cc_id.message_id).unwrap_or((
+            vec![0; 32].try_into().expect("conversion should not fail"),
```
cgorenflo, line 44: extract the zeroed ID into a variable, it's not very readable anymore

packages/signature-verifier-api/Cargo.toml
```diff
@@ -6,9 +6,11 @@ license = "MIT OR Apache-2.0"
 edition = { workspace = true }
 
 [dependencies]
+axelar-wasm-std = { workspace = true, features = ["derive"] }
```
cgorenflo, line 9: why is this necessary?

packages/signature-verifier-api/Cargo.toml
```diff
@@ -6,9 +6,11 @@ license = "MIT OR Apache-2.0"
...
 cosmwasm-schema = { workspace = true }
 cosmwasm-std = { workspace = true }
 error-stack = { workspace = true }
+msgs-derive = { workspace = true }
```
cgorenflo, line 13: in a followup PR, could you actually move this macro into axelar-wasm-std-derive? I don't think we need to have a separate package for this

contracts/axelarnet-gateway/src/clients/external.rs
andrew-interop: https://github.com/axelarnetwork/axelar-lints/pull/7

packages/signature-verifier-api/Cargo.toml
andrew-interop: Without it, `packages/signature-verifier-api/src/msg.rs` won't pass `cargo check`, since the permissions macro seems to need `axelar-wasm-std`.

integration-tests/src/lib.rs
```diff
@@ -1,3 +1,16 @@
...
+    dylint_lib = "amplifier_lints",
+    allow(execute_without_explicit_permissions)
+)]
+#![cfg_attr(dylint_lib = "amplifier_lints", allow(get_fn))]
+#![cfg_attr(
+    dylint_lib = "amplifier_lints",
+    allow(msg_without_explicit_permissions)
+)]
+#![cfg_attr(dylint_lib = "amplifier_lints", allow(ref_opt_type))]
+#![cfg_attr(dylint_lib = "amplifier_lints", allow(unwraps_outside_tests))]
```
cgorenflo, line 12: Would this work?
```suggestion
#![cfg_attr(dylint_lib = "amplifier_lints", allow(cosmwasm_addr_in_msg_struct, execute_without_explicit_permissions, get_fn, msg_without_explicit_permissions, ref_opt_type, unwraps_outside_tests))]
```
Are all of these even necessary? We want to allow unwrap because it makes tests more concise, but I don't think that we need to allow any of the others

packages/axelar-core-std/src/nexus/execute.rs
```diff
@@ -40,8 +40,9 @@ fn parse_message_id(message_id: &str) -> Result<(nonempty::Vec<u8>, u64), Error>
 impl From<router_api::Message> for Message {
     fn from(msg: router_api::Message) -> Self {
         // fallback to using all 0's as the tx ID if it's not in the expected format
+        let zeroed_tx_id = vec![0; 32].try_into().expect("conversion should not fail");
```
cgorenflo, line 43:
```suggestion
        let zeroed_message_id = vec![0; 32].try_into().expect("conversion should not fail");
```
message_id != tx_id (at least not all the time)

cgorenflo (approved):

PR #948, opened 2025-07-08
chore: add semantic version validation to release action

cgorenflo:
## Summary
- Add new validation step to ensure semantic version matches cargo release expectations
- Extract and parse version from cargo release output using sed regex with explanatory comments
- Validate version format supports semver with optional pre-release suffixes (alpha, beta, rc)
- Compare semantic version analysis with cargo release output to catch mismatches
- Provide detailed error messages for version mismatches and parsing failures

## Test plan
- [x] Verified regex works with different release types (patch, minor, major)
- [x] Verified regex works with pre-release versions (alpha, beta, rc)
- [x] Tested parsing with actual cargo release dry runs
- [x] Confirmed compatibility across macOS and Linux
- [x] Tested error scenarios (parsing failures, version mismatches)


chipshort (commented):
The script itself looks good and works on my machine, but wouldn't it be easier to directly provide cargo release with the version number determined by the semantic-version step? Like this:
`cargo release -p ${{ inputs.binary-to-release }}  ${{ steps.semantic-version.outputs.version }}` instead of using `version-type`?
That way there is no possibility for disagreement between the two.

PR #951, opened 2025-07-09
feat(ampd-sdk): add client connection pool

maancham:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros

### DIFF
--- a/ampd/src/broadcaster_v2/confirmer.rs
+++ b/ampd/src/broadcaster_v2/confirmer.rs
@@ -111,7 +111,11 @@ where
     }
 }
 
-async fn confirm_tx<T>(client: &T, tx_hash: String, retry_policy: RetryPolicy) -> Result<TxResponse>
+pub async fn confirm_tx<T>(
+    client: &T,
+    tx_hash: String,
+    retry_policy: RetryPolicy,
+) -> Result<TxResponse>
 where
     T: cosmos::CosmosClient + Clone,
 {
--- a/ampd/src/broadcaster_v2/mod.rs
+++ b/ampd/src/broadcaster_v2/mod.rs
@@ -30,7 +30,7 @@ mod tx;
 
 pub use broadcaster::Broadcaster;
 pub use config::Config;
-pub use confirmer::TxConfirmer;
+pub use confirmer::{confirm_tx, TxConfirmer};
 #[cfg(test)]
 pub use msg_queue::QueueMsg;
 pub use msg_queue::{MsgQueue, MsgQueueClient};
--- a/ampd/src/commands/bond_verifier.rs
+++ b/ampd/src/commands/bond_verifier.rs
@@ -7,24 +7,33 @@ use report::ResultCompatExt;
 use service_registry_api::msg::ExecuteMsg;
 use valuable::Valuable;
 
-use crate::commands::{broadcast_tx, verifier_pub_key};
+use crate::commands::{broadcast_tx, verifier_pub_key, BroadcastArgs};
 use crate::config::Config;
 use crate::{Error, PREFIX};
 
 #[derive(clap::Args, Debug, Valuable)]
 pub struct Args {
-    pub service_name: nonempty::String,
-    pub amount: u128,
-    pub denom: String,
+    service_name: nonempty::String,
+    amount: u128,
+    denom: String,
+    #[clap(flatten)]
+    broadcast: BroadcastArgs,
 }
 
 pub async fn run(config: Config, args: Args) -> Result<Option<String>, Error> {
-    let coin = Coin::new(args.amount, args.denom.as_str()).change_context(Error::InvalidInput)?;
+    let Args {
+        denom,
+        amount,
+        service_name,
+        broadcast,
+    } = args;
+
+    let coin = Coin::new(amount, denom.as_str()).change_context(Error::InvalidInput)?;
 
     let pub_key = verifier_pub_key(config.tofnd_config.clone()).await?;
 
     let msg = serde_json::to_vec(&ExecuteMsg::BondVerifier {
-        service_name: args.service_name.into(),
+        service_name: service_name.into(),
     })
     .expect("bond verifier msg should serialize");
 
@@ -37,7 +46,7 @@ pub async fn run(config: Config, args: Args) -> Result<Option<String>, Error> {
     .into_any()
     .expect("failed to serialize proto message");
 
-    let tx_hash = broadcast_tx(config, tx, pub_key).await?;
+    let tx_hash = broadcast_tx(config, tx, pub_key, broadcast.skip_confirmation).await?;
 
     Ok(Some(format!(
         "successfully broadcast bond verifier transaction, tx hash: {}",
--- a/ampd/src/commands/claim_stake.rs
+++ b/ampd/src/commands/claim_stake.rs
@@ -6,20 +6,27 @@ use report::ResultCompatExt;
 use service_registry_api::msg::ExecuteMsg;
 use valuable::Valuable;
 
-use crate::commands::{broadcast_tx, verifier_pub_key};
+use crate::commands::{broadcast_tx, verifier_pub_key, BroadcastArgs};
 use crate::config::Config;
 use crate::{Error, PREFIX};
 
 #[derive(clap::Args, Debug, Valuable)]
 pub struct Args {
-    pub service_name: nonempty::String,
+    service_name: nonempty::String,
+    #[clap(flatten)]
+    broadcast: BroadcastArgs,
 }
 
 pub async fn run(config: Config, args: Args) -> Result<Option<String>, Error> {
+    let Args {
+        service_name,
+        broadcast,
+    } = args;
+
     let pub_key = verifier_pub_key(config.tofnd_config.clone()).await?;
 
     let msg = serde_json::to_vec(&ExecuteMsg::ClaimStake {
-        service_name: args.service_name.into(),
+        service_name: service_name.into(),
     })
     .expect("claim stake msg should be serializable");
 
@@ -32,7 +39,7 @@ pub async fn run(config: Config, args: Args) -> Result<Option<String>, Error> {
     .into_any()
     .expect("failed to serialize proto message");
 
-    let tx_hash = broadcast_tx(config, tx, pub_key).await?;
+    let tx_hash = broadcast_tx(config, tx, pub_key, broadcast.skip_confirmation).await?;
 
     Ok(Some(format!(
         "successfully broadcast claim stake transaction, tx hash: {}",
--- a/ampd/src/commands/deregister_chain_support.rs
+++ b/ampd/src/commands/deregister_chain_support.rs
@@ -7,22 +7,30 @@ use router_api::ChainName;
 use service_registry_api::msg::ExecuteMsg;
 use valuable::Valuable;
 
-use crate::commands::{broadcast_tx, verifier_pub_key};
+use crate::commands::{broadcast_tx, verifier_pub_key, BroadcastArgs};
 use crate::config::Config;
 use crate::{Error, PREFIX};
 
 #[derive(clap::Args, Debug, Valuable)]
 pub struct Args {
-    pub service_name: nonempty::String,
-    pub chains: Vec<ChainName>,
+    service_name: nonempty::String,
+    chains: Vec<ChainName>,
+    #[clap(flatten)]
+    broadcast: BroadcastArgs,
 }
 
 pub async fn run(config: Config, args: Args) -> Result<Option<String>, Error> {
+    let Args {
+        service_name,
+        chains,
+        broadcast,
+    } = args;
+
     let pub_key = verifier_pub_key(config.tofnd_config.clone()).await?;
 
     let msg = serde_json::to_vec(&ExecuteMsg::DeregisterChainSupport {
-        service_name: args.service_name.into(),
-        chains: args.chains,
+        service_name: service_name.into(),
+        chains,
     })
     .expect("deregister chain support msg should serialize");
 
@@ -35,7 +43,7 @@ pub async fn run(config: Config, args: Args) -> Result<Option<String>, Error> {
     .into_any()
     .expect("failed to serialize proto message");
 
-    let tx_hash = broadcast_tx(config, tx, pub_key).await?;
+    let tx_hash = broadcast_tx(config, tx, pub_key, broadcast.skip_confirmation).await?;
 
     Ok(Some(format!(
         "successfully broadcast deregister chain support transaction, tx hash: {}",
--- a/ampd/src/commands/mod.rs
+++ b/ampd/src/commands/mod.rs
@@ -1,13 +1,16 @@
 use std::pin::Pin;
 
 use clap::Subcommand;
+use cosmrs::proto::cosmos::base::abci::v1beta1::TxResponse;
 use cosmrs::proto::Any;
 use cosmrs::AccountId;
 use error_stack::{Result, ResultExt};
+use futures::TryFutureExt;
 use serde::{Deserialize, Serialize};
 use valuable::Valuable;
 
-use crate::config::{Config as AmpdConfig, Config};
+use crate::asyncutil::future::RetryPolicy;
+use crate::config::Config;
 use crate::tofnd::{Multisig, MultisigClient};
 use crate::types::{CosmosPublicKey, TMAddress};
 use crate::{broadcaster_v2, cosmos, tofnd, Error, PREFIX};
@@ -22,6 +25,13 @@ pub mod send_tokens;
 pub mod set_rewards_proxy;
 pub mod unbond_verifier;
 pub mod verifier_address;
+
+#[derive(clap::Args, Debug, Valuable)]
+pub struct BroadcastArgs {
+    /// Skip transaction confirmation
+    #[arg(long, short)]
+    skip_confirmation: bool,
+}
 
 #[derive(Debug, Subcommand, Valuable)]
 pub enum SubCommand {
@@ -86,31 +96,12 @@ async fn verifier_pub_key(config: tofnd::Config) -> Result<CosmosPublicKey, Erro
 }
 
 async fn broadcast_tx(
-    config: AmpdConfig,
+    config: Config,
     tx: Any,
     pub_key: CosmosPublicKey,
+    skip_confirmation: bool,
 ) -> Result<String, Error> {
-    let mut broadcaster = instantiate_broadcaster(config, pub_key).await?;
-
-    Ok(broadcaster
-        .broadcast(vec![tx].try_into().expect("must be non-empty"))
-        .await
-        .change_context(Error::Broadcaster)?
-        .txhash)
-}
-
-async fn instantiate_broadcaster(
-    config: Config,
-    pub_key: CosmosPublicKey,
-) -> Result<
-    broadcaster_v2::BroadcasterTask<
-        cosmos::CosmosGrpcClient,
-        Pin<Box<broadcaster_v2::MsgQueue>>,
-        MultisigClient,
-    >,
-    Error,
-> {
-    let AmpdConfig {
+    let Config {
         tm_grpc,
         tm_grpc_timeout,
         broadcast,
@@ -121,6 +112,35 @@ async fn instantiate_broadcaster(
         .await
         .change_context(Error::Connection)
         .attach_printable(tm_grpc.clone())?;
+
+    let mut broadcaster = instantiate_broadcaster(
+        broadcast.clone(),
+        tofnd_config,
+        cosmos_client.clone(),
+        pub_key,
+    )
+    .await?;
+
+    broadcaster
+        .broadcast(vec![tx].try_into().expect("must be non-empty"))
+        .map_err(|err| err.change_context(Error::Broadcaster))
+        .and_then(|res| handle_tx_result(broadcast, cosmos_client, res, skip_confirmation))
+        .await
+}
+
+async fn instantiate_broadcaster(
+    broadcaster_config: broadcaster_v2::Config,
+    tofnd_config: tofnd::Config,
+    cosmos_client: cosmos::CosmosGrpcClient,
+    pub_key: CosmosPublicKey,
+) -> Result<
+    broadcaster_v2::BroadcasterTask<
+        cosmos::CosmosGrpcClient,
+        Pin<Box<broadcaster_v2::MsgQueue>>,
+        MultisigClient,
+    >,
+    Error,
+> {
     let multisig_client = MultisigClient::new(
         tofnd_config.party_uid,
         tofnd_config.url.as_str(),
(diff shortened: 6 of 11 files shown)


packages/ampd-sdk/src/grpc/client.rs
```diff
@@ -55,103 +58,65 @@ pub trait Client {
...
-}
-
-#[derive(Debug, Clone, PartialEq)]
-pub struct BroadcastClientResponse {
-    pub tx_hash: String,
-    pub index: u64,
+pub struct ManagedGrpcClient {
+    connection_handle: ConnectionHandle,
+    blockchain_service_client: Option<BlockchainServiceClient<transport::Channel>>,
+    crypto_service_client: Option<CryptoServiceClient<transport::Channel>>,
```
cgorenflo, line 64: I don't think these need to be fields anymore

packages/ampd-sdk/src/grpc/utils.rs
```diff
@@ -0,0 +1,114 @@
...
+
+impl PartialEq for ConnectionState {
+    fn eq(&self, other: &Self) -> bool {
+        matches!(
+            (self, other),
+            (ConnectionState::Connected(_), ConnectionState::Connected(_))
+                | (ConnectionState::Disconnected, ConnectionState::Disconnected)
+                | (ConnectionState::Reconnecting, ConnectionState::Reconnecting)
+        )
+    }
```
cgorenflo, line 32: use std::mem::discriminant for comparison instead, then you don't have to write a brittle check that will become wrong if we add different variants

packages/ampd-sdk/src/grpc/connection_manager.rs
```diff
@@ -0,0 +1,216 @@
...
+
+        let max_attempts = self.retry_policy.max_attempts();
+        let mut attempts = 0u64;
+
+        while attempts < max_attempts {
+            attempts = attempts.saturating_add(1);
+
+            if let Some(delay) = self.retry_policy.delay() {
+                tokio::time::sleep(delay).await;
+            }
```
cgorenflo, line 113: why do you insist on reimplementing the abilities of the retry policy? 😆 
just use `with_retry`

packages/ampd-sdk/src/future/mod.rs
```diff
@@ -3,21 +3,21 @@ use std::time::Duration;
...
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Debug)]
 pub enum RetryPolicy {
     RepeatConstant { sleep: Duration, max_attempts: u64 },
     NoRetry,
 }
 
 impl RetryPolicy {
-    fn max_attempts(&self) -> u64 {
+    pub fn max_attempts(&self) -> u64 {
```
cgorenflo, line 13: not necessary

packages/ampd-sdk/src/future/mod.rs
```diff
@@ -3,21 +3,21 @@ use std::time::Duration;
...
         match self {
             RetryPolicy::RepeatConstant { max_attempts, .. } => *max_attempts,
             RetryPolicy::NoRetry => 1,
         }
     }
 
-    fn delay(&self) -> Option<Duration> {
+    pub fn delay(&self) -> Option<Duration> {
```
cgorenflo, line 20: not necessary

packages/ampd-sdk/src/grpc/utils.rs
cgorenflo, line 1: try not to create util modules if you can help it

packages/ampd-sdk/src/grpc/error.rs
cgorenflo, line 1: seems like different errors here belong into different submodules

packages/ampd-sdk/src/grpc/connection_manager.rs
```diff
@@ -0,0 +1,257 @@
...
+
+/// Connection manager that handles connection state and retries.
+///
+/// This is a multi-producer single consumer pattern:
+/// - The manager maintains the connection state via a watch channel
+/// - Multiple clients can subscribe to connection state changes
+/// - Clients communicate back to the manager via mpsc channel
+/// - The manager handles connection retries and failure recovery
+#[derive(Debug)]
+pub struct ConnectionManager {
```
cgorenflo, line 32: event though it's currently a pool of one, I'd prefer the name `ConnectionPool`. We already have too many handlers, processors, and managers. Let's be more specific when we can

packages/ampd-sdk/src/grpc/connection_manager.rs
```diff
@@ -0,0 +1,257 @@
...
+            }
+            Err(status) => {
+                let status_error = Err(status).into_report();
+                if self
+                    .connection_state
+                    .send(ConnectionState::Disconnected)
+                    .is_err()
+                {
+                    return status_error
+                        .attach_printable("connection failed and could not notify clients");
```
cgorenflo, line 102: don't use attach_printable for something that you could just as well do with change_context. In fact, prefer not to use attach_printable at all anymore in favour of #[instrument] if you need access to additional data.

packages/ampd-sdk/src/grpc/connection_manager.rs
```diff
@@ -0,0 +1,257 @@
...
+                };
+                status_error
+            }
```
cgorenflo, line 105: you're making the match more complicated than it has to be
```suggestion
        match endpoint.connect().await {
            Ok(channel) => {
                self.connection_state
                    .send(ConnectionState::Connected(channel))
            }
            Err(status) => {
               self
                    .connection_state
                    .send(ConnectionState::Disconnected)?;
                status
            }
```
You need to change err context where appropriate to get the same error types

packages/ampd-sdk/src/grpc/connection_manager.rs
```diff
@@ -0,0 +1,257 @@
...
+        }
+    }
+
+    async fn handle_connection_failure(&self) -> Result<(), Error> {
+        self.connection_state
+            .send(ConnectionState::Reconnecting)
+            .into_report()?;
```
cgorenflo, line 112: can you document what the purpose of the states is supposed to be? Why is there a difference between Disconnected and Reconnecting?

packages/ampd-sdk/src/grpc/connection_manager.rs
```diff
@@ -0,0 +1,257 @@
...
+
+        match with_retry(|| self.connect(), self.retry_policy).await {
+            Ok(()) => {
+                info!("successfully reconnected after failure");
+                Ok(())
+            }
+            Err(error_report) => {
+                error!(err = ?error_report, "failed to reconnect after max attempts");
```
cgorenflo, line 120: don't log the error if you return it later

packages/ampd-sdk/src/grpc/connection_manager.rs
```diff
@@ -0,0 +1,257 @@
...
+                let retry_error = Err(error_report);
+                if self
+                    .connection_state
+                    .send(ConnectionState::Disconnected)
+                    .is_err()
+                {
+                    return retry_error
+                        .attach_printable("connection failed and could not notify clients");
+                };
```
cgorenflo, line 129: I believe this can only fail if the receivers are either down or there is too much back pressure. This seems like something that should just be a warning log and then suppressed.

packages/ampd-sdk/src/grpc/connection_pool.rs
```diff
@@ -0,0 +1,306 @@
...
+        let (_, handle, _) = test_setup().await;
+
+        assert!(matches!(
+            *handle.connection_receiver.borrow(),
+            ConnectionState::Disconnected
+        ));
+    }
+
+    #[tokio::test]
+    async fn connection_pool_should_remain_diconnected_with_wrong_url() {
```
eguajardo, line 220:
```suggestion
    async fn connection_pool_should_remain_disconnected_with_wrong_url() {
```

packages/ampd-sdk/src/grpc/client.rs
```diff
@@ -55,103 +57,65 @@ pub trait Client {
...
+        ),
+        Error,
+    > {
+        let connection_state = self.connection_handle.connection_receiver.borrow().clone();
+
+        match connection_state {
+            ConnectionState::Connected(channel) => {
+                let blockchain_service_client = BlockchainServiceClient::new(channel.clone());
+                let crypto_service_client = CryptoServiceClient::new(channel);
+                Ok((blockchain_service_client, crypto_service_client))
```
eguajardo, line 84: maybe extract this so you avoid code duplication below

cgorenflo: agreed, separate connecting from creating the clients. I'm not even sure that part needs to have a dedicated function, usually a call just needs one or the other, so prophylactically creating both seems like a bit of a waste

packages/ampd-sdk/src/grpc/client.rs
```diff
@@ -55,103 +57,65 @@ pub trait Client {
...
+                match connection_state {
+                    ConnectionState::Connected(channel) => {
+                        let blockchain_service_client =
+                            BlockchainServiceClient::new(channel.clone());
+                        let crypto_service_client = CryptoServiceClient::new(channel);
+                        Ok((blockchain_service_client, crypto_service_client))
+                    }
+                    _ => bail!(Error::from(AppError::ConnectionUnavailable)),
+                }
+            }
```
cgorenflo, line 103: I'm a bit worried about the client having to make all these decisions for how to deal with the connection. It seems to make it's own state a bit unpredictable. Maybe it's better to put this into the connection handle? That would justify its existence and encapsulate the interaction with the connection state. Then the client would just go "connection handle give me a connection", and all the back and forth with the pool would not be visible

packages/ampd-sdk/src/grpc/client.rs
```diff
@@ -55,103 +57,65 @@ pub trait Client {
...
         }
     }
 }
 
 #[async_trait]
-impl Client for GrpcClient {
+impl Client for ManagedGrpcClient {
     type Stream = Pin<Box<dyn Stream<Item = Result<Event, Error>> + Send>>;
 
     async fn subscribe(
```
cgorenflo, line 112: this will still need the stream indirection, but that can be done in a follow up

packages/ampd-sdk/src/grpc/client.rs
```diff
@@ -317,11 +292,15 @@ mod tests {
         }
     }
 
-    async fn setup_test_client(
+    async fn test_setup(
```
cgorenflo, line 295: I know we use this naming scheme throughout many parts of the codebase, but I'm not a big fan of it, because it tells you nothing about what's happening. Furthermore, it seems like at least most tests don't need the connection and abort handle, so the abstraction doesn't fit the need. Better to have a separate setup for the few tests that do need them

packages/ampd-sdk/src/grpc/connection_pool.rs
```diff
@@ -0,0 +1,306 @@
...
+pub struct ConnectionPool {
+    url: Url,
+    connection_state: watch::Sender<ConnectionState>,
+    message_receiver: mpsc::Receiver<ClientMessage>,
+    retry_policy: RetryPolicy,
+}
+
+impl ConnectionPool {
+    pub fn new(url: &str) -> Result<(Self, ConnectionHandle), Error> {
+        let url = Url::new_sensitive(url).change_context(Error::InvalidUrl)?;
```
cgorenflo, line 98: we should get the url as this type already

cgorenflo: then you also don't need to return a result

packages/ampd-sdk/src/grpc/connection_pool.rs
```diff
@@ -0,0 +1,306 @@
...
+                ClientMessage::ConnectionFailed(details) => {
+                    warn!("client reported connection failure: {}", details);
+                    if let Err(e) = self.handle_connection_failure().await {
+                        warn!(err = ?e, "reconnection attempt failed, will retry on next client request");
+                    }
+                }
+            }
+        }
+
+        Ok(())
```
cgorenflo, line 134: how can this be killed? This looks like it's potentially stuck waiting for messages

packages/ampd-sdk/src/grpc/connection_pool.rs
```diff
@@ -0,0 +1,306 @@
...
+                Ok(())
+            }
+            Err(error_report) => {
+                self.notify_clients(ConnectionState::Disconnected, true)?;
+                Err(error_report)
+            }
+        }
+    }
+
+    fn notify_clients(&self, state: ConnectionState, log_on_error: bool) -> Result<(), Error> {
```
cgorenflo, line 170: what's the reason for this bool flag?

packages/ampd-sdk/src/grpc/error.rs
```diff
@@ -41,12 +42,15 @@ pub enum AppError {
...
+    #[error("connection to the manager is unavailable")]
+    ConnectionUnavailable,
 }
 
 #[derive(Error, Debug)]
 pub enum Error {
-    #[error("failed to connect to the grpc endpoint")]
-    GrpcConnection(#[from] tonic::transport::Error),
+    #[error("failed to reconnect to the pool")]
+    ConnectionStateReceiveFailed(#[from] watch::error::RecvError),
```
cgorenflo, line 53: is there no relevant information in the RecvError? This will discard the inner error

eguajardo (commented):
would be good to document the architecture with some diagrams. It doesn't have to be in this PR though

packages/ampd-sdk/src/grpc/client.rs
```diff
@@ -55,103 +58,65 @@ pub trait Client {
...
-}
-
-#[derive(Debug, Clone, PartialEq)]
-pub struct ContractsAddresses {
-    pub voting_verifier: AccountId,
-    pub multisig_prover: AccountId,
-    pub service_registry: AccountId,
-    pub rewards: AccountId,
+pub struct ManagedGrpcClient {
+    connection_handle: ConnectionHandle,
```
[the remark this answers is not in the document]
maancham: It makes the code redundant in my opinion to have two blocks of 
```
connection_receiver: watch::Receiver<ConnectionState>,
message_sender: mpsc::Sender<ClientMessage>,
```
in sdk. Also, there is a good chunk of documentation for `ConnectionHandle` that will quickly show anyone how this structure works.

packages/ampd-sdk/src/grpc/connection_pool.rs
maancham: passing in CancellationToken for graceful shutdown.

packages/ampd-sdk/src/grpc/client_utils.rs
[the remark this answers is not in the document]
maancham: since parse address is being used in both `client_utils` and `client`, I'd much rather keep it here than to move it to client (the only other viable option). I can change the filename to types

packages/ampd-sdk/src/grpc/connection_pool.rs
```diff
@@ -0,0 +1,306 @@
...
+            ConnectionState::Disconnected
+        ));
+    }
+
+    #[tokio::test]
+    async fn connection_pool_should_remain_diconnected_with_wrong_url() {
+        let (pool, handle) = ConnectionPool::new("https://localhost:1").unwrap();
+        let pool_task = tokio::spawn(async move { pool.run().await });
+
+        tokio::time::sleep(Duration::from_millis(200)).await;
```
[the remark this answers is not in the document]
maancham: I am sending that on client tests

packages/ampd-sdk/src/grpc/connection_pool.rs
maancham: removed to log only

packages/ampd-sdk/src/grpc/connection_pool.rs
```diff
@@ -0,0 +1,306 @@
...
+    fn notify_clients(&self, state: ConnectionState, log_on_error: bool) -> Result<(), Error> {
+        match self.connection_state.send(state) {
+            Ok(()) => Ok(()),
+            Err(send_error) => {
+                if log_on_error {
+                    warn!(err = ?send_error, "failed to notify clients of state change");
+                    Ok(())
+                } else {
+                    Err(send_error).into_report()
+                }
```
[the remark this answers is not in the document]
maancham: changed to log only

packages/ampd-sdk/src/grpc/connection_pool.rs
```diff
@@ -74,8 +85,60 @@ impl PartialEq for ConnectionState {
...
+                let connection_state = self.connection_receiver.borrow_and_update().clone();
+
+                match connection_state {
+                    ConnectionState::Connected(channel) => {
+                        self.has_sent_disconnected_message = false;
+                        return Ok(channel);
+                    }
+                    ConnectionState::Disconnected => {
+                        // Only send ConnectionFailed message if we haven't already sent one
+                        if !self.has_sent_disconnected_message {
```
eguajardo, line 113: this seems a bit hacky, maybe the message channel could be a watch channel instead and in the pool just process the failures when there was a change in the state?

packages/ampd-sdk/src/grpc/connection_pool.rs
```diff
@@ -74,8 +85,60 @@ impl PartialEq for ConnectionState {
...
+
+impl ConnectionHandle {
+    /// This method encapsulates all the connection state management logic:
+    /// - If connected, returns the channel immediately
+    /// - If disconnected, inform clients
+    /// - If reconnecting, waits for the connection to complete and returns the channel
+    ///
+    /// Clients should use this method instead of manually handling connection states.
+    pub async fn get_connected_channel(&mut self) -> Result<transport::Channel, Error> {
+        timeout(DEFAULT_SERVER_TIMEOUT, async {
```
eguajardo, line 102: in which cases this would timeout?

packages/ampd-sdk/src/grpc/connection_pool.rs
```diff
@@ -145,39 +216,31 @@ impl ConnectionPool {
 
         match endpoint.connect().await {
             Ok(channel) => {
-                self.notify_clients(ConnectionState::Connected(channel), false)?;
+                self.notify_clients(ConnectionState::Connected(channel));
```
eguajardo, line 219: it's a bit confusing to have success notifications here and failure notifications outside, could you move this one to the `handle_connection_failure` function?. Also, when possible try to use `map` and `inspect` and their `_err` equivalent for these kind of cases instead of matching on results

packages/ampd-sdk/src/grpc/connection_pool.rs
```diff
@@ -295,12 +361,41 @@ mod tests {
...
+            *handle.connection_receiver.borrow(),
+            ConnectionState::Disconnected
         ));
+
+        task1.abort();
+        task2.abort();
+        task3.abort();
+
+        pool_task.abort();
     }
```
eguajardo, line 400: is this actually testing the message is not sent multiple times? it seems it's testing the clients are getting `Disconnected`

packages/ampd-sdk/src/grpc/connection_pool.rs
maancham: when server is out

packages/ampd-sdk/src/grpc/connection_pool.rs
```diff
@@ -74,8 +85,60 @@ impl PartialEq for ConnectionState {
...
+                        if !self.has_sent_disconnected_message {
+                            self.send_client_message(ClientMessage::ConnectionFailed(
+                                "client requested connection while disconnected".to_string(),
+                            ))
+                            .await?;
+                            self.has_sent_disconnected_message = true;
+                        }
+
+                        let mut receiver = self.connection_receiver.clone();
+                        receiver.changed().await.into_report()?;
```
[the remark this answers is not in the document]
maancham: marking as resolved based on a conversation I had. The `run` function of pool will take care of it.

packages/ampd-sdk/src/grpc/connection_pool.rs
maancham: marking as resolved based on previous conversation.

packages/ampd-sdk/src/grpc/connection_pool.rs
maancham: changed the name

packages/ampd-sdk/src/grpc/connection_pool.rs
```diff
@@ -0,0 +1,394 @@
...
+                    }
+                    ConnectionState::Reconnecting => {
+                        let mut receiver = self.connection_receiver.clone();
+                        receiver.changed().await.into_report()?;
+                    }
+                }
+            }
```
cgorenflo, line 122: can we simplify to this? (pseudo code)
```
loop{
    let state = receiver.borrow_and_update();
    match state{
       ConnectionState::Connected(channel) => return Connected<channel>,
       ConnectionState::Shutdown/Disconnected => return Shutdown
       _ =>     receiver.changed().await.into_report()?;
    }    
}
```
We will need another enum for the return from the handle to tell the client to disconnect when the pool shuts down (I don't want to shut down with errors)

packages/ampd-sdk/src/grpc/connection_pool.rs
```diff
@@ -0,0 +1,394 @@
...
+        match endpoint.connect().await {
+            Ok(channel) => {
+                self.notify_clients(ConnectionState::Connected(channel));
+                Ok(())
+            }
+            Err(status) => Err(status).into_report(),
+        }
+    }
+
+    async fn handle_connection_failure(&mut self) -> Result<(), Error> {
```
cgorenflo, line 222: simplify your naming
```suggestion
    async fn reconnect(&mut self) -> Result<(), Error> {
```

packages/ampd-sdk/src/grpc/connection_pool.rs
```diff
@@ -0,0 +1,394 @@
...
+            tokio::select! {
+                message = self.message_receiver.recv() => {
+                    match message {
+                        Some(ClientMessage::ConnectionFailed(details)) => {
+                            warn!("client reported connection failure: {}", details);
+                            if !self.is_reconnecting {
+                                if let Err(e) = self.handle_connection_failure().await {
+                                    warn!(err = ?e, "reconnection attempt failed, will retry on next client request");
+                                }
+                            }
```
cgorenflo, line 189: there is an issue here: let's say we have 10 clients. All of them fail to make a request call and notify the pool. The first message will reconnect and wait until that's done. After that we have a newly established connection, but still 9 failure notifications in the channel. That means the pool will kill and reconnect the connection another 9 times.
It needs to be something like this:
- receive failure notification
- set status to reconnecting
- drain failure notification channel
- actually reconnect while continuing to drain (`select!`)
- when connection is established change the state and stop draining

packages/ampd-sdk/src/grpc/connection_pool.rs
```diff
@@ -0,0 +1,394 @@
...
+/// This enum is used by the connection pool to communicate the connection status to clients.
+#[derive(Debug, Clone)]
+pub enum ConnectionState {
+    /// The connection is established and ready for service calls.
+    /// Contains the active gRPC transport channel.
+    Connected(transport::Channel),
+
+    /// The connection is not available and no reconnection attempt is in progress.
+    /// This is the initial state when the connection pool starts, or the final state after all reconnection attempts have failed.  
+    Disconnected,
```
cgorenflo, line 65: Why even make this state ambiguous? Just split it into unconnected and shut down (or keep the disconnected name for this one). Then it's clear that the handle should request a new connection for one, but not the other

packages/ampd-sdk/src/grpc/connection_pool.rs
```diff
@@ -0,0 +1,394 @@
...
+                }
+            }
+        })
+        .await
+        .into_report()?
+    }
+
+    pub async fn send_client_message(&self, message: ClientMessage) -> Result<(), Error> {
+        self.message_sender.send(message).await.into_report()
+    }
```
cgorenflo, line 130:
```suggestion
    pub async fn request_reconnect(&self) -> Result<(), Error> {
        self.message_sender.send(ConnectionFailed).await.into_report()
    }
```

packages/ampd-sdk/src/grpc/connection_pool.rs
```diff
@@ -0,0 +1,394 @@
...
+                        return Ok(channel);
+                    }
+                    ConnectionState::Disconnected => {
+                        self.send_client_message(ClientMessage::ConnectionFailed(
+                            "client requested connection while disconnected".to_string(),
+                        ))
+                        .await?;
+
+                        let mut receiver = self.connection_receiver.clone();
+                        receiver.changed().await.into_report()?;
```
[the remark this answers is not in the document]
maancham: removed.

packages/ampd-sdk/src/grpc/connection_pool.rs
```diff
@@ -0,0 +1,394 @@
...
+    /// This method encapsulates all the connection state management logic:
+    /// - If connected, returns the channel immediately
+    /// - If disconnected, inform clients
+    /// - If reconnecting, waits for the connection to complete and returns the channel
+    ///
+    /// Clients should use this method instead of manually handling connection states.
+    pub async fn get_connected_channel(&mut self) -> Result<transport::Channel, Error> {
+        timeout(DEFAULT_SERVER_TIMEOUT, async {
+            loop {
+                let connection_state = self.connection_receiver.borrow_and_update().clone();
```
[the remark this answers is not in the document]
maancham: If I don't clone, I will get the 
```
pub struct Ref<'a, T> {
    inner: RwLockReadGuard<'a, T>,
    has_changed: bool,
}
```
then, I can not simply dereference it because `ConnectionState` does not implement Copy trait, and it can not be added to it since channel does not implement Copy.

PR #952, opened 2025-07-10
refactor(voting-verifier): implement into-event macro for voting verifier events

kyanokashi:
## Description
Implemented IntoEvent macro for voting verifier events. 

This isn't a breaking change as event signatures remain the same. 

Needed to get rid of `PollMetadata` as it modifies the event signature. This is due to us not being able to manually flatten the `poll_metadata` field which was what we were doing previously. This would cause increased overhead for axelarons as they would need to implement the event with two different types in order to provide support for both version

```
// Expected event signature
{
messages
...poll_metadata
}

// with IntoEvent
{
messages
poll_metadata
}

```

## Todos

- [x] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros

## Notes

---

> [!NOTE]
> Introduces a new IntoEvent-driven `Event` enum in `voting-verifier`, deprecating `PollMetadata`/legacy event shapes, while preserving backward compatibility via `legacy_events`.
> 
> - Add `events::Event` (e.g., `Instantiated`, `MessagesPollStarted`, `VerifierSetPollStarted`, `QuorumReached`, `Voted`, `PollEnded`) and emit these in contract execute/instantiate paths
> - Add `legacy_events` module mirroring old event signatures for compatibility; tests compare new vs legacy parsing
> - Update all AMPd handlers (EVM, Sui, Solana, Starknet, Stacks, Stellar, MultiversX) and ampd-handlers to consume new `wasm-*` event types and flattened fields
> - Adjust tests and add golden files for new events; minor crate exports (`multisig::Event`) and `serde` dep for `voting-verifier`
> - No functional behavior change to event semantics; event keys remain stable
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4fcac034714a50ff6b51551d4e4dc62475175d11. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/contracts/voting-verifier/src/lib.rs
+++ b/contracts/voting-verifier/src/lib.rs
@@ -5,7 +5,6 @@ pub mod events;
 pub use events::Event;
 
 pub mod legacy_events;
-pub use events::Event as LegacyEvent;
 
 pub mod contract;
 pub mod error;


cgorenflo (approved):

contracts/voting-verifier/src/testdata/events_should_not_change.golden
```diff
@@ -1,164 +0,0 @@
...
-        "value": "\"hex_tx_hash_and_event_index\""
-      },
-      {
-        "key": "address_format",
-        "value": "\"eip55\""
-      }
-    ],
-    "type": "instantiated"
-  },
-  "event_messages_poll_started": {
```
cjcobb23, line 47: why do we need to remove this? This is concerning because the events should not change. Otherwise we risk breaking external consumers, particularly ampd. For ampd, the messages_poll_started and verifier_set_poll_started should not change.

cgorenflo: I believe it's just split into multiple different files now

ampd/src/handlers/evm_verify_msg.rs
```diff
@@ -371,6 +370,179 @@ mod tests {
         goldie::assert_debug!(event);
     }
 
+    #[test]
+    fn test_voting_verifier_event_vs_legacy_event_parsing() {
```
kyanokashi, line 374: @cgorenflo @cjcobb23 

Upon further inspection of the code I had a feeling that ampd would already support parsing both event versions into the same `PollStartedEvent` value.

Can you guys please sanity check this and ensure the test properly tests for this behavior?

PR #954, opened 2025-07-11
fix(ampd): prevent port collisions in monitoring server tests

cgorenflo:
## Summary
- Replace hardcoded ports with OS-assigned ports using `TcpListener::bind` with port 0
- Add `TcpConnector` abstraction to support both `SocketAddr` and existing `TcpListener`
- Update tests to use OS-selected ports to prevent conflicts during parallel test execution

## Test plan
- [x] All existing monitoring server tests pass
- [x] Tests can run in parallel without port conflicts
- [x] Both disabled and enabled server configurations work correctly
- [x] Documentation is comprehensive and accurate

### DIFF
--- a/ampd/src/monitoring/server.rs
+++ b/ampd/src/monitoring/server.rs
@@ -199,7 +199,6 @@ impl Server {
     }
 }
 
-
 /// TCP connection abstraction for the monitoring server
 ///
 /// This enum allows the server to be configured with either a socket address
@@ -269,7 +268,6 @@ impl From<TcpListener> for TcpConnector {
         TcpConnector::Listener(listener)
     }
 }
-
 
 /// HTTP server implementation for monitoring endpoints
 ///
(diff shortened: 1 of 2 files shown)


ampd/src/monitoring/server.rs
```diff
@@ -188,25 +194,95 @@ impl Server {
...
+    /// Establishes a TCP connection, returning a bound listener
+    ///
+    /// For `Address` variants, this will bind to the specified socket address.
+    /// For `Listener` variants, this returns the existing listener as-is.
+    ///
+    /// # Errors
+    ///
+    /// Returns an `io::Error` if binding to the address fails (only applicable
+    /// for `Address` variants).
+    pub async fn connect(self) -> io::Result<TcpListener> {
```
maancham, line 228: can we have connect and bind_address be more consistent with their return types? I mean is there a downside to have:
`pub async fn connect(self) -> Result<TcpListener, Error>`?

maancham (approved):

PR #957, opened 2025-07-15
feat(minor-service-registry): limit active verifiers based on override

cjcobb23:
## Description

We limit the max number of verifiers based on the base service params, but we don't limit verifiers based on the service override param (it's just ignored). This has come up in several integrations though, where chains can only support a certain number of verifiers (XRPL, Stacks).

The solution here simply sorts the active verifiers by weight and then address, and then truncates to the max specified in the override. Alternative approaches, such as rejecting chain registration when the max is exceeded, come with issues, because all of the verifiers registered for a chain may not be authorized and/or bonded at a given time, so you could end up with a situation where the contract restricts too much or too little. And even if the contract checks bonded/authorization status at chain registration, those states can later change.


contracts/service-registry/src/contract.rs
```diff
@@ -3172,4 +3172,138 @@ mod test {
...
+            expected_addresses[0]
+        );
+        assert_eq!(
+            active_verifiers[1].verifier_info.address.to_string(),
+            expected_addresses[1]
+        );
+        assert_eq!(
+            active_verifiers[2].verifier_info.address.to_string(),
+            expected_addresses[2]
+        );
```
cgorenflo, line 3257: why is this a requirement?

cgorenflo: I think a reasonable requirement is that the set is stable, but I don't think we need to guarantee alphabetical order

contracts/service-registry/src/contract/query.rs
```diff
@@ -41,10 +41,21 @@ pub fn active_verifiers(
...
+            .into_iter()
+            .sorted_by(|a, b| {
+                b.weight
+                    .cmp(&a.weight)
+                    .then_with(|| a.verifier_info.address.cmp(&b.verifier_info.address))
+            })
+            .take(max_verifiers as usize)
+            .collect(),
+        _ => verifiers,
+    })
```
cgorenflo, line 58:
```suggestion
    Ok(verifiers.into_iter().take(max_verifiers.unwrap_or(usize::Max) as usize).collect())
```

contracts/service-registry/src/contract/query.rs
```diff
@@ -41,10 +45,64 @@ pub fn active_verifiers(
...
+    );
+    let hash_b = hash_address_with_seed(
+        &b.verifier_info.address.to_string(),
+        &block_height.to_string(),
+    );
+    hash_a.cmp(&hash_b)
+}
+
+fn hash_address_with_seed(address: &str, seed: &str) -> u64 {
+    let mut hasher = DefaultHasher::new();
```
cgorenflo, line 102: pass hasher in so we don't have to initialize it fresh for each comparison?

contracts/service-registry/src/contract/query.rs
```diff
@@ -41,10 +45,64 @@ pub fn active_verifiers(
...
+    };
+
+    // Now sort by weight descending, but for verifiers with the cutoff weight, break ties via pick_random
+    verifiers.sort_by(|a, b| {
+        b.weight.cmp(&a.weight).then_with(|| {
+            if a.weight == cutoff_weight {
+                pick_random(a, b, block_height)
+            } else {
+                Ordering::Equal
+            }
```
cgorenflo, line 80: are you aware that this is not a fair pick? It's just fuzzing the position in the initial sequence a bit

contracts/service-registry/src/contract/query.rs
```diff
@@ -41,10 +45,64 @@ pub fn active_verifiers(
...
+        })
+    });
```
cgorenflo, line 82: alternative suggestion:
after figuring out the cutoff, collect all items with weight< cutoff. Then filter for only weight == cutoff and do something like
```
let seeded_rng = rand::SeedableRng::seed_from_u64(block_height)
let (selected, _) = verifiers_with_cutoff_weight.partial_shuffle(seeded_rng, free_spot_count);
active_verifiers.extend_from_slice(selected)
```

cgorenflo: this way, you never have to sort and the shuffling should be faster than the hashing and comparison. This process is about fairness and not security (all verifiers are authorized), so relying on a seeded rng should be no concern

contracts/service-registry/src/contract/query.rs
cgorenflo: (if you want to stick with this approach)

contracts/service-registry/src/contract/query.rs
cjcobb23: How so? The algorithm basically sorts by weight and then within the sub slice of verifiers with the cutoff weight, assigns each a random number in the 256 bit range, and sorts by that. So the order of that sub slice should be random, no?

The pick_random function is a bit misnamed, it's not picking a random element each time it's called, but is deterministic based on block height and verifier address.

PR #962, opened 2025-07-18
refactor: use client package for all clients

chipshort:
## Description

Replaces all the custom client implementations with ones using the `client` package and unifies the naming and exports of those contract clients.

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros


packages/signature-verifier-api/src/client.rs
```diff
@@ -1,48 +1,61 @@
...
+            } => Error::VerifySignature {
+                signature,
+                message,
+                public_key,
+                signer_address,
+                session_id,
+            },
         }
     }
+}
```
cgorenflo, line 37: could we change this into something like `Error::for(msg)`? I'm not a big fan of using `From` for msg to error transformations, because conceptually it's confusing. The error just "happens" to print out the information from the message, but it's not really an equivalent object

PR #965, opened 2025-07-22
Metrics/system2

yidan233:
Tracked:
`Current CPU usage percentage of the ampd process`
`Current memory usage in bytes of the ampd process`
https://github.com/GuillaumeGomez/sysinfo
The metrics are collected using the `sysinfo` library, which provides cross-platform access to` process` and `system information`.

 platform limitation, missing permission, or other failure will default the system metric to `0`.

The sysinfo crate supports the following platforms:
https://docs.rs/crate/sysinfo/latest?
Linux
macOS
Windows
Android
iOS
FreeBSD
Raspberry Pi

On unsupported platforms or restricted permissions, process system metrics may not be available. metrics will fallback with values of 0.
(such as android or sandboxed macOS/iOS)
https://docs.rs/sysinfo/latest/sysinfo/struct.Process.html?utm_source=chatgpt.com

### DIFF
--- a/ampd/src/monitoring/endpoints/metrics.rs
+++ b/ampd/src/monitoring/endpoints/metrics.rs
@@ -1,5 +1,5 @@
-use std::sync::Arc;
-use std::sync::Mutex;
+use std::sync::atomic::AtomicU64;
+use std::sync::{Arc, Mutex};
 
 use axelar_wasm_std::voting::Vote;
 use axum::body::Body;
@@ -14,10 +14,8 @@ use prometheus_client::encoding::EncodeLabelSet;
 use prometheus_client::metrics::counter::Counter;
 use prometheus_client::metrics::family::Family;
 use prometheus_client::metrics::gauge::Gauge;
-use std::sync::atomic::AtomicU64;
-use sysinfo::{get_current_pid, Pid, ProcessesToUpdate, ProcessRefreshKind};
-use sysinfo::System;
 use prometheus_client::registry::Registry;
+use sysinfo::{get_current_pid, Pid, ProcessRefreshKind, ProcessesToUpdate, System};
 use thiserror::Error;
 use tokio::sync::mpsc;
 use tokio::task::JoinHandle;
@@ -44,7 +42,9 @@ pub enum Msg {
     VerificationVote {
         vote_status: Vote,
         chain_name: String,
-    },  
+    },
+    /// Update the system metrics
+    UpdateSystemMetrics { cpu_usage: f64, memory_usage: f64 },
 }
 
 /// Errors that can occur in metrics processing
@@ -113,8 +113,8 @@ impl Client {
 #[derive(Clone)]
 struct MetricsState {
     registry: Arc<Registry>,
+    client: Client,
     system_metrics_collector: SystemMetricsCollector,
-    metrics: Arc<Metrics>,
 }
 
 /// Creates a metrics endpoint with processor and client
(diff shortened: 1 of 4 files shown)


cgorenflo (commented):
You have implemented the sysinfo metrics analogous to the rest of the metrics, and I believe that the solution is correct. However, due to the setup with the System Object that needs to be passed around as an ArcMutex, and the way collection is handled, the solution is very complex and a bit convoluted.
We already have a form of concurrency management through the metrics client, so I believe it is simpler to move the System into the SystemInfoMetrics, which are also only instantiated once. Then the UpdateSystemMetrics doesn't take any values, and simply sends the signal to update to the metrics. The sysinfo metrics can then be directly updated inside `system_info.update_metrics()`. This way you get around using Arc<Mutex>, you don't need a collector at all and MetricsState only needs the registry and the client.
There is one additional problem that I will have to think more about: Both in your current solution and my change suggestion, the metrics are updated asynchronously, so most likely by the time `encode` is called, the sysinfo has not been fully updated yet. This means cpu and memory are probably always delayed by a few seconds. I'll have to think more about a potential solution to that issue

ampd/src/monitoring/endpoints/metrics.rs
```diff
@@ -300,14 +348,153 @@ impl VerificationVoteMetrics {
...
+            memory_usage: 0.0,
+        }
+    }
+}
+
+fn refresh_process_metrics(system: &mut System, process_id: Pid) {
+    system.refresh_processes_specifics(
+        ProcessesToUpdate::Some(&[process_id]),
+        true,
+        ProcessRefreshKind::nothing().with_cpu().with_memory(),
```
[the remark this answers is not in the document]
yidan233: I think `ProcessRefreshKind`  doesn't have `cpu()`

ampd/src/monitoring/endpoints/metrics.rs
```diff
@@ -421,3 +554,21 @@ mod tests {
         result.join("\n") + "\n"
     }
 }
+
+// Filters out dynamic system metrics from Prometheus output for stable golden file tests
+/// System metrics like CPU and memory usage are inherently dynamic and change between test runs
+/// This function removes these metrics to create stable, predictable output for testing.
+#[cfg(test)]
+pub mod test_utils {
+    pub fn filter_system_metrics_output(text: &str) -> String {
```
cgorenflo, line 563:
```suggestion
    pub fn filter_out_system_metrics(text: &str) -> String {
```

ampd/src/monitoring/endpoints/metrics.rs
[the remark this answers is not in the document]
yidan233: ooooright!
but I think 
```
let r = r.with_cpu();
assert_eq!(r.cpu(), true);
```
`cpu()` is just returning a bool value checking if the `processRefreshKind` include cpu related metrics?

ampd/src/monitoring/endpoints/metrics.rs
```diff
@@ -308,6 +314,86 @@ impl VerificationVoteMetrics {
...
+
+        // Encode Memory Usage
+        let memory_gauge = ConstGauge::new(process_metrics.memory_usage);
+        let memory_encoder = encoder.encode_descriptor(
+            "ampd_memory_usage",
+            "Memory usage of the ampd process in bytes",
+            Some(&Unit::Bytes),
+            memory_gauge.metric_type(),
+        )?;
+        memory_gauge.encode(memory_encoder)?;
```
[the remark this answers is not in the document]
yidan233:
```
 let unit = Unit::Other("percent".to_string()); <= defined here 
        let cpu_encoder = encoder.encode_descriptor(
            "ampd_cpu_usage",
            "CPU usage of the ampd process in percentage",
            Some(&unit),
            cpu_gauge.metric_type(),
        )?;
```
the `unit` will be concatenated to the metrics name by `prometheus_client`

ampd/src/monitoring/endpoints/metrics.rs
```diff
@@ -308,6 +314,86 @@ impl VerificationVoteMetrics {
     }
 }
 
+/// System metrics collector that provides real-time CPU and memory usage for the AMPD process.
+/// Metrics are collected fresh on each scrape request using the sysinfo crate.
+/// when metrics is not available, no system metrics will be returned.
+/// this may happens because of permission issues, process termination, or platform limitations
```
cgorenflo, line 320:
```suggestion
/// this may happen because of permission issues, process termination, or platform limitations
```

ampd/src/monitoring/endpoints/metrics.rs
```diff
@@ -433,3 +564,22 @@ mod tests {
         result.join("\n") + "\n"
     }
 }
+
+/// System metrics like CPU and memory usage are inherently dynamic and change between test runs
+/// This function normalizes the system metrics to 0 for consistent output in golden file tests.
+#[cfg(test)]
+pub mod test_utils {
+    pub fn normalize_system_metrics(text: &str) -> String {
```
cgorenflo, line 572:
```suggestion
    pub fn zeroize_system_metrics(text: &str) -> String {
```
normalization is something else

ampd/src/monitoring/endpoints/metrics.rs
```diff
@@ -433,3 +564,22 @@ mod tests {
...
+        let mut result = Vec::new();
+        for line in text.lines() {
+            if line.starts_with("ampd_cpu_usage_percent") && !line.starts_with("#") {
+                result.push("ampd_cpu_usage_percent 0".to_string());
+            } else if line.starts_with("ampd_memory_usage_bytes") && !line.starts_with("#") {
+                result.push("ampd_memory_usage_bytes 0".to_string());
+            } else {
+                result.push(line.to_string());
+            }
+        }
```
cgorenflo, line 582: this completely overwrites the line, which won't show any discrepancies if something about the output changes. If my assumption that the last part of the line can be a float (or integer) is correct, I'd rather only replace that:

```suggestion
        let trailing_number = Regex::new(r"\d+(\.\d+)?$").unwrap();
        for line in text.lines() {
            if line.starts_with("ampd_cpu_usage_percent") || line.starts_with("ampd_memory_usage_bytes") {
                result.push(trailing_numer.replace(line, "BLANKED_FOR_TEST"));
            } else {
                result.push(line.to_string());
            }
        }
```
you don't need to check for `#` because it's incompatible with your other `starts_with` checks

PR #968, opened 2025-07-22
feat(ampd): stellar protocol v23 update

ahramy:
- Stellar Protocol V23 update
- `rust 1.85.0` is required due to stellar updates:
```
error: rustc 1.81.0 is not supported by the following packages:
  stellar-rpc-client@23.0.0-rc.5 requires rustc 1.85.0
  stellar-xdr@23.0.0-rc.2 requires rustc 1.84.0
  stellar-xdr@23.0.0-rc.2 requires rustc 1.84.0
  stellar-xdr@23.0.0-rc.2 requires rustc 1.84.0
Either upgrade rustc or select compatible dependency versions with
`cargo update <name>@<current-ver> --precise <compatible-ver>`
where `<compatible-ver>` is the latest version supporting rustc 1.81.0

error: rustc 1.84.0 is not supported by the following package:
  stellar-rpc-client@23.0.0-rc.5 requires rustc 1.85.0
Either upgrade rustc or select compatible dependency versions with
`cargo update <name>@<current-ver> --precise <compatible-ver>`
where `<compatible-ver>` is the latest version supporting rustc 1.84.0
```


ahramy:
There appear to be some differences in lint formatting between 1.85.0 and 1.86.0. I updated the code to pass the `Basic / Lints` checks.

packages/interchain-token-service-std/Cargo.toml
```diff
@@ -12,7 +12,7 @@ exclude = ["contract.wasm", "hash.txt"]
 optimize = """docker run --rm -v "$(pwd)":/code \
   --mount type=volume,source="$(basename "$(pwd)")_cache",target=/code/target \
   --mount type=volume,source=registry_cache,target=/usr/local/cargo/registry \
-  cosmwasm/optimizer:0.16.1
+  cosmwasm/optimizer:0.17.0
```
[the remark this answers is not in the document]
ahramy: I see. Removed the script from the `Cargo.toml` file and tested `just optimize`:
```
just optimize

docker run --rm -v "$(pwd)":/code --mount type=volume,source="$(basename "$(pwd)")_cache",target=/code/target --mount type=volume,source=registry_cache,target=/usr/local/cargo/registry cosmwasm/optimizer:0.17.0
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
1.86.0-x86_64-unknown-linux-musl (active, default)
cargo 1.86.0 (adf9b6ad1 2025-02-28)
Building project /code ...
Found workspace member entries: ["ampd", "ampd-handlers", "contracts/*", "external-gateways/*", "integration-tests", "packages/*", "tools/*"]
Package directories: ["ampd", "ampd-handlers", "contracts/axelarnet-gateway", "contracts/coordinator", "contracts/gateway", "contracts/interchain-token-service", "contracts/its-abi-translator", "contracts/multisig", "contracts/multisig-prover", "contracts/rewards", "contracts/router", "contracts/service-registry", "contracts/voting-verifier", "external-gateways/stellar", "integration-tests", "packages/ampd-proto", "packages/ampd-sdk", "packages/axelar-core-std", "packages/axelar-wasm-std", "packages/axelar-wasm-std-derive", "packages/client", "packages/events", "packages/events-derive", "packages/evm-gateway", "packages/gateway-api", "packages/interchain-token-service-std", "packages/into-inner-derive", "packages/its-msg-translator-api", "packages/msgs-derive", "packages/multisig-prover-api", "packages/report", "packages/router-api", "packages/service-registry-api", "packages/signature-verifier-api", "packages/starknet-checked-felt", "packages/sui-gateway", "packages/sui-types", "packages/voting-verifier-api", "packages/xrpl-types", "tools/migration-remover"]
Contracts to be built: ["contracts/axelarnet-gateway", "contracts/coordinator", "contracts/gateway", "contracts/interchain-token-service", "contracts/its-abi-translator", "contracts/multisig", "contracts/multisig-prover", "contracts/rewards", "contracts/router", "contracts/service-registry", "contracts/voting-verifier"]
Building "axelarnet_gateway" ...
Building axelarnet_gateway with features {}

...

Optimizing ...

Post-processing artifacts...
d28b4358647a2d74a6d4777f003c2c02ba0d791365c15f4527d4d92673371df0  axelarnet_gateway.wasm
8a993dfe8f6a766efd37cdc9529a4efca5495c52fb21cf150986f46a9f9a1cc0  coordinator.wasm
4f06c17135c87a227a778fc64a73d47cc27b57ad5b453c447196fb7991339a02  gateway.wasm
3a3b1247c24783509fc33fe30c0303ce12ca5cf7dde73f40dd8c235386e9ae7e  interchain_token_service.wasm
1b27527ca670ae5f7137731d846f24396a969fcefaa83c9f8d2027286f7dbe37  its_abi_translator.wasm
a62084b7ba1c733af1d9b7e4e1b9a6e2973517b14491965d6de00a6f3614d4ca  multisig.wasm
888beee2fabb465c43fef68414caa081380d41e8d113c22c5e86fcead5f47444  multisig_prover.wasm
2c83f0515bbceceb8ab9650f97548e2853106f9710bee10f57279e07e235837a  rewards.wasm
4683499d777071b894acbbef8c9c04d24c373607b57ddddc80639b1348dd81c8  router.wasm
418c2f7c8e37e23da44bb195224c2559b9212bcd1bc367d4ead20f1dc29e0cd3  service_registry.wasm
800071e354ac5d8bf3d0c5fafc53e5292b33a392d05919a231166d82804ee436  voting_verifier.wasm
Done.

```

ampd/src/stellar/rpc_client.rs
```diff
@@ -28,13 +28,22 @@ const STATUS_SUCCESS: &str = "SUCCESS";
...
-        let contract_events = response
-            .result_meta
-            .and_then(|meta| match meta {
-                TransactionMeta::V3(data) => data.soroban_meta.map(|meta| meta.events),
-                _ => None,
-            })
-            .unwrap_or_default();
+        // Protocol 23 (CAP-0067): Extract contract events from the unified events structure
+        // contract_events is Vec<Vec<ContractEvent>> (per operation), so we flatten it
+        let events_vec: Vec<ContractEvent> = response
```
milapsheth, line 33: Can this change the order of events emitted in txs pre-v23 upgrade? If so, that would allow replay attack since the same message can replayed under a new message id. Should look into it closer.

ampd/src/stellar/rpc_client.rs
```diff
@@ -28,13 +28,22 @@ const STATUS_SUCCESS: &str = "SUCCESS";
...
     fn from((transaction_hash, response): (Hash, GetTransactionResponse)) -> Self {
-        let contract_events = response
-            .result_meta
-            .and_then(|meta| match meta {
-                TransactionMeta::V3(data) => data.soroban_meta.map(|meta| meta.events),
-                _ => None,
-            })
-            .unwrap_or_default();
+        // Protocol 23 (CAP-0067): Extract contract events from the unified events structure
+        // contract_events is Vec<Vec<ContractEvent>> (per operation), so we flatten it
```
milapsheth, line 32: Stellar still allows a single actual soroban call in one tx, right? Though flattening is fine

ampd/src/stellar/rpc_client.rs
milapsheth, line 33:
```suggestion
        let events: Vec<ContractEvent> = response
```
avoid type suffix in variable names

ampd/src/stellar/rpc_client.rs
```diff
@@ -28,13 +28,22 @@ const STATUS_SUCCESS: &str = "SUCCESS";
...
+            .into_iter()
+            .flatten()
+            .collect();
+
+        let contract_events = events_vec.try_into().unwrap_or_else(|_| {
+            warn!(
+                tx_hash = %transaction_hash,
+                "Contract events exceed VecM capacity, returning empty list"
+            );
+            VecM::default()
```
milapsheth, line 45: hmm, silent failure. we should handle the failure explicitly, probably something we can do with the ampd handler approach better

ahramy: I see. I added a logic to handle the failure explicitly. please take a look

ampd/src/stellar/rpc_client.rs
ahramy: yeah, I think flattening is the way to go

ampd/src/stellar/rpc_client.rs
ahramy: Yeah, it's a valid concern, and it's stated in the [CAP-67 document](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0067.md), in the `Retroactively emitting events` section:
```
“There will be a configuration flag added that will result in [V20,V22] events being emitted in the V23 format for the scenarios just mentioned.
This does mean that the events can't be used to verify the hash of the events in InvokeHostFunctionResult, but this flag will be disabled by default.”
```
This confirms Pre-v23 ledgers can emit updated event formats (e.g. mint instead of transfer) if the flag is enabled, Enabling it changes event content and potentially order, breaking event hash verification. This could allow replay attacks if message IDs are derived from event hashes. 
By default, this behavior is off, so original event order and structure is preserved.

ampd/src/stellar/rpc_client.rs
cjcobb23: can we just use a regular Vec here? I don't see why we need to use this `VecM` type. Also, it looks like the capacity is u32::Max, doubt we will ever hit this case.

ampd/Cargo.toml
```diff
@@ -75,7 +75,7 @@ starknet-checked-felt = { workspace = true }
 starknet-core = { workspace = true }
 starknet-providers = { workspace = true }
 stellar = { workspace = true }
-stellar-rpc-client = "21.4.0"
+stellar-rpc-client = "23.0.0-rc.5"
```
cjcobb23, line 78: is this right, we are building off of an RC?

cjcobb23: also should be  { workspace = true}

Cargo.toml
```diff
@@ -98,7 +98,7 @@ starknet-core = "0.12.0"
 starknet-providers = "0.12.0"
 starknet-types-core = { version = "0.1.7" }
 stellar = { version = "^1.0.0", path = "external-gateways/stellar" }
-stellar-xdr = { version = "21.2.0" }
+stellar-xdr = { version = "23.0.0-rc.2" }
```
cjcobb23, line 101: does this need to be in the workspace? I don't think we are using this anywhere except ampd

ampd/Cargo.toml
ahramy: Yeah, no stable release yet, just RCs for now. 

This is the only place stellar-rpc-client is being used, so `{workspace = true}` doesn't work.
If I change it to use the workspace, I get an error: `dependency.stellar-rpc-client was not found in workspace.dependencies.`

Cargo.toml
ahramy: no need to be in the workspace

ampd/src/handlers/stellar_verify_msg.rs
```diff
@@ -304,10 +304,12 @@ mod tests {
             metadata: PollMetadata {
                 poll_id: "100".parse().unwrap(),
                 source_chain: "stellar".parse().unwrap(),
-                source_gateway_address: ScAddress::Contract(stellar_xdr::curr::Hash::from([1; 32]))
-                    .to_string()
-                    .try_into()
-                    .unwrap(),
+                source_gateway_address: ScAddress::Contract(stellar_xdr::curr::ContractId(
+                    stellar_xdr::curr::Hash::from([1; 32]),
+                ))
```
cgorenflo, line 309:
```suggestion
                source_gateway_address: ScAddress::Contract(stellar_xdr::curr::Hash::from([1; 32]).into())
```
adjust accordingly in other places
you could even do (maybe, I don't know if the compiler will be able to find this)
```suggestion
                source_gateway_address: ScAddress::Contract([1; 32].into().into())
```

ampd/src/lib.rs
```diff
@@ -471,10 +471,8 @@ impl App {
                 handlers::stellar_verify_msg::Handler::new(
                     verifier.clone(),
                     cosmwasm_contract.clone(),
-                    stellar::rpc_client::Client::new(
-                        rpc_url.to_string().trim_end_matches('/').into(),
-                    )
-                    .change_context(Error::Connection)?,
+                    stellar::rpc_client::Client::new(rpc_url.as_str().trim_end_matches('/').into())
+                        .change_context(Error::Connection)?,
```
cgorenflo, line 475: I'd rather change the Client definition to take in a Url, then you can't give it an invalid url, and we also don't have to do this conversion everywhere. Side note, the client expects a String just to convert it into a &str, so you convert Url into &str into String into &str here.
Is the trimming necessary?

cgorenflo: adjust other locations accordingly

ampd/src/stellar/rpc_client.rs
```diff
@@ -21,20 +21,21 @@ pub enum Error {
...
-            })
-            .unwrap_or_default();
+        // Protocol 23 (CAP-0067): Extract contract events from the unified events structure
+        // contract_events is Vec<Vec<ContractEvent>> (per operation), so we flatten it
+        let contract_events: Vec<ContractEvent> = response
+            .events
+            .contract_events
+            .into_iter()
+            .flatten()
+            .collect();
```
cgorenflo, line 38: this seems highly dangerous and I believe there have been discussion around it that I didn't have time to follow. Stellar is changing retroactively how events are emitted. This can open all kinds of double spending attacks (different order, different events). We need to make ABSOLUTELY SURE this cannot be exploited to verify the same message multiple times (either new or existing) before going ahead with it

external-gateways/stellar/src/lib.rs
```diff
@@ -84,7 +84,9 @@ impl TryFrom<Message> for ScVal {
         ];
 
         let vals: [ScVal; 5] = [
-            ScVal::Address(ScAddress::Contract(Hash(value.contract_address.0))),
+            ScVal::Address(ScAddress::Contract(stellar_xdr::curr::ContractId(Hash(
+                value.contract_address.0,
+            )))),
```
cgorenflo, line 89: Hash().into() to simplify

ampd/src/stellar/rpc_client.rs
cgorenflo: can we enforce a block height so you can't verify past txs before the upgrade?

cgorenflo: or maybe the tx metadata can signify if it is from pre or post upgrade?

ampd/src/handlers/stellar_verify_msg.rs
ahramy: got it, updated to `source_gateway_address: ScAddress::Contract(stellar_xdr::curr::Hash::from([1; 32]).into())`. the latter give a compile time error.

ampd/src/stellar/rpc_client.rs
ahramy: this was confirmed and resolved by the conversation with stellar team: https://interoplabs.slack.com/archives/C04545F3S59/p1754328250849909

external-gateways/stellar/src/lib.rs
ahramy: updated

ampd/src/lib.rs
ahramy: make sense. updated.

ampd/src/stellar/rpc_client.rs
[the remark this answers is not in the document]
ahramy: we confirmed this with the stellar team: https://interoplabs.slack.com/archives/C04545F3S59/p1754328250849909

packages/axelar-wasm-std/src/address.rs
```diff
@@ -188,8 +188,22 @@ mod tests {
             address::Error::InvalidAddress(..)
         );
 
-        // invalid
-        let invalid = "MA7QYNF7SOWQ3GLR2BGMZEHXAVIRZA4KVWLTJJFC7MGXUA74P7UJVAAAAAAAAAAAAAJLK";
+        // muxed account (Protocol 23 CAP-0067 support)
```
milapsheth, line 191: This is an unrelated breaking change to protocol behaviour. E.g. a previous message to such an address would now be considered valid. I don't think we need to support MUX addresses as valid. Let's only treat basic user accounts and contact addresses as valid unless there's a clear requirement to support these address types

ampd/src/lib.rs
```diff
@@ -478,10 +478,8 @@ impl App {
                 handlers::stellar_verify_msg::Handler::new(
                     verifier.clone(),
                     cosmwasm_contract.clone(),
-                    stellar::rpc_client::Client::new(
-                        rpc_url.to_string().trim_end_matches('/').into(),
```
milapsheth, line 482: Are you sure about removing the trimming? Have you tested the negative edge case? I think Haiyi might have added this for a reason

ampd/src/stellar/rpc_client.rs
```diff
@@ -21,20 +23,21 @@ pub enum Error {
...
     fn from((transaction_hash, response): (Hash, GetTransactionResponse)) -> Self {
-        let contract_events = response
-            .result_meta
-            .and_then(|meta| match meta {
-                TransactionMeta::V3(data) => data.soroban_meta.map(|meta| meta.events),
-                _ => None,
-            })
-            .unwrap_or_default();
+        // Protocol 23 (CAP-0067): Extract contract events from the unified events structure
+        // contract_events is Vec<Vec<ContractEvent>> (per operation), so we flatten it
```
milapsheth, line 34: While the Stellar team has mentioned that this shouldn't affect past events, this is still a dangerous and unintuitive behaviour. Better to think more about how to do this safely. E.g. Stellar team said that a Soroban contact call can only occur as a single operation in the tx. Using that information, why not just extract the first element, otherwise fail/return if length isn't 1

milapsheth, line 34: The context from our slack discussion with their team isn't documented here so someone else looking at this won't understand the reasoning

packages/axelar-wasm-std/src/address.rs
ahramy: I see. removed

ampd/src/cosmos.rs
```diff
@@ -771,9 +771,9 @@ mod tests {
     }
 
     #[test]
+    #[allow(deprecated)]
```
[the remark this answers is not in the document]
ahramy: running `cargo test` also gives the warning. do we want to remove warnings? by allowing this on the whole test?
```
warning: use of deprecated field `cosmos_sdk_proto::cosmos::tx::v1beta1::SimulateRequest::tx`
   --> ampd/src/cosmos.rs:777:13
    |
777 | /             tx: Some(Tx {
778 | |                 body: None,
779 | |                 auth_info: None,
780 | |                 signatures: vec![vec![5, 6, 7, 8]],
781 | |             }),
    | |______________^
    |
    = note: `#[warn(deprecated)]` on by default

warning: use of deprecated field `cosmos_sdk_proto::cosmos::base::abci::v1beta1::Result::data`
   --> ampd/src/cosmos.rs:791:17
    |
791 |                 data: vec![1, 2, 3],
    |                 ^^^^^^^^^^^^^^^^^^^

   Compiling ampd-sdk v0.1.0 (/Users/zenik/5_its_deploy/axelar-amplifier/packages/ampd-sdk)
   Compiling ampd-handlers v0.1.0 (/Users/zenik/5_its_deploy/axelar-amplifier/ampd-handlers)
warning: `ampd` (lib test) generated 2 warnings
    Finished `test` profile [unoptimized + debuginfo] target(s) in 50.94s
     Running unittests src/lib.rs (target/debug/deps/ampd-1072e12973f786fe)
```

ampd/src/stellar/rpc_client.rs
ahramy: make sense, updated

ahramy: yeah, updated

ampd/src/stellar/rpc_client.rs
```diff
@@ -21,24 +23,38 @@ pub enum Error {
...
+                "Transaction operation count does not match expected single operation for Soroban contract call"
+            );
+        }
 
         Self {
-            transaction_hash: transaction_hash.to_string(),
-            successful: response.status == STATUS_SUCCESS,
+            transaction_hash: tx_hash_str,
+            successful: (response.status == STATUS_SUCCESS)
+                && operation_count == EXPECTED_OPERATION_COUNT,
```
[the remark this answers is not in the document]
ahramy: the current approach is valid. reference: https://github.com/stellar/stellar-rpc/blob/protocol-23/cmd/stellar-rpc/internal/integrationtest/get_transactions_test.go#L104

ampd/src/stellar/rpc_client.rs
```diff
@@ -27,21 +27,58 @@ pub struct TxResponse {
...
+                // Protocol 23: Soroban contract calls can only occur as a single operation in the tx
+                // Therefore, we extract the first (and only) operation, otherwise fail.
+                let contract_events = response.events.contract_events;
+                let op_count = contract_events.len();
+                let events = contract_events.into_iter().flatten().collect();
+
+                if op_count != EXPECTED_OPERATION_COUNT_V4 {
+                    warn!(
+                        tx_hash = %transaction_hash,
+                        operation_count = op_count,
```
milapsheth, line 47:
```suggestion
                        operation_count = operation_count,
```
for clarity

ampd/src/stellar/rpc_client.rs
```diff
@@ -27,21 +27,58 @@ pub struct TxResponse {
...
+                );
+                (vec![], false)
+            }
+        };
 
         Self {
-            transaction_hash: transaction_hash.to_string(),
-            successful: response.status == STATUS_SUCCESS,
+            transaction_hash,
+            successful,
```
milapsheth, line 81: the meaning of `successful` variable is being changed here. It doesn't represent whether the tx was successful anymore. Change the structure of the code instead to do this in a more clear way. E.g. Implement `TryFrom` instead of `From` trait, so if there's an error (e.g. invalid op count), we return an Err instead of just making things sort of work with existing types

milapsheth: It feels weird to add logging in a From conversion util. Let the caller log the returned error instead

ampd/src/stellar/rpc_client.rs
```diff
@@ -27,21 +27,58 @@ pub struct TxResponse {
...
-            .collect();
+        let transaction_hash = transaction_hash.to_string();
+
+        let (contract_events, successful) = match response.result_meta.as_ref() {
+            Some(TransactionMeta::V4(_)) => {
+                // Protocol 23: Soroban contract calls can only occur as a single operation in the tx
+                // Therefore, we extract the first (and only) operation, otherwise fail.
+                let contract_events = response.events.contract_events;
+                let op_count = contract_events.len();
+                let events = contract_events.into_iter().flatten().collect();
```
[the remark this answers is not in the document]
ahramy: make sense, updated

ampd/src/cosmos.rs
```diff
@@ -771,9 +771,9 @@ mod tests {
     }
 
     #[test]
-    #[allow(deprecated)]
     fn ensure_simulate_req_res_serialization_do_not_change() {
         let req = SimulateRequest {
+            #[allow(deprecated)]
             tx: Some(Tx {
```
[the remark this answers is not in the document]
ahramy: updated, now the warning is gone!

ampd/src/cosmos.rs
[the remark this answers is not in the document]
ahramy: updated!

ampd/src/stellar/rpc_client.rs
```diff
@@ -16,31 +18,68 @@ pub enum Error {
...
+    ) -> Result<Self, Self::Error> {
+        let transaction_hash = transaction_hash.to_string();
+        if response.status != STATUS_SUCCESS {
+            return Ok(Self {
+                transaction_hash,
+                successful: false,
+                contract_events: vec![],
+            });
         }
+        let contract_events = match response.result_meta.as_ref() {
```
milapsheth, line 54:
```suggestion

        let contract_events = match response.result_meta.as_ref() {
```

ampd/src/stellar/rpc_client.rs
```diff
@@ -16,31 +18,68 @@ pub enum Error {
...
+                let operation_count = contract_events.len();
+
+                if operation_count != EXPECTED_SOROBAN_OPERATION_COUNT {
+                    return Err(TxParseError::InvalidOperationCount {
+                        expected: EXPECTED_SOROBAN_OPERATION_COUNT,
+                        actual: operation_count,
+                    });
+                }
+
+                contract_events.into_iter().next().unwrap_or_default()
```
milapsheth, line 66: Using default is dangerous in case the logic is moved around/changes
```suggestion
                contract_events.get(0).expect("soroban operation not found")
```
Expect on it

ampd/src/stellar/rpc_client.rs
```diff
@@ -16,31 +18,68 @@ pub enum Error {
...
+                successful: false,
+                contract_events: vec![],
+            });
         }
+        let contract_events = match response.result_meta.as_ref() {
+            Some(TransactionMeta::V4(_)) => {
+                let contract_events = response.events.contract_events;
+                let operation_count = contract_events.len();
+
+                if operation_count != EXPECTED_SOROBAN_OPERATION_COUNT {
```
cjcobb23, line 60: why is this check necessary? comments would be helpful

ampd/src/stellar/rpc_client.rs
cjcobb23: can't you just match on this as a slice:
```
match contract_events.as_slice() {
    [event] -> event,
     _ -> return Err(...)
}
```

ahramy: make sense. updated

ampd/src/stellar/rpc_client.rs
cjcobb23: also does this mean the event index is always 0?

ampd/src/stellar/rpc_client.rs
```diff
@@ -16,31 +18,72 @@
...
+            }
+            Some(TransactionMeta::V3(data)) => data
+                .soroban_meta
+                .as_ref()
+                .map(|meta| meta.events.to_vec())
+                .unwrap_or_default(),
+            _ => {
+                return Err(TxParseError::UnsupportedMetadataVersion);
+            }
+        };
```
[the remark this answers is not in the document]
ahramy: this is expected

ampd/src/stellar/rpc_client.rs
ahramy: lint error: ```error: accessing first element with `contract_events.get(0)`
Error:   --> ampd/src/stellar/rpc_client.rs:67:17
   |
67 | /                 contract_events
68 | |                     .get(0)
   | |___________________________^ help: try: `contract_events.first()`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#get_first
   = note: `-D clippy::get-first` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(clippy::get_first)]`
   ```
   thus, updated to use `first()`

ampd/src/stellar/rpc_client.rs
```diff
@@ -51,19 +51,26 @@ impl TryFrom<(Hash, GetTransactionResponse)> for TxResponse {
...
                         expected: EXPECTED_SOROBAN_OPERATION_COUNT,
                         actual: operation_count,
                     });
                 }
 
-                contract_events.into_iter().next().unwrap_or_default()
+                contract_events
+                    .first()
+                    .expect("soroban operation not found")
+                    .clone()
```
milapsheth, line 73: there are a bunch of lint warnings below, address those? any reason the gh actions aren't catching them, maybe some config needs to be tuned?

milapsheth (approved):

packages/axelar-wasm-std/src/address.rs
milapsheth: this wasn't addressed. we're still treating muxed addresses as valid

packages/axelar-wasm-std/src/address.rs
```diff
@@ -188,8 +188,8 @@ mod tests {
             address::Error::InvalidAddress(..)
         );
 
-        // invalid
-        let invalid = "MA7QYNF7SOWQ3GLR2BGMZEHXAVIRZA4KVWLTJJFC7MGXUA74P7UJVAAAAAAAAAAAAAJLK";
```
milapsheth, line 192: why are we treating this address as valid? take a look at my original comment again that got resolved
https://github.com/axelarnetwork/axelar-amplifier/pull/968#discussion_r2272147599

ampd/src/stellar/rpc_client.rs
```diff
@@ -111,11 +160,233 @@ impl Client {
         let tx_hash = Hash::from_str(tx_hash.as_str()).change_context(Error::TxHash)?;
 
         match self.0.get_transaction(&tx_hash).await {
-            Ok(response) => Ok(Some(TxResponse::from((tx_hash, response)))),
+            Ok(response) => match TxResponse::try_from((tx_hash, response)) {
+                Ok(tx_response) => Ok(Some(tx_response)),
+                Err(err) => {
+                    warn!(error = %err, "failed to parse transaction response");
+                    Ok(None)
+                }
```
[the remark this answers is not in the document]
ahramy: Thanks for the comment. now the code looks much cleaner. updated.

packages/axelar-wasm-std/src/address.rs
ahramy: got it. updated

packages/axelar-wasm-std/src/address.rs
ahramy: I think it was missed, updated to consider muxed tyep as invalid.

ampd/src/stellar/rpc_client.rs
```diff
@@ -65,13 +111,33 @@ pub struct Client(stellar_rpc_client::Client);
...
+    fn validate_tx_response(
+        result: Result<GetTransactionResponse, stellar_rpc_client::Error>,
+        hash: Hash,
+    ) -> Option<TxResponse> {
+        result
+            .map_err(|err| {
+                warn!(error = ?err, tx_hash = ?hash, "failed to get transaction response");
+                err
+            })
+            .ok()
```
milapsheth, line 130: Too much nesting, not very clean/readable. Return early using let else if it's None and break the next part into a separate statement?

milapsheth (approved):

ampd/src/stellar/rpc_client.rs
ahramy: cool, updated, now it looks cleaner

ampd/src/stellar/rpc_client.rs
```diff
@@ -65,13 +111,30 @@ pub struct Client(stellar_rpc_client::Client);
...
                 .map_err(|err_str| report!(Error::Client).attach_printable(err_str))?,
         ))
     }
 
+    fn validate_tx_response(
+        result: Result<GetTransactionResponse, stellar_rpc_client::Error>,
+        hash: Hash,
+    ) -> Option<TxResponse> {
+        let Ok(response) = result else {
+            warn!(error = ?result.unwrap_err(), tx_hash = ?hash, "failed to get transaction response");
```
milapsheth, line 126: Bad practice to unwraps. Use destructuring to ensure the enum is always handled safely. Since you need to use `err` here let else statement doesn't work as well compared to match or using map_err and then doing let else

milapsheth (approved):

ampd/src/stellar/rpc_client.rs
ahramy: I see. udated

packages/axelar-wasm-std/src/address.rs
```diff
@@ -188,8 +188,22 @@ mod tests {
...
+
+        let lower_case = muxed_addr.to_lowercase();
+        assert_err_contains!(
+            address::validate_address(&lower_case, &address::AddressFormat::Stellar),
+            address::Error,
+            address::Error::InvalidAddress(..)
+        );
+
+        // invalid - use clearly malformed address with invalid characters
+        let invalid = "INVALID_STELLAR_ADDRESS_WITH_BAD_CHARS_123!@#";
```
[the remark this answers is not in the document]
ahramy: resolved, made it invalid since we don't support it yet

PR #973, opened 2025-07-28
feat(ampd): track error returns by rpc nodes 

yidan233:
- Wrapped all RPC clients with `monitoring::Client` and a `handler_label`
- Recorded RPC errors via Msg::RpcError { handler_label } 
- **All errors** returned from RPC nodes are recorded, even if they represent expected states like `TransactionNotFound`.
-  `MVX`, `Solana`, `Starknet` do not distinguish between connection-level errors and valid “`transaction not found`” responses both are returned as `Err(...)`. comments are added at top to note the `TransactionNoteFound` returned error type I have found


ampd/src/json_rpc.rs
```diff
@@ -16,14 +19,20 @@ where
     P: JsonRpcClient,
 {
     provider: P,
+    monitoring_client: monitoring::Client,
+    handler_label: String,
```
cgorenflo, line 23: the json client shouldn't care about the handler. Given that the chain name looks like the important corresponding factor across multiple metrics, use `chain: ChainName` instead

cgorenflo: the rpc is the same across different handlers for the same chain anyway

ampd/src/monitoring/endpoints/metrics.rs
```diff
@@ -40,6 +40,8 @@ pub enum Msg {
         vote_status: Vote,
         chain_name: String,
     },
+    /// number of errors returned from RPC Calls
+    RpcError { handler_label: String },
```
cgorenflo, line 44: same here, go with chain (or chain_name, just use it everywhere consistently)

ampd/src/monitoring/endpoints/metrics.rs
```diff
@@ -202,12 +204,22 @@ async fn serve_metrics(
...
+struct HandlerLabel {
+    handler_label: String,
 }
 
 struct BlockReceivedMetrics {
     total: Counter,
 }
 
+struct RpcErrorMetrics {
+    total: Family<HandlerLabel, Counter>,
```
cgorenflo, line 220: doesn't seem necessary, just use `String` for the label?

ampd/src/monitoring/endpoints/metrics.rs
```diff
@@ -251,10 +266,34 @@ impl Metrics {
...
             }
+
+            Msg::RpcError { handler_label } => {
+                self.rpc_error.record_rpc_error(handler_label);
+            }
         }
     }
 }
 
+impl RpcErrorMetrics {
```
cgorenflo, line 277: keep impl blocks close to the struct

ampd/src/monitoring/endpoints/metrics.rs
```diff
@@ -251,10 +266,34 @@ impl Metrics {
...
+    fn new() -> Self {
+        let total = Family::<HandlerLabel, Counter>::default();
+        Self { total }
+    }
+
+    fn register(&self, registry: &mut Registry) {
+        registry.register(
+            "handler_rpc_errors",
+            "number of RPC errors per handler",
```
cgorenflo, line 286:
```suggestion
            "number of RPC errors per chain",
```

ampd/src/monitoring/endpoints/metrics.rs
```diff
@@ -251,10 +266,34 @@ impl Metrics {
...
+            self.total.clone(),
+        );
+    }
+
+    fn record_rpc_error(&self, handler_label: String) {
```
cgorenflo, line 291: chain name

ampd/src/stellar/rpc_client.rs
```diff
@@ -61,15 +64,48 @@ impl TxResponse {
...
+                warn!(
+                    handler_label = &self.handler_label,
+                    tx_hash = ?tx_hash,
+                    "received RPC error: {:?}", err
+                );
+                self.monitoring_client
+                    .metrics()
+                    .record_metric(Msg::RpcError {
+                        handler_label: self.handler_label.clone(),
+                    });
```
cgorenflo, line 107: move this into `transaction_response` and get rid of this function

ampd/src/xrpl/json_rpc.rs
```diff
@@ -15,9 +40,24 @@ pub trait XRPLClient {
...
+            .call(req)
+            .await
+            .map(Some)
+            .or_else(|err| match err {
+                error::Error::Api(reason) if reason == "txnNotFound" => Ok(None),
+                _ => {
+                    warn!(
+                        handler_label = &self.handler_label,
+                        "received RPC error: {:?}", err
+                    );
```
cgorenflo, line 53: remove

ampd/src/mvx/proxy.rs
```diff
@@ -6,11 +6,57 @@ use hex::ToHex;
...
+            .inspect_err(|err| {
+                warn!(
+                    handler_label = &self.handler_label,
+                    "received RPC error: {:?}", err
+                );
+                self.monitoring_client
+                    .metrics()
+                    .record_metric(Msg::RpcError {
+                        handler_label: self.handler_label.clone(),
+                    });
```
cgorenflo, line 55: just do this inside `transaction_info_with_results` and you don't even have to import anyhow

cgorenflo (commented):
I didn't consistently mark all the same issues across all clients because it's repeating everywhere:
- don't log warnings if the error gets passed on (or if it was transformed into an option and there was no warning log before)
- send the monitoring message inside the existing client transaction, no need to add another layer of abstraction
- use the chain name instead of tying the clients to handlers (ideally they shouldn't even know about that, but it does make some logical sense)

ampd/src/monitoring/endpoints/metrics.rs
cgorenflo, line 1: can you merge the metrics test into a single one in a follow up PR? Every time we add a new metric, all the golden files change. I'd rather have one test checking for all the metrics

ampd/src/mvx/proxy.rs
```diff
@@ -6,11 +6,34 @@ use hex::ToHex;
...
+use router_api::ChainName;
 
+use crate::monitoring;
+use crate::monitoring::metrics::Msg;
 use crate::types::Hash;
 
 const STATUS_SUCCESS: &str = "success";
 
+pub struct Client {
+    client: GatewayProxy,
```
cgorenflo, line 18:
```suggestion
    proxy: GatewayProxy,
```
to stay somewhat consistent with the naming

ampd/src/stacks/http_client.rs
```diff
@@ -134,9 +159,23 @@ impl Client {
             .get(endpoint.clone())
             .send()
             .await
+            .inspect_err(|_| {
```
cgorenflo, line 162: `transaction` is a private function that is only called by `valid_transaction` (and your test, which shouldn't call into a private function). If you record the metrics in `valid_transaction` you only have to define it once

ampd/src/stacks/http_client.rs
```diff
@@ -119,9 +130,23 @@ impl Client {
...
+                self.monitoring_client
+                    .metrics()
+                    .record_metric(Msg::RpcError {
+                        chain_name: self.chain_name.clone(),
+                    });
+            })
```
cgorenflo, line 139: with `use futures::TryFutureExt;` you can get around this double recording:
```
self.client
            .get(endpoint.clone())
            .send()
            .map_err(|_| Error::LatestBlock { endpoint })
            .and_then(|response| response.json::<Block>().map_err(|_| Error::Json))
            .await
            .inspect_err(|_| {
                self.monitoring_client
                    .metrics()
                    .record_metric(Msg::RpcError {
                        chain_name: self.chain_name.clone(),
                    });
            })
```

PR #974, opened 2025-07-29
feat(ampd): update grpc client api to match changes in proto files

eguajardo:
## Description
Should be merged together with https://github.com/axelarnetwork/ampd-proto/pull/10

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros


packages/axelar-wasm-std/src/chain_name.rs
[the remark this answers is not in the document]
eguajardo: it was mostly moved, this is the only thing that I had to add:
```
impl From<Error> for crate::error::ContractError {
    fn from(error: Error) -> Self {
        report!(error).into()
    }
}
```
I couldn't use `IntoContractError` directly because it was part of the `derive` package which is only accessible through an optional feature flag. I tried some other alternatives like returning `Report` but then a lot of other things started to break that were using `CrossChainId::new` and I couldn't figure out a way to fix them without changing like 20+ files. I might give it another shot at this in the future

packages/axelar-wasm-std/src/chain_name.rs
```diff
@@ -0,0 +1,461 @@
...
+pub enum Error {
+    #[error("chain name is invalid")]
+    InvalidChainName,
+}
+
+impl From<Error> for crate::error::ContractError {
+    fn from(error: Error) -> Self {
+        report!(error).into()
+    }
+}
```
cgorenflo, line 24:
```suggestion
use crate as axelar_wasm_std;
use crate::{FnExt, FIELD_DELIMITER};

#[derive(thiserror::Error, Debug, PartialEq, IntoContractError)]
pub enum Error {
    #[error("chain name is invalid")]
    InvalidChainName,
}
```

packages/axelar-wasm-std/src/chain_name.rs
```diff
@@ -0,0 +1,461 @@
...
+use cw_storage_plus::{Key, KeyDeserialize, Prefixer, PrimaryKey};
+use error_stack::report;
+use valuable::Valuable;
+
+use crate::{FnExt, FIELD_DELIMITER};
+
+#[derive(thiserror::Error, Debug, PartialEq)]
+pub enum Error {
+    #[error("chain name is invalid")]
+    InvalidChainName,
```
cgorenflo, line 17: InvalidChainName is still an error defined in router-api, mostly because of the constructor for CrossChainId. It's very difficult to get around that. Would it make more sense to just move the entire `primitives.rs` into this package? (separate PR)

It's kind of odd that the ChainName types specifically are done here while all the other core types live in the router-api

packages/axelar-wasm-std/src/chain_name.rs
eguajardo: can't use stuff from `axelar-wasm-std-derive` because it's only available to `axelar-wasm-std` using the `derive` feature flag, it would need to depend on itself using the `derive` feature flag

PR #987, opened 2025-08-01
ci: unify Rust setup across workflows

cgorenflo:
## Summary

This PR unifies Rust toolchain setup across all GitHub Actions workflows by creating a centralized configuration and shared action.

### 🎯 Centralized Rust Configuration
- Added `rust-toolchain.toml` to specify Rust 1.86.0 with required components and targets
- All workflows now use this single source of truth for Rust version

### 🔧 Shared Setup Action
- Created `.github/actions/setup-rust` composite action that:
  - Uses `actions-rust-lang/setup-rust-toolchain` which automatically reads `rust-toolchain.toml`
  - Configures `useblacksmith/rust-cache` for Blacksmith-optimized caching
  - Uses resolved toolchain hash for automatic cache separation

### 📝 Workflow Updates
- Replaced `dtolnay/rust-toolchain` with the new shared action
- Removed redundant `RUSTFLAGS` (now in `.cargo/config.toml`)
- Replaced deprecated `actions-rs/cargo` with direct commands
- Fixed double Rust installation in lints job
- Updated to `baptiste0928/cargo-install@v3` consistently
- Removed deprecated `save-always` flag from docs workflow

### 🐳 Docker Updates
- Updated Dockerfiles with comments referencing `rust-toolchain.toml`

## Why These Changes?

1. **Single Source of Truth**: No more hardcoded Rust versions scattered across workflows
2. **Automatic Target Installation**: `wasm32-unknown-unknown` is installed automatically
3. **Better Caching**: Blacksmith-optimized caching with automatic toolchain-based separation
4. **Reduced Maintenance**: Update Rust version in one place, all workflows follow
5. **Cleaner Workflows**: Removed deprecated actions and redundant configurations

## Test Plan

- [x] All existing CI checks pass
- [x] Rust toolchain is correctly installed from `rust-toolchain.toml`
- [x] Caching works as expected with new cache keys
- [x] No functional changes to the actual build/test process


.github/workflows/basic.yaml
```diff
@@ -148,56 +92,34 @@ jobs:
...
-          toolchain: 1.86.0
-          components: clippy
-          override: true
+        run: cargo +nightly fmt-check #alias
 
       - name: Run cargo sort
-        uses: actions-rs/cargo@v1
-        with:
-          command: sort-check # alias
+        run: cargo sort-check
```
maancham, line 119: can we bring back the `#alias` comment for this and clippy below?

.github/workflows/docs.yaml
```diff
@@ -13,24 +13,25 @@ concurrency:
...
+          cache: false
+          rustflags: ""
 
       - name: Cache cargo and tools
         id: cache-cargo
         uses: "actions/cache@v4"
         with:
           path: |
             ~/.cargo
-          save-always: true
```
maancham, line 33: I could not find the documentation for this field, why remove it?

cgorenflo: see PR description, it's deprecated. There is a workaround, but it's quite involved and the docs check is just not worth the effort for a small edge case. Worst case we have a few more cache misses

.github/workflows/basic.yaml
cgorenflo, line 119:
```suggestion
        run: cargo sort-check #alias
```

.github/workflows/basic.yaml
```diff
@@ -148,56 +92,34 @@ jobs:
...
-          command: fmt-check # alias
-
-      - name: Install stable toolchain
-        uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
-          toolchain: 1.86.0
-          components: clippy
-          override: true
+        run: cargo +nightly fmt-check #alias
```
cgorenflo, line 116:
```suggestion
        run: cargo +nightly fmt-check # alias
```

PR #998, opened 2025-08-08
feat(ampd): track ampd error in monitoring server

yidan233:
While several AMPD errors are already covered by existing stage and flow metrics, this PR introduces three new metrics to track key failure points that were previously unobserved:
`event_stream_timeout_total`
Tracked when the event stream does not emit any events within the expected duration.

`msg_enqueue_error_total`
Tracked when a message fails to enqueue into the broadcast queue, i.e.exceeding the gas cap,failed simulation...

`event_publisher_error_total`
Tracked when the EventPublisher fails to query or send events

### DIFF
--- a/ampd/src/broadcast/msg_queue.rs
+++ b/ampd/src/broadcast/msg_queue.rs
@@ -1032,16 +1032,13 @@ mod tests {
         let gas_price_amount = 0.025;
         let gas_price_denom = "uaxl";
 
-        let mut cosmos_client = setup_client(&TMAddress::random(PREFIX));
-        cosmos_client.expect_simulate().once().returning(move |_| {
-            Ok(SimulateResponse {
-                gas_info: Some(GasInfo {
-                    gas_wanted: gas_cost,
-                    gas_used: gas_cost,
-                }),
-                result: None,
-            })
+        let gas_info = Some(GasInfo {
+            gas_wanted: gas_cost,
+            gas_used: gas_cost,
         });
+
+        let cosmos_client = setup_client_with_simulate(&TMAddress::random(PREFIX), gas_info, 1);
+
         let broadcaster = broadcaster::Broadcaster::builder()
             .client(cosmos_client)
             .chain_id("chain-id".parse().unwrap())
--- a/ampd/src/broadcast/msg_queue.rs
+++ b/ampd/src/broadcast/msg_queue.rs
@@ -19,8 +19,9 @@ use tracing::{instrument, warn};
 use valuable::Valuable;
 
 use super::{broadcaster, Error, Result};
-use crate::cosmos;
+use crate::monitoring::metrics::Msg;
 use crate::types::TMAddress;
+use crate::{cosmos, monitoring};
 
 type TxResult = std::result::Result<(String, u64), Arc<Report<Error>>>;
 
@@ -247,6 +248,7 @@ impl MsgQueue {
         msg_cap: usize,
         gas_cap: Gas,
         duration: time::Duration,
+        monitoring_client: monitoring::Client,
     ) -> (Pin<Box<MsgQueue>>, MsgQueueClient<T>)
     where
         T: cosmos::CosmosClient,
@@ -257,7 +259,7 @@ impl MsgQueue {
             Box::pin(MsgQueue {
                 stream: ReceiverStream::new(rx).fuse(),
                 deadline: time::sleep(duration),
-                queue: Queue::new(gas_cap),
+                queue: Queue::new(gas_cap, monitoring_client),
                 duration,
             }),
             MsgQueueClient { broadcaster, tx },
@@ -333,14 +335,16 @@ struct Queue {
     msgs: Vec<QueueMsg>,
     gas_cost: Gas,
     gas_cap: Gas,
+    monitoring_client: monitoring::Client,
 }
 
 impl Queue {
-    pub fn new(gas_cap: Gas) -> Self {
+    pub fn new(gas_cap: Gas, monitoring_client: monitoring::Client) -> Self {
         Queue {
             msgs: vec![],
             gas_cost: Gas::default(),
             gas_cap,
+            monitoring_client,
         }
     }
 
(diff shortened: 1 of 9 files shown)


ampd/src/broadcast/msg_queue.rs
```diff
@@ -355,6 +359,10 @@ impl Queue {
                 gas: msg.gas,
                 gas_cap: self.gas_cap,
             };
+            self.monitoring_client
+                .metrics()
+                .record_metric(Msg::MessageEnqueueError);
+
```
maancham, line 365: is this the only type of error that can happen in message queue?

ampd/src/broadcast/msg_queue.rs
```diff
@@ -980,4 +1023,55 @@ mod tests {
...
+        let handle = tokio::spawn(async move {
+            assert!(msg_queue.next().await.is_none());
+        });
+
+        drop(msg_queue_client);
+        handle.await.unwrap();
+
+        let metric = receiver.recv().await.unwrap();
+        assert_eq!(metric, Msg::MessageEnqueueError);
+        assert!(receiver.try_recv().is_err());
```
maancham, line 1075: why use this last assert?

ampd/src/event_processor.rs
```diff
@@ -899,6 +911,134 @@ mod tests {
...
+            "handler".to_string(),
+            MockEventHandler::new(),
+            stream::pending(),
+            event_config,
+            token.child_token(),
+            msg_queue_client,
+            monitoring_client,
+        ));
+
+        tokio::time::sleep(Duration::from_secs(2)).await;
```
maancham, line 959: why are you using `start_paused = true` and `sleep`? Is there a way to use `advance` and not add more wait time to the test?

ampd/src/event_processor.rs
```diff
@@ -899,6 +911,134 @@ mod tests {
...
+        token.cancel();
+        let _ = task.await.unwrap();
+
+        let metric = receiver.recv().await.unwrap();
+        assert_eq!(metric, metrics::Msg::EventPollingTimeout);
+        assert!(receiver.try_recv().is_err());
+    }
+
+    #[tokio::test(start_paused = true)]
+    async fn should_record_msg_enqueue_error_when_msg_enqueue_failed() {
```
maancham, line 970: anyway we could extract some parts (or even use existing test setup functions) to refactor this test into a more concise version? (like gas setup or broadcaster builder stages)

ampd/src/event_sub/mod.rs
```diff
@@ -318,4 +337,66 @@ mod tests {
...
+        // 1st call return LatestBlockQuery error -> record metric
+        assert_err_contains!(stream.next().await.unwrap(), Error, Error::LatestBlockQuery);
+        tokio::time::advance(Duration::from_secs(1)).await;
+
+        // Second call should succeed and emit events
+        assert!(matches!(
+            stream.next().await.unwrap(),
+            Ok(Event::BlockBegin(_))
+        ));
+        tokio::time::advance(Duration::from_secs(1)).await;
```
maancham, line 392: nice

ampd/src/event_sub/mod.rs
```diff
@@ -318,4 +337,66 @@ mod tests {
...
+        });
+
+        let token = CancellationToken::new();
+        let (monitoring_client, mut receiver) = test_utils::monitoring_client();
+        let (event_publisher, subscriber) =
+            EventPublisher::new(tm_client, 100, Duration::from_secs(1), monitoring_client);
+        let mut stream = subscriber.subscribe();
+        let handle = tokio::spawn(event_publisher.run(token.child_token()));
+
+        // 1st call return LatestBlockQuery error -> record metric
```
maancham, line 383: I guess it is better to remove comments from the test here

ampd/src/monitoring/endpoints/metrics.rs
```diff
@@ -59,6 +59,12 @@ pub enum Msg {
         success: bool,
         duration: Duration,
     },
+    /// Record the number of error in msg enqueue
```
maancham, line 62:
```
/// Record the number of errors in message enqueue operations
MessageEnqueueError,
/// Record the number of event stream polling timeouts
EventPollingTimeout,
/// Record the number of errors that occur in the event publisher
EventPublisherError,
```
minor changes

ampd/src/event_sub/mod.rs
```diff
@@ -318,4 +337,66 @@ mod tests {
             finalize_block_events: vec![],
         }
     }
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn should_record_event_publisher_err_successfully() {
```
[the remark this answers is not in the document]
yidan233: the other error happens when 
` let _ = self.tx.send(event).map_err(Report::new).inspect_err(|err| {`
is quite hard to test, I think the original test sets did not include test for it either

ampd/src/grpc/blockchain_service.rs
```diff
@@ -76,6 +78,11 @@ where
         self.msg_queue_client
             .clone()
             .enqueue(msg)
+            .inspect_err(|_| {
+                self.monitoring_client
+                    .metrics()
+                    .record_metric(Msg::MessageEnqueueError);
+            })
```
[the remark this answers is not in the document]
yidan233: no, this is different from the logging inside the queue
The one inside the queue is happens only when a single message gas estimation is larger than the queue capacity and the message is dropped

ampd/src/broadcast/msg_queue.rs
```diff
@@ -345,17 +356,23 @@ impl Queue {
     }
 
     #[instrument(skip(handle_error))]
-    pub fn push_or<F>(&mut self, msg: QueueMsg, handle_error: F) -> Option<nonempty::Vec<QueueMsg>>
+    pub fn push_or<F>(
+        &mut self,
+        msg: QueueMsg,
+        handle_error: F,
+        monitoring_client: &monitoring::Client,
```
cgorenflo, line 363: there is probably no way around adding the monitoring client to MsgQueue, but I definitely don't want to add it here. you can use a closure inside MsgQueue to create a handle_error function that already closes over the monitoring client, so the Queue doesn't have to know about it

PR #1011, opened 2025-08-15
fix(minor-coordinator): coordinator v2 migration to add missing chains

SDavidson1177:
## Description

The previous implementation of the coordinator v2.0.0 migration required the list of `{chain_name, gateway, prover, verifier}` in the migration message to have a 1:1 mapping with the list of `{chain_name, prover}` registered with the coordinator. This requirement may be unnecessarily strict, as migration scripts would need to check for edge cases such as if additional chains are registered on the router, or if a chain does not have a prover or verifier (like `axelar`).

In this PR, I tweak this restriction.  This migration allows missing chain contracts to be added if that chain exists in the router. Chains that do not have valid verifiers, provers or gateways will be ignored. **It is expected that the migration script will be used to generate the migration message, as the script verifies all these conditions.**

## Edge Cases for Migrating Provers
The only state that this migration modifies from Coordinator v1.1.0 is the `chain -> prover` mapping. We want to create a new `chain -> (prover, gateway, verifier)` mapping. There are several edge cases that must be carefully accounted for. I have outlined them below:

### Goal
If a `(chain name, prover)` pair exists in Coordinator v1.1.0, and there exists a `gateway` and `verifier` for that `(chain name, prover)` pair, Coordinator v2 stores `(chain name, prover, gateway, verifier)`. We can find these `gateway`s and `verifiers` by consulting the router. 

AND

If a `(chain name, gateway)` pair exists in the Router, and there exists a valid `prover` for that `chain name`, Coordinator v2 stores `(chain name, prover, gateway, verifier)`. The prover can be queried from the multisig contract.

### Edge Cases

1. `(chain name, prover)` in Coordinator v1.1.0. Gateway and verifier found for `chain name` in router: Add to Coordinator v2
2. `(chain name, prover)` in Coordinator v1.1.0. Gateway and verifier **not** found for `chain name` in router: Delete from Coordinator v2 (ignore incomplete registrations)
3. `(chain name, prover)` **not** in Coordinator v1.1.0. Gateway and verifier found for `chain name` in router: Add to Coordinator v2 if we can query prover from multisig contract (fill in missing data)

This conditions will be checked automatically when constructing the migration message using the migration script: https://github.com/axelarnetwork/axelar-contract-deployments/pull/937

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros


contracts/multisig/src/contract/migrations/mod.rs
```diff
@@ -34,9 +36,39 @@ pub fn migrate(
...
+            }
+            Ok(None) => {
+                let _ = CALLERS_FOR_CHAIN.save(
+                    deps.storage,
+                    &chain_name,
+                    &HashSet::from([contract_address]),
+                );
+            }
+            _ => {}
+        }
```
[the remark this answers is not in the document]
cgorenflo: why do you discard errors here?

contracts/multisig/src/state.rs
```diff
@@ -26,6 +26,7 @@ pub const SIGNING_SESSION_COUNTER: Item<Uint64> = Item::new("signing_session_cou
 pub const SIGNING_SESSIONS: Map<u64, SigningSession> = Map::new("signing_sessions");
 // The keys represent the addresses that can start a signing session.
 pub const AUTHORIZED_CALLERS: Map<&Addr, ChainName> = Map::new("authorized_callers");
+pub const CALLERS_FOR_CHAIN: Map<&ChainName, HashSet<Addr>> = Map::new("callers_for_chain");
```
cgorenflo, line 29: this is not ideal, because you always have to load all addresses per chain. I suggest using a `Map<(ChainName, Addr), ()>` and then using prefix + range if you need to lookup multiple addresses

contracts/multisig/src/state.rs
```diff
@@ -26,6 +26,7 @@ pub const SIGNING_SESSION_COUNTER: Item<Uint64> = Item::new("signing_session_cou
 pub const SIGNING_SESSIONS: Map<u64, SigningSession> = Map::new("signing_sessions");
 // The keys represent the addresses that can start a signing session.
 pub const AUTHORIZED_CALLERS: Map<&Addr, ChainName> = Map::new("authorized_callers");
```
cgorenflo, line 28: this does not have a good name, it's not storing the callers, but rather the chain a caller is authorized for (can one caller be authorized for multiple chains?)

contracts/multisig/src/state.rs
SDavidson1177: Makes sense, I have changed it

contracts/multisig/src/state.rs
SDavidson1177: I have renamed to `CHAIN_CALLER_PAIRS` to better match the structure change

contracts/multisig/src/contract/execute.rs
```diff
@@ -213,9 +216,10 @@ pub fn unauthorize_callers(
     deps: DepsMut,
     contracts: HashMap<Addr, ChainName>,
 ) -> error_stack::Result<Response, ContractError> {
-    contracts.iter().for_each(|(contract_address, _)| {
-        AUTHORIZED_CALLERS.remove(deps.storage, contract_address)
-    });
+    for contract in contracts.keys() {
+        remove_authorized_caller(deps.storage, contract.clone())
+            .map_err(|_| error_stack::report!(ContractError::CallerNotFound(contract.clone())))?;
+    }
```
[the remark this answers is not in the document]
SDavidson1177: `remove_authorized_caller` can cause an error if we can not parse the caller information. So, I'll change the error name to make it less missleading.

contracts/multisig/src/state.rs
```diff
@@ -26,6 +26,7 @@ pub const SIGNING_SESSION_COUNTER: Item<Uint64> = Item::new("signing_session_cou
...
+pub const CHAIN_CALLER_PAIRS: Map<(ChainName, Addr), ()> = Map::new("callers_for_chain");
```
cgorenflo, line 29: make these private. They are tightly coupled and it's dangerous to give other modules the ability to manipulate them independently

cgorenflo, line 29: could you check if (&ChainName, &Addr) work as key as well? Then we wouldn't have to do that much cloning

contracts/multisig/src/contract.rs
```diff
@@ -98,7 +98,10 @@ pub fn execute(
             signed_sender_address,
         } => execute::register_pub_key(deps, info, public_key, signed_sender_address),
         ExecuteMsg::AuthorizeCallers { contracts } => {
-            let contracts = validate_contract_addresses(&deps, contracts)?;
+            let (addrs, names): (Vec<_>, Vec<_>) = contracts.into_iter().unzip();
+            let contracts = validate_contract_addresses(&deps, addrs)?;
+            let contracts = std::iter::zip(contracts, names).collect::<HashMap<Addr, ChainName>>();
+
```
eguajardo, line 104: could you use `map` instead of unzipping then zipping again?

contracts/coordinator/src/contract/migrations/mod.rs
```diff
@@ -412,12 +386,12 @@ mod tests {
         )
         .is_ok());
 
-        let chain_name1 = chain_name!(CHAIN_1);
-        let chain_name2 = chain_name!(CHAIN_2);
+        let chain_name1 = ChainName::try_from(CHAIN_1).unwrap();
```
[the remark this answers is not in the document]
SDavidson1177: Makes sense, just fixed

contracts/multisig/src/state.rs
```diff
@@ -114,6 +115,48 @@ pub fn save_pub_key(
     Ok(pub_keys().save(store, (signer, pub_key.key_type()), &pub_key.into())?)
 }
 
+pub fn save_prover_to_registry(
```
eguajardo, line 118: a bit nitpicky, but could we not use `_registry` suffix for these functions and variables? the reason is that I find it a bit confusing because we have a contract called service registry, maybe just call the map `PROVER_CHAIN` and adjust the function names accordingly? what do you think?

contracts/multisig/src/contract/execute.rs
```diff
@@ -211,19 +214,25 @@ pub fn authorize_callers(
 
 pub fn unauthorize_callers(
     deps: DepsMut,
-    contracts: HashMap<Addr, ChainName>,
+    contracts: Vec<Addr>,
```
eguajardo, line 217:
```suggestion
    callers: Vec<Addr>,
```
probably better if we use `callers` to match the function name?

contracts/multisig/src/contract/execute.rs
```diff
@@ -211,19 +214,25 @@ pub fn authorize_callers(
...
-    });
+    let callers_and_chains = contracts
+        .into_iter()
+        .map(|contract| {
+            remove_prover_from_registry(deps.storage, contract.clone())
+                .map(|cc| cc.map(|chain| (contract, chain)))
+                .change_context(ContractError::InvalidCaller)
+        })
+        .filter_map_ok(|contract| contract)
+        .collect::<error_stack::Result<Vec<(Addr, ChainName)>, ContractError>>()?;
```
eguajardo, line 227:
```suggestion
    let callers_and_chains = contracts
        .into_iter()
        .filter_map(|contract| {
            remove_prover_from_registry(deps.storage, contract.clone())
                .map(|cc| cc.map(|chain| (contract, chain)))
                .change_context(ContractError::InvalidCaller)
                .transpose()
        })
        .collect::<error_stack::Result<Vec<(Addr, ChainName)>, ContractError>>()?;
```

contracts/multisig/src/contract/execute.rs
```diff
@@ -181,11 +184,11 @@ pub fn require_authorized_caller(
...
-) -> Result<bool, ContractError> {
-    Ok(AUTHORIZED_CALLERS
-        .may_load(storage, sender_addr)
-        .map_err(ContractError::from)?
-        == Some(chain_name.clone()))
+) -> error_stack::Result<bool, ContractError> {
+    let chain = chain_by_prover(storage, sender_addr.clone())
+        .change_context(ContractError::InvalidCaller)?;
+
+    Ok(chain == Some(chain_name.clone()))
```
eguajardo, line 191: could you use `query::caller_authorized` instead?

contracts/multisig/src/contract/execute.rs
```diff
@@ -214,16 +212,16 @@ pub fn authorize_callers(
...
-    let callers_and_chains = contracts
+    let callers_and_chains = callers
         .into_iter()
-        .map(|contract| {
-            remove_prover_from_registry(deps.storage, contract.clone())
+        .filter_map(|contract| {
+            remove_prover(deps.storage, &contract)
                 .map(|cc| cc.map(|chain| (contract, chain)))
-                .change_context(ContractError::InvalidCaller)
+                .change_context(ContractError::InvalidProver)
```
eguajardo, line 222: this error seems off too, because from current context we can't tell if the error was caused by an invalid prover. So, maybe use something like `ContractError:Storage` or `ProverRemovalFailed`?

contracts/multisig/src/client.rs
```diff
@@ -172,6 +178,13 @@ impl Client<'_> {
             .query(&msg)
             .change_context_lazy(|| Error::for_query(msg))
     }
+
+    pub fn authorized_callers(&self, chain_name: ChainName) -> Result<HashSet<Addr>, Error> {
+        let msg = QueryMsg::AuthorizedCallers { chain_name };
+        self.client
+            .query(&msg)
+            .change_context_lazy(|| Error::for_query(msg))
+    }
```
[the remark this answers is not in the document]
SDavidson1177: Fixed

contracts/multisig/src/contract/query.rs
```diff
@@ -30,12 +30,18 @@ pub fn public_key(deps: Deps, verifier: Addr, key_type: KeyType) -> StdResult<Pu
...
+    Ok(chain_by_prover(storage, &address)?
         .filter(|c| c == &chain_name)
         .is_some())
 }
 
-pub fn callers_for_chain(deps: Deps, chain_name: ChainName) -> HashSet<Addr> {
-    provers_by_chain(deps.storage, chain_name.clone()).collect()
+pub fn prover_for_chain(deps: Deps, chain_name: ChainName) -> Result<Addr, ContractError> {
+    prover_by_chain(deps.storage, chain_name.clone())
+        .change_context(ContractError::InvalidChainName)?
```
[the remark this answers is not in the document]
SDavidson1177: I'll change it to be a storage error as you suggested

eguajardo (approved):

PR #1023, opened 2025-08-27
feat(coordinator): add instantiate2 address getter

SDavidson1177:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros

## Notes

Requested here: https://github.com/axelarnetwork/arcs/pull/18#discussion_r2282132488

### DIFF
--- a/contracts/coordinator/src/contract/execute.rs
+++ b/contracts/coordinator/src/contract/execute.rs
@@ -1,11 +1,12 @@
 use std::collections::{HashMap, HashSet};
 
 use axelar_wasm_std::nonempty;
-use cosmwasm_std::{Addr, Binary, DepsMut, Env, Response, Storage, WasmMsg, WasmQuery};
+use cosmwasm_std::{Addr, Binary, DepsMut, Env, Response, Storage, WasmMsg};
 use error_stack::{Result, ResultExt};
 use router_api::ChainName;
 
 use crate::contract::errors::Error;
+use crate::contract::query::instantiate2_addr;
 use crate::events::{ContractInstantiation, Event};
 use crate::msg::{DeploymentParams, ProverMsg, VerifierMsg};
 use crate::state;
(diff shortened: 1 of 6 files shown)


contracts/coordinator/src/contract/query.rs
```diff
@@ -94,3 +94,24 @@ pub fn chain_contracts_info(
...
+                &deps
+                    .api
+                    .addr_canonicalize(&env.contract.address.to_string().clone())
+                    .change_context(Error::Instantiate2Address)?,
+                salt,
+            )
+            .change_context(Error::Instantiate2Address)?,
+        )
+        .change_context(Error::Instantiate2Address)
+}
```
cgorenflo, line 117: try to keep the error specificity to what caused the error, not the overall call. When someone calls `instantiate2_addr` and get an error, it's clear that this call failed. More useful is the information why the call failed. Then in turn, when for example instantiate_chain_contracts fails, that's the correct level to change to `Error::Instantiate2Address` to provide the information why that call failed

integration-tests/tests/coordinator_one_click_deployment.rs
cgorenflo, line 1: as discussed, the tests don't quite capture the use case

contracts/coordinator/src/contract/execute.rs
```diff
@@ -55,27 +56,6 @@ pub fn set_active_verifier_set(
...
-                    .change_context(Error::Instantiate2Address)?,
-                salt,
-            )
-            .change_context(Error::Instantiate2Address)?,
-        )
-        .change_context(Error::Instantiate2Address)
-}
-
 fn launch_contract(
     deps: &DepsMut,
```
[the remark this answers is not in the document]
SDavidson1177: Changed to Deps

contracts/coordinator/src/contract/query.rs
SDavidson1177: Makes sense, changed the error message to "failed to canonicalize address"

cgorenflo (approved):

PR #1027, opened 2025-08-29
feat(packages): add limit type for paginating queries

SDavidson1177:
## Description

It is a good idea to paginate queries that return iterable objects, such as `Vec`s or `HashSet`s, because returning long lists can lead to high gas costs. In those cases, we allow the caller to limit the amount of returned values by supplying an unsigned integer `n`. The query endpoint then uses the iterator's `.take(n)` method to limit the amount of returned entries to `n`.

I propose that all queries that return iterable objects take as argument `limit: Limit`. The query should then use `.take(limit.into())` to return at most `limit` elements. This prohibits the query from returning unlimited results. Furthermore, `Limit` objects are prohibited from having a value of 0.

## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,26 @@
 
 ## [Unreleased](https://github.com/axelarnetwork/axelar-amplifier/tree/HEAD)
 
-[Full Changelog](https://github.com/axelarnetwork/axelar-amplifier/compare/ampd-v1.11.0..HEAD)
+[Full Changelog](https://github.com/axelarnetwork/axelar-amplifier/compare/ampd-v1.12.0..HEAD)
+
+## [v1.12.0](https://github.com/axelarnetwork/axelar-amplifier/tree/ampd-v1.12.0) (2025-09-19)
+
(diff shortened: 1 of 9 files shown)


packages/limit/src/lib.rs
cgorenflo, line 1: is it worth it to create a separate type instead of using nonempty::u32 (or u64)?

packages/router-api/src/msg.rs
```diff
@@ -60,7 +67,8 @@ pub enum QueryMsg {
     #[returns(Vec<ChainEndpoint>)]
     Chains {
         start_after: Option<ChainName>,
-        limit: Option<u32>,
+        #[serde(default = "default_pagination_limit")]
+        limit: u32,
```
cgorenflo, line 71: how does the default work when the type is not an option?

packages/limit/src/lib.rs
SDavidson1177: Good point. Since `.take()` takes a usize as an argument, I'll use `nonempty::Usize`. I'll add a `try_from(u32)` method to `nonempty::Usize`.

PR #1032, opened 2025-09-04
chore(ampd): deserialize cosmrs message values

maancham:
## Todos

- [ ] Unit tests
- [ ] Manual tests
- [ ] Documentation
- [ ] Connect epics/issues

## Convention Checklist
- [ ] Derive macros

### DIFF
--- a/ampd/src/event_processor.rs
+++ b/ampd/src/event_processor.rs
@@ -137,6 +137,68 @@ fn is_avalanche_evm_handler(handler_label: &str) -> bool {
     evm_handlers.iter().any(|handler| handler_label == *handler)
 }
 
+fn show_protobuf_structure(bytes: &[u8]) -> Result<String, Box<dyn std::error::Error>> {
+    use prost::encoding::{decode_key, decode_varint, WireType};
+
+    let mut result = Vec::new();
+    let mut buf = bytes;
+    let mut field_count = 0;
+
+    while !buf.is_empty() && field_count < 20 {
+        if let Ok((field_num, wire_type)) = decode_key(&mut buf) {
+            match wire_type {
+                WireType::Varint => {
+                    if let Ok(value) = decode_varint(&mut buf) {
+                        result.push(format!("Field {}: Varint = {}", field_num, value));
+                    }
+                }
+                WireType::LengthDelimited => {
+                    if let Ok(len) = decode_varint(&mut buf) {
+                        if buf.len() >= len as usize {
+                            let data = &buf[..len as usize];
+                            buf = &buf[len as usize..];
+
+                            if let Ok(s) = String::from_utf8(data.to_vec()) {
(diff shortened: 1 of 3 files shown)


ampd/src/event_processor.rs
```diff
@@ -137,6 +138,70 @@ fn is_avalanche_evm_handler(handler_label: &str) -> bool {
     evm_handlers.iter().any(|handler| handler_label == *handler)
 }
 
+fn show_protobuf_structure(bytes: &[u8]) -> Result<String, Box<dyn std::error::Error>> {
```
eguajardo, line 141: could you add a TODO to indicate that this code should be removed at some point?

ampd/src/event_processor.rs
```diff
@@ -137,6 +138,70 @@ fn is_avalanche_evm_handler(handler_label: &str) -> bool {
...
+                            } else {
+                                break;
+                            }
+                        } else {
+                            break;
+                        }
+                    }
+                }
+                WireType::SixtyFourBit => {
+                    if buf.len() >= 8 {
```
eguajardo, line 179: is it even possible to have len < 8? what happens then?

ampd/src/event_processor.rs
```diff
@@ -137,6 +138,70 @@ fn is_avalanche_evm_handler(handler_label: &str) -> bool {
...
+                                } else {
+                                    result.push(format!(
+                                        "Field {}: Bytes = {} bytes",
+                                        field_num,
+                                        data.len()
+                                    ));
+                                }
+                            } else {
+                                break;
+                            }
```
eguajardo, line 172: when does it falls in this case? do we want to break or continue?

ampd/src/event_processor.rs
```diff
@@ -164,11 +229,14 @@ where
         Ok(msgs) => {
             if is_avalanche_evm_handler(handler_label) {
                 for (i, msg) in msgs.iter().enumerate() {
+                    let deserialized_values = show_protobuf_structure(&msg.value).unwrap();
```
eguajardo, line 232: I don't think we should panic or error here, because if there is a bug or unexpected error on this function this will break event handling. Maybe log the error message instead?

ampd/src/event_processor.rs
```diff
@@ -137,6 +138,70 @@ fn is_avalanche_evm_handler(handler_label: &str) -> bool {
...
+                }
+                WireType::ThirtyTwoBit => {
+                    if buf.len() >= 4 {
+                        let bytes = &buf[..4];
+                        let value = u32::from_le_bytes(bytes.try_into().unwrap());
+                        buf = &buf[4..];
+                        result.push(format!("Field {}: 32-bit = {}", field_num, value));
+                    }
+                }
+                _ => break,
```
eguajardo, line 194: should this really break or keep going to the next field?

cgorenflo (approved):

PR #1042, opened 2025-09-15
fix(ampd): retry when there is a sequence mismatch error

eguajardo:
## Convention Checklist
- [ ] Derive macros

### DIFF
--- a/ampd/src/broadcast/broadcaster.rs
+++ b/ampd/src/broadcast/broadcaster.rs
@@ -202,9 +202,7 @@ where
     ///
     /// * `Error::EstimateGas` - If the gas estimation fails
     pub async fn estimate_gas(&mut self, msgs: Vec<Any>) -> Result<Gas> {
-        let acc_sequence = self.acc_sequence.read().await;
-
-        cosmos::estimate_gas(&mut self.client, msgs, self.pub_key, *acc_sequence)
+        cosmos::estimate_gas(&mut self.client, msgs, self.pub_key)
             .await
             .change_context(Error::EstimateGas)
     }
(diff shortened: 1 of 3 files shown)


ampd/src/broadcast/broadcaster.rs
```diff
@@ -202,11 +195,27 @@ where
...
+            {
+                Ok(gas) => Ok(gas),
+                Err(e) => match parse_sequence_error(e.to_string()) {
+                    // Retry with the expected sequence number
+                    Some(expected_seq) => {
+                        *acc_sequence = expected_seq;
+
+                        cosmos::estimate_gas(&mut self.client, msgs, self.pub_key, *acc_sequence)
+                            .await
+                    }
```
cjcobb23, line 212: what is the actual root cause of the issue? is there a possibility that the expected is also not correct, because something changes before the next broadcast? would be helpful to describe the problem in the comment here

eguajardo: added a description of possible reasons why a mismatch could happen. The root cause was that the previous fix was resetting the sequence each time a message was added to the queue, but there were still unconfirmed transactions, so the account sequence expected for simulation didn't match the account sequence that was fetch when querying the account

cjcobb23: how is the sequence determined in usual operations?

eguajardo: when the broadcaster is instantiated, the account is queried to get the sequence number:
https://github.com/axelarnetwork/axelar-amplifier/blob/07027683949a25a59956b54787721eb788e0b9c8/ampd/src/broadcast/broadcaster.rs#L138-L148
Then after each successful broadcast, the sequence number is incremented:
https://github.com/axelarnetwork/axelar-amplifier/blob/07027683949a25a59956b54787721eb788e0b9c8/ampd/src/broadcast/broadcaster.rs#L259-L264

cjcobb23 (approved):

PR #1055, opened 2025-09-29
chore(ampd): add name to task runs

maancham:
## Description

This PR adds task name to ampd tasks to help with debugging sub-tasks inside a group.

## Convention Checklist
- [ ] Derive macros

### DIFF
--- a/ampd/src/asyncutil/task.rs
+++ b/ampd/src/asyncutil/task.rs
@@ -2,11 +2,11 @@ use std::future::Future;
 use std::pin::Pin;
 
 use axelar_wasm_std::error::extend_err;
-use error_stack::{Context, Result, ResultExt};
+use error_stack::{Context, Result};
 use thiserror::Error;
 use tokio::task::JoinSet;
 use tokio_util::sync::CancellationToken;
-use tracing::info;
+use tracing::{info, warn};
 
 type PinnedFuture<T> = Pin<Box<dyn Future<Output = T> + Send>>;
 /// This type represents an awaitable action that can be cancelled. It abstracts away the necessary boxing and pinning
@@ -36,7 +36,7 @@ where
     E: From<TaskError> + Context,
 {
     name: String,
-    tasks: Vec<CancellableTask<Result<(), E>>>,
+    tasks: Vec<(String, CancellableTask<Result<(), E>>)>,
 }
 
 impl<E> TaskGroup<E>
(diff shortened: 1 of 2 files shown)


ampd/src/asyncutil/task.rs
```diff
@@ -51,75 +52,116 @@ where
...
     use tokio_util::sync::CancellationToken;
 
-    use crate::asyncutil::task::{CancellableTask, TaskError, TaskGroup};
+    use crate::asyncutil::task::{CancellableTask, TaskError, TaskGroup, TaskGroupError};
+
+    impl From<TaskGroupError> for TaskError {
+        fn from(_: TaskGroupError) -> Self {
+            TaskError {}
+        }
+    }
```
cgorenflo, line 164: still needed?

ampd/src/asyncutil/task.rs
```diff
@@ -135,53 +177,85 @@ mod test {
...
+                "error_task_4",
+                CancellableTask::create(|token| async move {
+                    token.cancelled().await;
+                    Err(report!(TaskError {}))
+                }),
+            );
         let result = tasks.run(CancellationToken::new()).await;
         let err = result.unwrap_err();
-        assert_eq!(err.current_frames().len(), 4);
+        assert_eq!(err.current_frames().len(), 1);
```
cgorenflo, line 231: while it succeeds, this test isn't very useful anymore. The point of assertions is to make sure our requirements are met. We don't care that there is only one top frame. What we do care about is that the entire error history is captured. Since all the errors are the same we don't have to worry about reordering and can use goldie to check the output

ampd/src/asyncutil/task.rs
```diff
@@ -135,53 +177,85 @@ mod test {
...
+                CancellableTask::create(|_| async { Ok(()) }),
+            )
+            .add_task(
+                "error_task_2",
+                CancellableTask::create(|_| async { Err(report!(TaskError {})) }),
+            );
         let result = tasks.run(CancellationToken::new()).await;
         let err = result.unwrap_err();
-        assert_eq!(err.current_frames().len(), 3);
+        assert_eq!(err.current_frames().len(), 1);
```
cgorenflo, line 259: similarly here, find a deterministic way to check the error is what we expect

ampd/src/lib.rs
```diff
@@ -823,6 +945,8 @@ pub enum Error {
     Connection,
     #[error("task execution failed")]
     Task(#[from] TaskError),
+    #[error("App (task group) execution failed")]
```
cgorenflo, line 948:
```suggestion
    #[error("app failed")]
```
the error will look like this: `app failed: task group failed: task failed: ...`

cgorenflo (approved):

PR #1173, opened 2026-06-01
chore: fix clippy warnings + bump `cosmwasm-check`

anstylian:
### why
This PR:

1. Fixes Clippy lints introduced by newer Clippy versions.
1. Updates the `cosmwasm-check` version used in CI.
1. Fixes the cargo wasm subcommand to exclude crates that should not be compiled for CosmWasm.


rista404 (approved):
lgtm, thanks!

